PR #1, opened 2017-06-09
Compute and Disk fixes

catsby:
This PR fixes a handful of failing tests related to Disks. @danawillow can you take a peak? Related to https://github.com/hashicorp/terraform/pull/15095#issuecomment-307188973

### DIFF
--- a/google/resource_compute_disk_test.go
+++ b/google/resource_compute_disk_test.go
@@ -292,7 +292,7 @@ func testAccComputeDisk_deleteDetach(instanceName, diskName string) string {
 	return fmt.Sprintf(`
 resource "google_compute_disk" "foo" {
 	name = "%s"
-	image = "debian-8"
+	image = "debian-8-jessie-v20170523"
 	size = 50
 	type = "pd-ssd"
 	zone = "us-central1-a"
--- a/google/resource_compute_disk.go
+++ b/google/resource_compute_disk.go
@@ -69,6 +69,7 @@ func resourceComputeDisk() *schema.Resource {
 			"size": &schema.Schema{
 				Type:     schema.TypeInt,
 				Optional: true,
+				Computed: true,
 			},
 
 			"self_link": &schema.Schema{
@@ -85,6 +86,7 @@ func resourceComputeDisk() *schema.Resource {
 			"type": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
+				Default:  "pd-standard",
 				ForceNew: true,
 			},
 			"users": &schema.Schema{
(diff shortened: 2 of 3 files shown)


catsby:
```
TF_ACC=1 go test ./google -v -run=TestAccComputeDisk_ -timeout 120m
=== RUN   TestAccComputeDisk_importBasic
--- PASS: TestAccComputeDisk_importBasic (24.23s)
=== RUN   TestAccComputeDisk_basic
--- PASS: TestAccComputeDisk_basic (23.71s)
=== RUN   TestAccComputeDisk_updateSize
--- PASS: TestAccComputeDisk_updateSize (35.45s)
=== RUN   TestAccComputeDisk_fromSnapshotURI
--- PASS: TestAccComputeDisk_fromSnapshotURI (128.16s)
=== RUN   TestAccComputeDisk_encryption
--- PASS: TestAccComputeDisk_encryption (23.91s)
=== RUN   TestAccComputeDisk_deleteDetach
--- PASS: TestAccComputeDisk_deleteDetach (99.33s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 334.824s
```

google/resource_compute_disk_test.go
```diff
@@ -292,7 +292,7 @@ func testAccComputeDisk_deleteDetach(instanceName, diskName string) string {
 	return fmt.Sprintf(`
 resource "google_compute_disk" "foo" {
 	name = "%s"
-	image = "debian-8"
+	image = "debian-8-jessie-v20170523"
```
catsby: It seems that before https://github.com/hashicorp/terraform/pull/14573 the `google_compute_disk` resource didn't READ much back from the API. I suppose the API allows `debian-8` to initialize as a convince for the user, but the API seems to give it a more concrete value in the backend. 14573 introduced reading these values back, thus we see the diff shown in the test failures. 

I believe the new behavior is the correct behavior, and we should discourage users from simply using the `debian-8` shorthand going forward.

danawillow: > we should discourage users from simply using the debian-8 shorthand going forward.

Given that these tests fail, I think we've inadvertently already started discouraging users from using it by breaking them 😞 

Mind filing an issue to make our lack of support for that shorthand more explicit? I think there are a few places in the code where that needs to be changed.

Once that's done, feel free to merge this when you're ready 👍

catsby: > Mind filing an issue to make our lack of support for that shorthand more explicit? I think there are a few places in the code where that needs to be changed

I'm not as familiar with the GCP provider as I'd like, what other places need changing?

It's true, we should not support that shorthand anymore, I can update the docs

PR #119, opened 2017-06-14
Update existing resources import docs

rileykarson:
We can improve the bottom-up discovery of importable GCP resources by improving the documentation on the page of resources that can be imported.

In each `Import` block;


stack72:
Hi @rileykarson 

This LGTM! Thanks for this :)

Paul

PR #120, opened 2017-06-14
Add Import blocks to docs for 5 various resources

rileykarson:
Add docs specifying how to import 5 more of our 21 importable resources, bringing the total to 11. Specifically:

- google_bigquery_dataset
- google_compute_address
- google_compute_autoscaler
- google_compute_disk
- google_compute_firewall

Like in #119:

In each Import block;


stack72:
Great addition @rileykarson :) Thanks for this!

PR #122, opened 2017-06-14
Add `boot_disk` property to `google_compute_instance`

danawillow:
This is change 1 in a series of changes to provide a better experience for attaching/detaching disks to/from instances. The end goal is to have a separate resource for attached disks, in order to have some attached disks that are managed by terraform side-by-side with disks that are managed some other way (such as by k8s), and to allow users to attach/detach disks (see #33)

The order of the changes will look roughly like:
1. (this) Add `boot_disk` property
2. Add `scratch_disk` property and deprecate `disk`
3. Remove `disk` property, make `boot_disk` Required
4. Add `google_compute_attached_disk` resource and deprecate the `attached_disk` property

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -45,6 +45,7 @@ func resourceComputeInstance() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 							Computed: true,
+							ForceNew: true,
 						},
 
 						"disk_encryption_key_raw": &schema.Schema{
@@ -52,6 +53,11 @@ func resourceComputeInstance() *schema.Resource {
 							Optional:  true,
 							ForceNew:  true,
 							Sensitive: true,
+						},
+
+						"disk_encryption_key_sha256": &schema.Schema{
+							Type:     schema.TypeString,
+							Computed: true,
 						},
 
 						"initialize_params": &schema.Schema{
@@ -1351,6 +1357,9 @@ func flattenBootDisk(d *schema.ResourceData, disk *compute.AttachedDisk) []map[s
 		"source":      sourceUrl[len(sourceUrl)-1],
 		// disk_encryption_key_raw is not returned from the API, so don't store it in state.
 		// If necessary in the future, this can be copied from what the user originally specified.
+	}
+	if disk.DiskEncryptionKey != nil {
+		result["disk_encryption_key_sha256"] = disk.DiskEncryptionKey.Sha256
 	}
 	if v, ok := d.GetOk("boot_disk.0.initialize_params.#"); ok {
 		result["initialize_params.#"] = v.(int)
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -484,9 +484,9 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 	// Build up the list of disks
 
 	disks := []*compute.AttachedDisk{}
-	var bootDisk *compute.AttachedDisk
-	if _, ok := d.GetOk("boot_disk"); ok {
-		bootDisk, err = expandBootDisk(d, config, zone, project)
+	var hasBootDisk bool
+	if _, hasBootDisk = d.GetOk("boot_disk"); hasBootDisk {
+		bootDisk, err := expandBootDisk(d, config, zone, project)
 		if err != nil {
 			return err
 		}
@@ -496,7 +496,7 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 	disksCount := d.Get("disk.#").(int)
 	attachedDisksCount := d.Get("attached_disk.#").(int)
 
-	if disksCount+attachedDisksCount == 0 && bootDisk == nil {
+	if disksCount+attachedDisksCount == 0 && !hasBootDisk {
 		return fmt.Errorf("At least one disk, attached_disk, or boot_disk must be set")
 	}
 	for i := 0; i < disksCount; i++ {
@@ -508,7 +508,7 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 		var disk compute.AttachedDisk
 		disk.Type = "PERSISTENT"
 		disk.Mode = "READ_WRITE"
-		disk.Boot = i == 0 && bootDisk == nil
+		disk.Boot = i == 0 && !hasBootDisk
 		disk.AutoDelete = d.Get(prefix + ".auto_delete").(bool)
 
 		if _, ok := d.GetOk(prefix + ".disk"); ok {
@@ -599,7 +599,7 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 			AutoDelete: false, // Don't allow autodelete; let terraform handle disk deletion
 		}
 
-		disk.Boot = i == 0 && disksCount == 0 && bootDisk == nil
+		disk.Boot = i == 0 && disksCount == 0 && !hasBootDisk
 
 		if v, ok := d.GetOk(prefix + ".device_name"); ok {
 			disk.DeviceName = v.(string)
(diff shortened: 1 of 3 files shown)


danawillow:
Also note that tests are currently failing, but will be fixed in #1.

google/resource_compute_instance.go
```diff
@@ -394,12 +469,23 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 	}
 
 	// Build up the list of disks
+
+	disks := []*compute.AttachedDisk{}
+	var bootDisk *compute.AttachedDisk
```
rileykarson: We keep the `bootdisk` object in scope for the entire function but only use it to perform `nil` checks outside of the `if` block below. What if we moved it inside the `if` block, and pulled up the `ok` result of `d.GetOk("boot_disk");`, giving it a name like `hasBootDisk`?

That way, we know that code later in the function can only interact with the `bootDisk` as part of `disks`, and we can make lines like

```go
disk.Boot = i == 0 && bootDisk == nil
```

look like:

```go
disk.Boot = i == 0 && !hasBootDisk
```

google/resource_compute_instance.go
```diff
@@ -1142,3 +1235,79 @@ func resourceInstanceTags(d *schema.ResourceData) *compute.Tags {
 
 	return tags
 }
+
+func expandBootDisk(d *schema.ResourceData, config *Config, zone *compute.Zone, project string) (*compute.AttachedDisk, error) {
```
rileykarson: It looks like we only use schema elements that are properties of `boot_disk.0`; we can reduce the information this function needs by passing in a limited set of data, like [BackendService's expandBackends()](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_backend_service.go#L354)

google/resource_compute_instance.go
```diff
@@ -1142,3 +1235,79 @@ func resourceInstanceTags(d *schema.ResourceData) *compute.Tags {
...
+			}
+
+			disk.InitializeParams.SourceImage = imageUrl
+		}
+	}
+
+	return disk, nil
+}
+
+func flattenBootDisk(d *schema.ResourceData, disk *compute.AttachedDisk) []map[string]interface{} {
```
rileykarson: Same as expandBootDisk - do we need access to all of `schema.ResourceData` here? Or just a subset?

google/resource_compute_instance_test.go
```diff
@@ -1187,6 +1248,49 @@ func testAccComputeInstance_attachedDisk(disk, instance string) string {
 	}`, disk, instance)
 }
 
+func testAccComputeInstance_bootDisk(instance string) string {
+	return fmt.Sprintf(`
+	resource "google_compute_instance" "foobar" {
```
rileykarson: Nit: Can you de-indent both of these configs so that the top level is against the left side of the file? That way, they are easier to copy out and modify/test interactively, or for users who browse the source to pull out pre-tested .tf configurations.

google/resource_compute_instance_test.go
```diff
@@ -1187,6 +1248,49 @@ func testAccComputeInstance_attachedDisk(disk, instance string) string {
...
+			initialize_params {
+				image = "debian-8-jessie-v20160803"
+			}
+			disk_encryption_key_raw = "SGVsbG8gZnJvbSBHb29nbGUgQ2xvdWQgUGxhdGZvcm0="
+		}
+
+		network_interface {
+			network = "default"
+		}
+	}`, instance)
```
rileykarson: Nit: can the ` be moved to the next line? This would also make using these tests in .tf configs easier.

google/resource_compute_instance.go
danawillow: Done.

google/resource_compute_instance.go
danawillow: The problem with that approach is that it auto-populates all the map keys, and sets them to default values. So if you want to check the existence of something, you have to check against "", 0, false, and nil, rather than being able to use a consistent `d.GetOk("whatever"); ok`. It _seems_ like you should be able to just do `v, ok := data["whatever"]; ok`, but that will return true for elements not set in the config, which is a problem when you're then trying to use that value in another call, like in line 1274. This way feels less bug-prone.

google/resource_compute_instance_test.go
danawillow: I'll do it in a separate PR so it's the same for all of them since there are a lot of tests in this file.

google/resource_compute_instance_test.go
danawillow: likewise

google/resource_compute_instance.go
rileykarson: That makes sense, thanks!

rileykarson:
LGTM :+1:

google/resource_compute_instance_test.go
danawillow: Separate PR is submitted, updated here (and other comment as well)

google/resource_compute_instance.go
```diff
@@ -25,6 +26,80 @@ func resourceComputeInstance() *schema.Resource {
...
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
+							ForceNew: true,
+						},
+
+						"device_name": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
```
paddycarver: Does this need to be ForceNew?

google/resource_compute_instance.go
```diff
@@ -394,12 +469,23 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 	}
 
 	// Build up the list of disks
+
+	disks := []*compute.AttachedDisk{}
+	var hasBootDisk bool
+	if _, hasBootDisk = d.GetOk("boot_disk"); hasBootDisk {
```
paddycarver: Nit: `if _, hasBootDisk :=` means you don't need `var hasBootDisk bool`.

google/resource_compute_instance.go
```diff
@@ -1142,3 +1235,79 @@ func resourceInstanceTags(d *schema.ResourceData) *compute.Tags {
...
+}
+
+func flattenBootDisk(d *schema.ResourceData, disk *compute.AttachedDisk) []map[string]interface{} {
+	sourceUrl := strings.Split(disk.Source, "/")
+	result := map[string]interface{}{
+		"auto_delete": disk.AutoDelete,
+		"device_name": disk.DeviceName,
+		"source":      sourceUrl[len(sourceUrl)-1],
+		// disk_encryption_key_raw is not returned from the API, so don't store it in state.
+		// If necessary in the future, this can be copied from what the user originally specified.
```
paddycarver: I think in other places we do the fingerprint of the key. Is that returned from the API?

google/resource_compute_instance.go
danawillow: Yup. Fixed.

google/resource_compute_instance.go
danawillow: Yup, added

paddycarver:
Rebase and LGTM. 👍

google/resource_compute_instance.go
pydevops: The Forcenew:true would cause the current resource such as a db instance to be recreated. is this desired? maybe I am asking a terraform basic question instead of a google provider one here. thanks.

danawillow: It means that if the field in question changes in value then it would, yes.

johnmehler:
This seems like the best place to ask without opening a new ticket-- why does resizing the boot disk force creation of a new resource? Example code below:

```
boot_disk {
    initialize_params {
      image = "${var.machine_image_id}"
      type  = "${var.boot_disk_type}"
      size  = "${var.boot_disk_size_gb}"
    }
  }
allow_stopping_for_update = true
```

If the size param is changed and `terraform apply` is run again, the build plan will include something like

`boot_disk.0.initialize_params.0.size:                "10" => "200" (forces new resource)`

In the console, you can resize (or at least increase the size of) a the boot disk without destroying and recreating it. Is it possible to do so in Terraform?

Issue #24, opened 2017-06-13
Support for interface on disk in google_compute_instance


hashibot:
_This issue was originally opened by @devth as hashicorp/terraform#5598. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

I don't see an option for `interface`: https://www.terraform.io/docs/providers/google/r/compute_instance.html#disk

> interface
> Optional. The kind of disk interface exposed to the VM for this SSD. Valid values are SCSI and NVME. SCSI is the default and is supported by more guest operating systems. NVME may provide higher performance.

From https://cloud.google.com/sdk/gcloud/reference/compute/instances/create

rileykarson:
Closed with #123.

PR #123, opened 2017-06-14
Add `scratch_disk` property to `google_compute_instance` and deprecate `disk`

danawillow:
This is change 1 in a series of changes to provide a better experience for attaching/detaching disks to/from instances. The end goal is to have a separate resource for attached disks, in order to have some attached disks that are managed by terraform side-by-side with disks that are managed some other way (such as by k8s), and to allow users to attach/detach disks (see #33)

The order of the changes will look roughly like:
1. Add `boot_disk` property
2. (this) Add `scratch_disk` property and deprecate `disk`
3. Remove `disk` property, make `boot_disk` Required
4. Add `google_compute_attached_disk` resource and deprecate the `attached_disk` property

Note that this change branches off of #122. It can still be reviewed by only looking at the final two commits.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -107,10 +107,27 @@ func resourceComputeInstance() *schema.Resource {
 				},
 			},
 
-			"disk": &schema.Schema{
+			"scratch_disk": &schema.Schema{
 				Type:     schema.TypeList,
 				Optional: true,
 				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"interface": &schema.Schema{
+							Type:         schema.TypeString,
+							Optional:     true,
+							Default:      "SCSI",
+							ValidateFunc: validation.StringInSlice([]string{"SCSI", "NVME"}, false),
+						},
+					},
+				},
+			},
+
+			"disk": &schema.Schema{
+				Type:       schema.TypeList,
+				Optional:   true,
+				ForceNew:   true,
+				Deprecated: "Use boot_disk, scratch_disk, and attached_disk instead",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						// TODO(mitchellh): one of image or disk is required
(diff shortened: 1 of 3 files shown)


google/resource_compute_instance.go
```diff
@@ -480,6 +497,15 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 		disks = append(disks, bootDisk)
 	}
 
+	scratchDisks := []*compute.AttachedDisk{}
```
rileykarson: Same as https://github.com/terraform-providers/terraform-provider-google/pull/122#discussion_r122057157, we can keep this element in block scope instead of function scope.

That would mean changing the one other use of `scratchDisks` from

```go
if len(scratchDisks) > 0 {
```

to

```go
if d.Get("scratch_disk.#").(int) > 0
```

which looks a bit grosser, but holds better semantics to me - we are giving a schema error by comparing schema elements.

google/resource_compute_instance_test.go
```diff
@@ -1331,6 +1380,34 @@ func testAccComputeInstance_local_ssd(instance string) string {
 	}`, instance)
 }
 
+func testAccComputeInstance_scratchDisk(instance string) string {
+	return fmt.Sprintf(`
+	resource "google_compute_instance" "scratch" {
```
rileykarson: Same nit as https://github.com/terraform-providers/terraform-provider-google/pull/122#discussion_r122062396 and https://github.com/terraform-providers/terraform-provider-google/pull/122#discussion_r122062678;

danawillow:
(will address comments later)

Note that this also ends up fixing https://github.com/terraform-providers/terraform-provider-google/issues/24.

google/resource_compute_instance.go
danawillow: I just went ahead and used a bool again.

google/resource_compute_instance_test.go
danawillow: Done.

google/resource_compute_instance.go
rileykarson: That works even better! :+1:

rileykarson:
Thanks for the changes! LGTM.

google/resource_compute_instance_test.go
```diff
@@ -749,7 +813,7 @@ func testAccCheckComputeInstanceDisk(instance *compute.Instance, source string,
 		}
 
 		for _, disk := range instance.Disks {
-			if strings.LastIndex(disk.Source, "/"+source) == len(disk.Source)-len(source)-1 && disk.AutoDelete == delete && disk.Boot == boot {
+			if strings.HasSuffix(disk.Source, source) && disk.AutoDelete == delete && disk.Boot == boot {
```
paddycarver: Won't stripping the slash prefixing source mean that a disk with name `bar` would be matched against a disk named `foobar`? Seems a stretch, but...

danawillow: Yup. Fixed.

paddycarver:
Rebase and LGTM. 👍

PR #125, opened 2017-06-14
Google appengine

ctavan:
From: https://github.com/hashicorp/terraform/pull/14293

This adds a resource `google_appengine_app` to enable Google App Engine on a project. My major intent for this is to be able to enable Google Datastore which requires App Engine to be enabled.

Would be used like this:

```hcl
provider "google" {
  project = "my-testing-project-123"
  region  = "europe-west1"
}

resource "google_appengine_app" "default" {
  project = "my-testing-project-123"
  region  = "europe-west"
}
```

The current implementation works for me but so far it is still very bare-bones since I would love to get your feedback on this first before continuing (e.g. with writing proper tests and more docs).

The strange thing about Google App Engine is that it can only be activated once for a gcloud project but cannot be de-activated: During activation one has to choose a region (which as a side-effect also determines the data-locality of the Google Datastore instance of that gcloud project) and this cannot be changed afterwards.

How would I best formulate these conditions in terms of terraform? E.g. terraform requires me to implement the Delete method but I don't see any sensible "delete" operation for Google App Engine.

@danawillow @paddycarver @radeksimko thoughts?

### DIFF
--- a/google/config.go
+++ b/google/config.go
@@ -16,6 +16,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"golang.org/x/oauth2/jwt"
+	"google.golang.org/api/appengine/v1"
 	"google.golang.org/api/bigquery/v2"
 	"google.golang.org/api/cloudbilling/v1"
 	"google.golang.org/api/cloudresourcemanager/v1"
@@ -42,6 +43,7 @@ type Config struct {
 	Project     string
 	Region      string
 
+	clientAppengine              *appengine.APIService
 	clientBilling                *cloudbilling.Service
 	clientCompute                *compute.Service
 	clientComputeBeta            *computeBeta.Service
@@ -126,6 +128,13 @@ func (c *Config) loadAndValidate() error {
 		"(%s %s) Terraform/%s", runtime.GOOS, runtime.GOARCH, versionString)
 
 	var err error
+
+	log.Printf("[INFO] Instantiating Google App Engine Client...")
+	c.clientAppengine, err = appengine.New(client)
+	if err != nil {
+		return err
+	}
+	c.clientAppengine.UserAgent = userAgent
 
 	log.Printf("[INFO] Instantiating GCE client...")
 	c.clientCompute, err = compute.New(client)
--- a/google/provider.go
+++ b/google/provider.go
@@ -61,6 +61,7 @@ func Provider() terraform.ResourceProvider {
 		},
 
 		ResourcesMap: map[string]*schema.Resource{
+			"google_appengine_app":                         resourceAppengineApp(),
 			"google_bigquery_dataset":                      resourceBigQueryDataset(),
 			"google_bigquery_table":                        resourceBigQueryTable(),
 			"google_bigtable_instance":                     resourceBigtableInstance(),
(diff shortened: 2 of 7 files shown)


ctavan:
I'd still be happy to bring this PR into a mergeable state (i.e. adding tests etc.) so it would be great to get your feedback on the questions from the initial comment.

ctavan:
Sorry for bumping this. I keep on using an own patched build of this plugin because of this feature. If there's any chance I would love to see this merged at some point.

Is there any chance to receive some feedback any time soon such that I could bring this PR into a mergeable state?

paddycarver:
Hey @ctavan,

This is really delayed, and I apologise for that. Bumps are _totally_ reasonable.

The reason we took so long to respond to this is because the code looks _more or less fine_, but the reasoning behind writing it is unusual, and we were investigating that. My impression was that Datastore doesn't _need_ an App Engine app to work, and that's true; it is straightforward to enable Datastore in the UI. Unfortunately, this UI functionality does not seem to be reflected in the API, and so you're right that creating an App Engine app is the only way to programmatically enable the Datastore API on a project.

That being said, I'm still hesitant, because I'd like to have App Engine support designed from day 1 and ship with a good story behind it, to ensure that the workflow of using App Engine with Terraform gets the attention it deserves. To that end, the Googlers working on this provider have been talking to the developers of the Datastore API to see if we can get a programmatic way to do this, which we'd then support with Terraform. That's the ideal solution here. Unfortunately, I don't know the current status of those talks.

That said, in the interest of getting this closed (you've been _incredibly_ patient, and I'm super grateful for it) I'm going to move that discussion into Google's public issue tracker: https://issuetracker.google.com/issues/65422801.

Here's what I propose: if I can beg your patience for a bit longer, I want to see if we can get this supported at the API level. If we can't, or if we don't hear back in the next week or two, or it doesn't look like this will happen, I'll take on this issue and get it merged myself.

I know that's not an ideal answer, and I do apologise for the lack of transparency on this up til now. I was hoping to come back to you with a final answer, but that kept moving just beyond reach.

Thanks again for your patience, and your work in contributing to Terraform. It's deeply appreciated.

ctavan:
Thanks for the detailed response @paddycarver and no worries about the delay, I prefer a solid answer over a rushed fuzzy one!

I have posted some further observations in https://issuetracker.google.com/issues/65422801#comment4

I believe there simply is no such thing as "Create a Datastore Instance". I believe that Datastore is technically simply a feature that _comes with_ App Engine, it's just that the Cloud Console UI pretends that the two services are independent, while it performs the same operations (i.e. create an App Engine instance) under the hood (maybe it's just better for marketing not to bury Datastore somewhere in the App Engine feature list?!).

I'm still be happy to assist in bringing this PR forward however I'm equally happy if you take it over (since I'm currently really just interested in activating app engine and don't have a use case for any of the other app engine features I can't promise to find the time to implement other app engine features).

paddycarver:
Hey @ctavan, here's a long-overdue update on this: I just opened PR #1503 to add support. I took a bit of a different approach, because App Engine apps can't be deleted, and embedded it in the project resource, so they can share lifecycles. But I really appreciate your patience and code here, it was super helpful as I worked my way around App Engine. Sorry for the delay on this, but I'm glad we finally have a resolution.

ctavan:
@paddycarver no worries. After all my use case initially was only to be able to use Datastore which I have stopped using quite a while ago in favor of a self-hosted Redis.

Still glad to see true appengine support to emerge, will help me with other projects that are currently still being managed manually.

PR #126, opened 2017-06-14
Refactor resource_compute_instance_group_manager for multi version support

rileykarson:
This is the first step towards adding support for a beta field to a resource - converting it to being able to support multiple API fields. This means it needs to take in an enum specifying the API version, and switch which Compute API client it uses based on that.

`resource_compute_instance_group_manager` still only has the ability to support the GA api after this PR - it's refactoring the existing functionality only. Support for Beta operations and adding a Beta field will come later to keep it (moderately) reviewable.

### DIFF
--- a/google/api_levels.go
+++ b/google/api_levels.go
@@ -1,0 +1,8 @@
+package google
+
+type ApiLevel uint8
+
+const (
+	// PRODUCTION is the GA API.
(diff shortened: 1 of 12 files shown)


rileykarson:
I'm going to modify this and resubmit - I found a way of adding the ApiLevel to CRUD methods that I like more & that produces way better diffs.

PR #40, opened 2017-06-13
 Add compute_backend_service import support

selmanj:
Does what it says on the tin.

/cc @danawillow

PR #127, opened 2017-06-14
Change underscores to hyphens in google_compute_backend_service docs example

rileykarson:
#40 got merged before I noticed the change from hyphens to underscores in the example in the docs; GCP resources names only allow hyphens as special characters and not underscores.

Update the .tf example and `Import` example to use hyphens.


stack72:
nice catch @rileykarson :)

PR #129, opened 2017-06-15
Refactor resource_compute_instance_group_manager for multi version support

rileykarson:
This is the first step towards adding support for a beta field to a resource - converting it to being able to support multiple API fields. This means it needs to take in an enum specifying the API version, and switch which Compute API client it uses based on that.

`resource_compute_instance_group_manager` still only has the ability to support the GA api after this PR - it's refactoring the existing functionality only. Support for Beta operations and adding a Beta field will come later to keep the changes contained.

I moved from a function generator in #126 to adding a "curryApiLevel" method here. This will keep smaller diffs as the entire resource file isn't indented a block anymore.


google/shared/instance_group_manager.go
```diff
@@ -0,0 +1,156 @@
...
+	// include in API requests with the JSON null value. By default, fields
+	// with empty values are omitted from API requests. However, any field
+	// with an empty value appearing in NullFields will be sent to the
+	// server as null. It is an error if a field in this list has a
+	// non-empty value. This may be used to include null fields in Patch
+	// requests.
+	NullFields []string `json:"-"`
+}
+
+func (s *InstanceGroupManager) ToProduction() *compute.InstanceGroupManager {
```
danawillow: Let's swap out `Production` with `v1` everywhere, since there could be more than one production version.

google/resource_compute_instance_group_manager.go
```diff
@@ -99,8 +102,11 @@ func resourceComputeInstanceGroupManager() *schema.Resource {
 			"target_pools": &schema.Schema{
 				Type:     schema.TypeSet,
 				Optional: true,
-				Elem:     &schema.Schema{Type: schema.TypeString},
-				Set:      schema.HashString,
+				Elem: &schema.Schema{
+					Type:             schema.TypeString,
+					DiffSuppressFunc: compareSelfLinkRelativePaths,
```
rileykarson: Done.

google/self_link_relative_path_helpers.go
```diff
@@ -0,0 +1,42 @@
+package google
```
rileykarson: Done

google/self_link_relative_path_helpers.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// Hash the relative path of a self link.
+func selfLinkRelativePathHash(selfLink interface{}) int {
+	path, _ := getRelativePath(selfLink.(string))
+	return hashcode.String(path)
+}
+
+func getRelativePath(selfLink string) (string, error) {
+	stringParts := strings.SplitAfterN(selfLink, "/", 6)
```
rileykarson: Done - splitting on "projects/".

google/shared/instance_group_manager.go
rileykarson: Done

google/api_version_converters.go
```diff
@@ -0,0 +1,80 @@
...
+
+	err = json.Unmarshal(bytes, out)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func convertInstanceGroupManagerToV1(self interface{}) (*compute.InstanceGroupManager, error) {
```
danawillow: We're going to end up needing a whole bunch of these. Do you think it's worth it to have a separate function for each one instead of inlining it with the code? The majority of this is just error handling, which needs to be done by the caller anyway.

google/resource_compute_instance_group_manager.go
```diff
@@ -200,6 +217,7 @@ func flattenNamedPorts(namedPorts []*compute.NamedPort) []map[string]interface{}
 }
 
 func resourceComputeInstanceGroupManagerRead(d *schema.ResourceData, meta interface{}) error {
+	computeApiVersion := v1
```
danawillow: Might be worth it to add a comment here that this will eventually be figured out based on what's set in the request with a link to the issue about multiple api versions

google/resource_compute_instance_group_manager.go
```diff
@@ -212,36 +230,46 @@ func resourceComputeInstanceGroupManagerRead(d *schema.ResourceData, meta interf
...
+			d.SetId("")
+			return nil
+		}
 
-	if manager == nil {
-		log.Printf("[WARN] Removing Instance Group Manager %q because it's gone", d.Get("name").(string))
-		// The resource doesn't exist anymore
-		d.SetId("")
-		return nil
+		manager, err = convertInstanceGroupManagerToV1(v1Manager)
```
danawillow: I don't have a good suggestion for you, but it feels weird to me to be taking something called `v1Manager`, running it through a function that converts it to v1 (which it already is), and then storing that in a variable called `manager`

google/resource_compute_instance_group_manager.go
```diff
@@ -313,21 +353,40 @@ func resourceComputeInstanceGroupManagerUpdate(d *schema.ResourceData, meta inte
...
+			var managedInstances *compute.InstanceGroupManagersListManagedInstancesResponse
+			switch computeApiVersion {
+			case v1:
+				managedInstancesV1, err := config.clientCompute.InstanceGroupManagers.ListManagedInstances(
+					project, d.Get("zone").(string), d.Id()).Do()
+				if err != nil {
+					return fmt.Errorf("Error getting instance group managers instances: %s", err)
+				}
+
+				managedInstances, err = convertInstanceGroupManagersListManagedInstancesResponseToV1(managedInstancesV1)
```
danawillow: missing error handling

google/compute_shared_operation.go
```diff
@@ -0,0 +1,22 @@
+package google
```
rileykarson: The shared directory and the files inside it were removed when we moved to json serialising/deserialising to change struct versions.

google/resource_compute_instance_group_manager.go
rileykarson: Done.

google/resource_compute_instance_group_manager.go
rileykarson: This is a bit weird because there is still only a single version - `v1` is the canonical version right now, but won't be once there is another version supported. Removed until necessary.

google/resource_compute_instance_group_manager.go
rileykarson: Done

google/api_version_converters.go
rileykarson: We could probably eliminate this - I like hiding the empty structs you need to instantiate from the caller, but it does add a lot error handling LOC.

Another thing we could consider is upgrading the errors here to panics - they are unrecoverable for the caller, since we are really just using json to do a cast; fixing it would require a recompile. Then, the return would only be the new struct, and we would have a much cleaner interface for the caller.

danawillow: As discussed offline, let's at least see what it looks like to do it without the separate functions.

rileykarson: Done.

danawillow:
I'm happy with this! If you are too, feel free to merge.

danawillow:
Also, for posterity:
```
TF_ACC=1 go test ./google -v -run=TestAccInstanceGroupManager -timeout 120m
=== RUN   TestAccInstanceGroupManager_importBasic
--- PASS: TestAccInstanceGroupManager_importBasic (214.22s)
=== RUN   TestAccInstanceGroupManager_importUpdate
--- PASS: TestAccInstanceGroupManager_importUpdate (212.41s)
=== RUN   TestAccInstanceGroupManager_basic
--- PASS: TestAccInstanceGroupManager_basic (211.67s)
=== RUN   TestAccInstanceGroupManager_update
--- PASS: TestAccInstanceGroupManager_update (266.77s)
=== RUN   TestAccInstanceGroupManager_updateLifecycle
--- PASS: TestAccInstanceGroupManager_updateLifecycle (114.19s)
=== RUN   TestAccInstanceGroupManager_updateStrategy
--- PASS: TestAccInstanceGroupManager_updateStrategy (210.64s)
=== RUN   TestAccInstanceGroupManager_separateRegions
--- PASS: TestAccInstanceGroupManager_separateRegions (212.19s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	1442.259s
```

PR #135, opened 2017-06-16
Update 5 more importable resources' docs

rileykarson:
Update the documentation of 5 more of our importable resources to include `Import` blocks specifying how to import them.

We now have explicit documentation for:

- google_compute_forwarding_rule
- google_compute_global_address
- google_compute_http_health_check
- google_compute_instance_group_manager

As in #119, to review:

In each Import block;


stack72:
LGTM! thanks for this :)

PR #136, opened 2017-06-16
Make the org_id optional when creating a project.

ubschmidt2:


paddycarver:
Thanks for opening the PR! We'll get a review for this as soon as it has acceptance tests. :)

ubschmidt2:
Thanks ;-). Will add tests ...

danawillow:
@evandbrown

ubschmidt2:
Added an acceptance test. PTAL.
I didn't want to rename/touch the existing tests (all requiring the org), but it might make sense. PLMK.

danawillow:
```
make testacc TEST=./google TESTARGS='-run=TestAccGoogleProject_createWithoutOrg' GOOGLE_USE_DEFAULT_CREDENTIALS=true
TF_ACC=1 go test ./google -v -run=TestAccGoogleProject_createWithoutOrg -timeout 120m
=== RUN   TestAccGoogleProject_createWithoutOrg
--- PASS: TestAccGoogleProject_createWithoutOrg (18.58s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	18.756s
```

Since it needs the extra environment variable for using default credentials, this test will always fail in the nightly acceptance test suite, which run under a service account. @catsby or @stack72 is there anything that we'd be able to do to get this to pass or just ignore this test in TeamCity?

stack72:
Hi @danawillow 

DM me on slack about the new Env Var and we can get it added 

Thanks 

Paul

ubschmidt2:
Hmm. I'd expect the test to succeed when run under the same service account as the other tests. I haven't tests this, though, I don't own an org.

danawillow:
Wouldn't it need to be running on a machine that has application default credentials?

ubschmidt2:
No, the test doesn't depend on the default credentials at all.
When run with the org's service service account, the test should be able to create the project as usual. My expectation is that the new project will be automatically associated with the org. (not tested)

danawillow:
It can't be run with a service account:
`google_project.acceptance: Error creating project terraform-p1vao20pwe (Terraform Acceptance Tests): googleapi: Error 403: Service accounts cannot create projects without a parent., forbidden.`

ubschmidt2:
Thanks, Dana! Too bad, I didn't expect this outcome, sorry.

paddycarver:
Sorry, catching up after some time off. My understanding of this (at the moment) is that when authenticating with Application Default Credentials (presumably created from a full user account, not a service account) the organization can be inferred, and when authenticating through a service account, the organization can't be inferred. This PR would then only affect those authenticating Terraform using full user accounts. Is that correct?

ubschmidt2:
@paddycarver: yes, this is correct. (In the end, this change is only about making TF usable for human users that do not have an org.)

danisla:
@paddycarver, the [docs](https://www.terraform.io/docs/providers/google/#authentication-json-file) only show how to authenticate with a Compute Engine Service Account or using a JSON file for a service account.

How would I tell terraform to use the credentials from the full user account?

ubschmidt2:
Hmm. I think we should update the docs in this regard (will send a PR in the next days, unless someone else beats me to it): if no explicit credentials are specified, the http client code falls back to using the default credentials, i.e
 your own credentials if you've previously authenticated with gcloud, or the compute engine default credentials if running in GCE.

ubschmidt2:
Going to add a few lines to skip the new test if $GOOGLE_ORG is set, i.e. when running in your hosted acceptance test environment. Sounds good?

danawillow:
Yup, that sounds reasonable to me.

ubschmidt2:
Done.

paddycarver:
Sorry, I lost this thread. :(

My hesitation here is only that the test can't be run as part of our nightly integration tests, which means we rely on users running it locally to ensure that it continues working, which makes me nervous, at best.

This could be fixed by adding a user account to our CI, which is probably something we could do, but has its own problems and is a bit more of a hurdle.

To answer @danisla's question, I believe the way the user account credentials are getting picked up is by using the `gcloud` command to configure application default credentials, which Terraform will respect.

So, in summary: I think this is fine as a feature, it's just the testing story that's making me nervous. I'm checking in with the rest of the Terraform team now to see how complicated it is to get an actual user running requests from CI.

ubschmidt2:
Re https://github.com/terraform-providers/terraform-provider-google/pull/136#issuecomment-316214847 and the error `Service accounts cannot create projects without a parent.`:
It occurs to me that it should be possible to whitelist your service account for project creation without a parent. @paddycarver and @danawillow: PLMK if this would work for you.

ubschmidt2:
Forgot to say that the whitelisting would happen on the Google side, of course.

paddycarver:
I think whitelisting a service account for project creation would be (if not an ideal solution) the best solution available to us at the moment.

Failing that, I'm looking into registering a GMail account and trying to set up some way for the CI server to populate its application default credentials, but every way I look at it, that looks messy and prone to breakage (because Terraform doesn't speak OAuth2, and can't refresh the token).

Whitelisting a service account is only an issue in that:

1. It's harder for people outside HashiCorp to run the tests (although, they may be able to with a GMail account and Application Default Credentials).
2. If there's ever any change to the auth strategy for the tests (e.g., a separate service account per test, or splitting tests between multiple service accounts) or in which project runs the tests, we may need to switch the service account that's whitelisted, which seems not-awesome.

At the moment, though, it definitely sounds like the best path forward.

danawillow:
Agreed, let's go with that.

(To Paddy's point 1, I'd say that any test that requires org-level ownership permissions is already hard for people outside HashiCorp to run)

paddycarver:
(That is a fair point and a much-needed reminder that not everyone has 4 GSuite orgs they admin and can easily test things on just laying around. I feel much better about point 1 now. :) )

ubschmidt2:
Great, let me file an internal request for whitelisting then. What's the "service account ID" (...@...gserviceaccount.com, cf. https://pantheon.corp.google.com/iam-admin/serviceaccounts/project) that TeamCity is using for running the acceptance tests?  Please feel free to mail to stschmidt@google.com or to Dana.

ubschmidt2:
Internal ref b/64597820.

rosbo:
The `org_id` is now Optional since #438. Closing.

Issue #144, opened 2017-06-20
google_compute_backend_service only supports ONE health check


jwgoh:
### Terraform Version
v0.9.6

### Affected Resource(s)
- google_compute_backend_service

Argument reference `health_checks` is a bit misleading because only on health check is supported

Also the document mentions the below:

> health_checks - (Required) Specifies a list of HTTP health check objects for checking the health of the backend service.

EDIT: I can open a PR to update the doc

stack72:
Closed via #145

PR #145, opened 2017-06-20
Restrict the number of health_checks in Backend Service resources to exactly 1.

rileykarson:
Fix for #144.

Affects `google_compute_backend_service` and `google_compute_region_backend_service`.
- Update documentation to specify that exactly 1 health check is required.
- Perform a plan-time validation that there is exactly 1 element in the health check list.

### DIFF
--- a/google/resource_compute_backend_service.go
+++ b/google/resource_compute_backend_service.go
@@ -40,8 +40,10 @@ func resourceComputeBackendService() *schema.Resource {
 			"health_checks": &schema.Schema{
 				Type:     schema.TypeSet,
 				Elem:     &schema.Schema{Type: schema.TypeString},
-				Required: true,
 				Set:      schema.HashString,
+				Required: true,
+				MinItems: 1,
+				MaxItems: 1,
 			},
 
 			"backend": &schema.Schema{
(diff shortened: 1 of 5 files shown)


google/resource_compute_backend_service.go
```diff
@@ -40,8 +40,10 @@ func resourceComputeBackendService() *schema.Resource {
 			"health_checks": &schema.Schema{
 				Type:     schema.TypeSet,
 				Elem:     &schema.Schema{Type: schema.TypeString},
-				Required: true,
 				Set:      schema.HashString,
+				Required: true,
+				MinItems: 1,
```
stack72: If you have a `required: true` then you won't need `MinItems: 1` :)

google/resource_compute_region_backend_service.go
```diff
@@ -37,8 +37,10 @@ func resourceComputeRegionBackendService() *schema.Resource {
 			"health_checks": &schema.Schema{
 				Type:     schema.TypeSet,
 				Elem:     &schema.Schema{Type: schema.TypeString},
-				Required: true,
 				Set:      schema.HashString,
+				Required: true,
+				MinItems: 1,
```
stack72: Not needed :)

Issue #17, opened 2017-06-13
Support labels for google cloud resources


hashibot:
_This issue was originally opened by @JorritSalverda as hashicorp/terraform#3546. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

According to http://googlecloudplatform.blogspot.nl/2015/10/using-labels-to-organize-Google-Cloud-Platform-resources.html you can now use key/value pairs to label your resources.

It seem to be tags that have key=value format, but if I try that I get the following error, so apparently it's handled differently through the gcloud api.

```
Reason: invalid, Message: Invalid value for field 'tags.items[0]': 'environment=prod'.  Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)'
```

Please add support for this as it will allow us to use this to label our resources for better cost reporting.

hashibot:
_This comment was originally opened by @JorritSalverda as https://github.com/hashicorp/terraform/issues/3546#issuecomment-149206013. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original comment is below._

<hr>

Judging from the info in the article it's a replacement for tags, but tags that previously existed are shown as value-less keys, which first led me to believe it used the tags system, but it's separately set with

```
curl --request POST --header "Authorization: Bearer your-oauth-token" --header 'Content-Type: application/json' --data '{"labels": {"env": "canary"}, "labelFingerprint": "abcdefghij="}' https://www.googleapis.com/compute/alpha/projects/your-project-id/zones/your-zone/instances/your-instance/setLabels
```

hashibot:
_This comment was originally opened by @lwander as https://github.com/hashicorp/terraform/issues/3546#issuecomment-149247134. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original comment is below._

<hr>

Unfortunately this seems to only be available in the [beta API](https://cloud.google.com/compute/docs/reference/beta/instances/setLabels). I'll have to see what might break if we move from v1 to beta. 

Have you considered using the `metadata` instance attribute? It supports key value pairs and is documented [here](https://www.terraform.io/docs/providers/google/r/compute_instance.html#metadata)

hashibot:
[...]

PR #150, opened 2017-06-21
Add support for using labels on compute_instance

selmanj:
Note that the configuration for `testAccComputeInstance_update` was modified as it was unsetting a value that forced a new instance to be created instead of updated (specifically the `metadata_startup_script` key).

Part of #17.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -343,6 +343,18 @@ func resourceComputeInstance() *schema.Resource {
 				Computed: true,
 			},
 
+			"labels": &schema.Schema{
+				Type:     schema.TypeMap,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
+
+			"label_fingerprint": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
 			"create_timeout": &schema.Schema{
 				Type:     schema.TypeInt,
 				Optional: true,
@@ -676,6 +688,7 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 		Name:              d.Get("name").(string),
 		NetworkInterfaces: networkInterfaces,
 		Tags:              resourceInstanceTags(d),
+		Labels:            resourceInstanceLabels(d),
 		ServiceAccounts:   serviceAccounts,
 		Scheduling:        scheduling,
 	}
@@ -845,6 +858,14 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("tags_fingerprint", instance.Tags.Fingerprint)
 	}
 
+	if len(instance.Labels) > 0 {
+		d.Set("labels", instance.Labels)
+	}
+
+	if instance.LabelFingerprint != "" {
+		d.Set("label_fingerprint", instance.LabelFingerprint)
+	}
+
 	disksCount := d.Get("disk.#").(int)
 	attachedDisksCount := d.Get("attached_disk.#").(int)
 	disks := make([]map[string]interface{}, 0, disksCount)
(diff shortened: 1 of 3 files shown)


tpoindessous:
Hi

first, I want to thank you for your work. I'm trying to use labels in gcp_compute_snapshot data resource and I will look at your code.

At first sight, it seems that your function `func resourceInstanceLabels` could be more generic and maybe be useful for other resources.

Is it possible for you, please, to look at this and try to add this function in a more generic file ?

Maybe your other function in _test file could be also used by other people ?

What do you think about this ?

Thank you.

Thomas

google/resource_compute_instance.go
```diff
@@ -1127,6 +1166,18 @@ func resourceInstanceMetadata(d *schema.ResourceData) (*compute.Metadata, error)
 	return m, nil
 }
 
+func resourceInstanceLabels(d *schema.ResourceData) map[string]string {
+	labels := map[string]string{}
+	if v, ok := d.GetOk("labels"); ok {
+		labelMap := v.(map[string]interface{})
+		for k, v := range labelMap {
+			vstr := v.(string)
```
danawillow: nit: I think it's just as readable to do `labels[k] = v.(string)`

google/resource_compute_instance_test.go
```diff
@@ -795,6 +797,24 @@ func testAccCheckComputeInstanceTag(instance *compute.Instance, n string) resour
 	}
 }
 
+func testAccCheckComputeInstanceLabel(instance *compute.Instance, key string, value string) resource.TestCheckFunc {
```
danawillow: Hmm, this checks that a single label exists but doesn't check that there aren't labels we don't expect to be there. Do you think this is sufficient or should it take the key/value map and compare that?

website/docs/r/compute_instance.html.markdown
```diff
@@ -46,6 +46,7 @@ resource "google_compute_instance" "default" {
     foo = "bar"
   }
 
+
```
danawillow: nit: this extra newline got added

danawillow:
@tpoindessous I mostly agree with you, although I'm not sure where those functions should live. Honestly I think they should be in core somewhere, since unpacking a map[string]interface{} into a map[string]string (or an []interface{} into a []string, which is another one I've seen around a bunch) is something that would be useful across providers.

As for the test one, that particular function itself can't be used by other tests because it's relying specifically on an `instance` being passed in, but the idea behind it could probably be generalized, sure. I'm not going to block merging on either of these changes since they could pretty easily be done in a follow-up PR by you or @selmanj, but Joe if you want to do them in this one I'm not going to stop you either.

google/resource_compute_instance.go
selmanj: whoops, yeah, not sure why there's an extra variable there. Will fix

google/resource_compute_instance_test.go
selmanj: I could add that but I'm not sure it's something that could happen... Although it might make the test syntax cleaner later down the road.

At some point I think our tests would benefit from building a terraform config up programmatically via a builder or something similar; this would end up making the tests a lot more explicit (currently a lot of information is hidden in the 'config' function). If we had that, then testing the entire map would be a lot easier...

Something like: 
```
	resource.Test(t, resource.TestCase{
		PreCheck:     func() { testAccPreCheck(t) },
		Providers:    testAccProviders,
		CheckDestroy: testAccCheckComputeInstanceDestroy,
		Steps: []resource.TestStep{
			resource.TestStep{
				Config: NewInstanceBuilder(instanceName).WithLabels(myLabels).Build(),
					testAccCheckComputeInstanceExists(
						"google_compute_instance.foobar", &instance),
					testAccCheckComputeInstanceHasLabels(&instance, myLabels),
				),
			},
		},
```

tpoindessous:
Hi

yes, I agree with you, features first :)

We could change that later, when we will have several usages of Labels.

Thanks for your work, it will be very useful !

PR #152, opened 2017-06-21
allow updating additional_zones, turn it into a set

### LINKED ISSUE #104: Upgrade a GKE cluster from single zone to multi zone without recreating it
_This issue was originally opened by @devth as hashicorp/terraform#15004. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

This is somewhat related to terraform-providers/terraform-provider-google#103

The `gcloud` CLI [provides a way](https://cloud.google.com/container-engine/docs/clusters/operations#update_a_running_cluster_to_multi_zone) to convert a single-zone cluster to multizone.

In Terraform perhaps adding zones to `additional_zones` could use this API to non-destructively add a zone to the cluster. /cc @danawillow

danawillow:
Fixes #104.

`additional_zones` had to be turned into a set because GCP returns them in alphabetical order regardless of the order they're set in the request.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -667,7 +667,11 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 	}
 
 	if d.HasChange("additional_zones") {
-		azs := convertStringArr(d.Get("additional_zones").(*schema.Set).List())
+		azSet := d.Get("additional_zones").(*schema.Set)
+		if azSet.Contains(zoneName) {
+			return fmt.Errorf("additional_zones should not contain the original 'zone'.")
+		}
+		azs := convertStringArr(azSet.List())
 		locations := append(azs, zoneName)
 		req := &container.UpdateClusterRequest{
 			Update: &container.ClusterUpdate{
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -106,10 +106,9 @@ func resourceContainerCluster() *schema.Resource {
 			},
 
 			"additional_zones": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Optional: true,
 				Computed: true,
-				ForceNew: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 
@@ -386,7 +385,7 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 	}
 
 	if v, ok := d.GetOk("additional_zones"); ok {
-		locationsList := v.([]interface{})
+		locationsList := v.(*schema.Set).List()
 		locations := []string{}
 		for _, v := range locationsList {
 			location := v.(string)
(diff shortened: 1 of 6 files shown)


google/resource_container_cluster.go
```diff
@@ -625,27 +627,60 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
...
+
+		d.SetPartial("node_version")
 	}
 
-	// Wait until it's updated
-	waitErr := containerOperationWait(config, op, project, zoneName, "updating GKE cluster", 10, 2)
-	if waitErr != nil {
-		return waitErr
+	if d.HasChange("additional_zones") {
+		azs := convertStringArr(d.Get("additional_zones").(*schema.Set).List())
```
selmanj: `convertStringArr` is defined in `resource_compute_target_pool.go` - is there a better place to put it?

google/resource_container_cluster.go
```diff
@@ -625,27 +627,60 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
...
+		locations := append(azs, zoneName)
```
selmanj: I'm guessing there's no problem if a location is repeated multiple times (i.e. the API doesn't throw an error)? Otherwise it might make sense to do this call before converting it to a list

google/resource_container_cluster_migrate.go
```diff
@@ -0,0 +1,74 @@
...
+	v int, is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
+	if is.Empty() {
+		log.Println("[DEBUG] Empty InstanceState; nothing to migrate.")
+		return is, nil
+	}
+
+	switch v {
+	case 0:
+		log.Println("[INFO] Found Container Cluster State v0; migrating to v1")
+		is, err := migrateClusterStateV0toV1(is)
```
selmanj: I think this can be shorted to just
```
return migrateCLusterStateV0toV1(is)
```

google/resource_container_cluster_migrate.go
```diff
@@ -0,0 +1,74 @@
...
+		if err != nil {
+			return is, err
+		}
+		return is, nil
+	default:
+		return is, fmt.Errorf("Unexpected schema version: %d", v)
+	}
+}
+
+func migrateClusterStateV0toV1(is *terraform.InstanceState) (*terraform.InstanceState, error) {
```
selmanj: nit: capital 't' in the `to` part of `migrateClusterStateV0toV1`?

google/resource_container_cluster_test.go
```diff
@@ -219,6 +231,10 @@ func testAccCheckContainerClusterDestroy(s *terraform.State) error {
 	return nil
 }
 
+var setFields map[string]struct{} = map[string]struct{}{
```
selmanj: Could this just be a list/set of strings rather than having an unused struct{} value?

google/resource_container_cluster_migrate_test.go
```diff
@@ -0,0 +1,75 @@
...
+	cases := map[string]struct {
+		StateVersion int
+		Attributes   map[string]string
+		Expected     map[string]string
+		Meta         interface{}
+	}{
+		"change additional_zones from list to set": {
+			StateVersion: 0,
+			Attributes: map[string]string{
+				"additional_zones.#": "1",
```
selmanj: Should this be 2?

google/resource_container_cluster.go
danawillow: I put it in provider.go, I think that's the best option we have righ tnow.

google/resource_container_cluster.go
danawillow: Good catch! The API does throw an error. I just added a check like we have in `Create`

google/resource_container_cluster_migrate.go
danawillow: Done

google/resource_container_cluster_migrate.go
danawillow: Leaving it lowercase to match the other files

google/resource_container_cluster_migrate_test.go
danawillow: Yup! Done.

google/resource_container_cluster_test.go
selmanj: :(  OK

PR #157, opened 2017-06-26
Add support for node pool autoscaling

### LINKED ISSUE #44: [GKE] Support autoscaling
_This issue was originally opened by @thecodeassassin as hashicorp/terraform#8605. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

hey all,

Google cloud supports autoscaling for nodes which is very useful in high traffic kubernetes deployments. Could you add this feature to terraform? Currently i use local exec like this:

``` bash
  # enable autoscaling
  provisioner "local-exec" {
      command = "gcloud container clusters update ${var.google_container_cluster.name} --enable-autoscaling --min-nodes=${var.min_nodes} --max-nodes=${var.max_nodes} --project='${var.project_name}'"
  }
```

Documentation is here
https://cloud.google.com/container-engine/docs/clusters/operations

danawillow:
Fixes #44.

This is essentially the same code as https://github.com/hashicorp/terraform/pull/12779, except I added the `enabled` attribute and support for update.

```
TF_ACC=1 go test ./google -v -run=TestAccContainerNodePool_autoscaling -timeout 120m
=== RUN   TestAccContainerNodePool_autoscaling
--- PASS: TestAccContainerNodePool_autoscaling (1056.29s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	1056.417s
```

### DIFF
--- a/google/resource_container_node_pool.go
+++ b/google/resource_container_node_pool.go
@@ -63,12 +63,6 @@ func resourceContainerNodePool() *schema.Resource {
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"enabled": &schema.Schema{
-							Type:     schema.TypeBool,
-							Optional: true,
-							Default:  true,
-						},
-
 						"min_node_count": &schema.Schema{
 							Type:     schema.TypeInt,
 							Required: true,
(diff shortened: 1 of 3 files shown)


google/resource_container_node_pool.go
```diff
@@ -129,10 +180,70 @@ func resourceContainerNodePoolRead(d *schema.ResourceData, meta interface{}) err
 
 	d.Set("name", nodePool.Name)
 	d.Set("initial_node_count", nodePool.InitialNodeCount)
+	if nodePool.Autoscaling != nil && nodePool.Autoscaling.Enabled {
```
paddycarver: How do we detect if autoscaling has been turned on out of band and needs to be switched back on?

danawillow: Ah, that's a good point.

I went ahead and checked what the cloud console does- when you turn autoscaling off and then back on, you lose the min/max node count you had originally set. Based on that, I removed the enabled attribute so we can easily detect out-of-band changes (this way we don't need to differentiate between set to false and not set at all, although in retrospect maybe a `DiffSuppressFn` on the `autoscaling` attribute would work? not sure). Let me know what you think.

PR #163, opened 2017-06-28
Get a single bigtable instance when performing reads

rileykarson:
With https://github.com/GoogleCloudPlatform/google-cloud-go/issues/682 resolved, we don't need to loop over the list of instances in the Admin api client anymore.

### DIFF
--- a/google/resource_bigtable_instance.go
+++ b/google/resource_bigtable_instance.go
@@ -8,6 +8,7 @@ import (
 
 	"cloud.google.com/go/bigtable"
 	"golang.org/x/net/context"
+	"log"
 )
 
 func resourceBigtableInstance() *schema.Resource {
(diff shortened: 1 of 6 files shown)


google/resource_bigtable_instance.go
```diff
@@ -132,27 +132,13 @@ func resourceBigtableInstanceRead(d *schema.ResourceData, meta interface{}) erro
 
 	defer c.Close()
 
-	instances, err := c.Instances(ctx)
+	instance, err := c.InstanceInfo(ctx, d.Id())
 	if err != nil {
-		return fmt.Errorf("Error retrieving instances. %s", err)
+		return fmt.Errorf("Error retrieving instance. Could not find %s.", d.Id())
```
danawillow: You'll want to also remove the instance from state, like in https://github.com/terraform-providers/terraform-provider-google/blob/master/google/provider.go#L270 (realized I forgot about that in the first review, whoops!)

Also, might it be worth it to still pass along the error message from the bigtable client?

rileykarson: Done.

PR #164, opened 2017-06-28
Add name validation function, standardise validation functions used

rileykarson:
We have several validation functions that are just reimplementations of ones that live in core's `helper/validation` package; let's use those instead.

We define a separate validation function each time we want to validate the `name` field of a resource. Let's use a single one instead.

These two resources had validation functions for `name` that validated only length and not format, but both follow the same rules as most names, a 1-63 character [RFC1035](https://www.ietf.org/rfc/rfc1035.txt) string. Errors will now surface at plan time instead of apply time for them.
- resource_compute_instance_template.go
- resource_compute_ssl_certificate.go

### DIFF
--- a/google/data_source_storage_object_signed_url.go
+++ b/google/data_source_storage_object_signed_url.go
@@ -23,6 +23,7 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/pathorcontents"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 	"golang.org/x/oauth2/google"
 	"golang.org/x/oauth2/jwt"
 )
@@ -68,7 +69,7 @@ func dataSourceGoogleSignedUrl() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Default:      "GET",
-				ValidateFunc: validateHttpMethod,
+				ValidateFunc: validation.StringInSlice([]string{"GET", "HEAD", "PUT", "DELETE"}, true),
 			},
 			"path": &schema.Schema{
 				Type:     schema.TypeString,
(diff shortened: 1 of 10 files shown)


google/validation.go
```diff
@@ -0,0 +1,16 @@
+package google
+
+import (
+	"fmt"
+	"regexp"
+)
+
+func validateGCPName(v interface{}, k string) (ws []string, errors []error) {
```
selmanj: Good candidate for unit testing here

Also what do you think about splitting it out into two functions - `validateGCPName` and `validateRegexp` (where `validateGCPName` is implemented in terms of `validateRegexp`)? I can see us needing regexs for other things later (for example, metadata keys)

google/data_source_storage_object_signed_url.go
```diff
@@ -68,7 +69,7 @@ func dataSourceGoogleSignedUrl() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Default:      "GET",
-				ValidateFunc: validateHttpMethod,
+				ValidateFunc: validation.StringInSlice([]string{"GET", "HEAD", "PUT", "DELETE"}, true),
```
selmanj: Could add our own validation function that is 'isHTTPVerb' (although given not all the http verbs are here, maybe that would be more confusing?) What do you think?

rileykarson: Given that not all of the http verbs are here, I don't think we should add a custom validator yet;  if there are several other validators with identical parameters in the future, that is something we should consider, though.

selmanj: SGTM

google/validation.go
rileykarson: Done.

google/validation.go
```diff
@@ -0,0 +1,24 @@
+package google
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+	"regexp"
+)
+
+func validateGCPName(v interface{}, k string) (ws []string, errors []error) {
```
selmanj: Shouldn't this also return schema.SchemaValidationFunc?

selmanj: Or, better suggestion - make the other function return (ws []string, errors []error) (not a func)

rileykarson: `validateGCPName` is implicitly a `schema.SchemaValidationFunc` because of its signature; we want `validateRegexp` to validate a against a variable regex and not a fixed one like `validateGCPName`, so we pass that regex into a function that makes a `schema.SchemaValidationFunc` for us.

There are similar functions to `validateRegexp` living in the core Terraform repo's [helper/validation package](https://github.com/hashicorp/terraform/blob/master/helper/validation/validation.go).

selmanj: Spoke in person; I understand the differing signatures here. I'm fine with the code as is.

PR #172, opened 2017-07-05
Allow for running acceptance tests using the default credentials.

ubschmidt2:
This change adds another option for supplying authentication credentials
to acceptance tests: If GOOGLE_USE_DEFAULT_CREDENTIALS is set, the default
credentials are used. When run from a compute engine instance, the compute
engine default credentials are used. When run from the user's workstation,
the user's credentials are used, if the user has authenticated with the
GCloud CLI beforehand.

### DIFF
--- a/google/provider_test.go
+++ b/google/provider_test.go
@@ -53,6 +53,7 @@ func testAccPreCheck(t *testing.T) {
 		"GOOGLE_CREDENTIALS",
 		"GOOGLE_CLOUD_KEYFILE_JSON",
 		"GCLOUD_KEYFILE_JSON",
+		"GOOGLE_USE_DEFAULT_CREDENTIALS",
 	}
 	if v := multiEnvSearch(creds); v == "" {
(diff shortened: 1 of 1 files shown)


google/provider_test.go
```diff
@@ -53,6 +53,7 @@ func testAccPreCheck(t *testing.T) {
 		"GOOGLE_CREDENTIALS",
 		"GOOGLE_CLOUD_KEYFILE_JSON",
 		"GCLOUD_KEYFILE_JSON",
+		"GOOGLE_USE_DEFAULT_CREDENTIALS",
```
catsby: Does this also need to be added to `provider.go` so that the credentials will be picked up? Unless I'm mistaken this section here just pre-validates that we have some kind of credentials to move forward with. Are they picked up somewhere else that I'm not seeing?

ubschmidt2: No further changes are required, the code is already there.
providerConfigure calls loadAndValidate, which uses the DefaultClient if no credentials are supplied.

PR #173, opened 2017-07-05
Add Bigtable sidebar entries

rileykarson:
Add `google_bigtable_instance` and `google_bigtable_table` to the docs sidebar.

### DIFF
--- a/website/google.erb
+++ b/website/google.erb
@@ -21,6 +21,17 @@
     </ul>
     </li>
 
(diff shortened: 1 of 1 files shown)


danawillow:
While you're here- I think it probably makes sense to actually have all the data sources at the top and the resources below. If you agree, mind making that change now? Otherwise 👍  (and feel free to make the change + merge without re-review)

PR #176, opened 2017-07-05
Add support for google_compute_project_metadata_item

selmanj:
This allows terraform users to manage single key/value items within the
project metadata map, rather than the entire map itself.

### DIFF
--- a/google/provider.go
+++ b/google/provider.go
@@ -86,6 +86,7 @@ func Provider() terraform.ResourceProvider {
 			"google_compute_instance_template":      resourceComputeInstanceTemplate(),
 			"google_compute_network":                resourceComputeNetwork(),
 			"google_compute_project_metadata":       resourceComputeProjectMetadata(),
+			"google_compute_project_metadata_item":  resourceComputeProjectMetadataItem(),
 			"google_compute_region_backend_service": resourceComputeRegionBackendService(),
 			"google_compute_route":                  resourceComputeRoute(),
 			"google_compute_router":                 resourceComputeRouter(),
--- a/google/resource_compute_project_metadata_item.go
+++ b/google/resource_compute_project_metadata_item.go
@@ -1,0 +1,178 @@
+package google
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/compute/v1"
+)
+
+func resourceComputeProjectMetadataItem() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceComputeProjectMetadataItemCreate,
+		Read:   resourceComputeProjectMetadataItemRead,
+		Update: resourceComputeProjectMetadataItemUpdate,
+		Delete: resourceComputeProjectMetadataItemDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"key": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"value": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"project": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceComputeProjectMetadataItemCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	projectID, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	key := d.Get("key").(string)
+	val := d.Get("value").(string)
+
+	err = updateComputeCommonInstanceMetadata(config, projectID, key, &val)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(key)
+
+	return nil
+}
+
+func resourceComputeProjectMetadataItemRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	projectID, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] Loading project metadata: %s", projectID)
+	project, err := config.clientCompute.Projects.Get(projectID).Do()
+	if err != nil {
+		return fmt.Errorf("Error loading project '%s': %s", projectID, err)
+	}
+
+	md := flattenComputeMetadata(project.CommonInstanceMetadata.Items)
+	val, ok := md[d.Id()]
+	if !ok {
+		// Resource no longer exists
+		d.SetId("")
+		return nil
+	}
+
+	d.Set("key", d.Id())
+	d.Set("value", val)
+
+	return nil
+}
+
+func resourceComputeProjectMetadataItemUpdate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	projectID, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	if d.HasChange("value") {
+		key := d.Get("key").(string)
+		_, n := d.GetChange("value")
+		new := n.(string)
+
+		err = updateComputeCommonInstanceMetadata(config, projectID, key, &new)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func resourceComputeProjectMetadataItemDelete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	projectID, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	key := d.Get("key").(string)
+
+	err = updateComputeCommonInstanceMetadata(config, projectID, key, nil)
+	if err != nil {
+		return err
+	}
+
+	d.SetId("")
+	return nil
+}
+
+func updateComputeCommonInstanceMetadata(config *Config, projectID string, key string, afterVal *string) error {
+	updateMD := func() error {
+		log.Printf("[DEBUG] Loading project metadata: %s", projectID)
+		project, err := config.clientCompute.Projects.Get(projectID).Do()
+		if err != nil {
+			return fmt.Errorf("Error loading project '%s': %s", projectID, err)
+		}
+
+		md := flattenComputeMetadata(project.CommonInstanceMetadata.Items)
+
+		val, ok := md[key]
+
+		if !ok {
+			if afterVal == nil {
+				// Asked to set no value and we didn't find one - we're done
+				return nil
+			}
+		} else {
+			if afterVal != nil && *afterVal == val {
+				// Asked to set a value and it's already set - we're done.
+				return nil
+			}
+		}
+
+		if afterVal == nil {
+			delete(md, key)
+		} else {
+			md[key] = *afterVal
+		}
+
+		// Attempt to write the new value now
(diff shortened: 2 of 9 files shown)


selmanj:
Oops, forgot to `gfmt`

website/docs/r/compute_project_metadata.html.markdown
```diff
@@ -8,7 +8,10 @@ description: |-
 
 # google\_compute\_project\_metadata
 
-Manages metadata common to all instances for a project in GCE.
+Manages metadata common to all instances for a project in GCE. If you
```
rileykarson: What if we state that this resource authoritatively manages all of the project's metadata before we explain `google_compute_project_metadata_item` as an alternative? That way it will be more clear what the potential gotchas with using this resource are, and how the alternative solves them.

website/docs/r/compute_project_metadata.html.markdown
```diff
@@ -8,7 +8,10 @@ description: |-
...
+want to manage only single key/value pairs within the project metadata
+rather than the entire map, then use
```
rileykarson: Is it being a map an implementation detail? We describe it as a series below in the `metadata` field docs, for example.

google/provider.go
```diff
@@ -286,3 +287,28 @@ func linkDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 	}
 	return false
 }
+
+func computeMetadataToMap(metadata []*compute.MetadataItems) map[string]string {
```
rileykarson: We can name these functions flattenComputeMetadata (computeMetadataToMap) and expandComputeMetadata (mapToComputeMetadata) to match similar functions like in [`google_compute_backend_service`](https://github.com/terraform-providers/terraform-provider-google/blob/c5877743843791c4f8cdd8834e8c13594e45c349/google/resource_compute_backend_service.go#L331).

google/resource_compute_project_metadata_item.go
```diff
@@ -0,0 +1,191 @@
...
+	return &schema.Resource{
+		Create: resourceComputeProjectMetadataItemCreate,
+		Read:   resourceComputeProjectMetadataItemRead,
+		Update: resourceComputeProjectMetadataItemUpdate,
+		Delete: resourceComputeProjectMetadataItemDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		SchemaVersion: 0,
```
rileykarson: We don't need to set a SchemaVersion until we perform a migration.

website/docs/r/compute_project_metadata.html.markdown
selmanj: I almost documented exactly this, but unfortunately there's a bug right now in how `google_compute_project_metadata` works; right now it works similar to this PR, except upon delete (which deletes the entire map). We have a work item to go fix this so that it manages the entire metadata, but until then I'm hesitant to document that behavior (hence the current carefully worded description).

website/docs/r/compute_project_metadata.html.markdown
selmanj: Hmm, good point. I see in the documentation about it being a set of metadata field docs, but nowhere about it being a map. Will update.

google/provider.go
selmanj: I can do that for consistency's sake, but I am not sure the naming is improved.

google/resource_compute_project_metadata_item.go
selmanj: Will remove

website/docs/r/compute_project_metadata.html.markdown
rileykarson: SGTM then :+1:

google/provider.go
```diff
@@ -297,3 +298,39 @@ func convertStringArr(ifaceArr []interface{}) []string {
...
+// to value.
+func flattenComputeMetadata(metadata []*compute.MetadataItems) map[string]string {
+	m := map[string]string{}
+
+	for _, val := range metadata {
+		// check for duplicates
+		if val.Value == nil {
+			continue
+		}
+		if v2, ok := m[val.Key]; ok {
```
danawillow: nit: should `v2` just be `v`?

google/resource_compute_project_metadata_item.go
```diff
@@ -0,0 +1,189 @@
...
+	err = updateComputeCommonInstanceMetadata(config, projectID, key, &val, nil)
+	if err != nil {
+		return err
+	}
+
+	d.SetId("")
+	return nil
+}
+
+func updateComputeCommonInstanceMetadata(config *Config, projectID string, key string, beforeVal *string, afterVal *string) error {
```
danawillow: Oh is it because of differentiating between unset and set to the empty string? If so can you add a comment for clarity?

google/provider.go
```diff
@@ -297,3 +298,39 @@ func convertStringArr(ifaceArr []interface{}) []string {
...
+}
+
+// expandComputeMetadata transforms a map representing computing metadata into a list of compute.MetadataItems suitable
+// for the GCP client.
+func expandComputeMetadata(m map[string]string) []*compute.MetadataItems {
+	metadata := make([]*compute.MetadataItems, len(m))
+
+	idx := 0
+	for key, value := range m {
+		metadata[idx] = &compute.MetadataItems{Key: key, Value: &value}
```
danawillow: `idx++`? (you can also do `metadata = append(metadata, &compute.MetadataItems{Key: key, Value: &value})`. I think since the backing array won't need to be resized, it should have the same performance but I'm not 100% sure.)

google/resource_compute_project_metadata_item.go
```diff
@@ -0,0 +1,189 @@
...
+		md := flattenComputeMetadata(project.CommonInstanceMetadata.Items)
+
+		val, ok := md[key]
+
+		if !ok {
+			if afterVal == nil {
+				// Asked to set no value and we didn't find one - we're done
+				return nil
+			}
+			if beforeVal != nil {
```
selmanj: When I wrote this, it wasn't clear to me when a refresh occurred. Does it _always_ occur before a write?

If so, then the race condition you listed above is the only one I can think of. If that's not worth checking then I'm not sure it's worth enforcing a `beforeVal` at all, right?

google/resource_compute_project_metadata_item.go
selmanj: Yes; will do

(ps do we have a maybe type or something better than `*string`)?

google/resource_compute_project_metadata_item.go
```diff
@@ -0,0 +1,189 @@
...
+
+func updateComputeCommonInstanceMetadata(config *Config, projectID string, key string, beforeVal *string, afterVal *string) error {
+	updateMD := func() error {
+		log.Printf("[DEBUG] Loading project metadata: %s", projectID)
+		project, err := config.clientCompute.Projects.Get(projectID).Do()
+		if err != nil {
+			return fmt.Errorf("Error loading project '%s': %s", projectID, err)
+		}
+
+		// Should the ID contain the project name?
```
selmanj: This was supposed to have a TODO/NOTE in front of it but i foolishly forgot here - sorry, removing.

google/resource_compute_project_metadata_item.go
```diff
@@ -0,0 +1,189 @@
...
+		Create: resourceComputeProjectMetadataItemCreate,
+		Read:   resourceComputeProjectMetadataItemRead,
+		Update: resourceComputeProjectMetadataItemUpdate,
+		Delete: resourceComputeProjectMetadataItemDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"project": {
```
selmanj: Fixed

google/resource_compute_project_metadata_item_test.go
```diff
@@ -0,0 +1,136 @@
...
+				Config: testAccProjectMetadataItem_basic("myKey", "myUpdatedValue"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckProjectMetadataItem_hasMetadata("myKey", "myUpdatedValue"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccComputeProjectMetadataItem_basicImport(t *testing.T) {
```
selmanj: OK, will do

google/provider.go
selmanj: Probably, although `val` should have a better name too. Will fix names.

google/provider.go
selmanj: I am undone

will fix

google/resource_compute_project_metadata_item_test.go
```diff
@@ -0,0 +1,136 @@
...
+			return fmt.Errorf("Unable to find a value for key '%s'", key)
+		}
+		if val != value {
+			return fmt.Errorf("Value for key '%s' does not match. Expected '%s' but found '%s'", key, value, val)
+		}
+		return nil
+	}
+}
+
+func testAccProjectMetadataItem_basic(key, val string) string {
```
selmanj: Done

website/docs/r/compute_project_metadata_item.html.markdown
```diff
@@ -0,0 +1,48 @@
+---
+layout: "google"
+page_title: "Google: google_compute_project_metadata_item"
+sidebar_current: "docs-google-compute-project-metadata_item"
```
selmanj: AGH, fixed

website/docs/r/compute_project_metadata_item.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+Manages a single key/value pair on metadata common to all instances for
+a project in GCE. Using `google_compute_project_metadata_item` lets you
+manage a single key/value setting in Terraform rather than the entire
+project metadata map.
+
+## Example Usage
+
+```hcl
+resource "google_compute_project_metadata_item" "default" {
+  key: "my_metadata"
```
selmanj: Fixed

google/provider.go
```diff
@@ -297,3 +298,39 @@ func convertStringArr(ifaceArr []interface{}) []string {
 	}
 	return arr
 }
+
+// flattenComputeMetadata transforms a list of MetadataItems (as returned via the GCP client) into a simple map from key
```
selmanj: I moved it to metadata.go as it works with both project metadata and instance metadata

danawillow:
```
make testacc TEST=./google TESTARGS='-run=TestAccComputeProjectMetadataItem'
TF_ACC=1 go test ./google -v -run=TestAccComputeProjectMetadataItem -timeout 120m
=== RUN   TestAccComputeProjectMetadataItem_importBasic
--- PASS: TestAccComputeProjectMetadataItem_importBasic (38.07s)
=== RUN   TestAccComputeProjectMetadataItem_basic
--- PASS: TestAccComputeProjectMetadataItem_basic (22.98s)
=== RUN   TestAccComputeProjectMetadataItem_basicWithEmptyVal
--- PASS: TestAccComputeProjectMetadataItem_basicWithEmptyVal (22.58s)
=== RUN   TestAccComputeProjectMetadataItem_basicUpdate
--- PASS: TestAccComputeProjectMetadataItem_basicUpdate (36.65s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	120.403s
```

👍

PR #177, opened 2017-07-05
Add support for Google Cloud Bigtable.

rileykarson:
Merge the `bigtable-support` branch into `master`.

Add support for two new resources:
- `google_bigtable_instance` - A container for a single Bigtable cluster.
- `google_bigtable_table` - A Bigtable table contained within an instance.

### DIFF
--- a/google/bigtable_client_factory.go
+++ b/google/bigtable_client_factory.go
@@ -8,15 +8,15 @@ import (
 	"google.golang.org/api/option"
 )
 
-type ClientFactoryBigtable struct {
+type BigtableClientFactory struct {
 	UserAgent   string
 	TokenSource oauth2.TokenSource
 }
 
-func (s *ClientFactoryBigtable) NewInstanceAdminClient(project string) (*bigtable.InstanceAdminClient, error) {
+func (s *BigtableClientFactory) NewInstanceAdminClient(project string) (*bigtable.InstanceAdminClient, error) {
 	return bigtable.NewInstanceAdminClient(context.Background(), project, option.WithTokenSource(s.TokenSource), option.WithUserAgent(s.UserAgent))
 }
 
-func (s *ClientFactoryBigtable) NewAdminClient(project, instance string) (*bigtable.AdminClient, error) {
+func (s *BigtableClientFactory) NewAdminClient(project, instance string) (*bigtable.AdminClient, error) {
 	return bigtable.NewAdminClient(context.Background(), project, instance, option.WithTokenSource(s.TokenSource), option.WithUserAgent(s.UserAgent))
 }
--- a/google/bigtable_client_factory.go
+++ b/google/bigtable_client_factory.go
@@ -13,10 +13,10 @@ type BigtableClientFactory struct {
 	TokenSource oauth2.TokenSource
 }
 
-func (s *BigtableClientFactory) NewInstanceAdminClient(project string) (*bigtable.InstanceAdminClient, error) {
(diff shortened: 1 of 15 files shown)


danawillow:
Assuming nothing's changed since the other reviews, 👍

google/resource_bigtable_instance.go
```diff
@@ -0,0 +1,193 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+
+			"num_nodes": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				ForceNew:     true,
```
selmanj: This value can't be updated? Seems like a natural thing to want to adjust later

google/bigtable_client_factory.go
```diff
@@ -0,0 +1,22 @@
...
+	"golang.org/x/oauth2"
+	"google.golang.org/api/option"
+)
+
+type BigtableClientFactory struct {
+	UserAgent   string
+	TokenSource oauth2.TokenSource
+}
+
+func (s *BigtableClientFactory) NewInstanceAdminClient(project string) (*bigtable.InstanceAdminClient, error) {
```
selmanj: If you remove the pointer on the receiver in the signature (e.g. 
```
func (s BigtableClientFactory) NewInstanceAdminClient(project string) (*bigtable.InstanceAdminClient, error) {
```
), then this function is much more explicit that it is immutable (and is therefore thread safe, etc)

google/bigtable_client_factory.go
```diff
@@ -0,0 +1,22 @@
...
+	return bigtable.NewInstanceAdminClient(context.Background(), project, option.WithTokenSource(s.TokenSource), option.WithUserAgent(s.UserAgent))
+}
+
+func (s *BigtableClientFactory) NewAdminClient(project, instance string) (*bigtable.AdminClient, error) {
```
selmanj: Same comment as above

google/resource_bigtable_instance.go
```diff
@@ -0,0 +1,193 @@
...
+		StorageType: storageType,
+		Zone:        d.Get("zone").(string),
+	}
+
+	c, err := config.bigtableClientFactory.NewInstanceAdminClient(project)
+	if err != nil {
+		return fmt.Errorf("Error starting instance admin client. %s", err)
+	}
+
+	defer c.Close()
```
selmanj: I think golang tooling gets mad if you don't check the return result here in case it returns an error

google/resource_bigtable_instance.go
rileykarson: We can't update it with the client as-is, or even read it for that matter. When the client supports updating this value, we can remove `ForceNew`.

google/bigtable_client_factory.go
rileykarson: Done

google/bigtable_client_factory.go
rileykarson: Done

google/resource_bigtable_instance.go
rileykarson: I haven't had Gogland get mad at me, and I don't think we can do anything useful with an error from this - any interesting errors would have come from operations we performed.

selmanj: Yeah agree, the only thing we can do is log it but if we're closing it I doubt we're interested in the log message 👍

google/resource_bigtable_instance.go
selmanj: OK

PR #185, opened 2017-07-11
Add support for auto_healing_policies to google_compute_instance_group_manager.

rileykarson:
Add support for `auto_healing_policies` to `google_compute_instance_group_manager`.

Walking through the below list of scenarios, I saw the expected results:

- `CREATE v1`
- `REFRESH v1`
- `DESTROY v1`
- `IMPORT v1`
- `UPGRADE v1 to Beta`
- `CREATE Beta`
- `REFRESH Beta`
- `DESTROY Beta`
- `IMPORT Beta`
- `UPGRADE Beta to v1`

The weird cases were, as we expected:

- `UPGRADE v1 to Beta` - We can't read at Beta during the pre-apply refresh, so we don't catch out of band changes to Beta fields
- `IMPORT Beta` - All resources are currently imported at `v1`
- `UPGRADE Beta to v1` - The post-apply read will be performed at Beta and not `v1`, so the state file will have Beta self links. The next refresh will correct this.

### DIFF
--- a/google/resource_compute_instance_group_manager.go
+++ b/google/resource_compute_instance_group_manager.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 
 	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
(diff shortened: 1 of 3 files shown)


rileykarson:
After the change:

```
TF_ACC=1 go test ./google -v -run=TestAccInstanceGroupManager_autoHealingPolicies -timeout 120m
=== RUN   TestAccInstanceGroupManager_autoHealingPolicies
--- PASS: TestAccInstanceGroupManager_autoHealingPolicies (226.61s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	226.800s
```

PR #187, opened 2017-07-12
[WIP] Multiversioned Compute Service & Versioned Operations

rileykarson:
I'm looking for high level comments and discussion with this PR - it isn't on track to be merged. It is a combination of a refactoring of `Operation` and the addition of `compute_multiversion_service.go` as-is, both of which should land as separate changes.

Right now, we directly invoke the `compute` service. This means that we need to add a lot of awkward conditionals and switches to resource files in order to support multiple versions of the same resource; we also need to wait on several different kinds of `Operation`, all of which have identical data.

What if we owned the `compute` service, though? Then we could;

- Abstract away the exact service object we perform a request with.
- Return the same kind of `Operation` struct regardless of version.
- Allow resource files to exclusively use the highest version of structs, converting them in the service.
- Clean up most version information from struct files, making them easier to read

And in the future;

- We can provide interfaces for the subset of methods required by individual resources, and potentially mock those in tests

We would ultimately own more lines of code as a result of this change - in turn, though, much of it is boilerplate converting between versions as we need to do that a bit more often than we currently need to in just resource files, such as with `Operation` structs.

This PR highlights what I see some of the changes to `compute_operation.go` looking like, and what a single API method being added to the service does to the current versioned resource code.


danawillow:
Initial thoughts: refactoring operation is a good idea, regardless. There's a lot of stuff duplicated between the region/zone/global versions of operation that doesn't really need to be duplicated.

I'm a bit more cautious on the compute multiversion service just for upkeep purposes, though I wonder whether it could be autogenerated.

rileykarson:
Closing since this is superseded by #198 and #240.

PR #191, opened 2017-07-14
Refactor Compute Operations to work identically & introduce a scope-less API

rileykarson:
We can use the `Region` and `Zone` properties of a compute `Operation` to infer it's scope; we don't need to call scope-specific methods.

- Refactor `compute_operation.go` to use the same code paths for all 3 operation types
- Introduce a universal operation type that new resources should use going forward.

We can migrate existing resources over time with a series of followup CLs, eliminating the functions for each scope type as we go.

Smoke testing this with a resource of each scope:

Global:
```
TF_ACC=1 go test ./google -v -run=TestAccComputeGlobalAddress -timeout 120m
=== RUN   TestAccComputeGlobalAddress_importBasic
--- PASS: TestAccComputeGlobalAddress_importBasic (22.76s)
=== RUN   TestAccComputeGlobalAddress_basic
--- PASS: TestAccComputeGlobalAddress_basic (22.31s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	45.236
```

Regional:
```
TF_ACC=1 go test ./google -v -run=TestAccComputeAddress -timeout 120m
--- PASS: TestAccComputeAddress_importBasic (22.98s)
--- PASS: TestAccComputeAddress_basic (23.04s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	46.226s
```

Zonal:
```
TF_ACC=1 go test ./google -v -run=TestAccComputeInstance_basic1 -timeout 120m
=== RUN   TestAccComputeInstance_basic1
--- PASS: TestAccComputeInstance_basic1 (61.90s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	62.100s
```

### DIFF
--- a/google/compute_operation.go
+++ b/google/compute_operation.go
@@ -7,27 +7,24 @@ import (
 	"time"
 
 	"github.com/hashicorp/terraform/helper/resource"
+
 	"google.golang.org/api/compute/v1"
 )
 
-// OperationWaitType is an enum specifying what type of operation
-// we're waiting on.
-type ComputeOperationWaitType byte
+type ScopeType uint8
 
 const (
-	ComputeOperationWaitInvalid ComputeOperationWaitType = iota
-	ComputeOperationWaitGlobal
-	ComputeOperationWaitRegion
-	ComputeOperationWaitZone
+	Global ScopeType = iota
+	Region
+	Zone
 )
 
 type ComputeOperationWaiter struct {
-	Service *compute.Service
-	Op      *compute.Operation
-	Project string
-	Region  string
-	Type    ComputeOperationWaitType
-	Zone    string
+	Service   *compute.Service
+	Op        *compute.Operation
+	Project   string
+	Scope     string
+	ScopeType ScopeType
 }
 
 func (w *ComputeOperationWaiter) RefreshFunc() resource.StateRefreshFunc {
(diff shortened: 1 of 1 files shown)


google/compute_operation.go
```diff
@@ -4,58 +4,64 @@ import (
...
+		}
+
+		switch scopeType {
+		case Global:
+			op, err = w.Service.GlobalOperations.Get(w.Project, w.Op.Name).Do()
+		case Region:
+			op, err = w.Service.RegionOperations.Get(w.Project, scope, w.Op.Name).Do()
+		case Zone:
+			op, err = w.Service.ZoneOperations.Get(w.Project, scope, w.Op.Name).Do()
 		default:
```
danawillow: It doesn't seem like this condition could possibly be hit. What if the `op` assignments were all done in the previous if/else statements?

google/compute_operation.go
```diff
@@ -74,93 +80,83 @@ type ComputeOperationError compute.OperationError
...
 		Op:      op,
 		Project: project,
-		Zone:    zone,
-		Type:    ComputeOperationWaitZone,
 	}
+
+	return waitComputeOperationWaiter(w, timeoutMin, activity)
+}
+
+func waitComputeOperationWaiter(w *ComputeOperationWaiter, timeoutMin int, activity string) error {
```
danawillow: I don't have a great suggestion for you but I don't like that "wait" occurs twice in this fn name. `waitCompute`? `waitComputeOperation` could work but I don't like that there's a `computeOperationWait` already. Or we can just leave it be since I don't think it'll need to be a separate fn once all the others are gone.

google/compute_operation.go
rileykarson: Done.

google/compute_operation.go
rileykarson: Added a todo to make sure that we inline the method once the others are eliminated.

danawillow:
Thanks! Merge at will.

PR #193, opened 2017-07-14
Refresh google_compute_autoscaler using the zone set in state when possible

### LINKED ISSUE #192: Incorrect zones for google_compute_autoscaler in tfstate
### Terraform Version
Terraform v0.9.11

### Affected Resource(s)
- google_compute_autoscaler

https://gist.github.com/pst/aed987feba7b9514a9887783ce975028

### Debug Output
 1. First successful `terraform apply`:
     https://gist.github.com/pst/5d096cd8d720a83b23d81c4e09516aeb
 1. `terraform show` with incorrect state:
     https://gist.github.com/pst/35a110ec5249ad69b6bb7b39ebff812f
 1. `terraform plan` (forcing new autoscaler resources):
      https://gist.github.com/pst/9ab7340f185342fa4106db641faa5596
 1. Second `terraform apply` that fails (not all that interesting because the plan is already wrong):
      https://gist.github.com/pst/00d1839fc5c1e1cc3915d4feb5919c06

### Expected Behavior
Applying above hcl, terraform creates the resources. Re-applying the unchanged hcl, does not change the resources.

### Actual Behavior
Applying above hcl, terraform correctly creates the resources on GCP. Most importantly the `google_compute_instance_group_manager` and `google_compute_autoscaler` resources are correctly distributed across the three zones specified.

However, the `terraform.ftstate` file and `terraform show` claim all `google_compute_autoscaler` resources are in a single zone (see [autoscaler.1](https://gist.github.com/pst/35a110ec5249ad69b6bb7b39ebff812f#file-tf_debug_show_after_first_apply-log-L47) and [autoscaler.2](https://gist.github.com/pst/35a110ec5249ad69b6bb7b39ebff812f#file-tf_debug_show_after_first_apply-log-L62)). Hence a `terraform plan` and `terraform apply` try to assign the correct zone which forces a new resource (see [plan output](https://gist.github.com/pst/9ab7340f185342fa4106db641faa5596#file-tf_debug_plan_after_first_apply-log-L4707:L4730)). This then also fails with a `Error 400` from the API. Optimally the internal state should correctly reflect the reality on GCP and not try to recreate the the resources when the hcl is applied again.

### Steps to Reproduce
 1. `terraform apply`
 1. `terraform show`  # not required to reproduce, just for clarification
 1. `terraform plan`  # not required to reproduce, just for clarification
 1. `terraform apply`

rileykarson:
Previously, we refreshed `google_compute_autoscaler` by scanning the `region`'s `zone`s every time, instead of only in the import path, when we didn't know the `zone`. This brings `google_compute_autoscaler`'s `Read` path in line with both other importable zonal resources, `google_compute_disk` and `google_compute_instance_group_manager`.

Fixes #192.

### DIFF
--- a/google/resource_compute_autoscaler.go
+++ b/google/resource_compute_autoscaler.go
@@ -289,20 +289,37 @@ func resourceComputeAutoscalerRead(d *schema.ResourceData, meta interface{}) err
 	if err != nil {
 		return err
 	}
+
 	var getAutoscaler = func(zone string) (interface{}, error) {
 		return config.clientCompute.Autoscalers.Get(project, zone, d.Id()).Do()
 	}
 
-	resource, err := getZonalResourceFromRegion(getAutoscaler, region, config.clientCompute, project)
-	if err != nil {
-		return err
+	var scaler *compute.Autoscaler
+	var e error
+	if zone, ok := d.GetOk("zone"); ok {
+		scaler, e = config.clientCompute.Autoscalers.Get(project, zone.(string), d.Id()).Do()
+		if e != nil {
+			return handleNotFoundError(e, d, fmt.Sprintf("Autoscaler %q", d.Id()))
+		}
+	} else {
+		// If the resource was imported, the only info we have is the ID. Try to find the resource
+		// by searching in the region of the project.
+		var resource interface{}
+		resource, e = getZonalResourceFromRegion(getAutoscaler, region, config.clientCompute, project)
+
+		if e != nil {
+			return e
+		}
+
+		scaler = resource.(*compute.Autoscaler)
 	}
-	if resource == nil {
+
+	if scaler == nil {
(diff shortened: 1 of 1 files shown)


rileykarson:
```
TF_ACC=1 go test ./google -v -run=TestAccComputeAutoscaler_ -timeout 120m
=== RUN   TestAccComputeAutoscaler_importBasic
--- PASS: TestAccComputeAutoscaler_importBasic (71.13s)
=== RUN   TestAccComputeAutoscaler_basic
--- PASS: TestAccComputeAutoscaler_basic (69.71s)
=== RUN   TestAccComputeAutoscaler_update
--- PASS: TestAccComputeAutoscaler_update (82.64s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	223.591s
```

Issue #182, opened 2017-07-09
Add import functionality for Google Compute Engine resources


jvrplmlmn:
### Terraform Version

```bash
$ terraform --version
Terraform v0.10.0-beta2-dev
```

### Affected Resource(s)

Most Google Compute Engine resources lack the import functionality. An example: `google_compute_instance`.

danawillow:
Hey @jvrplmlmn, just for prioritization: which resources in particular would you like import support on? Obviously we'd like to include it for all of them but this will at least help us come up with an ordering.

jvrplmlmn:
Hi @danawillow, right now I'm mainly interested in `google_compute_instance`.

Is there any guide on how to write an importer function for Terraform? I might be able to contribute a patch if I get some guidance 😃

danawillow:
Self-assigning but @rosbo will be working on this.

@jvrplmlmn there isn't a guide per se, but take a look at https://github.com/terraform-providers/terraform-provider-google/pull/40 for an example. `google_compute_instance` isn't quite as straightforward though, because it currently reads from state in the read function (you can tell from the `d.Get(...)`s which won't work for import, so it'll need to stop doing that first.

rosbo:
Most compute resources now have import support. `google_compute_instance` doesn't have import support yet but there is a specific issue to address it (#269).

If there is a specific resource for which you would like to have import support, please open a new issue for that specific resource.

Closing this issue.

PR #194, opened 2017-07-14
Add import support for google_compute_image.

rosbo:
issue #182 
/cc @danawillow

### DIFF
--- a/google/resource_compute_image.go
+++ b/google/resource_compute_image.go
@@ -14,6 +14,7 @@ func resourceComputeImage() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceComputeImageCreate,
 		Read:   resourceComputeImageRead,
+		Update: resourceComputeImageUpdate,
 		Delete: resourceComputeImageDelete,
 		Importer: &schema.ResourceImporter{
 			State: resourceComputeImageImportState,
@@ -87,7 +88,6 @@ func resourceComputeImage() *schema.Resource {
 				Type:     schema.TypeInt,
 				Optional: true,
 				Default:  computeImageCreateTimeoutDefault,
-				ForceNew: true,
 			},
 		},
 	}
@@ -186,6 +186,12 @@ func resourceComputeImageRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("family", image.Family)
 	d.Set("self_link", image.SelfLink)
 
+	return nil
+}
+
+func resourceComputeImageUpdate(d *schema.ResourceData, meta interface{}) error {
+	// Pass-through for updates to Terraform-specific `create_timeout` field.
+	// The Google Cloud Image resource doesn't support update.
 	return nil
 }
 
--- a/google/resource_compute_image.go
+++ b/google/resource_compute_image.go
@@ -8,11 +8,16 @@ import (
 	"google.golang.org/api/compute/v1"
 )
 
+const computeImageCreateTimeoutDefault = 4
+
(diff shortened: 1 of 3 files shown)


google/resource_compute_image.go
```diff
@@ -195,3 +213,9 @@ func resourceComputeImageDelete(d *schema.ResourceData, meta interface{}) error
 	d.SetId("")
 	return nil
 }
+
+func resourceComputeImageImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	d.Set("create_timeout", computeImageCreateTimeoutDefault)
```
danawillow: Can you explain why this is necessary instead of setting `create_timeout` on read every time?

rosbo: Added a comment explaining why.

Another option would be to set the `create_timeout` in the read method if it is not set but I prefer not cluttering the read method logic with logic specific to import.

danawillow: Hmm, so actually I think we should stop this from being ForceNew. Just from thinking about it logically, it's the amount of time we're willing to wait for the instance to create. IMO, changing it doesn't actually mean I want to create a new one, it just means that should it recreate for some other reason, we would want to wait that long. Does that seem right? If so we can remove ForceNew and just never do a `d.Set` on it.

rosbo: Agree for the ForceNew part, I will do it.

However, if we don't set `create_timeout` to the default value on import, then, after importing, the user will always have a "plan diff" with:
  ~ google_compute_image.web-image
      create_timeout: "" => "4"

The update is fast (only update the value locally). I am wondering if this is the expected behavior that after an update, we may have a diff that we can't fix without either running apply or fiddling with the state file.

Thoughts?

danawillow: Oh, you're right. In this case the way you've done it is probably correct, then.

rosbo: I removed the ForceNew from the create_timeout field.

website/docs/r/compute_image.html.markdown
```diff
@@ -83,3 +83,11 @@ In addition to the arguments listed above, the following computed attributes are
 exported:
 
 * `self_link` - The URI of the created resource.
+
+## Import
+
+VM image can be imported using the `name`, e.g.
+
+```
+$ terraform import google_compute_image.web-image image1
```
danawillow: Can you change `image1` to `my-custom-image` since that's the name in the example?

rosbo: Done

danawillow:
Looks good, thanks!

```
make testacc TEST=./google TESTARGS='-run=TestAccComputeImage'
TF_ACC=1 go test ./google -v -run=TestAccComputeImage -timeout 120m
=== RUN   TestAccComputeImage_resolveImage
--- PASS: TestAccComputeImage_resolveImage (83.84s)
=== RUN   TestAccComputeImage_importFromRawDisk
--- PASS: TestAccComputeImage_importFromRawDisk (67.65s)
=== RUN   TestAccComputeImage_importFromSourceDisk
--- PASS: TestAccComputeImage_importFromSourceDisk (80.38s)
=== RUN   TestAccComputeImage_basic
--- PASS: TestAccComputeImage_basic (77.72s)
=== RUN   TestAccComputeImage_basedondisk
--- PASS: TestAccComputeImage_basedondisk (81.10s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	390.833s
```

PR #195, opened 2017-07-14
Store non-v1 resources' self links as v1

rileykarson:
Store non-v1 resources' self links as v1 so that dependent single-version resources don't see diffs.

### DIFF
--- a/google/resource_compute_instance_group_manager.go
+++ b/google/resource_compute_instance_group_manager.go
@@ -373,7 +373,7 @@ func resourceComputeInstanceGroupManagerRead(d *schema.ResourceData, meta interf
 	d.Set("fingerprint", manager.Fingerprint)
 	d.Set("instance_group", manager.InstanceGroup)
 	d.Set("target_size", manager.TargetSize)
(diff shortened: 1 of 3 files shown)


danawillow:
```
make testacc TEST=./google TESTARGS='-run=TestAccInstanceGroupManager_selfLinkStability'
TF_ACC=1 go test ./google -v -run=TestAccInstanceGroupManager_selfLinkStability -timeout 120m
=== RUN   TestAccInstanceGroupManager_selfLinkStability
--- PASS: TestAccInstanceGroupManager_selfLinkStability (113.59s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	113.728s
```
Looks good. Can you add a comment somewhere in the test (possibly in both the test definition part and the config part) about what the test is actually doing (basically just that because autohealing is set, the igm is going to be created with beta apis and thus have beta self links). And also maybe a TODO or something to update the test to use something else once autohealing isn't in beta anymore.

rileykarson:
Done

Issue #93, opened 2017-06-13
provider/google: support alpha/beta compute APIs in addition to GA


hashibot:
_This issue was originally opened by @danawillow as hashicorp/terraform#14742. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Currently, only features available in the v1 version of the compute API are able to be used in Terraform. This is a general issue for support of alpha and beta APIs as well.

(assigning myself but @rileykarson will be working on this)

rileykarson:
Given that we've closed out most of the issues in this list, I'm going to close the issue. There are a few issues left outstanding;

- #132 has gone GA :fireworks: Feel free to :+1:  the issue if it's an important feature for you!
- #79 and #80 are both in Container Engine, which only has a single client version.
- #106, #212, #288 are all for `google_compute_instance`; that resource is being tracked by #311.
- #51, #91 are both in Alpha. As with all of our issues; add a :+1: if it's important to you, and feel free to bump the issue when it has gone Beta!

davidquarles:
@rileykarson Forgive me, I'm new here, but is the [container/v1beta1](https://godoc.org/google.golang.org/api/container/v1beta1) API an extremely recent addition, or are you saying multi-version support for the GKE API won't be supported?  Specifically, I'd like to use/add the [NodeTaint](https://godoc.org/google.golang.org/api/container/v1beta1#NodeTaint) field to `google_container_node_pool`, but I'm imagining multi-version API support is a rather intricate prerequisite.

If multi-version support for the GKE client is desirable, does it merit opening a new/separate issue?  Any insight is appreciated.  Thanks!

rileykarson:
Hi @davidquarles!

We created this issue to track multi-version support for just the `compute` API, but that had more to do with most/all of the alpha/beta feature requests being against `compute` at the time, and so that we could consolidate information about supporting multiple versions of those APIs in one place instead of having parts of it live in separate issues.

I would recommend filing an issue against `google_container_node_pool` to support NodeTaint, and to add a note in the issue body that the feature is only present at `v1beta1`  and not `v1`.

[...]

PR #198, opened 2017-07-14
Add support for auto_healing_policies to google_compute_instance_group_manager

### LINKED ISSUE #48: provider/google: Support for managed instance group autohealing
_This issue was originally opened by @premist as hashicorp/terraform#10360. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Managed instance group on Google Compute Engine has a feature where you can monitor and automatically re-create instances based on associated health check. You can use `instanceGroupManagers.setAutoHealingPolicies` to apply the health check to the managed instance group, like the following:

```http
POST https://www.googleapis.com/compute/beta/projects/[PROJECT_ID]/zones/[ZONE]/instanceGroupManagers/[INSTANCE_GROUP]/setAutoHealingPolicies

{
 "autoHealingPolicies": [
  {
   "healthCheck": "global/httpHealthChecks/example-check",
   "initialDelaySec": 120
  }
 ]
}
```

Note that this API is currently in beta.

### Documentations

- [API Reference](https://cloud.google.com/compute/docs/reference/beta/instanceGroupManagers/setAutoHealingPolicies)
- [Setting up health checking for managed instance groups](https://cloud.google.com/compute/docs/instance-groups/creating-groups-of-managed-instances#monitoring_groups)

rileykarson:
Add support for `auto_healing_policies` to `google_compute_instance_group_manager`; Fixes #48 
Add support for a beta GCP feature to a Google Provider resource; part of #93 

Terraform manages `v0beta` and `v0alpha` Compute resources by using non-`v1` APIs to create and manage resources when necessary, automatically using the most restrictive API it can. Users don't need to think about what version they are using, or perform complicated operations to manage their state files.

For reviewing, the best way to view the merge changes is by filtering for only the last two commits in the change view; Github is able to show just those changes and hides the merged commits.


danawillow:
Assuming this is just merging that branch into master with all the different changes I've reviewed, I'm ok with this. However, since it's a pretty big infrastructure change, it may be worth it to check in slack whether anyone from HashiCorp wants to review it first or whether they're ok with our judgment.

google/api_versions.go
```diff
@@ -0,0 +1,99 @@
+package google
```
evandbrown: Need license header

google/compute_beta_operation.go
```diff
@@ -0,0 +1,167 @@
...
+	state := w.Conf()
+	state.Delay = 10 * time.Second
+	state.Timeout = 4 * time.Minute
+	state.MinTimeout = 2 * time.Second
+	opRaw, err := state.WaitForState()
+	if err != nil {
+		return fmt.Errorf("Error waiting for %s: %s", activity, err)
+	}
+
+	op = opRaw.(*computeBeta.Operation)
```
evandbrown: Worth checking that the type assertion succeeds here? Like `op, ok := opRaw.(...)`

google/compute_beta_operation.go
```diff
@@ -0,0 +1,167 @@
...
+	if op.Error != nil {
+		return ComputeBetaOperationError(*op.Error)
+	}
+
+	return nil
+}
+
+func computeBetaOperationWaitZone(config *Config, op *computeBeta.Operation, project string, zone, activity string) error {
+	return computeBetaOperationWaitZoneTime(config, op, project, zone, 4, activity)
```
evandbrown: Should we make the time vars more configurable/obvious?

google/compute_beta_operation.go
```diff
@@ -0,0 +1,167 @@
...
+
+	op = opRaw.(*computeBeta.Operation)
+	if op.Error != nil {
+		return ComputeBetaOperationError(*op.Error)
+	}
+
+	return nil
+}
+
+func computeBetaOperationWaitZone(config *Config, op *computeBeta.Operation, project string, zone, activity string) error {
```
evandbrown: `project, zone, activity string` (remove the `string` after project)

google/compute_beta_operation.go
```diff
@@ -0,0 +1,167 @@
...
+func computeBetaOperationWaitZoneTime(config *Config, op *computeBeta.Operation, project string, zone string, minutes int, activity string) error {
+	w := &ComputeBetaOperationWaiter{
+		Service: config.clientComputeBeta,
+		Op:      op,
+		Project: project,
+		Zone:    zone,
+		Type:    ComputeBetaOperationWaitZone,
+	}
+	state := w.Conf()
+	state.Delay = 10 * time.Second
```
evandbrown: See previous about configuring timeouts in one place.

google/compute_shared_operation.go
```diff
@@ -0,0 +1,25 @@
...
+	computeBeta "google.golang.org/api/compute/v0.beta"
+	"google.golang.org/api/compute/v1"
+)
+
+func computeSharedOperationWaitZone(config *Config, op interface{}, project string, zone, activity string) error {
+	return computeSharedOperationWaitZoneTime(config, op, project, zone, 4, activity)
+}
+
+func computeSharedOperationWaitZoneTime(config *Config, op interface{}, project string, zone string, minutes int, activity string) error {
+	if op == nil {
```
evandbrown: nil check shouldn't be necessary as the type switch's `default` will catch it.

google/compute_beta_operation.go
```diff
@@ -0,0 +1,167 @@
...
+		case ComputeBetaOperationWaitGlobal:
+			op, err = w.Service.GlobalOperations.Get(
+				w.Project, w.Op.Name).Do()
+		case ComputeBetaOperationWaitRegion:
+			op, err = w.Service.RegionOperations.Get(
+				w.Project, w.Region, w.Op.Name).Do()
+		case ComputeBetaOperationWaitZone:
+			op, err = w.Service.ZoneOperations.Get(
+				w.Project, w.Zone, w.Op.Name).Do()
+		default:
```
rileykarson: `ComputeBetaOperation` is a copy of the existing `ComputeOperation` where we sub `compute/v0beta`'s `Operation` type for `compute/v1`'s `Operation` type. These changes would also need to be applied to `compute_operation.go` as well as `compute_beta_operation.go`. and I don't think that kind of change makes sense in the context of this PR.

Improving Operations is something that I've been thinking about with #187, which would introduce a standard `Operation` type for all versioned compute resources (such as storing them all as `v1` Operations), and a refactoring of `compute_operation.go` that was pulled into #191.

google/compute_beta_operation.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// ComputeBetaOperationError wraps computeBeta.OperationError and implements the
+// error interface so it can be returned.
+type ComputeBetaOperationError computeBeta.OperationError
+
+func (e ComputeBetaOperationError) Error() string {
+	var buf bytes.Buffer
+
+	for _, err := range e.Errors {
```
rileykarson: See above. If it's something you think would be valuable for `compute_operation.go` as well, feel free to comment with it on #191.

google/compute_beta_operation.go
rileykarson: See above.

google/compute_beta_operation.go
rileykarson: See above/changes made in #191.

google/compute_beta_operation.go
rileykarson: See above. If we want to expose a better interface, that's something we should consider for #191.

google/compute_beta_operation.go
rileykarson: See above. Change made in https://github.com/terraform-providers/terraform-provider-google/pull/191/files#diff-ef92ceb02a5a2c0c742317597b0bc870R146

google/api_versions.go
rileykarson: Are license headers something that we include in any of these files? Looking over `provider.go`, `config.go`, and a few resources, I can't see one included elsewhere.

google/compute_shared_operation.go
rileykarson: Added a `nil` check as a case in #234 instead of having it at the top.

rileykarson:
Closing this PR in favour of splitting it in 2, adding Beta support to `google_compute_instance_group_manager` in #234 followed by adding `auto_healing_policies` in a later PR.

google/api_versions.go
evandbrown: Ahh, I guess not. But we should. I'll file an issue.

PR #201, opened 2017-07-14
Support import for google_compute_instance_group

rosbo:
issue #182

cc/ @danawillow

### DIFF
--- a/google/resource_compute_instance_group_migrate_test.go
+++ b/google/resource_compute_instance_group_migrate_test.go
@@ -8,24 +8,43 @@ import (
 
 func TestComputeInstanceGroupMigrateState(t *testing.T) {
 	cases := map[string]struct {
-		StateVersion int
-		Attributes   map[string]string
-		Expected     map[string]string
-		Meta         interface{}
+		StateVersion       int
+		Attributes         map[string]string
+		ExpectedAttributes map[string]string
+		ExpectedId         string
+		Meta               interface{}
 	}{
-		"change instances from list to set": {
+		"v1 to v2": {
+			StateVersion: 1,
+			Attributes: map[string]string{
+				"zone": "us-central1-c",
+				"name": "instancegroup-test",
+			},
+			ExpectedAttributes: map[string]string{
+				"zone": "us-central1-c",
+				"name": "instancegroup-test",
+			},
+			ExpectedId: "us-central1-c/instancegroup-test",
+			Meta:       &Config{},
+		},
+		"v0 to v2": {
 			StateVersion: 0,
 			Attributes: map[string]string{
+				"zone":        "us-central1-c",
+				"name":        "instancegroup-test",
 				"instances.#": "1",
 				"instances.0": "https://www.googleapis.com/compute/v1/projects/project_name/zones/zone_name/instances/instancegroup-test-1",
 				"instances.1": "https://www.googleapis.com/compute/v1/projects/project_name/zones/zone_name/instances/instancegroup-test-0",
 			},
-			Expected: map[string]string{
+			ExpectedAttributes: map[string]string{
+				"zone":                 "us-central1-c",
+				"name":                 "instancegroup-test",
 				"instances.#":          "1",
 				"instances.764135222":  "https://www.googleapis.com/compute/v1/projects/project_name/zones/zone_name/instances/instancegroup-test-1",
 				"instances.1519187872": "https://www.googleapis.com/compute/v1/projects/project_name/zones/zone_name/instances/instancegroup-test-0",
 			},
-			Meta: &Config{},
+			ExpectedId: "us-central1-c/instancegroup-test",
+			Meta:       &Config{},
 		},
 	}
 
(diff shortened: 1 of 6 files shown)


website/docs/r/compute_instance_group.html.markdown
```diff
@@ -97,3 +97,11 @@ exported:
 * `self_link` - The URI of the created resource.
 
 * `size` - The number of instances in the group.
+
+## Import
+
+Instance group can be imported using the `zone` and `name`, e.g.
```
rileykarson: Just wondering if you and @danawillow have talked about what we want to do for `region` and `zone` across all importable resources - we've checked against the provider region, or against all the provider region's zones in the past to find objects by Id alone.

I like this approach a lot - it means we don't clutter read with parsing logic by using an import func, we don't need to write migration functions, and it lets users import resources from more than 1 region/1 set of zones inside a single provider. Are we going to use the same approach across all importable resources going forward and do we have a plan to bring it to older ones?

rosbo: I decided to import using zone/name because it is a valid case to have multiple instance groups with the same name across different zones. 

I will sync up with @danawillow to ensure consistency on our approach for all importable resources.

rosbo: After talking with @danawillow and Evan, we decided that import method should:
- Always supports the fully specified resource identifier (implemented in this PR).
- For convenience, we can add support for partial specification by `name` only if guessing the other parameters is feasible. The logic for guessing should be in the ImportFunc and not the read method to avoid polluting the read method's logic. If more than one resource match the name (e.g. two instance groups in two different zones), fail with an error message to avoid surprising behavior.
- We will go ahead and submit this PR as is. I will add support for the convenient import by name only in a separate PR.
- We should update the import method for all the other resources to support the fully specified resource. Otherwise, in the case where two resources have the same name in two different zones or regions, the user cannot import the resource.

danawillow:
```
make testacc TEST=./google TESTARGS='-run=TestAccComputeInstanceGroup_'
TF_ACC=1 go test ./google -v -run=TestAccComputeInstanceGroup_ -timeout 120m
=== RUN   TestAccComputeInstanceGroup_import
--- PASS: TestAccComputeInstanceGroup_import (136.04s)
=== RUN   TestAccComputeInstanceGroup_basic
--- PASS: TestAccComputeInstanceGroup_basic (92.69s)
=== RUN   TestAccComputeInstanceGroup_update
--- PASS: TestAccComputeInstanceGroup_update (134.13s)
=== RUN   TestAccComputeInstanceGroup_outOfOrderInstances
--- PASS: TestAccComputeInstanceGroup_outOfOrderInstances (93.11s)
=== RUN   TestAccComputeInstanceGroup_network
--- PASS: TestAccComputeInstanceGroup_network (154.72s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	610.832s
```

website/docs/r/compute_instance_group.html.markdown
rileykarson: SGTM! :+1:

Thanks for taking the time to flesh out a concrete plan for this.

rosbo:
Added ID state migration.

There was already a MigrateState function from version 0 to 1 but it wasn't properly setup. I have wired it properly. Let me know if you have any concerns about this.

google/resource_compute_instance_group_migrate_test.go
```diff
@@ -41,7 +60,15 @@ func TestComputeInstanceGroupMigrateState(t *testing.T) {
 			t.Fatalf("bad: %s, err: %#v", tn, err)
 		}
 
-		for k, v := range tc.Expected {
+		if is.ID != tc.ExpectedId {
+			t.Fatalf("bad: %s\n\n expected: %s\n got: %s", tn, tc.ExpectedId, is.ID)
+		}
+
+		for k, v := range tc.ExpectedAttributes {
+			if k == "id" {
```
danawillow: empty if statement?

google/resource_compute_instance_group_migrate_test.go
```diff
@@ -8,24 +8,43 @@ import (
...
-		Expected     map[string]string
-		Meta         interface{}
+		StateVersion       int
+		Attributes         map[string]string
+		ExpectedAttributes map[string]string
+		ExpectedId         string
+		Meta               interface{}
 	}{
-		"change instances from list to set": {
+		"v1 to v2": {
```
danawillow: It doesn't look like this test is actually running:
```
=== RUN   TestComputeInstanceGroupMigrateState
2017/07/19 10:00:58 [INFO] Found Compute Instance Group State v0; migrating to v1
2017/07/19 10:00:58 [DEBUG] Attributes before migration: map[string]string{"instances.#":"1", "instances.0":"https://www.googleapis.com/compute/v1/projects/project_name/zones/zone_name/instances/instancegroup-test-1", "instances.1":"https://www.googleapis.com/compute/v1/projects/project_name/zones/zone_name/instances/instancegroup-test-0"}
2017/07/19 10:00:58 [DEBUG] Attributes after migration: map[string]string{"instances.#":"1", "instances.1519187872":"https://www.googleapis.com/compute/v1/projects/project_name/zones/zone_name/instances/instancegroup-test-0", "instances.764135222":"https://www.googleapis.com/compute/v1/projects/project_name/zones/zone_name/instances/instancegroup-test-1"}
--- PASS: TestComputeInstanceGroupMigrateState (0.00s)
```
It's not showing anything migrating to v2

danawillow: wait just kidding, I forgot to actually pull your latest commit. Nothing to see here.

google/resource_compute_instance_group_migrate_test.go
rosbo: Oupps. Removed. The id assertions is done above.

google/resource_compute_instance_group_migrate_test.go
```diff
@@ -108,6 +104,10 @@ func TestComputeInstanceGroupMigrateState_empty(t *testing.T) {
 		is = &terraform.InstanceState{}
 		is, err = resourceComputeInstanceGroupMigrateState(tc.StateVersion, is, meta)
 
+		// should handle non-nil but empty
```
danawillow: I think these few lines can be removed (they're the same as what's directly above them)

danawillow:
👍 , merge at will

PR #202, opened 2017-07-16
vendor: Remove unused dependencies

radeksimko:
This is probably just leftover from the initial provider split. 🤷‍♂️


radeksimko:
@catsby Basically as I was working my way through https://github.com/hashicorp/terraform/issues/15562 and I ran through the vendor.json and saw some dependencies that made no sense to me at first sight. AWS being one of them, so I just dug deeper and tried to `govendor remove` each one I thought doesn't need to be there and then ran `make test` to see if I'm right or not.

I wish there was more sophisticated way to do this 🤷‍♂️

PR #206, opened 2017-07-17
Change automatic_restart to be computed if not specified

### LINKED ISSUE #107: Provide better default scheduling behaviour for Google Cloud preemptible instances in Instance Templates
_This issue was originally opened by @danielcompton as hashicorp/terraform#15055. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
Terraform v0.9.6

### Affected Resource(s)

```hcl
resource "google_compute_instance_template" "instance_template" {
  name_prefix = "my-dev-template-"
  description = "template description"

  tags = ["dev"]

  instance_description = "desc "
  can_ip_forward       = false

  scheduling {
    preemptible       = true
  }

  disk {
    source_image = "debian-cloud/debian-8"
    auto_delete  = true
    boot         = true
  }

  network_interface {
    network = "default"
  }

  service_account {
    scopes = ["sql-admin", "servicecontrol", "service-management", "logging-write", "monitoring-write", "trace-append", "devstorage-rw"]
  }

  lifecycle {
    create_before_destroy = true
  }
}
```

### Log output
https://gist.github.com/danielcompton/6f29b9ae9e0507be2a1c81849a234591

### Expected Behavior
The instance template using a preemptible VM should have been created.

### Actual Behavior
I got an error 

```
Error 400: Invalid value for field 'resource.properties.scheduling.preemptible': 'true'. Scheduling must have preemptible be false when AutomaticRestart is true., invalid
```

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
This is happening because `scheduling.automatic_restart` defaults to true, which is useful and matches the behaviour for standard instances, but isn't a valid option for preemptible instances. This may also apply to standard compute instances, but I haven't checked.

A few possible options I can see would be to remove the automatic_restart default entirely and let it be computed, use DefaultFunc, or maybe something else.

I also noticed that the base `automatic_restart` value causes resources to be deleted and then recreated if you have specified `scheduling.automatic_restart`, as it defaults to true, which is then treated as a change to the resource. Providing an explicit false for both of them works, but then you get warnings about using a deprecated attribute.

selmanj:
Fixes #107

### DIFF
--- a/google/resource_compute_instance_template.go
+++ b/google/resource_compute_instance_template.go
@@ -134,7 +134,7 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 			"automatic_restart": &schema.Schema{
 				Type:       schema.TypeBool,
 				Optional:   true,
-				Default:    true,
+				Computed:   true,
 				ForceNew:   true,
 				Deprecated: "Please use `scheduling.automatic_restart` instead",
 			},
@@ -262,8 +262,8 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 						"automatic_restart": &schema.Schema{
 							Type:     schema.TypeBool,
 							Optional: true,
-							Default:  true,
 							ForceNew: true,
+							Computed: true,
 						},
 
 						"on_host_maintenance": &schema.Schema{


rileykarson:
I definitely agree that the deprecated value should be computed & optional, if not removed entirely.

My concern with setting the second value to computed as well means that when we move from

```hcl
scheduling {
  automatic_restart = true
}
```

to

```hcl
scheduling {
}
```

Terraform won't see any diffs, and the resource won't get recreated. This means that the second scheduling block doesn't guarantee that the server-side resource's property is false; it can be true if there was an already existing resource and we modified the config.

What happens if the deprecated value is computed and the new one isn't? Especially in the case where users were relying on the old default of true in their configs. I expect we'll have enough diffs/recreates that we need to use Computed like this, but being able to stop that corner case before it happens would be nice as well.

rileykarson:
Also, linking #224 since this should fix it as well as #107.

google/resource_compute_instance_template.go
```diff
@@ -262,8 +262,8 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 						"automatic_restart": &schema.Schema{
 							Type:     schema.TypeBool,
 							Optional: true,
-							Default:  true,
 							ForceNew: true,
+							Computed: true,
```
rosbo: If you set this field to Computed, it means a user cannot specify a value for `automatic_restart`. GCP default value for this field is true. It means a user could never create an instance with terraform with `automatic_restart = false`. I am confused about why that would be something desirable

selmanj:
@rosbo The docs on `Computed` say that you can override the value by setting it in your config:

```
	// If Computed is true, then the result of this value is computed
	// (unless specified by config) on creation.
```

selmanj:
A main problem here is that there's no way to infer whether or not `automatic_restart` was set explicitly or not. Ideally we want to not specify it all when creating a resource, as GCP will automatically choose the correct value whether or not `is_preemptible` is specified. We can't do that here unfortunately...

google/resource_compute_instance_template.go
rileykarson: To my understanding, `Computed` means that it can't be set and can only be read, but `Computed` and `Optional` means that it acts like it is `Computed` until a value is set in config, at which point it is treated as `Optional`.

rosbo: Thanks for the clarification, I wasn't aware of this "trick"

dgolja:
What about just making sure that you can not define `automatic_restart` twice in the config file ? This would for sure fix the #224

selmanj:
Good point @dgolja; this PR is not complete and is off-topic from the original issue. I'll close this for now and submit a new one specifically around removing the deprecated parameter (which has been deprecated since 2015) and fixing scheduling.automatic_restart.

PR #208, opened 2017-07-17
Persist state from the API for google_sql_database_instance regardless of what attributes the user has set

### LINKED ISSUE #114: provider/google: google_sql_database_instance doesn't set state in read
_This issue was originally opened by @paddycarver as hashicorp/terraform#15201. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Like a lot of the older resources in the GCP provider, `google_sql_database_instance` only stores in state the properties that the user has defined in configs. This makes detecting drift unreliable, and also makes it impossible for us to support importing this resource type. We should update it to persist state from the API regardless of what is in the config.

danawillow:
Fixes #114.

~Currently a WIP because tests are failing. Debug logs for `TestAccGoogleSqlDatabaseInstance_slave ` at https://gist.github.com/danawillow/f5647fc7a78417eb99c536599742870b.~ Tests are no longer failing, this is now ready for review.

### DIFF
--- a/google/resource_sql_database_instance.go
+++ b/google/resource_sql_database_instance.go
@@ -48,6 +48,7 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 						"activation_policy": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
+							Computed: true,
 						},
 						"authorized_gae_applications": &schema.Schema{
 							Type:     schema.TypeList,
@@ -57,6 +58,7 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 						"backup_configuration": &schema.Schema{
 							Type:     schema.TypeList,
 							Optional: true,
+							Computed: true,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"binary_log_enabled": &schema.Schema{
@@ -70,6 +72,7 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 									"start_time": &schema.Schema{
 										Type:     schema.TypeString,
 										Optional: true,
+										Computed: true,
 									},
 								},
 							},
@@ -97,21 +100,26 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 						},
 						"disk_autoresize": &schema.Schema{
 							Type:             schema.TypeBool,
-							Default:          true,
 							Optional:         true,
 							DiffSuppressFunc: suppressFirstGen,
+							// Set computed instead of default because this property is for second-gen only.
+							Computed: true,
 						},
 						"disk_size": &schema.Schema{
 							Type:     schema.TypeInt,
 							Optional: true,
+							Computed: true,
 						},
 						"disk_type": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
+							// Set computed instead of default because this property is for second-gen only.
+							Computed: true,
 						},
 						"ip_configuration": &schema.Schema{
 							Type:     schema.TypeList,
 							Optional: true,
+							Computed: true,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"authorized_networks": &schema.Schema{
@@ -137,6 +145,7 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 									"ipv4_enabled": &schema.Schema{
 										Type:     schema.TypeBool,
 										Optional: true,
+										Default:  false,
 									},
 									"require_ssl": &schema.Schema{
 										Type:     schema.TypeBool,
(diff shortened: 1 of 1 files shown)


google/resource_sql_database_instance.go
```diff
@@ -640,243 +655,13 @@ func resourceSqlDatabaseInstanceRead(d *schema.ResourceData, meta interface{}) e
...
-		_ipAddress := make(map[string]interface{})
-
-		_ipAddress["ip_address"] = ip.IpAddress
-		_ipAddress["time_to_retire"] = ip.TimeToRetire
-
-		_ipAddresses[i] = _ipAddress
-	}
-
-	d.Set("ip_address", _ipAddresses)
+	d.Set("settings", flattenSettings(instance.Settings))
```
catsby: We generally recommend checking for errors from `d.Set` when setting non scalar values like slices and maps.

```
if err := d.Set("settings", flattenSettings(instance.Settings)); err != nil {
  log.Printf("[WARN] failed to set <...>")
}
```

Generally you want to log a `WARN` and not actually error out here

google/resource_sql_database_instance.go
```diff
@@ -48,6 +48,7 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 						"activation_policy": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
+							Computed: true,
```
rileykarson: It looks like we're setting a lot of values to `Computed`, which worries me a bit because of what happens when you remove a value in config to get the default behaviour; where the resource will keep the last-set value for the field. How come we've had to make so much of this computed?

danawillow: A lot of the fields have defaults that come back from the API or have defaults that differ between first and second generation instances

google/resource_sql_database_instance.go
```diff
@@ -276,6 +290,7 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 							Type:     schema.TypeBool,
 							Optional: true,
 							ForceNew: true,
+							Computed: true,
 						},
 						"master_heartbeat_period": &schema.Schema{
 							Type:     schema.TypeInt,
```
danawillow: Done

google/resource_sql_database_instance.go
danawillow: Done

google/resource_sql_database_instance.go
rileykarson: Thanks for clarifying! :+1:

danawillow:
@catsby not really, but I think that's a good thing for a refactoring change like this. I did notice that we had no ability to remove blocks in an update so I made that change, although because of the bug with Computed+Optional (see the discussion in #committers-terraform from yesterday) they don't really work the way they're supposed to anyway :-/

@rileykarson Yeah let's take that into a followup since this is big enough as it is.

rileykarson:
Following up with Postgres tests SGTM

LGTM, merge at will :+1: 

```
TF_ACC=1 go test ./google -v -run=TestAccGoogleSqlDatabaseInstance_ -timeout 120m
=== RUN   TestAccGoogleSqlDatabaseInstance_importBasic
--- PASS: TestAccGoogleSqlDatabaseInstance_importBasic (49.47s)
=== RUN   TestAccGoogleSqlDatabaseInstance_importBasic3
--- PASS: TestAccGoogleSqlDatabaseInstance_importBasic3 (370.99s)
=== RUN   TestAccGoogleSqlDatabaseInstance_basic
--- PASS: TestAccGoogleSqlDatabaseInstance_basic (57.92s)
=== RUN   TestAccGoogleSqlDatabaseInstance_basic2
--- PASS: TestAccGoogleSqlDatabaseInstance_basic2 (57.79s)
=== RUN   TestAccGoogleSqlDatabaseInstance_basic3
--- PASS: TestAccGoogleSqlDatabaseInstance_basic3 (373.19s)
=== RUN   TestAccGoogleSqlDatabaseInstance_settings_basic
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_basic (37.22s)
=== RUN   TestAccGoogleSqlDatabaseInstance_slave
--- PASS: TestAccGoogleSqlDatabaseInstance_slave (806.62s)
=== RUN   TestAccGoogleSqlDatabaseInstance_diskspecs
--- PASS: TestAccGoogleSqlDatabaseInstance_diskspecs (361.09s)
=== RUN   TestAccGoogleSqlDatabaseInstance_maintenance
--- PASS: TestAccGoogleSqlDatabaseInstance_maintenance (390.77s)
=== RUN   TestAccGoogleSqlDatabaseInstance_settings_upgrade
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_upgrade (54.81s)
=== RUN   TestAccGoogleSqlDatabaseInstance_settings_downgrade
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_downgrade (44.18s)
=== RUN   TestAccGoogleSqlDatabaseInstance_authNets
--- PASS: TestAccGoogleSqlDatabaseInstance_authNets (69.94s)
=== RUN   TestAccGoogleSqlDatabaseInstance_multipleOperations
--- PASS: TestAccGoogleSqlDatabaseInstance_multipleOperations (81.35s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	2755.451s
```

Issue #207, opened 2017-07-17
Crash when updating BigQuery Schema


jimmy-btn:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
2017/07/17 14:39:31 [INFO] Terraform version: 0.9.11
2017/07/17 14:39:31 [INFO] Go runtime version: go1.8.3

### Affected Resource(s)
Please list the resources as a list, for example:
- google_bigquery_table

### Terraform Configuration Files
```hcl
// Configure the Google Cloud provider
provider "google" {
  credentials = "${file("creds.json")}"
  project     = "..."
  region      = "us"
}

resource "google_bigquery_dataset" "default" {
  dataset_id                  = "testid"
  friendly_name               = "test"
  description                 = "This is a test description"
  location                    = "US"
  default_table_expiration_ms = 3600000

  labels {
    env = "default"
  }
}

resource "google_bigquery_table" "default" {
  dataset_id = "${google_bigquery_dataset.default.dataset_id}"
  table_id   = "tableid"

  time_partitioning {
    type = "DAY"
  }

  labels {
    env = "default"
  }

  schema = "${file("schema.json")}"
}
```

Schema (for create):
```json
[
  {
    "name": "date",
    "type": "TIMESTAMP",
    "mode": "REQUIRED",
    "description": "Time of record"
  },
  {
    "name": "event",
    "type": "STRING",
    "mode": "REQUIRED",
    "description": "This is an event"
  }
]
```

Schema (for update):
```json
[
  {
    "name": "date",
    "type": "TIMESTAMP",
    "mode": "REQUIRED",
    "description": "Time of record"
  },
  {
    "name": "event",
    "type": "STRING",
    "mode": "REQUIRED",
    "description": "This is an event"
  },
  {
    "name": "new",
    "type": "STRING",
    "mode": "NULLABLE",
    "description": "I should not be allowed to add this"
  }
]
```

### Panic Output
https://gist.github.com/jimmy-btn/6cccce524a2dd80ad98114a95cfa9c3d

### Expected Behavior
After modifying schema.json (in this case, to add the 'new' column) the BigQuery table should be updated to include the new column.

### Actual Behavior
Crashes
(Panics on conversion from {} to int64)

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
[...]

PR #209, opened 2017-07-17
Update expected type of expiration time

jimmy-btn:
schema.ResourceData encodes expiration_time as an int, while our Table object expects it as an int64. Unpack appropriately.

Resolves Issue #207


danawillow:
Thanks for the fix, @jimmy-btn!

PR #215, opened 2017-07-18
Schema Generator script

danawillow:
Provides a script that can be used to generate a schema for a new GCP resource.

This script is based on https://github.com/radeksimko/terraform-gen, but uses GCP's discovery service, which has more information about the various types than can be gathered from just the structs.

Example output for `go run ./scripts/schemagen.go -api dataproc -resource Cluster`: https://gist.github.com/danawillow/f937d1009b1ced785da7d3d09ced7e3d

```
$ go test ./scripts/
ok  	github.com/terraform-providers/terraform-provider-google/scripts	0.095s
```

@selmanj @rileykarson @rosbo @radeksimko

### DIFF
--- a/scripts/schemagen.go
+++ b/scripts/schemagen.go
@@ -87,7 +87,7 @@ func main() {
 
 	fmtd, err := format.Source(buf.Bytes())
 	if err != nil {
-		log.Fatal(err)
+		log.Printf("Formatting error: %s", err)
 	}
 
 	if _, err := f.Write(fmtd); err != nil {
--- a/scripts/schemagen.go
+++ b/scripts/schemagen.go
@@ -9,6 +9,10 @@
 // 	* No way to differentiate between fields that are/are not updateable.
 // 	* Required/Optional/Computed are set based on keywords in the description.
 //
+// Usage requires credentials. Obtain via gcloud:
+//
+//   gcloud auth application-default login
+//
 // Usage example (from root dir):
 //
 //   go run ./scripts/schemagen.go -api pubsub -resource Subscription -version v1
@@ -69,17 +73,21 @@ func main() {
 		log.Fatal(err)
 	}
 
-	fields := generateFields(resp.Schemas, *resource)
+	required, optional, computed := generateFields(resp.Schemas, *resource)
 
 	buf := &bytes.Buffer{}
 	err = googleTemplate.Execute(buf, struct {
-		TypeName string
-		Fields   map[string]string
+		TypeName  string
+		ReqFields map[string]string
+		OptFields map[string]string
+		ComFields map[string]string
 	}{
 		// Capitalize the first letter of the api name, then concatenate the resource name onto it.
 		// e.g. compute, instance -> ComputeInstance
-		TypeName: strings.ToUpper((*api)[0:1]) + (*api)[1:] + *resource,
-		Fields:   fields,
+		TypeName:  strings.ToUpper((*api)[0:1]) + (*api)[1:] + *resource,
+		ReqFields: required,
+		OptFields: optional,
+		ComFields: computed,
 	})
 	if err != nil {
 		log.Fatal(err)
@@ -95,19 +103,29 @@ func main() {
 	}
 }
 
-func generateFields(jsonSchemas map[string]discovery.JsonSchema, property string) map[string]string {
-	fields := make(map[string]string, 0)
+func generateFields(jsonSchemas map[string]discovery.JsonSchema, property string) (required, optional, computed map[string]string) {
+	required = make(map[string]string, 0)
+	optional = make(map[string]string, 0)
+	computed = make(map[string]string, 0)
 
 	for k, v := range jsonSchemas[property].Properties {
 		content, err := generateField(jsonSchemas, k, v, false)
 		if err != nil {
 			log.Printf("ERROR: %s", err)
 		} else {
-			fields[underscore(k)] = content
+			if strings.Contains(content, "Required:") {
+				required[underscore(k)] = content
+			} else if strings.Contains(content, "Optional:") {
+				optional[underscore(k)] = content
+			} else if strings.Contains(content, "Computed:") {
+				computed[underscore(k)] = content
+			} else {
+				log.Println("ERROR: Found property that is neither required, optional, nor computed")
+			}
 		}
 	}
 
-	return fields
+	return
 }
 
 func generateField(jsonSchemas map[string]discovery.JsonSchema, field string, v discovery.JsonSchema, isNested bool) (string, error) {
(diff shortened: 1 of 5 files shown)


scripts/schemagen.go
```diff
@@ -0,0 +1,277 @@
...
+		return "", err
+	}
+
+	return buf.String(), nil
+}
+
+// shamelessly borrowed from https://github.com/radeksimko/terraform-gen/blob/master/internal/util/util.go
+func underscore(name string) string {
+	var words []string
+	var camelCase = regexp.MustCompile("(^[^A-Z]*|[A-Z]*)([A-Z][^A-Z]+|$)")
```
radeksimko: FYI: One bug/caveat to be aware of is this https://github.com/radeksimko/terraform-gen/blob/master/internal/util/util_test.go#L12-L13
I didn't get to fixing this regexp to accommodate such case.

scripts/schemagen.go
```diff
@@ -0,0 +1,277 @@
...
+		// Capitalize the first letter of the api name, then concatenate the resource name onto it.
+		// e.g. compute, instance -> ComputeInstance
+		TypeName: strings.ToUpper((*api)[0:1]) + (*api)[1:] + *resource,
+		Fields:   fields,
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	err = exec.Command("gofmt", "-w", fileName).Run()
```
radeksimko: Wouldn't it be "cleaner" to do the formatting internally instead of calling out via `exec.Command`?
i.e. basically what this PR proposes: https://github.com/radeksimko/terraform-gen/pull/1/files

scripts/schemagen.go
```diff
@@ -0,0 +1,277 @@
...
+			log.Printf("ERROR: %s", err)
+		} else {
+			fields[underscore(k)] = content
+		}
+	}
+
+	return fields
+}
+
+func generateField(schemas map[string]discovery.JsonSchema, field string, v discovery.JsonSchema, isNested bool) (string, error) {
```
radeksimko: It feels a little bit confusing from my perspective to see two different variables, each called "schema". 
I guess it's technically correct name, just confusing if you come from Terraform where schema always means `github.com/hashicorp/terraform/helper/schema`.

Maybe calling this one `jsonSchemas` or something like that would help?

scripts/schemagen.go
danawillow: Yes it would, thanks for the pointer! Done.

scripts/schemagen.go
danawillow: Done.

scripts/schemagen.go
danawillow: I found a different regexp that's a bit more robust so I replaced this with that. The `externalIPs` case is a pretty extreme corner-case because of the plural, so I just hardcoded in handling of plurals.

scripts/schemagen.go
```diff
@@ -0,0 +1,277 @@
...
+)
+
+func main() {
+	api := flag.String("api", "", "api to query")
+	resource := flag.String("resource", "", "resource to generate")
+	version := flag.String("version", "v1", "api version to query")
+	flag.Parse()
+
+	if *api == "" || *resource == "" {
+		log.Fatal("usage: go run schemagen.go -api $API -resource $RESOURCE -version $VERSION")
```
danawillow: Done.

scripts/schemagen.go
```diff
@@ -0,0 +1,277 @@
...
+		}
+		if submatch[2] != "" {
+			words = append(words, submatch[2])
+		}
+	}
+
+	return strings.ToLower(strings.Join(words, "_"))
+}
+
+var schemaTemplate = template.Must(template.New("schema").Parse(`{{if .IsNested}}&schema.Schema{{end}}{{"{"}}{{if not .IsNested}}
```
danawillow: Hmm, I wouldn't say most. I think it's only for URL parameters, so it misses pretty much all of the resource object ones. It would definitely be a cool thing to have in a future version of this though.

scripts/schemagen.go
```diff
@@ -83,17 +85,21 @@ func main() {
 		log.Fatal(err)
 	}
 
-	err = exec.Command("gofmt", "-w", fileName).Run()
+	fmtd, err := format.Source(buf.Bytes())
 	if err != nil {
 		log.Fatal(err)
```
radeksimko: I can't pull up specific examples right now, but I'm pretty sure the generator may sometimes produce invalid code that will fail to format and therefore fail to be even written in a file. I think it's more desirable to actually let the generator write invalid code into the file and eventually fix it manually.

TL;DR I would turn the `log.Fatal()` here into `log.Print()` or something like that.

danawillow: Done

scripts/schemagen.go
```diff
@@ -0,0 +1,281 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fields := generateFields(resp.Schemas, *resource)
+
+	buf := &bytes.Buffer{}
+	err = googleTemplate.Execute(buf, struct {
+		TypeName string
+		Fields   map[string]string
```
rileykarson: We print fields in whatever order Go wants, but that's not how we tend to structure them in hand-written files. This is probably out of scope for this version of the generator, but what if we ordered it so that `Required` fields appeared first, followed by `Optional`, and finally `Computed`?

scripts/schemagen.go
```diff
@@ -0,0 +1,281 @@
...
+// This script draws heavily from https://github.com/radeksimko/terraform-gen,
+// but uses GCP's discovery API instead of the struct definition to generate
+// the schemas.
+//
+// This is not meant to be a definitive source of truth for resource schemas,
+// just a starting point. It has some notable deficiencies, such as:
+// 	* No way to differentiate between fields that are/are not updateable.
+// 	* Required/Optional/Computed are set based on keywords in the description.
+//
+// Usage example (from root dir):
```
rileykarson: Can you add a note that users should use Google default credentials derived from `gcloud`?

scripts/schemagen.go
radeksimko: We don't really have a convention around this, but it would be great to have one 👍  I like the idea. Conventions I've seen applied to existing resources include either what you described (Required -> Optional -> Computed) or alphabetical order.

scripts/schemagen.go
danawillow: Done

scripts/schemagen.go
danawillow: Done

scripts/schemagen.go
```diff
@@ -0,0 +1,281 @@
...
+		s.Elem = elem
+	default:
+		return "", fmt.Errorf("Unable to process: %s %s", field, v.Type)
+	}
+
+	return schemaCode(s, isNested)
+}
+
+func setProperties(v discovery.JsonSchema, s *schema.Schema) {
+	if v.ReadOnly || strings.HasPrefix(v.Description, "Output-only") {
```
danawillow: I added `[Output Only]` as an option, but I'm not going to go through the docs now and check for all possibilities (especially since many APIs don't even document required/optional/output only)

PR #217, opened 2017-07-19
Fix bug where range variable is improperly dereferenced

selmanj:


danawillow:
Thanks for the fix! merge at will.

PR #220, opened 2017-07-19
google_compute_ssl_certificate: mark private_key as sensitive

renchap:
This prevents the private key to be displayed in `terraform plan` and `terraform apply`.
It does not impact anything else.

I considered marking `certificate` sensitive as well to avoid displaying a full certificate (+ intermediates) as it clutters the output and does not bring very readable information, but as it is not really sensitive data I did not do it. Feel free to tell me if you think this is a good idea.


selmanj:
Thanks for this change! I'm not eager to suppress certificate as sensitive just because it's verbose, but this private key labeling is absolutely correct.

PR #183, opened 2017-07-10
Add charset and collation to google_sql_database.

ubschmidt2:
Fixes https://github.com/terraform-providers/terraform-provider-google/issues/179.

Issue #179, opened 2017-07-07
Feature: Support setting charset and collation for google_sql_database resource


ryancurrah:
Per the Google API docs for creating a [CloudSQL database](https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/databases) it is possible to provide the `charset` and `collation` to the API. My suggestion is to be able to provide these options to the Terraform [google_sql_database](https://www.terraform.io/docs/providers/google/r/sql_database.html) provider.

Example input params...

```
resource "google_sql_database_instance" "master" {
  name = "master-instance"

  settings {
    tier = "D0"
  }
}

resource "google_sql_database" "users" {
  name      = "image-store-bucket"
  instance  = "${google_sql_database_instance.master.name}"
  charset   = "utf8"
  collation = "utf8_bin"
}
```

ryancurrah:
Woo hoo!

PR #229, opened 2017-07-20
Mark google_sql_database.{charset,collation} as computed instead of having defaults

ubschmidt2:
This change is required to avoid the following scenario:
When upgrading from a previous version of the Google provider, TF will change
the charset/collation of existing (TF-managed) databases to utf8/utf8_general_ci
(if the user hasn't added different config values before running TF apply),
potentially overriding any non-default settings that the user my have applied
through the Cloud SQL admin API. This violates POLA.

This is an addendum to PR #183 and issue #179.

### DIFF
--- a/website/docs/r/sql_database.html.markdown
+++ b/website/docs/r/sql_database.html.markdown
@@ -44,10 +44,9 @@ The following arguments are supported:
 * `project` - (Optional) The project in which the resource belongs. If it
     is not provided, the provider project is used.
 
-* `charset` - (Optional) The MySQL charset value (default "utf8").
+* `charset` - (Optional) The MySQL charset value.
 
-* `collation` - (Optional) The MySQL collation value (default
-    "utf8_general_ci").
+* `collation` - (Optional) The MySQL collation value.
 
 ## Attributes Reference
 
--- a/website/docs/r/sql_database.html.markdown
+++ b/website/docs/r/sql_database.html.markdown
@@ -44,9 +44,19 @@ The following arguments are supported:
 * `project` - (Optional) The project in which the resource belongs. If it
     is not provided, the provider project is used.
 
-* `charset` - (Optional) The MySQL charset value.
+* `charset` - (Optional) The charset value. See MySQL's [Supported Character
+    Sets and
+    Collations](https://dev.mysql.com/doc/refman/5.6/en/charset-charsets.html)
+    and PostgreSQL's [Character Set
+    Support](https://www.postgresql.org/docs/9.6/static/multibyte.html)
+    for more details and supported values.
 
-* `collation` - (Optional) The MySQL collation value.
+* `collation` - (Optional) The collation value. See MySQL's [Supported Character
+    Sets and
+    Collations](https://dev.mysql.com/doc/refman/5.6/en/charset-charsets.html)
+    and PostgreSQL's [Collation
+    Support](https://www.postgresql.org/docs/9.6/static/collation.html) for
+    more details and supported values.
 
 ## Attributes Reference
 
--- a/website/docs/r/sql_database.html.markdown
+++ b/website/docs/r/sql_database.html.markdown
@@ -46,14 +46,14 @@ The following arguments are supported:
 
 * `charset` - (Optional) The charset value. See MySQL's [Supported Character
     Sets and
-    Collations](https://dev.mysql.com/doc/refman/5.6/en/charset-charsets.html)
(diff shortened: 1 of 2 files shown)


ubschmidt2:
PLKM if this makes sense. I'm not sure how this could be properly tested for in an acceptance test.

website/docs/r/sql_database.html.markdown
```diff
@@ -44,10 +44,19 @@ The following arguments are supported:
 * `project` - (Optional) The project in which the resource belongs. If it
     is not provided, the provider project is used.
 
-* `charset` - (Optional) The MySQL charset value (default "utf8").
-
-* `collation` - (Optional) The MySQL collation value (default
-    "utf8_general_ci").
+* `charset` - (Optional) The charset value. See MySQL's [Supported Character
+    Sets and
+    Collations](https://dev.mysql.com/doc/refman/5.6/en/charset-charsets.html)
```
ubschmidt2: Done.

paddycarver:
My 2 cents on the update problem:

* Filing an upstream bug is definitely where I'd start with this.
* We should similarly file a bug for (what sounds like?) our inability to tell whether it's MySQL or Postgres
* My concern with the `Create` workaround is how it will work with import. If you can successfully create a database outside of Terraform (e.g., through the console) and then can import it, and the type still gets set correctly, I'm happy.
* Silly question, but there's no way for the database to _change_ type, right? Otherwise, `Create` is really the wrong spot for it, as we need to detect drift _somehow_.

I think if we can't get a workaround working without too much hacky behaviour, not allowing it for MySQL is better than silently failing in Postgres.

rileykarson:
This is all the context I have on what's going on with these values in Postgres right now, + with the Cloud SQL documentation for it.  Feel free to correct anything I got wrong; I don't know what I want to do to fix how these fields work for it yet.

There exist two side by side api docs; [MySQL](https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/databases) and [Postgres](https://cloud.google.com/sql/docs/postgres/admin-api/v1beta4/databases) that are essentially the same, and describe the exact same API client; they are written for interacting with MySQL databases and not Postgres.

This becomes really clear when looking at [Creating and Managing PostgreSQL Databases](https://cloud.google.com/sql/docs/postgres/create-manage-databases#create), where it's specified that when creating a database in Cloud Console;

> In the New database dialog, specify the name of the database, and optionally the character set and collation.

This operation is impossible as you can only specify these values [charset and collation] for a MySQL database in the Cloud Console UI; they are automatically always `UTF8` and `en_US.UTF8` for Postgres.

Why becomes clear when `Create`-ing a database in a Postgres instance with custom values; even though `charset` becomes the proper Postgres value, `ENCODING`, and `collation` becomes `LC_COLLATE`, it still doesn't work. The issue is that the creation script is:

```sql
CREATE DATABASE "db-name" WITH OWNER cloudsqlsuperuser ENCODING `UTF8` LC_COLLATE 'en_GB.UTF8';
```

Because we didn't explicitly specify a "template" database, Postgres implicitly uses `template1`, a template that has it's charset and collation set to the same as the system defaults. These are `UTF8` and `en_US.UTF8`. There exists another default template, `template0`, which has no charset/collation, and would work. But since we used `template1` and don't match the template, the `CREATE` fails. The error on the Go side has no text so users won't know what's wrong if they encounter this error, so we would need to make an error ourselves.

`Read` works as expected, reflecting the correct value from each type of database, as does `Delete`.

`Update` for Postgres does nothing; the `Operation` reports success, and no changes are made. So we will never converge if the value is changed if the fields are not `ForceNew`. There should be an API-side error here, because changing these values for Postgres has a different meaning than for MySQL. (see #260 for context on that)

ubschmidt2:
PTAL. I think it would be worthwhile to merge this PR. It's an improvement for MySQL, but doesn't address the larger PostgreSQL issues yet.

rileykarson:
Would you mind adding a note that Postgres doesn't support a charset other than `UTF8` and a collation other than `en_US.UTF8` at `Create`?

Otherwise this LGTM - we don't have a strong case for preferring `Default` over `Computed`, especially when there can be two values like in this case.

ubschmidt2:
Apologies for the delay. Now back on this. I've added a small note about the restriction for PostgreSQL databases.

ubschmidt2:
Sent upstream FR for supporting user-specified charset and collation values for Cloud SQL PostgreSQL databases.

Issue #190, opened 2017-07-13
Feature: Support for Google BigQuery Views (google_bigquery_view)


poiuytrez:
I would be interested to have BigQuery Views support in Terraform. 
https://cloud.google.com/bigquery/docs/views

PR #230, opened 2017-07-21
Support Bigquery Views

jimmy-btn:
Extend the Google provider for Terraform to support Bigquery Views.

Make a few drive by cleanups in the process:

- Amend the Basic Table test to actually verify that a valid table has been created
- Support modifying Table Schema by parsing computed IDs correctly
- Assign a default location to Tables created before Bigquery supported two zones


google/resource_bigquery_dataset.go
```diff
@@ -238,6 +238,12 @@ func resourceBigQueryDatasetRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("dataset_id", res.DatasetReference.DatasetId)
 	d.Set("default_table_expiration_ms", res.DefaultTableExpirationMs)
 
+	// Older Tables in BigQuery have no Location set in the API response. We can safely assume that these tables
```
danawillow: This is only going to come up during import, right? If so I wouldn't mind a comment that mentions that just to be extra clear. Also mind moving the other d.Set for location into an else with this clause?

google/resource_bigquery_table.go
```diff
@@ -92,6 +92,23 @@ func resourceBigQueryTable() *schema.Resource {
...
 			},
 
+			// View: [Experimental] If specified, configures this table as a view.
+			"view": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						// Type: [Required] The only type supported is DAY, which will generate
```
danawillow: I think this comment was accidentally pasted from a different one

google/resource_bigquery_table.go
```diff
@@ -92,6 +92,23 @@ func resourceBigQueryTable() *schema.Resource {
 				},
 			},
 
+			// View: [Experimental] If specified, configures this table as a view.
+			"view": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
```
danawillow: Do you think it makes sense to add the other `view` attributes, or do they not make sense in terraform? (useLegacySql, userDefinedFunctionResources)

google/resource_bigquery_table.go
```diff
@@ -202,12 +219,17 @@ func resourceTable(d *schema.ResourceData, meta interface{}) (*bigquery.Table, e
 		},
 	}
 
+	// TODO(jmcgill): How do I do nested objects in a terraform configuration.
```
danawillow: do you still need this TODO?

google/resource_bigquery_table_test.go
```diff
@@ -64,11 +85,40 @@ func testAccBigQueryTableExists(n string) resource.TestCheckFunc {
 			return fmt.Errorf("No ID is set")
 		}
 		config := testAccProvider.Meta().(*Config)
-		_, err := config.clientBigQuery.Tables.Get(config.Project, rs.Primary.Attributes["dataset_id"], rs.Primary.Attributes["name"]).Do()
+		table, err := config.clientBigQuery.Tables.Get(config.Project, rs.Primary.Attributes["dataset_id"], rs.Primary.Attributes["table_id"]).Do()
```
danawillow: Good catch, thanks! Want to fix that up in `testAccCheckBigQueryTableDestroy` too?

danawillow:
For tracking purposes: this fixes #190

google/resource_bigquery_dataset.go
jimmy-btn: Great idea - Done!

google/resource_bigquery_table_test.go
jimmy-btn: Done!

google/resource_bigquery_table.go
```diff
@@ -394,3 +428,24 @@ func flattenTimePartitioning(tp *bigquery.TimePartitioning) []map[string]interfa
 
 	return []map[string]interface{}{result}
 }
+
+func expandView(configured interface{}) *bigquery.ViewDefinition {
+	raw := configured.([]interface{})[0].(map[string]interface{})
+	vd := &bigquery.ViewDefinition{Query: raw["query"].(string)}
+
+	if v, ok := raw["use_legacy_sql"]; ok {
+		vd.UseLegacySql = v.(bool)
```
danawillow: You'll also need to set `vd.ForceSendFields = append(vd.ForceSendFields, "UseLegacySql"` so if it gets updated to false, it still gets sent along the wire (since false values end up defaulting to empty)

google/resource_bigquery_table.go
```diff
@@ -394,3 +428,24 @@ func flattenTimePartitioning(tp *bigquery.TimePartitioning) []map[string]interfa
...
+	}
+
+	return vd
+}
+
+func flattenView(vd *bigquery.ViewDefinition) []map[string]interface{} {
+	result := map[string]interface{}{"query": vd.Query}
+
+	if vd.UseLegacySql {
```
danawillow: Fun fact: this already actually gets stored in state whether it's true or false because of the way terraform handles objects, so we may as well just simplify the logic here and always set it anyway.

google/resource_bigquery_table.go
jimmy-btn: Good catch - thank you!

Would you like me to add an end to end test for this particular case (UseLegaclySQL = false)? Running these AF tests costs $$$ so I wasn't sure on how to balance coverage vs cost.

google/resource_bigquery_table.go
jimmy-btn: Done! Should I skip the check on line 436 (in expandView) too?

jimmy-btn: Actually nevermind; I need to expand the value anyway. That extra error check is harmless.

google/resource_bigquery_table.go
danawillow: Sure! You'd want to do a test specifically for updating it, since the default is false so if you set it to false explicitly on create it won't make a difference whether you forced it to send. If cost is an issue for you let me know and I'm happy to run it and post the results.

jimmy-btn: Perfect! Tests added for converting useLegacySQL from true to false - The API is kind enough to reject views with incorrect syntax when using new-SQL so I am now extra confident that this is working as expected!

PR #234, opened 2017-07-24
Add versioned Beta support to google_compute_instance_group_manager

rileykarson:
Split of #198 - Part 1/2

Add support for multiple versions to `google_compute_instance_group_manager`, by adding simultaneous support for `v1` and `v0beta`. No fields are `v0beta` yet, so it will not enter that state. Setting `InstanceGroupManagerBaseVersion` to `v0beta` instead of `v1` will allow you to test the beta code.

This is the first versioned resource for #93 

@danawillow


rileykarson:
I just caught an issue that wasn't failing before and is now; `ForceSendFields` is lost when converting between versions, and `target_size` needs to be force sent if it has a `0` value. The rest of my local tests are still running but I expect them to pass, `target_size` was updated on master and had to be merged.

The last 2 commits still need to be reviewed, the rest were covered in past pull requests.

rileykarson:
Just finished my local test run and all of them are passing now.

rileykarson:
Hi @catsby!

I'm not sure where this kind of documentation would live yet; if you have a good idea of where it should go, I'd love the advice. It doesn't seem to fit with the [main provider page](https://www.terraform.io/docs/providers/google/index.html), which is about authentication, for example.  For now, to go over your questions; 

### 1 - Enrolling in Beta Features
Users of the Terraform GCP Provider will enroll in Beta features implicitly by using fields that are marked Beta in the documentation. Resources that are considered to be in Beta will use the v0beta endpoints of the Compute API, instead of the v1 endpoints.

While this has not been realised yet, some resources may begin supporting the Update operation only in the Beta API for fields that are otherwise GA. When this is the case, or if it has explicit demand, support for a provider-level “GA-only” field will be added that a user can set to cause Terraform to return an error if it would have used a non-GA Compute API (in case the customer has a strict need for GA-only features/APIs).

When the support for custom diffs has evolved in Terraform’s core/helper libraries, we will investigate populating diffs with information about what API is being used.

### 2 - Beta Feature Reliability
GCP does not offer SLAs for features in Beta; they are not necessarily going to be as reliable as GA resources. As highlighted in the [GCP Launch Stages documentation](https://cloud.google.com/terms/launch-stages), Beta releases tend to be feature-complete but may be more unstable than a release in GA, with some open outstanding issues still present.

In practice, this should mean that Beta releases are stable & reliable releases, especially if they have been in Beta for some amount of time.

### 3 - Feature Promotion
When a feature is promoted to GA from Beta, we will remove the “Beta” tag in documentation and remove the field from the list of Beta fields in code. The next time Terraform performs a refresh, it will implicitly be read as a GA resource, and operations will be performed against GA APIs again.

### 4 - Breaking Changes
Breaking changes are less of a concern for Beta releases from GCP than the label Beta might suggest; as shown in Launch Stages, any Beta release will be feature complete and should not surface breaking changes on the path to GA outside of extraordinary circumstances. Beta releases are the time that a GCP feature has entered the public eye.

While we don’t present an SLA or deprecation policy for Beta features, that doesn’t necessarily mean that they will go without a deprecation period. For example, Managed Instance Group Updates, an Alpha feature, has gone through a deprecation period beginning in September 2016 that is still ongoing: from [rolling updates to managed instance groups](https://cloud.google.com/compute/docs/instance-groups/creating-groups-of-managed-instances#starting_an_update) to [updating managed instance groups](https://cloud.google.com/compute/docs/instance-groups/updating-managed-instance-groups).

Major breaking changes for Beta features can be handled on a case by case basis; if a feature is removed entirely, we will need to deprecate and remove it quickly, generating a major version out of schedule if necessary. If a feature is renamed/replaced, we will prioritise adding support for the new version, deprecating the old one, and removing it when appropriate.

google/resource_compute_instance_group_manager.go
```diff
@@ -121,10 +127,27 @@ func getNamedPorts(nps []interface{}) []*compute.NamedPort {
...
+		namedPorts = append(namedPorts, &computeBeta.NamedPort{
+			Name: np["name"].(string),
+			Port: int64(np["port"].(int)),
+		})
+	}
+
+	return namedPorts
+}
+
+var InstanceGroupManagerBaseVersion = v1
```
catsby: Can we bump this up to above the schema declaration? It's kind of tucked away here on line 147, all by itself. 

Also, but not required, I'd name it something that describes it's function/intent better, so specifically calling out that we're setting the base API version that the resource is talking to. Without getting too wordy, something like `igmBaseAPIVersion`. Not required though.

rileykarson: Done.

rileykarson:
@catsby: Will do!

Tests:
```bash
TF_ACC=1 go test ./google -v -run=TestAccInstanceGroupManager_ -timeout 120m
=== RUN   TestAccInstanceGroupManager_importBasic
--- PASS: TestAccInstanceGroupManager_importBasic (265.09s)
=== RUN   TestAccInstanceGroupManager_importUpdate
--- PASS: TestAccInstanceGroupManager_importUpdate (212.84s)
=== RUN   TestAccInstanceGroupManager_basic
--- PASS: TestAccInstanceGroupManager_basic (221.84s)
=== RUN   TestAccInstanceGroupManager_targetSizeZero
--- PASS: TestAccInstanceGroupManager_targetSizeZero (45.23s)
=== RUN   TestAccInstanceGroupManager_update
--- PASS: TestAccInstanceGroupManager_update (277.79s)
=== RUN   TestAccInstanceGroupManager_updateLifecycle
--- PASS: TestAccInstanceGroupManager_updateLifecycle (105.11s)
=== RUN   TestAccInstanceGroupManager_updateStrategy
--- PASS: TestAccInstanceGroupManager_updateStrategy (211.10s)
=== RUN   TestAccInstanceGroupManager_separateRegions
--- PASS: TestAccInstanceGroupManager_separateRegions (212.83s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	1552.027s
```

Issue #211, opened 2017-07-18
google_compute_instance resource does not read scheduling field


selmanj:
### Terraform Version
Terraform v0.10.0-dev (be900e808506efb018bd6adbe457cfcbaf6281a9)

### Affected Resource(s)
- google_compute_instance

### Terraform Configuration Files
```hcl
provider "google" {
  credentials = "${file("~/google/terraform-devel.json")}"
  project     = "terraform-devel"
  region      = "us-central1"
}
resource "google_compute_instance" "default" {
  name         = "sometestname"
  machine_type = "n1-standard-1"
  zone         = "us-central1-a"

  disk {
    image = "debian-cloud/debian-8"
  }

  disk {
    type    = "local-ssd"
    scratch = true
  }

  network_interface {
    network = "default"

    access_config {
    }
  }


  scheduling {
    automatic_restart = true
  }
}
```

### Expected Behavior
Terraform should notice changes to options underneath scheduling and enforce them.

### Actual Behavior
Terraform will apply changes made to the config file but never reads values from GCP.

### Steps to Reproduce
1. `terraform apply`
2. Go into Cloud console and change `automatic_restart` behavior to off or false
3. `terraform plan` shows no changes, but it should show changes needed to `automatic_restart`.

PR #237, opened 2017-07-24
google_compute_instance reads `scheduling` fields from GCP.

rosbo:
issue #211 

Marked `preemptible` as ForceNew because this field can only be set at the creation of an instance.
Reads `scheduling` fields from GCP. Make sure this does not introduce a diff.
There should be at most one `scheduling` block. Enforcing it with MaxItems=1

cc/ @selmanj @danawillow

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -378,22 +378,39 @@ func resourceComputeInstance() *schema.Resource {
 
 			"scheduling": &schema.Schema{
 				Type:     schema.TypeList,
+				MaxItems: 1,
 				Optional: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					// The `scheduling` block is optional in the config but implicitly has a length
+					// of 1 with the default values for `automatic_restart` and `preemptible`.
+					// This suppress the diff `scheduling.#: "1" => "0"` when the `scheduling` block
+					// is not specified by the user.
+					if k == "scheduling.#" {
+						return true
+					}
+
+					// Don't suppress the diff for the scheduling fields.
+					return false
+				},
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"on_host_maintenance": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
+							Computed: true,
 						},
 
 						"automatic_restart": &schema.Schema{
 							Type:     schema.TypeBool,
 							Optional: true,
+							Computed: true,
 						},
 
 						"preemptible": &schema.Schema{
 							Type:     schema.TypeBool,
 							Optional: true,
+							Default:  false,
+							ForceNew: true,
 						},
 					},
 				},
@@ -1051,6 +1068,9 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("disk", disks)
 	d.Set("attached_disk", attachedDisks)
 	d.Set("scratch_disk", scratchDisks)
+
+	scheduling, _ := flattenScheduling(instance.Scheduling)
+	d.Set("scheduling", scheduling)
 
 	d.Set("self_link", instance.SelfLink)
 	d.SetId(instance.Name)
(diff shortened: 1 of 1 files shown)


rosbo:
I realized that the default value for `automatic_restart` depends also on whether an instance is preemptible or not. Please wait before reviewing.

rosbo:
This PR is now ready for review. I am now handling all the `scheduling` fields.

google/resource_compute_instance.go
```diff
@@ -378,22 +378,39 @@ func resourceComputeInstance() *schema.Resource {
...
 			"scheduling": &schema.Schema{
 				Type:     schema.TypeList,
+				MaxItems: 1,
 				Optional: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					// The `scheduling` block is optional in the config but implicitly has a length
+					// of 1 with the default values for `automatic_restart` and `preemptible`.
+					// This suppress the diff `scheduling.#: "1" => "0"` when the `scheduling` block
+					// is not specified by the user.
+					if k == "scheduling.#" {
```
rosbo: Not exactly,

If the user don't add the `scheduling` block to the config because they are fine with the default values. Then, the desired state as scheduling.# = 0.

But the current state gets refreshed (calls the read method), the populates the values for the current state scheduling. Hence, `scheduling.#` will always be 0 for the current state.

This means a user gets a perpetual diff if they don't specify the scheduling block (which is an okay thing to do if you are fine with the default values):
 `scheduling.#: "1" => "0"`

google/resource_compute_instance.go
```diff
@@ -378,22 +378,39 @@ func resourceComputeInstance() *schema.Resource {
...
 					Schema: map[string]*schema.Schema{
 						"on_host_maintenance": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
+							Computed: true,
 						},
 
 						"automatic_restart": &schema.Schema{
 							Type:     schema.TypeBool,
 							Optional: true,
```
rosbo: `automatic_restart` default value depends on the `preemptible` field value. DefaultFunc cannot look at the value of another field to determine its default value.

The Optional & Computed means that unless the user specify it in the config, take the GCP value.

google/resource_compute_instance.go
danawillow: Is there a possibility of this over-suppressing? Like if an element is added/removed/changed but `scheduling.#` is still 1? Why not `Computed`?

google/resource_compute_instance.go
danawillow: yeah, optional+computed is generally fine but booleans get weird since we can't tell the difference between `false` and unset. We need the user to be able to set `automatic_restart` to false, so we `ForceSendFields` it, which means that we'll always be providing a value to GCP and the computedness of it doesn't really end up mattering.

google/resource_compute_instance.go
rosbo: Thanks. Computed true on the TypeList itself effectively does the same thing. This is the elegant solution I was looking for.

google/resource_compute_instance.go
rosbo: Done

google/resource_compute_instance.go
```diff
@@ -1160,14 +1170,11 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
 		if val, ok := d.GetOk(prefix + ".automatic_restart"); ok {
 			scheduling.AutomaticRestart = googleapi.Bool(val.(bool))
 		}
-
-		if val, ok := d.GetOk(prefix + ".preemptible"); ok {
-			scheduling.Preemptible = val.(bool)
-		}
-
+		// The Preemptible field cannot be updated. The instance must be recreated.
```
danawillow: Hmm, so I think the following scenario would fail:
- preemptible is true
- I want to change one of the other attributes
because I think this would try to update preemptible to false.

However, we can't actually change either of the other attributes when preemptible is true- it would fail server-side. So two options for fixing this:
1. put scheduling.Preemptible back in the request- we know that we won't hit this if it changes since it's marked `ForceNew`, so this is just to propagate the current value
2. throw an error when trying to update other parts of scheduling if preemptible is true

I think #1 is clearer in the code and more future-proof, so I recommend that one- up to you though.

rosbo: I did suggestion `#1`

Issue #218, opened 2017-07-19
google_compute_disk created from snapshot, forces new resource then apply once more


z1nkum:
Looks like google_compute_disk was broken since v0.9.6 (or v0.9.7)

Create disk snapshot and then disk from this snapshot (clone disk operation)
If we execute `terraform apply` once more newly created disk will be destroyed and recreated

### Terraform Version
Terraform v0.9.11

### Affected Resource(s)
google_compute_disk

### Terraform Configuration Files
```hcl
resource "google_compute_snapshot" "db-shard1-4-root-snapshot" {
  name  = "db-shard1-4-root-snapshot"
  source_disk  = "db-shard1-4"
  zone  = "${google_compute_instance.db-shard1-4.zone}"
}

resource "google_compute_disk" "db-shard1-4-root-disk-from-snapshot" {
  name  = "db-shard1-4-root-cloned-disk-${count.index + 1}"
  type  = "pd-standard"
  zone  = "${var.zone-chooser[count.index]}"
  snapshot = "${google_compute_snapshot.db-shard1-4-root-snapshot.name}"
  count = 1
}
```

### Debug Output
```
-/+ google_compute_disk.db-shard1-4-root-disk-from-snapshot
    disk_encryption_key_sha256: "" => "<computed>"
    name:                       "db-shard1-4-root-cloned-disk-1" => "db-shard1-4-root-cloned-disk-1"
    self_link:                  "https://www.googleapis.com/compute/v1/projects/projectnamehere/zones/us-east1-d/disks/db-shard1-4-root-cloned-disk-1" => "<computed>"
    size:                       "10" => "0"
    snapshot:                   "https://www.googleapis.com/compute/v1/projects/projectnamehere/global/snapshots/db-shard1-4-root-snapshot" => "db-shard1-4-root-snapshot" (forces new resource)
    type:                       "pd-standard" => "pd-standard"
    users.#:                    "0" => "<computed>"
    zone:                       "us-east1-d" => "us-east1-d"


Plan: 1 to add, 0 to change, 1 to destroy.
```

### Expected Behavior
No changes expected on 2nd and others "terraform apply"

### Actual Behavior
Terraform decide to recreate disks (why??)

### Steps to Reproduce
List the steps required to reproduce the issue:
Execute `terraform apply`
And once more: `terraform apply`

nodesocket:
[...]

PR #238, opened 2017-07-25
Fix google_compute_disk created from snapshot, forces new resource then apply once more

z1nkum:
fix issue with recreation loop:
https://github.com/terraform-providers/terraform-provider-google/issues/218

### DIFF
--- a/google/resource_compute_disk.go
+++ b/google/resource_compute_disk.go
@@ -55,9 +55,9 @@ func resourceComputeDisk() *schema.Resource {
 			},
 
 			"image": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+				Type:             schema.TypeString,
+				Optional:         true,
+				ForceNew:         true,
 				DiffSuppressFunc: linkDiffSuppress,
 			},
 
@@ -79,9 +79,9 @@ func resourceComputeDisk() *schema.Resource {
 			},
 
 			"snapshot": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+				Type:             schema.TypeString,
+				Optional:         true,
+				ForceNew:         true,
 				DiffSuppressFunc: linkDiffSuppress,
 			},
 
--- a/google/resource_compute_disk.go
+++ b/google/resource_compute_disk.go
@@ -58,6 +58,7 @@ func resourceComputeDisk() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
+				DiffSuppressFunc: linkDiffSuppress,
 			},
 
 			"project": &schema.Schema{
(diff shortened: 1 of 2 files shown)


z1nkum:
@paddycarver please, take a look. Maybe I did not quite understand the idea.

z1nkum:
I really tried to figure out how to trigger and then catch ForceNew event in tests - but unsuccessfully
Test `TestAccComputeDisk_fromSnapshotURI` (_google/resource_compute_disk_test.go_) looks like suitable for this case. Perhaps we should add here one more `resource.TestStep` with the same config. But I can not imagine what should be inside `ComposeTestCheckFunc` test :(
Could you please help me with it, @paddycarver ?

I've already built provider with fix and check against config (#218) - works fine (at least, does not attempt to recreate the disk)

paddycarver:
Hey @z1nkum, absolutely no worries. There are a lot of tools in the test framework, and it's not documented as well as it could be.

In this case, I don't think the bug is that it's ForceNew--and that doesn't look like it's changing in your code. So I'd forget about trying to verify the ForceNew aspect of it, and rather just try to verify that there isn't a diff after the apply.

Fortunately, the test framework does that for us automatically; it will run another plan after all your test steps are run, and if there's a diff, it fails the test. So a good way to test this would be to use the snapshot name (the input that was causing the problem) in a test. If the test passes, things are good, and that's all it takes! If you want to be extra sure, run the test against the master branch (without the fix applied) and make sure it fails, with something like "non-empty plan" or "non-empty diff". Then you know you've recreated the problem, and fixed it.

You could even copy & paste the current snapshot test (https://github.com/terraform-providers/terraform-provider-google/blob/ff8bdc9b52f379b7cb42a2c2fe6dcb49e0c130c3/google/resource_compute_disk_test.go#L63-L85) and update the copy to use just the name, instead of the full URL. That would demonstrate that the problem was resolved. And having it automated means if the API ever changes, or we ever break Terraform's framework in a way that breaks this, or we ever introduce a regression, we'll know about it. :)

Thanks for all the hard work!

z1nkum:
Thank you, @paddycarver - I've got it.

I wrote test (actually, update _TestAccComputeDisk_fromSnapshotURI_) and tested it against master/branch. As expected, test passed on branch and fail on master:
```
--- FAIL: TestAccComputeDisk_fromSnapshot (152.56s)
        testing.go:428: Step 1 error: After applying this step, the plan was not empty:
```

z1nkum:
😱  sorry, I just wanted to rebase branch from upstream/master

paddycarver:
No worries! I've created a fixed branch (`terraform-providers:paddy_z1nkum_238`) that is up-to-date with master. You can: 

* pull that, mirror it to your repo, and use it as the basis of the PR
* pull that, force push it to the `z1nkum:snapshot-fix-url-to-name` branch you opened this PR from, which will update this PR
* just redo the steps I did to make that branch on your branch and force push it, which will update this PR:
  1. `git checkout snapshot-fix-url-to-name`
  2. `git checkout d0b7fbc`
  3. `git merge master`
  4. `git push -f`

Note that using the branch I created or the steps above will both delete anything you've done on the branch since d0b7fbc, which (as far as I can tell) is the last intentional commit. :)

z1nkum:
Thanks a lot! Now it seems all right

PR #240, opened 2017-07-25
Google API multiversioned service generator.

rileykarson:
In #187 I proposed a refactoring of `Operation` (split to #191), and introducing a versioned Compute service that simplifies the logic in the resource itself. This PR introduces a generator for that service; it's able to handle `Create`, `Read`, `Delete`, and full `Update` methods for resources; it does not generate partial update methods.

The full service & usage can be seen in #187, especially the [service file itself](https://github.com/rileykarson/terraform-provider-google/blob/80a0441aea1871b641dada38038e289b481539a1/google/compute_multiversion_service.go).

These commands can be used to show what the tool generates;

```bash
go run scripts/generate_service.go -type "Address" -scope "region"
```

```bash
go run scripts/generate_service.go -type "InstanceGroupManager" -scope "region" -updateversion "v0beta"
```

```bash
go run scripts/generate_service.go -type "BetaOnlyResource" -scope "region" -updateversion "v0beta" -lowestversion "v0beta"
```

```bash
go run scripts/generate_service.go -type "ResourceWithInstanceParent" -scope "region" -parent "instance"
```

@danawillow @selmanj @evandbrown @rosbo

### DIFF
--- a/scripts/generate_service.go
+++ b/scripts/generate_service.go
@@ -91,7 +91,7 @@ func main() {
 	}
 
 	// This is the information we would expect users to provide
-	// TODO: Support higher versions than v1, support Update, support more extra params
+	// TODO: investigate whether we need more params than scope+parent
 	normalisedData := &NormalisedData{
 		Type:          *clientTypeFlag,
 		LowestVersion: lowestVersion,
--- a/scripts/generate_service.go
+++ b/scripts/generate_service.go
@@ -1,0 +1,305 @@
+package main
+
+import (
+	"bytes"
+	"flag"
+	"fmt"
+	"go/format"
+	"log"
+	"os"
+	"strings"
+	"text/template"
+)
+
+type ComputeApiGenVersion int
+
+const (
+	none ComputeApiGenVersion = iota
+	v1
+	v0beta
+)
+
+var OrderedComputeApiVersions = []ComputeApiGenVersion{
+	v0beta,
+	v1,
+}
+
+type VersionInfo struct {
+	ClientNameSuffix string
+	ServiceName      string
+}
+
+type NormalisedData struct {
+	Type          string
+	LowestVersion ComputeApiGenVersion
+	UpdateVersion ComputeApiGenVersion
+	ExtraParams   []string
+}
+
+type TemplateData struct {
+	// Name for the resource in the new service API
+	ResourceName string
+
+	// Type of the resource object in the client. Generally but not always the ResourceName.
+	ClientType string
+
+	// Name of the Service of the client. Generally but not always the ResourceName pluralised.
+	ServiceType string
+
+	// Versions supported by this resource.
+	Versions []VersionInfo
+
+	// Versions supported by this resource for update.
+	UpdateVersions []VersionInfo
+
+	// Parameters to add to signatures/calls denormalised.
+	ExtraParamsSignature string
+	ExtraParamsCall      string
+}
+
+func main() {
+	clientTypeFlag := flag.String("type", "", "resource to generate a service for.")
+	scopeFlag := flag.String("scope", "", "scope if non-global. one of `region` or `zone`.")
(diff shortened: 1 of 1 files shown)


scripts/generate_service.go
```diff
@@ -0,0 +1,305 @@
...
+	"strings"
+	"text/template"
+)
+
+type ComputeApiGenVersion int
+
+const (
+	none ComputeApiGenVersion = iota
+	v1
+	v0beta
```
paddycarver: I think it may make sense to prefix these with `apiv`-- `apivNone` `apiv1`, `apiv0beta`, etc.--just to avoid claiming identifiers that may be shadowed within functions on accident.

scripts/generate_service.go
```diff
@@ -0,0 +1,305 @@
...
+)
+
+var OrderedComputeApiVersions = []ComputeApiGenVersion{
```
paddycarver: If the constants aren't ordered, we should avoid using iota. If the constants are ordered, I don't understand the need for the slice. :)

scripts/generate_service.go
```diff
@@ -0,0 +1,305 @@
...
+	clientTypeFlag := flag.String("type", "", "resource to generate a service for.")
+	scopeFlag := flag.String("scope", "", "scope if non-global. one of `region` or `zone`.")
+	parentFlag := flag.String("parent", "", "name of parent resource, if any.")
+	lowestVersionFlag := flag.String("lowestversion", "v1", "lowest version of the resource. one of `v1` or `v0beta`. defaults to v1.")
+	updateVersionFlag := flag.String("updateversion", "", "version at which update is supported. one of `v1` or `v0beta`.")
+	flag.Parse()
+
+	if *clientTypeFlag == "" {
+		flag.PrintDefaults()
+		log.Fatal("usage: go run generate_service.go -type $TYPE -scope $SCOPE -parent $PARENT -lowestversion $VERSION -updateversion $VERSION")
```
paddycarver: You can also use `flag.Usage()` here.

scripts/generate_service.go
```diff
@@ -0,0 +1,305 @@
...
+
+	f, err := os.Create(fmt.Sprintf("gen-%s.go", *clientTypeFlag))
+	defer f.Close()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	buf := &bytes.Buffer{}
+	t := template.Must(template.New("temp").Parse(multiversionServiceTemplate))
+	t.Execute(buf, templateData)
```
paddycarver: Ignoring an error here.

rileykarson: Done.

scripts/generate_service.go
rileykarson: `flag.usage()` prints some weird output about the binary for me; I prefer the message from `PrintDefaults` w/o that.

scripts/generate_service.go
rileykarson: Done.

scripts/generate_service.go
rileykarson: I'm not sure of another way to not have to attach explicit values to enum values w/o iota; the ordered list is the source of truth for ordering and the `iota` just conveniently means I don't need to set values.

paddycarver:
@danawillow @rileykarson did we decide to not move forward with this? Should it be closed? Sorry, lost some context here.

rosbo:
I like the current documentation where fields are marked as BETA.

danawillow:
We decided that @rileykarson was going to focus on other things while he was here, but taking a quick glance at it I'm actually pretty ok with having something like this- it's generating code that we're just copying and pasting in a bunch of different files. It would need to be updated though, and I don't expect @rileykarson to get around to it any time soon (and I think the rest of us have other things on our plates that this isn't going to be a priority yet). So I'm fine leaving it open in case someone else wants to take it over.

rosbo:
Since we are moving forward with MM, this is not relevant anymore.

Issue #178, opened 2017-07-06
Feature: google_compute_vpc_network_peer


mtohmaz:
Anyone know when/if Terraform will support creating [VPC network peers](https://cloud.google.com/compute/docs/peering-networks)? I know it's a Beta feature in GCP, but it's a very simple solution in connecting separate networks either inside or outside GCP.

danawillow:
Thanks for the FR, @mtohmaz! No timeline right now, but I'm going to link this to https://github.com/terraform-providers/terraform-provider-google/issues/93 so we can easily keep track of all the beta FRs.

danawillow:
Still no timeline, but this is in GA now: https://cloud.google.com/compute/docs/reference/latest/networks/addPeering

rosbo:
Hi Mahmoud, I added a new resource named `google_compute_network_peering` to support this new feature. The code has been committed and will be release soon in the next version of the Terraform Google provider.

Here is an example usage:
```hcl
resource "google_compute_network_peering" "peering1" {
  name = "peering1"
  network = "${google_compute_network.default.self_link}"
  peer_network = "${google_compute_network.other.self_link}"
}

resource "google_compute_network_peering" "peering2" {
  name = "peering2"
  network = "${google_compute_network.other.self_link}"
  peer_network = "${google_compute_network.default.self_link}"
}

resource "google_compute_network" "default" {
  name                    = "foobar"
  auto_create_subnetworks = "false"
}

resource "google_compute_network" "other" {
  name                    = "other"
  auto_create_subnetworks = "false"
}
```

PR #242, opened 2017-07-25
Add peering feature to google_compute_network.

rosbo:
issue #178

I added a few comments on my own PR to ask for the opinions of the reviewers on a few things

cc/ @danawillow @rileykarson

### DIFF
--- a/google/resource_compute_network.go
+++ b/google/resource_compute_network.go
@@ -6,12 +6,17 @@ import (
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
+	"regexp"
 )
+
+const peerNetworkLinkRegex = "projects/(" + projectRegex + ")/global/networks/((?:[a-z](?:[-a-z0-9]*[a-z0-9])?))$"
 
 func resourceComputeNetwork() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceComputeNetworkCreate,
 		Read:   resourceComputeNetworkRead,
+		Update: resourceComputeNetworkUpdate,
 		Delete: resourceComputeNetworkDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
@@ -63,6 +68,39 @@ func resourceComputeNetwork() *schema.Resource {
 			"self_link": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
+			},
+
+			"peering": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": &schema.Schema{
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validateGCPName,
+						},
+						"network": &schema.Schema{
+							Type:             schema.TypeString,
+							Required:         true,
+							ValidateFunc:     validateRegexp(peerNetworkLinkRegex),
+							DiffSuppressFunc: peerNetworkLinkDiffSuppress,
+						},
+						"auto_create_routes": &schema.Schema{
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
+						},
+						"state": &schema.Schema{
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"state_details": &schema.Schema{
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
 			},
 		},
 	}
@@ -117,6 +155,14 @@ func resourceComputeNetworkCreate(d *schema.ResourceData, meta interface{}) erro
 		return err
 	}
 
+	// Add peering. Network peerings cannot be added using the Insert method.
+	if d.Get("peering.#").(int) > 0 {
+		err = addPeering(config, project, network.Name, convertSchemaArrayToMap(d.Get("peering").([]interface{})))
+		if err != nil {
+			return err
+		}
+	}
+
 	return resourceComputeNetworkRead(d, meta)
 }
 
@@ -134,6 +180,7 @@ func resourceComputeNetworkRead(d *schema.ResourceData, meta interface{}) error 
 		return handleNotFoundError(err, d, fmt.Sprintf("Network %q", d.Get("name").(string)))
 	}
 
+	d.Set("peering", flattenNetworkPeerings(network.Peerings))
 	d.Set("gateway_ipv4", network.GatewayIPv4)
 	d.Set("self_link", network.SelfLink)
 	d.Set("ipv4_range", network.IPv4Range)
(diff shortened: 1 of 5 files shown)


google/resource_compute_network.go
```diff
@@ -64,6 +69,39 @@ func resourceComputeNetwork() *schema.Resource {
...
+						"name": &schema.Schema{
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validateGCPName,
+						},
+						"network": &schema.Schema{
+							Type:             schema.TypeString,
+							Required:         true,
+							ValidateFunc:     validateRegexp(peerNetworkLinkRegex),
+							DiffSuppressFunc: peerNetworkLinkDiffSuppress,
```
rosbo: I could achieve the same thing by using a StateFunc that always saves the value to a canonical format. Any preferences?

google/resource_compute_network.go
```diff
@@ -6,12 +6,17 @@ import (
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
+	"regexp"
 )
 
+const peerNetworkLinkRegex = "projects/(" + projectRegex + ")/global/networks/((?:[a-z](?:[-a-z0-9]*[a-z0-9])?))$"
```
rosbo: For the peer network, full URL and partial URL are supported:

1.  `https://www.googleapis.com/compute/v1/projects/my-project/global/networks/network-1` OK 
2. `projects/my-project/global/networks/network-1` OK
3. `global/networks/network-1` ACCEPTED by GCP **but REJECTED** by my implementation 
4. `network-1` WRONG

The reason why I am rejecting (3) even though GCP considers it valid is because I can't properly detect changes if I don't require the project in the URL. GCP, if the project is missing from the URL uses the default project.

If I have this configuration:
```hcl
resource "google_compute_network" "default" {
  name = "default"  

  peering {
    name = "foo"
    network = "global/networks/other-network"
  }
}
```

After the first `terraform apply`, the network is created and the `peering.0.network` field contains`https://www.googleapis.com/compute/v1/projects/my-project/global/networks/other-network` (this value is read from the GCP and GCP always return a full url.

If I call `terraform plan`. I should not see a diff. However, I have the following two strings for the network field:
- old `https://www.googleapis.com/compute/v1/projects/my-project/global/networks/other-network`
- new `global/networks/other-network`
I could suppress the diff by only comparing the network name. However, this would introduce a new bug, if I change the network from `/my-project/global/networks/other-network` to `ANOTHER-PROJECT/global/networks/other-network`, I would detect the diff because both network have the same name.

If I try using the StateFunc, I don't have access to the current project in a StateFunc in order to transform `global/networks/other-network` to its canonical form.

Anybody have a solution for this?
Or are we fine with not supporting the "project-less" version. This is a new field so we are not breaking anybody. We are only forcing them to be more explicit from the start.

google/resource_compute_network.go
```diff
@@ -143,6 +190,55 @@ func resourceComputeNetworkRead(d *schema.ResourceData, meta interface{}) error
...
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	network := d.Get("name").(string)
+
+	d.Partial(true)
+
+	if d.HasChange("peering") {
```
rileykarson: We have a lot of code dedicated to working with changing a set of peerings, but our tests only have 1 or 0 on each network. Let's test that code by adding a test that has several peerings, and updates keeping some but removing/adding others.

google/resource_compute_network.go
```diff
@@ -166,3 +262,102 @@ func resourceComputeNetworkDelete(d *schema.ResourceData, meta interface{}) erro
...
+		}
+		if !found {
+			remove = append(remove, oldPeering)
+		}
+	}
+
+	return add, remove
+}
+
+func addPeering(config *Config, project, network string, add []map[string]interface{}) error {
```
rileykarson: s/addPeering/addPeerings

google/resource_compute_network.go
```diff
@@ -143,6 +190,55 @@ func resourceComputeNetworkRead(d *schema.ResourceData, meta interface{}) error
...
+
+	network := d.Get("name").(string)
+
+	d.Partial(true)
+
+	if d.HasChange("peering") {
+		old, new := d.GetChange("peering")
+		add, remove := calcAddRemoveNetworkPeerings(old, new)
+
+		if len(remove) > 0 {
```
rileykarson: Let's factor this out into `removePeerings` as well, to mirror `addPeering`

google/resource_compute_network.go
rileykarson: Personally, I'm OK with only accepting the first two forms. Terraform self links will always be using full or relative partial URLs, and while omitting the project is a nice shorthand, it's information that users will always have.

I'd be interested if we could file/bump an issue against Terraform itself for getting more information inside a `StateFunc`/`DiffSuppressFunc`; if we knew the `project` in either, we could allow the shorthand as well.

google/resource_compute_network.go
rileykarson: I have no preferences here personally. I think that we could migrate between the two at will without even using a `MigrateFunc` if we ever wanted to, too.

website/docs/r/compute_network.html.markdown
```diff
@@ -17,6 +17,16 @@ resource "google_compute_network" "default" {
   name                    = "foobar"
   auto_create_subnetworks = "true"
 }
+
+resource "google_compute_network" "other" {
```
rileykarson: Same as with the tests - let's make this example a valid network peering.

If we can have one network automatically create subnetworks and we manually define them for another & manage to peer successfully, we'd be able to show even more of how to use this resource to readers.

website/docs/r/compute_network.html.markdown
```diff
@@ -43,8 +53,21 @@ The following arguments are supported:
...
     is not provided, the provider project is used.
+- - -
+
+The `peering` block supports:
+
+* `name` - (Required) Name of the peering.
+* `network` - (Required) Resource link of the peer network.
+* `auto_create_routes` - (Optional) If set to true, the routes between the two networks will
+  be created and managed automatically.
+* `state` - (Computed) State for the peering.
```
rileykarson: Output-only attributes typically live under Attributes Reference, but I'm not sure exactly what it should look like because these are part of the `peering` block.

google/resource_compute_network.go
rosbo: Right now, since I am changing the name of the peering, it does one add and one remove. I will add more coverage though. One problem is that I can't add more than 3 networks because by default, GCD account have a default network and a cap of 5 on the maximum number of networks. I like to leave at least one open to be able to allow some level of manual testing while my test run.

website/docs/r/compute_network.html.markdown
rosbo: Yeah, when I wrote these I wasn't sure where to put it because it is part of the `peering` block. @danawillow any preference?

rosbo:
After thinking more about it and also discussing with Riley, it may makes more sense to have the vpc peering feature be its own resource.

One problem comes with circular dependency when creating peer network:
```hcl
resource "google_compute_network" "network1" {
  name = "network-1" 

  auto_create_subnetworks = false

  peering {
    name = "one-two-peering"
    # Error: cycle network1 depends on network2 and vice versa
    network = "${google_compute_network.network2.self_link}"
  } 
}

resource "google_compute_network" "network2" {
  name = "network-2"  

  auto_create_subnetworks = false

  peering {
    name = "one-two-peering"
    # Error: cycle network1 depends on network2 and vice versa
    network = "${google_compute_network.network1.self_link}"
  }
}
```

To go around this a user must set manually (w/o interpolations) the self link:
```hcl
peering {
    name = "one-two-peering"
    network = "https://www.googleapis.com/compute/v1/projects/rosbo-personal/global/networks/network-2"
 } 
```

If we create a separate resource for peering. The config would look like:

```hcl
resource "google_compute_network" "network1" {
  name = "network-1" 
  auto_create_subnetworks = false
}

resource "google_compute_network" "network2" {
  name = "network-2" 
  auto_create_subnetworks = false
}

resource "google_compute_network_peering" "peering1-2" {
   name = "one-two-peering"
   network = "${google_compute_network.network1.self_link}"
   peer_network = "${google_compute_network.network2.self_link}"
}

resource "google_compute_network_peering" "peering2-1" {
   name = "one-two-peering"
   network = "${google_compute_network.network2.self_link}"
   peer_network = "${google_compute_network.network1.self_link}"
}
```

And alternative for the `google_compute_network_peering` resource would be to support bidirectionality

```hcl
resource "google_compute_network_peering" "peering" {
   name = "one-two-peering"
   network = "${google_compute_network.network2.self_link}"
   peer_network = "${google_compute_network.network1.self_link}"
   bidirectional = true  # Defaults to false. False means only a peering from `network` to `peer_network` is created.
}
```

rosbo:
Closed in favor of #259 using more fine grained resources.

PR #248, opened 2017-07-26
Fix scheduling automatic restart, remove deprecated options

### LINKED ISSUE #224: automatic_restart false in scheduling/google_compute_instance_template forces new resource
Hi there,

If in `google_compute_instance_template` you define `scheduling` block `automatic_restart` = `false`, terraform will always recreate the resource trying to update `automatic_restart` field to `true`.

### Terraform Version
Terraform v0.9.11

### Affected Resource(s)

```hcl
...
resource "google_compute_instance_template" "instance_template" {
  name_prefix = "dgolja-instance-template-tf-bug-"

  region = "${var.region}"

  lifecycle {
    create_before_destroy = true
  }

  can_ip_forward       = false

  scheduling {
    automatic_restart   = false
  }

  network_interface {
    network = "default"
  }

  disk {
    source_image = "projects/ubuntu-os-cloud/global/images/ubuntu-1604-xenial-v20170619a"
    disk_type    = "pd-ssd"
    auto_delete  = true
    boot         = true
  }

  service_account {
    scopes = ["userinfo-email"]
  }
}
...

```

### Debug Output
https://gist.github.com/dgolja/1c80d16db90d9bbea5fa1cce21d9f8f9

### Expected Behavior
Since there is no change terraform shouldn't recreate the resource.

### Actual Behavior
Keep updating the resource, however the new resource is the same.

### Steps to Reproduce
1. `terraform apply`

selmanj:
Trying this again... This is a reattempt of #206, which relates to #107 and fixes #224 . Unfortunately it's not a complete fix as we don't have a way to assign a reasonable default for `automatic_restart` (as its a boolean, we cannot distinguish between a default value or an explicitly specified value).


google/resource_compute_instance_template.go
```diff
@@ -131,14 +133,6 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"automatic_restart": &schema.Schema{
```
rileykarson: Removals are done by using the `Removed` property and not removing the field entirely, so that any users that are left see a descriptive error and not one that looks like a syntax error.

selmanj: Good call -- added

google/resource_compute_instance_template_migrate.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// automatic_restart is stored in two places. The top-level automatic_restart value is deprecated, so let's delete
+	// it from the state map for now. For paranoia's sake, we compare it to the value stored in scheduling as well.
+	ar := is.Attributes["automatic_restart"]
+	delete(is.Attributes, "automatic_restart")
+
+	if is.Attributes["scheduling.#"] != "1" {
+		return nil, fmt.Errorf("Found non-singular scheduling block in state; unsure how to proceed")
+	}
+	sched_ar := is.Attributes["scheduling.0.automatic_restart"]
```
rosbo: nit: we use camelcase instead of snake case for variable names

google/resource_compute_instance_template_migrate.go
```diff
@@ -0,0 +1,50 @@
...
+
+func migrateComputeInstanceTemplateStateV0toV1(is *terraform.InstanceState) (*terraform.InstanceState, error) {
+	log.Printf("[DEBUG] Attributes before migration: %#v", is.Attributes)
+
+	// automatic_restart is stored in two places. The top-level automatic_restart value is deprecated, so let's delete
+	// it from the state map for now. For paranoia's sake, we compare it to the value stored in scheduling as well.
+	ar := is.Attributes["automatic_restart"]
+	delete(is.Attributes, "automatic_restart")
+
+	if is.Attributes["scheduling.#"] != "1" {
```
rosbo: 0 is valid isn't, since it is optional to specify a scheduling block?

selmanj: It is optional, but it is always written to the state file (as we always read a value).

google/resource_compute_instance_template_migrate.go
selmanj: damn these former ruby fingers

google/resource_compute_instance_template.go
```diff
@@ -529,10 +523,10 @@ func resourceComputeInstanceTemplateCreate(d *schema.ResourceData, meta interfac
...
 		_scheduling := _schedulings[0].(map[string]interface{})
 
-		if vp, okp := _scheduling["automatic_restart"]; okp {
-			instanceProperties.Scheduling.AutomaticRestart = googleapi.Bool(vp.(bool))
-			forceSendFieldsScheduling = append(forceSendFieldsScheduling, "AutomaticRestart")
-		}
+		// "automatic_restart" has a default value and is always safe to dereference
+		automaticRestart := _scheduling["automatic_restart"].(bool)
+		instanceProperties.Scheduling.AutomaticRestart = googleapi.Bool(automaticRestart)
+		forceSendFieldsScheduling = append(forceSendFieldsScheduling, "AutomaticRestart")
```
rosbo: Do you really need the forceSend?

In my PR #237. I am dealing also with the `scheduling` blocks for `google_compute_instance` and I don't need to forceSendFields in order to send boolean with a false value...

Have you tested that it is really needed?

google/resource_compute_instance_template_test.go
```diff
@@ -333,6 +354,28 @@ func testAccCheckComputeInstanceTemplateTag(instanceTemplate *compute.InstanceTe
...
+		}
+		return nil
+	}
+}
+
+func testAccCheckComputeInstanceTemplateAutomaticRestart(instanceTemplate *compute.InstanceTemplate, automaticRestart bool) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		ar := instanceTemplate.Properties.Scheduling.AutomaticRestart
+		if ar == nil {
+			return fmt.Errorf("Expected to see a value for AutomaticRestart, but got nil? This shouldn't happen")
```
danawillow: I think "this shouldn't happen" is implied for an error 😛

google/resource_compute_instance_template_migrate.go
selmanj: Yes, the API should always return a value.

For paranoia's sake though, I could add some additional code just-in-case it's not there, as right now it'll just throw an error if it's missing the scheduling information.

google/resource_compute_instance_template.go
selmanj: I don't think you do in this case, as the boolean is of type `*bool`. I can remove it in a subsequent PR

google/resource_compute_instance_template_test.go
selmanj: Fair enough; I guess I'm trying to differentiate between an error that can sometimes happen (we get a 500 or the instance wasn't actually created) vs something that is an absolute illegal state (the nil value for example).

selmanj: Removed "this shouldn't happen".

google/resource_compute_instance_template_migrate.go
selmanj: Added an extra if block to handle empty or non-existant scheduling blocks

PR #249, opened 2017-07-26
Add Beta support for auto_healing_policies to instance group manager.

rileykarson:
Add support for the first Beta feature, `auto_healing_policies`, to `google_compute_instance_group_manager`. Beta support was added to the resource in #234.

Part of #93.
Fixes #48.

Beta features are documented by adding a third divider to the list of fields; in addition to `Required`, and `Optional`, there is now a `Beta` section, where Beta fields are listed.

When users begin using Beta fields, their resources will automatically "upgrade" to Beta, and begin performing operations using Beta APIs instead of v1. If they remove the Beta fields from their resource, it will then automatically "upgrade" back to v1.

By inferring the version from the fields a user uses, we;
- Don't need to perform validation that we match different schema versions in CRUD methods.
- Preserve existing Terraform state files.
- Allow seamless "upgrades" back to v1 as features enter GA.

We do not currently have a means of explicitly notifying users what version an operation is being performed at;  I'd like to investigate means of how to show the version to users and especially whether we can surface that information at the plan phase before committing to an implementation. 

I don't believe that it should be blocking for this change, as both our documentation and the GCP documentation will communicate that a feature is in Beta to users.

@catsby

Issue #245, opened 2017-07-26
Feature request: IPV6 Loadbalancing support


renchap:
Overview: https://cloud.google.com/compute/docs/load-balancing/ipv6

This will require changes to `google_compute_address` and `google_compute_global_forwarding_rule ` with a new `ipVersion` attribute.

rileykarson:
Linking to #93 since it looks like `ipVersion` is in Beta for each.

rileykarson:
Closed with #250 and #265.

renchap:
@rileykarson this feature is now in GA, should I create a separate ticket to support it?
https://cloudplatform.googleblog.com/2017/09/announcing-ipv6-global-load-balancing-ga.html

rosbo:
@renchap I created a new ticket tracking this work. #445. Thank you

PR #250, opened 2017-07-26
Add Beta support & Beta feature ip_version to google_compute_global_address

rileykarson:
Waiting on #249 to merge, but otherwise able to be reviewed.

- Convert `google_compute_global_address` to a versioned resource with Beta support.
- Add support for the Beta field `ip_version` to `google_compute_global_address`

Explicitly setting `IPV4` on a `v1` resource forces a recreate as you're going from `""` to `IPV4` and this resource can't update; this is working as expected based on #185. We don't have enough information at refresh time to perform the read at `v0beta`.

Part 1/2 of #245.
Related to #93

### DIFF
--- a/google/resource_compute_global_address_test.go
+++ b/google/resource_compute_global_address_test.go
@@ -25,6 +25,7 @@ func TestAccComputeGlobalAddress_basic(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckComputeGlobalAddressExists(
 						"google_compute_global_address.foobar", &addr),
+					testAccCheckComputeBetaGlobalAddressIPV4("google_compute_global_address.foobar"),
 				),
 			},
 		},
@@ -44,14 +45,11 @@ func TestAccComputeGlobalAddress_ipv6(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckComputeBetaGlobalAddressExists(
 						"google_compute_global_address.foobar", &addr),
+					testAccCheckComputeBetaGlobalAddressIPV6("google_compute_global_address.foobar"),
 				),
 			},
 		},
 	})
-
-	if addr.IpVersion != "IPV6" {
-		t.Errorf("Expected IP version to be IPV6, got %s", addr.IpVersion)
-	}
 }
 
 func testAccCheckComputeGlobalAddressDestroy(s *terraform.State) error {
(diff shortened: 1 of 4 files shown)


google/resource_compute_global_address_test.go
```diff
@@ -29,6 +31,29 @@ func TestAccComputeGlobalAddress_basic(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeGlobalAddressDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccComputeGlobalAddress_ipv6,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeBetaGlobalAddressExists(
```
rosbo: Consider asserting on the format of the computed address to make sure an IPv6 is used.
Additionally, to prevent regression, make sure the ipv4 address follows the ipv4 pattern.

rileykarson: I'm not sure I agree that we need to validate that the API isn't lying to us when it says an IP address is at some version; especially when we would need to use a regex or vendor a new package to verify it. 

Added a test for IPV4 on `_basic`, and moved checks to a TestCheckFunc.

rosbo: You could use https://golang.org/pkg/net/#ParseIP

ip := ParseIP(address)
ip.To16 != nil // means it is a IPv6 address.
ip.To4 != nil // means this is a IPv4 address.

Checking if the IpVersion match what we expect is probably fine for the IPv6 case. My main goal was not to test if the API "lies" to us but rather make sure we don't create an ipv6 address when we don't use the beta feature (and since we don't have the ip version field in that case, the only way to check is looking at address itself to validate it is ipv4).

rileykarson: Ah, ok - we can actually just read the resource using a Beta API in this test and see the `IpVersion`; that's what the TestCheckFunc does.

rileykarson:
```
TF_ACC=1 go test ./google -v -run=TestAccComputeGlobalAddress -timeout 120m
=== RUN   TestAccComputeGlobalAddress_importBasic
--- PASS: TestAccComputeGlobalAddress_importBasic (22.65s)
=== RUN   TestAccComputeGlobalAddress_basic
--- PASS: TestAccComputeGlobalAddress_basic (22.71s)
=== RUN   TestAccComputeGlobalAddress_ipv6
--- PASS: TestAccComputeGlobalAddress_ipv6 (22.45s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	67.973s
```

PR #231, opened 2017-07-22
(OLD) Google Cloud Dataproc support

nickithewatt:
Addresses https://github.com/terraform-providers/terraform-provider-google/issues/31 and provides support for Google Cloud Dataproc. @danawillow @radeksimko 

At this stage I have added support for the resources below which all work and have acceptance tests as well - Not sure where / when you run the acceptance tests as they are not run as part of the travis build? There is one acceptance test I added (`TestAccDataprocCluster_withServiceAcc`) which needs a new environment variable to be set in order to work. At present if it doesn't detect the variable it skips the test.   

The jobs are pretty much fleshed out for those wanting to run PySpark and Spark jobs on clusters. I thought I should first get some validation / feedback to see if you are happy with the approach I have taken. Then if necessary I can make any changes before completing the rest. 
**UPDATE** - I have implemented them all now... 

_**resources**_
  - [x] `google_dataproc_cluster`
  - [x] `google_dataproc_job`
    - [x] via `pyspark_config` support for PySpark jobs
    - [x] via `spark_config` support for Spark jobs
    - [x] via `sparksql_config` support for Spark-SQL jobs
    - [x] via `hadoop_config` support for Hadoop jobs
    - [x] via `hive_config` support for Hive jobs
    - [x] via `pig_config` support for Pig jobs

Specifically for `google_dataproc_job`, _create_ essentially _submits_ a job to the cluster to let it run. It does not wait for it to finish. Updating of jobs doesn't really make much sense, and for _delete_ I am genuinely deleting the job. Under normal circumstances, Dataproc won't let you delete active jobs. I have therefore added a `force_destroy` option which if true, will first cancel the job before deleting.

In any case, looking forward to your comments.
Thanks
Nicki

Issue #31, opened 2017-06-13
Google Cloud Dataproc resource


hashibot:
_This issue was originally opened by @shpsec-dennis as hashicorp/terraform#6312. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,

Feature request for having a dataproc\* resource(s) for Terraform. 

https://cloud.google.com/dataproc/

Regards,

Dennis O.

nickithewatt:
Hi, this is good timing as I am actually in the process of looking to add / contribute this functionality and would like to submit a PR in the not too distant future. Current work is happening in my branch here: https://github.com/nickithewatt/terraform-provider-google/tree/dataproc-support

Let me know if you would like me to submit a PR - would still be in a WIP progress state as busy fixing / putting in more work around the acceptance tests but happy to submit what I currently have. :)

danawillow:
Sure, that would be great! No guarantees that we'll get around to reviewing it before it's 100% done but it can't hurt :)

nickithewatt:
Great, I will see how it goes, might try and flesh out a few more things first! Thanks

nickithewatt:
Hi @danawillow 

Just wanted to know if there was anything more I can do on these PR's and what the ETA might be for a review?

Thanks!
Nicki

danawillow:
I'll get to them this week! Have a few things in the queue before these but they're close to the top :)

nickithewatt:
Awesome, thanks! 👍

saleh-sedighi-ck:
👍

PR #252, opened 2017-07-26
Add google_dataproc_cluster resource

nickithewatt:
@danawillow As requested ...
This is PR 1 of 2 (splitting https://github.com/terraform-providers/terraform-provider-google/pull/231 up) and looking to address https://github.com/terraform-providers/terraform-provider-google/issues/31 (Add support for Google Cloud Dataproc) specifically adding the `google_dataproc_cluster` resource.

To Recap:
This resource has acceptance tests however one of them (TestAccDataprocCluster_withServiceAcc  needs a new environment variable to be set in order to work. At present if it doesn't detect the variable it skips the test.


google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+			// Bucket may be gone already ignore
+			return nil
+		}
+		if ok && gerr.Code == 429 {
+			return resource.RetryableError(gerr)
+		}
+		return resource.NonRetryableError(err)
+	})
+	if err != nil {
+		fmt.Printf("[ERROR] Attempting to delete autogen bucket (for dataproc cluster) if exists: Error deleting bucket %s: %v\n\n", bucket, err)
```
danawillow: Let's use the full word (autogenerated) just for extra clarity. Also I don't think you need to put down "if exists" since at this point we know it exists.

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+		return err
+	}
+	log.Printf("[DEBUG] Attempting to delete autogen bucket (for dataproc cluster) if exists: Deleted bucket %v\n\n", bucket)
```
danawillow: same

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+	return emptyAndDeleteStorageBucket(config, bucket)
+}
+
+func emptyAndDeleteStorageBucket(config *Config, bucket string) error {
+	err := deleteStorageBucketContents(config, bucket)
+	if err != nil {
+		return err
+	}
+
+	err = deleteEmptyBucket(config, bucket)
```
danawillow: Just to confirm- you can't delete a bucket unless it's empty?

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+		// purge the bucket...
+		log.Printf("[DEBUG] Attempting to delete autogen bucket (for dataproc cluster) if exists. \n\n")
+
+		for _, object := range res.Items {
+			log.Printf("[DEBUG] Attempting to delete autogen bucket (for dataproc cluster) if exists. Found %s", object.Name)
+
+			err := config.clientStorage.Objects.Delete(bucket, object.Name).Do()
+			if err != nil {
+				if gerr, ok := err.(*googleapi.Error); ok && gerr.Code != 404 {
+					// Object is now gone ... ignore
```
danawillow: ignore what?

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+					log.Printf("[DEBUG] Attempting to delete autogen bucket (for dataproc cluster) if exists: Error trying to delete object: %s %s\n\n", object.Name, err)
+					return err
+				}
+			}
+			log.Printf("[DEBUG] Attempting to delete autogen bucket (for dataproc cluster) if exists: Object deleted: %s \n\n", object.Name)
+		}
+	} else {
```
danawillow: do you need this else?

google/resource_dataproc_cluster_test.go
```diff
@@ -0,0 +1,780 @@
...
+	return fmt.Sprintf(`
+resource "google_dataproc_cluster" "single_node_cluster" {
+	name   = "dproc-cluster-test-%s"
+	region = "us-central1"
+
+    properties = {
+        "dataproc:dataproc.allow.zero.workers" = "true"
+    }
+
+    # Because of current restrictions with computed AND default
```
danawillow: can you be explicit on what the restrictions are?

google/resource_dataproc_cluster_test.go
```diff
@@ -0,0 +1,780 @@
...
+  name           = "dproc-cluster-test-%s-init-script.sh"
+  bucket         = "${google_storage_bucket.init_bucket.name}"
+  content        = <<EOL
+#!/bin/bash
+ROLE=$$(/usr/share/google/get_metadata_value attributes/dataproc-role)
+if [[ "$${ROLE}" == 'Master' ]]; then
+  echo "on the master" >> /tmp/%s
+  gsutil cp /tmp/%s ${google_storage_bucket.init_bucket.url}
+else
+  echo "on the worker" >> /tmp/msg.txt
```
danawillow: should this msg.txt also be set to objName?

google/resource_dataproc_cluster_test.go
```diff
@@ -0,0 +1,780 @@
...
+    force_destroy = "true"
+}
+
+resource "google_storage_bucket_object" "init_script" {
+  name           = "dproc-cluster-test-%s-init-script.sh"
+  bucket         = "${google_storage_bucket.init_bucket.name}"
+  content        = <<EOL
+#!/bin/bash
+ROLE=$$(/usr/share/google/get_metadata_value attributes/dataproc-role)
+if [[ "$${ROLE}" == 'Master' ]]; then
```
danawillow: do we actually use the difference between being on the master/worker anywhere in the test?

google/resource_dataproc_cluster_test.go
```diff
@@ -0,0 +1,780 @@
...
+		if !ok {
+			return fmt.Errorf("Terraform resource Not found: %s", n)
+		}
+
+		if cluster.Config.MasterConfig == nil || cluster.Config.WorkerConfig == nil || cluster.Config.SecondaryWorkerConfig == nil {
+			return fmt.Errorf("Master/Worker/SecondaryConfig values not set in GCP, expecting values")
+		}
+
+		clusterTests := []tfAndGCPTestField{
+			{"master_config.0.num_masters", "3", strconv.FormatInt(cluster.Config.MasterConfig.NumInstances, base10)},
```
danawillow: you can use strconv.Itoa here so you don't need all the base10s

google/resource_dataproc_cluster_test.go
```diff
@@ -0,0 +1,780 @@
...
+    worker_config { }
+	master_config {
+		machine_type      = "n1-standard-1"
+		boot_disk_size_gb = 10
+	}
+
+	service_account = "%s"
+
+	service_account_scopes = [
+        #    The following scopes necessary for the cluster to function properly are
```
danawillow: Can you fix the indentation on these comments? (some use tabs and some use spaces)

google/resource_dataproc_cluster_test.go
```diff
@@ -0,0 +1,780 @@
...
+		#	always added, even if not explicitly specified:
+		#		useraccounts-ro: https://www.googleapis.com/auth/cloud.useraccounts.readonly
+		#		storage-rw:      https://www.googleapis.com/auth/devstorage.read_write
+		#		logging-write:   https://www.googleapis.com/auth/logging.write
+        #
+		#	So user is expected to add these explicitly (in this order) otherwise terraform
```
danawillow: Instead of requiring the user to add them in a specific order, can you just use a set?

google/resource_dataproc_cluster_test.go
```diff
@@ -0,0 +1,780 @@
...
+}
+
+func testAccDataprocCluster_withNetworkRefs(rnd, netName string) string {
+	return fmt.Sprintf(`
+resource "google_compute_network" "dataproc_network" {
+	name = "%s"
+	auto_create_subnetworks = true
+}
+
+resource "google_compute_firewall" "dataproc_network_firewall" {
```
danawillow: Can you add a comment explaining why we need firewalls for this test?

google/resource_dataproc_cluster_test.go
```diff
@@ -0,0 +1,780 @@
...
+
+func validateClusterDeleted(project, region, clusterName string, config *Config) error {
+	_, err := config.clientDataproc.Projects.Regions.Clusters.Get(
+		project, region, clusterName).Do()
+
+	if err != nil {
+		if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 404 {
+			return nil
+		} else if ok {
+			return fmt.Errorf("Error make GCP platform call to verify dataproc cluster deleted: http code error : %d, http message error: %s", gerr.Code, gerr.Message)
```
danawillow: conciseness suggestion: `"Error verifying dataproc cluster deletion. Code: %d. Message: %s"`

google/resource_dataproc_cluster_test.go
```diff
@@ -0,0 +1,780 @@
...
+		boot_disk_size_gb = 10
+	}
+
+	network = "${google_compute_network.dataproc_network.self_link}"
+}
+
+`, netName, rnd, rnd, rnd)
+}
+
+func testAccPreCheckWithServiceAccount(t *testing.T) {
```
danawillow: generally we put all the configs together at the bottom of the test file, so let's move these fns above them

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+layout: "google"
+page_title: "Google: google_dataproc_cluster"
+sidebar_current: "docs-google-dataproc-cluster"
+description: |-
+  Manages a Cloud Dataproc cluster resource.
+---
+
+# google\_dataproc\_cluster
+
+Manages a Cloud Dataproc cluster resource within GCE. For more information see
```
danawillow: GCE -> GCP

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+[the official dataproc documentation](https://cloud.google.com/dataproc/).
+
+
+!> **Warning:** Due to limitations of the API, all arguments except
+`labels`,`worker_config.num_workers` and `worker_config.preemptible_num_workers` are non-updateable. Changing any will cause recreation of the
```
danawillow: for clarity, "Changing any" -> "Changing others"

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+
+    master_config {
+        num_masters       = 1
+        machine_type      = "n1-standard-1"
+        boot_disk_size_gb = 10
+        num_local_ssds    = 1
+    }
+
+    worker_config {
+    	num_workers             = 2
```
danawillow: nit: indentation

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+    zone.
+
+- - -
+
+* `region` - (Optional) The region that the cluster and associated nodes will be created in.
+   Defaults to `global`.
+
+* `zone` - (Optional) The GCP zone where your data is stored and used (i.e. where
+    the master and the worker nodes will be created in). If region is set to 'global'
+    then `zone` IS mandatory, otherwise GCP is able to make use of [Auto Zone Placement](https://cloud.google.com/dataproc/docs/concepts/auto-zone)
```
danawillow: nit: IS -> is

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+    to determine this automatically for you.
+    Note: This setting additionally determines and restricts
+    which computing resources are available for use with other config such as
```
danawillow: nit: configs

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+    `master_config.machine_type` and `worker_config.machine_type`.
+
+* `staging_bucket` - (Optional) The Cloud Storage staging bucket used to stage files,
+   such as Hadoop jars, between client machines and the cluster.
+   Note: If you don't explicitly specify a `staging_bucket`
+   then GCP will auto create / assign one for you, however you are NOT guaranteed to get
```
danawillow: nit: rephrase to `then GCP will auto create / assign one for you. However, you are not guaranteed`

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+   with other clusters in the same region/zone also choosing to use the auto generation
+   option.
+
+* `delete_autogen_bucket` (Optional) If this is set to true, upon destroying the cluster,
+   if no explicit `staging_bucket` was specified (i.e. an auto generated bucket was relied
+   upon) then this auto generated bucket will also be deleted as part of the cluster destroy.
+   By default this is set to false.
+
+* `image_version` - (Optional) The Cloud Dataproc image version to use
+   for the cluster - this essentially controls the sets of software versions
```
danawillow: essentially?

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+
+* `network` - (Optional) The name or self_link of the Google Compute Engine
+    network to the cluster will be part of. Conflicts with `subnetwork`.
+    If neither is specified, this defaults to the "default" network.
+
+* `subnetwork` - (Optional) The name or self_link of the Google Compute Engine
+   subnetwork the cluster will be part of. Conflicts with `network`.
+
+* `initialization_actions` - (Optional) A list of scripts to be executed during
+   initialisation of the cluster. Each must be a GCS file with a gs:// prefix.
```
danawillow: for consistency, let's use the spelling `initialization`

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+
+* `metadata` - (Optional) The metadata key/value pairs assigned to instances in
+    the cluster.
+
+* `labels` - (Optional) The list of labels (key/value pairs) to be applied to
+   instances in the cluster.
+
+* `properties` - (Optional) A list of override and additional properties (key/value pairs)
+   used to modify various aspects of the common configuration files used when creating
+   a cluster. For a list of valid please see
```
danawillow: valid -> valid properties

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+    in GB. The primary disk contains the boot volume and system libraries, and the
+    smallest allowed disk size is 10GB, but defaults to 500GB. Note: If SSDs are not
+    attached, it also contains the HDFS data blocks and Hadoop working directories.
+
+* `num_local_ssds` - (Optional) The amount of local SSD disks that will be
+    attached to each cluster node. Defaults to 0.
+
+The **worker_config** supports:
+
+* `num_worker`- (Optional) Specifies the number of worker nodes to create.
```
danawillow: num_workers

website/docs/r/dataproc_cluster.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+   it will be the auto generated name.
+
+<a id="timeouts"></a>
+## Timeouts
+
+`google_dataproc_cluster` provides the following
+[Timeouts](/docs/configuration/resources.html#timeouts) configuration options:
+
+- `create` - (Default `10 minutes`) Used for creating clusters.
+- `update` - (Default `5 minutes`) Used for updates to clusters
```
danawillow: nit: "updates to" -> "updating"

danawillow:
Thanks @nickithewatt for the new resource! I'm excited to have dataproc functionality soon! I've left a bunch of comments, but a fair number of them are just phrasing suggestions. I appreciate the thorough tests, too!

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+		Update: resourceDataprocClusterUpdate,
+		Delete: resourceDataprocClusterDelete,
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(10 * time.Minute),
+			Update: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		SchemaVersion: 1,
```
nickithewatt: Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"delete_autogen_bucket": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				ForceNew: false,
```
nickithewatt: Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+			},
+
+			"staging_bucket": {
```
nickithewatt: I was trying to keep the config simple from the terraform user's perspective. Logically, most of these can simply be seen as global config as it applies to the whole cluster. Instead of creating an artificial extra config block, I decided to rather just define it at the top level. The sections where specific differences are required (i.e. `master_config` and `worker_config`) do however have their own configs

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"bucket": {
```
nickithewatt: Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
```
nickithewatt: True

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+				ConflictsWith: []string{"subnetwork"},
+				StateFunc: func(s interface{}) string {
+					return extractLastResourceFromUri(s.(string))
+				},
+			},
+
+			"subnetwork": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
```
nickithewatt: No, the network may be (if not specified) but the only way to set the subnetwork is explicitly. Changed

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+				Type:     schema.TypeMap,
+				Optional: true,
+				ForceNew: false,
+				Elem:     schema.TypeString,
+			},
+
+			"properties": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				ForceNew: false,
```
nickithewatt: True :) changed

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+							ForceNew: true,
+						},
+
+						"num_local_ssds": {
+							Type:     schema.TypeInt,
+							Optional: true,
+							Computed: true,
```
nickithewatt: Technically no. I had originally added it as computed to ensure it was listed as output for people to potentially use should they want to use terraform to build up a picture of what the cluster is comprised of, but this is probably overkill, removed.

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+							ForceNew: true,
+						},
+
+						"boot_disk_size_gb": {
+							Type:     schema.TypeInt,
+							Optional: true,
+							Computed: true,
```
nickithewatt: GCP itself provides defaults for these values if not specified. I was trying to maintain GCP making the decision rather than explicitly defaulting in Terraform. This way if GCP chooses to change the default there is no change required in Terraform.

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+							},
+						},
+					},
+				},
+			},
+
+			"service_account": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
```
nickithewatt: The service_scopes are, but not the account. Changed

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+				ForceNew: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+					StateFunc: func(v interface{}) string {
+						return canonicalizeServiceScope(v.(string))
+					},
+				},
+			},
+
+			"initialization_action_timeout_sec": {
```
nickithewatt: I went with this as an initial implementation to get something out there. That said people may well want to do this, so have adapted it to make this possible. You now need to specify zero or more `initialization_action` blocks with `script` and optional `timeout_sec` fields.

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
```
nickithewatt: It is, GCP defaults to 300s. Explicitly set the default value to 300 though to make it more explicit.

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+				ForceNew: true,
+			},
+
+			"initialization_actions": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: false,
```
nickithewatt: Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+			},
+
+			"metadata": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				ForceNew: true,
+				Elem:     schema.TypeString,
+			},
+
+			"master_instance_names": {
```
nickithewatt: Not necessarily no, had just added them to top level to make it easier to get hold of, however given there are other computed values within the block, it probably makes better sense to pop them inside. Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+
+func resourceDataprocClusterCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	timeoutInMinutes := int(d.Timeout(schema.TimeoutCreate).Minutes())
```
nickithewatt: Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+		gceConfig.SubnetworkUri = extractLastResourceFromUri(v.(string))
+	}
+
+	if v, ok := d.GetOk("tags"); ok {
+		tagsList := v.([]interface{})
+		tags := []string{}
+		for _, v := range tagsList {
+			tags = append(tags, v.(string))
+		}
+		gceConfig.Tags = tags
```
nickithewatt: Thanks, Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+	}
+
+	if v, ok := d.GetOk("service_account"); ok {
+		gceConfig.ServiceAccount = v.(string)
+	}
+
+	if v, ok := d.GetOk("service_account_scopes"); ok {
+		scopesList := v.([]interface{})
```
nickithewatt: This one is slightly different as it also does a transformation on the string. That said, created another helper function `convertAndMapStringArr` to do similar

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+		scopes := []string{}
+		for _, v := range scopesList {
+			scopes = append(scopes, canonicalizeServiceScope(v.(string)))
+		}
+
+		sort.Strings(scopes)
+		gceConfig.ServiceAccountScopes = scopes
+	}
+
+	gceConfig.ZoneUri = zone.(string)
```
nickithewatt: Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+
+	clusterConfig := &dataproc.ClusterConfig{
+		GceClusterConfig: gceConfig,
+		SoftwareConfig:   &dataproc.SoftwareConfig{},
+	}
+
+	if v, ok := d.GetOk("initialization_actions"); ok {
+
```
nickithewatt: Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+		clusterConfig.InitializationActions = actions
+	}
+
+	if v, ok := d.GetOk("staging_bucket"); ok {
+		clusterConfig.ConfigBucket = v.(string)
+	}
+
+	if v, ok := d.GetOk("master_config"); ok {
+		masterConfigs := v.([]interface{})
+		if len(masterConfigs) > 1 {
```
nickithewatt: Removed

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+			clusterConfig.MasterConfig.DiskConfig.BootDiskSizeGb = int64(v.(int))
+		}
+		if v, ok = masterConfig["num_local_ssds"]; ok {
+			clusterConfig.MasterConfig.DiskConfig.NumLocalSsds = int64(v.(int))
+		}
+	}
+
+	if v, ok := d.GetOk("worker_config"); ok {
+		configs := v.([]interface{})
+		if len(configs) > 1 {
```
nickithewatt: Removed

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+		}
+		if v, ok = config["num_local_ssds"]; ok {
+			clusterConfig.WorkerConfig.DiskConfig.NumLocalSsds = int64(v.(int))
+		}
+
+		clusterConfig.SecondaryWorkerConfig = &dataproc.InstanceGroupConfig{
+			DiskConfig: &dataproc.DiskConfig{},
+		}
+
+		if v, ok = config["preemptible_num_workers"]; ok {
```
nickithewatt: Is seems to work even without this, however I have set it to be explicit

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+	// Wait until it's created
+	waitErr := dataprocClusterOperationWait(config, op, "creating Dataproc cluster", timeoutInMinutes, 3)
+	if waitErr != nil {
+		// The resource didn't actually create
+		d.SetId("")
+		return waitErr
+	}
+
+	log.Printf("[INFO] Dataproc cluster %s has been created", clusterName)
+	d.SetId(clusterName)
```
nickithewatt: Fair enough, Done.

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+
+	e := resourceDataprocClusterRead(d, meta)
+	if e != nil {
```
nickithewatt: Changed - returned result directly

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+
+func resourceDataprocClusterUpdate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	// The only items which are currently able to be updated, without a
+	// forceNew in place are the labels and/or the number of worker nodes in a cluster
+	if !(d.HasChange("labels") ||
+		d.HasChange("worker_config.0.num_workers") ||
+		d.HasChange("worker_config.0.preemptible_num_workers")) {
+		return errors.New("*** programmer issue - update resource called however item is not allowed to be changed - investigate ***")
```
nickithewatt: No, I was probably just being paranoid ;) removed

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+
+		wconfigs := d.Get("worker_config").([]interface{})
+		conf := wconfigs[0].(map[string]interface{})
+
+		desiredNumWorks := conf["num_workers"].(int)
+		cluster.Config.WorkerConfig = &dataproc.InstanceGroupConfig{
+			NumInstances: int64(desiredNumWorks),
+		}
+
+		if len(updMask) > 0 {
```
nickithewatt: I agree, changed.

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+			updMask = updMask + ","
+		}
+		updMask = updMask + "config.secondary_worker_config.num_instances"
+	}
+
+	patch.UpdateMask(updMask)
+
+	op, err := patch.Do()
```
nickithewatt: This is slightly different as I need to set the UpdateMask which cannot be set in the constructor as with the others. If you chain it it looks a bit messy so did at least move it down so its all together.

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+		if len(cluster.Config.GceClusterConfig.ServiceAccountScopes) > 0 {
+			sort.Strings(cluster.Config.GceClusterConfig.ServiceAccountScopes)
+			d.Set("service_account_scopes", cluster.Config.GceClusterConfig.ServiceAccountScopes)
+		}
+
+	}
+
+	if cluster.Config.SoftwareConfig != nil {
+		d.Set("image_version", cluster.Config.SoftwareConfig.ImageVersion)
+		//We only want our overriden values here for now
```
nickithewatt: Because GCP returns ALL properties, both overridden by user and the default ones. If it we set the values here, the user would get a diff on a plan the next time as it would suddenly now include the 100's of default options.

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+			d.Set("preemptible_instance_names", cluster.Config.SecondaryWorkerConfig.InstanceNames)
+		}
+
+		d.Set("worker_config", workerConfig)
+	}
+
+	return nil
+}
+
+func extractInitTimeout(t string) (int, error) {
```
nickithewatt: Yes, Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+	d.SetId("")
+	return nil
+}
+
+func deleteAutogenBucketIfExists(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	// Determine if the user specified a specific override staging bucket, if so
+	// let it be ...  otherwise GCP will have created one for them somewhere along
+	// the way, we auto delete it as this will be dangling around otherwise
```
nickithewatt: Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+
+	v, specified := d.GetOk("staging_bucket")
+	if specified {
```
nickithewatt: Done

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,870 @@
...
+		log.Printf("[DEBUG] staging bucket %s (for dataproc cluster) has explicitly been set, leaving it...", v)
+		return nil
+	}
+	bucket := d.Get("bucket").(string)
+
+	log.Printf("[DEBUG] Attempting to delete autogen bucket %s (for dataproc cluster) ...", bucket)
```
nickithewatt: Done

google/resource_dataproc_cluster.go
nickithewatt: Confirmed, I got an error when I first tried to do this

google/resource_dataproc_cluster.go
nickithewatt: OK

google/resource_dataproc_cluster.go
nickithewatt: Done

google/resource_dataproc_cluster.go
nickithewatt: Nothing, redundant comment, removed

google/resource_dataproc_cluster.go
nickithewatt: No, removed

google/resource_dataproc_cluster_test.go
nickithewatt: Done

google/resource_dataproc_cluster_test.go
nickithewatt: Not in the test no, script is a typical init scipt. Simplified it to align with test.

google/resource_dataproc_cluster_test.go
nickithewatt: N/A simplified as per above

google/resource_dataproc_cluster_test.go
nickithewatt: Done

google/resource_dataproc_cluster_test.go
nickithewatt: This comment does not apply anymore, removed

google/resource_dataproc_cluster_test.go
nickithewatt: Done in test

google/resource_dataproc_cluster_test.go
nickithewatt: OK

google/resource_dataproc_cluster_test.go
nickithewatt: Can't use `strconv.Itoa` directly as its an int64 however as the type of fields being used are really not going to get that big, in reality I can probably downcast without loss of precision

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Done

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Done

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Done

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Done

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Done

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Done

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Removed

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Done

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Done

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Done

website/docs/r/dataproc_cluster.html.markdown
nickithewatt: Done

nickithewatt:
@danawillow Changes made

google/resource_dataproc_cluster_test.go
nickithewatt: Actually, I recall now, I couldn't get Terraform to store the state in the same order and it kept generating diff's even when nothing had changed. I don't believe the Set guarantees ordering (only uniqueness) and will have the same problem

danawillow:
In general, we'd really prefer to keep as closely as possible to the REST API for resources. GCP already produces 3 different ways of specifying resources (REST API, gcloud, and cloud console) and so adding a 4th, whether it's actually easier for the user or not, adds complexity. Since we're calling the REST API from Terraform, modeling our resource after that means that there's an easy place for users to go for extra documentation, the examples in the GCP documentation can easily be transferred over to a Terraform configuration, and we know that we won't accidentally introduce changes that aren't forwards-compatible. It also makes maintenance easier, and allows us to create standard functions that you might see across other resources, like the expand/flatten functions (here's one example, but they're all over: https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_bigquery_table.go) which make the resource code really easy to read for someone familiar with terraform but unfamiliar with a specific resource. Obviously there are going to be some places where it's best to differ from the API, but I think maintaining its original structure as much as possible is going to be the smartest way forward. Does that make sense?

nickithewatt:
Hi Dana, yes that makes sense. I do think that there may be times when mirroring the REST API exactly is slight overkill (e.g. the `cluster_config` containing block below). That said, I do appreciate the arguments and am happy to adjust the config to align closer with REST API.

Just to be clear ... is the config below closer to what you are expecting? If so, I shall adjust the code to reflect this.

```
resource "google_dataproc_cluster" "mycluster" {
	name           = "my-cluster"
	project        = "my-project"
	region         = "${var.my_region}"
	labels         = {
		"mykey" = "myvalue"
	}
	
	cluster_config {
	
		staging_bucket = "${google_storage_bucket.dproc_bucket.name}"
		
		gce_cluster_config {
			zone                   = "us-central1-f"
			network                = "${google_compute_network.poc_region.name}"
			subnetwork             = "${google_compute_subnetwork.poc_region.name}"
			tags                   = [ "aaa", "bbb"]
			service_account        = "something"
			service_account_scopes = [ "useraccounts-ro","storage-rw","logging-write" ]
		}
		
		worker_config {
			num_workers       = 2
			machine_type      = "n1-standard-2"
			disk_config {
				boot_disk_size_gb = 50
				num_local_ssds    = 0
			}
		}	      
	
		master_config {
			num_masters = 1
			machine_type = "n1-standard-2"
			disk_config {
				boot_disk_size_gb = 50
				num_local_ssds = 0
			}
		}
	
		[secondary|preemptible]_worker_config {
			num_workers = 2
			disk_config {
				boot_disk_size_gb = 50
				num_local_ssds = 0
			}
		}
		
		software_config {
			image_version  = "preview"
			properties = {
				"dataproc:dataproc.allow.zero.workers" = "true"
			}
		}
	
		# Can have multiple versions of these
		initialization_action {
			script      = "gs://blah/blah.sh"
			timeout_sec = 500
		}
	}
}
```

danawillow:
Yup, that looks right. I'd also rename `num_masters` and `num_workers` to `num_instances`, that way the schema can be shared between the three. Thanks!

nickithewatt:
@danawillow Hi! So I eventually got round to this although it turned out to be a bit more work than I had originally anticipated. The multiple nested config is there now as per previous comment, and also merged last master in. Let me know if you have any more comments. For the moment I have left the job PR (https://github.com/terraform-providers/terraform-provider-google/pull/253) until hopefully this one gets through, then I will address that one :) Thanks!

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,936 @@
...
+		d.SetId("")
+		return waitErr
+	}
+
+	log.Printf("[INFO] Dataproc cluster %s has been created", clusterName)
+	return resourceDataprocClusterRead(d, meta)
+
+}
+
+func preemptibleInstanceGroupConfigCreate(cfg map[string]interface{}) *dataproc.InstanceGroupConfig {
```
danawillow: We usually call functions like this expand functions, so this would be called `expandPreemptibleInstanceGroupConfig`. Mind renaming these?

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,936 @@
...
+		cluster.Labels = m
+
+		updMask = append(updMask, "labels")
+	}
+
+	if d.HasChange("cluster_config.0.worker_config.0.num_instances") {
+		wconfigs := d.Get("cluster_config.0.worker_config").([]interface{})
+		conf := wconfigs[0].(map[string]interface{})
+
+		desiredNumWorks := conf["num_instances"].(int)
```
danawillow: I think you can shorten these few lines just by doing `desiredNumWorks := d.Get("cluster_config.0.worker_config.0.num_instances").(int)`

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,936 @@
...
+		return err
+	}
+
+	d.Set("cluster_config", cfg)
+	return nil
+}
+
+func flattenClusterConfig(d *schema.ResourceData, cfg *dataproc.ClusterConfig) ([]map[string]interface{}, error) {
+
+	data := getOrCreateNewMapSchema(d, "cluster_config")
```
danawillow: I'm trying to figure out what this is used for. Is it because some fields aren't returned from the API? If so, the way we usually handle that is to do a d.Get() for those fields specifically. This makes me a bit nervous that we might at some point forget to populate a new field that gets added and not notice because it just kept whatever was already in state.

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,936 @@
...
+				Optional: true,
+				Default:  "global",
+				ForceNew: true,
+			},
+
+			"labels": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				Elem:     schema.TypeString,
+				// GCP automatically adds a 'goog-dataproc-cluster-name' label
```
danawillow: Can you add a test that includes labels? Because of the automatically added one, I think we're going to end up with a diff once we create the cluster and then read it back (since it'll have one more label than we have in our config)

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,936 @@
...
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"image_version": {
+										Type:     schema.TypeString,
+										Optional: true,
+										Computed: true,
+										ForceNew: true,
+									},
+
+									"override_properties": {
```
danawillow: Can you add a comment as to why we need two separate properties fields?

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,936 @@
...
+									},
+
+									"network": {
+										Type:          schema.TypeString,
+										Optional:      true,
+										Computed:      true,
+										ForceNew:      true,
+										ConflictsWith: []string{"cluster_config.gce_cluster_config.subnetwork"},
+										StateFunc: func(s interface{}) string {
+											return extractLastResourceFromUri(s.(string))
```
danawillow: I think this will be ok, but it does make me a bit nervous to only store the end portion of the network since hypothetically it could be in a different project. Do other resources have good ways to handle the name vs url problem? I took a quick look and didn't find anything great, but maybe you'll have better luck :) (I think we usually just use a diffsuppress)

google/resource_dataproc_cluster.go
```diff
@@ -0,0 +1,936 @@
...
+
+	if gcc.NetworkUri != "" {
+		gceConfig["network"] = extractLastResourceFromUri(gcc.NetworkUri)
+	}
+	if gcc.SubnetworkUri != "" {
+		gceConfig["subnetwork"] = extractLastResourceFromUri(gcc.SubnetworkUri)
+	}
+
+	if len(gcc.ServiceAccountScopes) > 0 {
+		sort.Strings(gcc.ServiceAccountScopes)
```
danawillow: Won't this lead to diffs if the user didn't originally specify them alphabetically?

nickithewatt:
Thanks Dana,

Will try to get round to these as soon as time permits!

nickithewatt:
@danawillow Have updated and addressed suggestions and recommended changes, and rerun the cluster acceptance tests which all pass as well.

danawillow:
Thanks so much @nickithewatt! This was a huge change and I really appreciate you seeing it through to the end :)

Merging now.

Issue #255, opened 2017-07-27
Feature: Support for Google Source Repositories


alex1x:
Support for https://cloud.google.com/source-repositories/

PR #256, opened 2017-07-27
Add Support for Google source repos

alex1x:
Addresses #255 
The naming of the resource was not self evident because it's not very consistent on GCP either. I went for "source repos" as in the API (google.golang.org/api/sourcerepo/v1) to avoid the length and repetitiveness of calling it a "Google Source Repositories Repository".


website/docs/r/sourcerepos_repository.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+layout: "google"
+page_title: "Google: google_sourcerepos_repository"
+sidebar_current: "docs-google-sourcerepos_repository"
+description: |-
+  Manages repositories within Google Cloud Source Repositories.
+---
+
+# google\_sourcerepos\_repository
+
+Manages repositories within Google Cloud Source Repositories.
```
rosbo: Can you add links to the Google Cloud documentation. I suggest appending the following to this line:
```markdown
For more information, see [the official
documentation](https://cloud.google.com/compute/docs/source-repositories) and
[API](https://cloud.google.com/source-repositories/docs/reference/rest/v1/projects.repos)
```

website/docs/r/sourcerepos_repository.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+
+## Example Usage
+
+This example is the common case of creating a repository within Google Cloud Source Repositores:
```
rosbo: typo: s/Repositores/Repositories

google/resource_source_repos_repository.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+func resourceSourceReposRepository() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceSourceReposRepositoryCreate,
+		Read:   resourceSourceReposRepositoryRead,
+		Delete: resourceSourceReposRepositoryDelete,
+		//Update: not supported,
+
+		Schema: map[string]*schema.Schema{
```
rosbo: Consider adding the `size` as a Computed field. I can see the value for a user who might want to add an output block to their Terraform config to print out the current repo size.

google/resource_source_repos_repository.go
```diff
@@ -0,0 +1,108 @@
...
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceSourceReposRepositoryCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	if _, ok := d.GetOk("project"); ok {
```
rosbo: Remove this block, I guess you have been using this for debugging :)

google/resource_source_repos_repository.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	repoName := d.Get("name").(string)
+	name := buildRepositoryName(project, repoName)
+
+	repo := &sourcerepo.Repo{
+		Name: name,
+	}
+
+	project = "projects/" + project
```
rosbo: Change for parent := "projects/" + project to use the same nomenclature than the official docs.

google/resource_source_repos_repository.go
```diff
@@ -0,0 +1,108 @@
...
+
+	job, err := config.clientSourceRepos.Projects.Repos.Create(project, repo).Do()
```
rosbo: For consistency with the other resources, we call the `job` var `op`.

google/resource_source_repos_repository.go
```diff
@@ -0,0 +1,108 @@
...
+
+func resourceSourceReposRepositoryRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	//project = "projects/" + project
```
rosbo: Remove these comments

alex1x:
Thanks for the comments, I think I addressed everything. Good assist on the "size". I can now get this:

```
$ terraform output
reposize = 12614
```

I didn't change "repos" to "repo" in every single part of the code but I think it should be fine.

rosbo:
Thank you Alex for your contribution!

I pushed a commit to your branch to fix the acceptance test (you forgot to rename the resource there after changing the name.

The tests are now passing.

TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccSourceRepoRepository_basic -timeout 120m
=== RUN   TestAccSourceRepoRepository_basic
--- PASS: TestAccSourceRepoRepository_basic (2.47s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	2.609s

@paddycarver, we will need to Enable the Cloud Repositories API on the HashiCorp CI server. I don't have the permissions to do that. We also need to make sure that the credentials we are using either have a "Owner" role (grants all permissions) or have the "Cloud Repositories Admin" role.

I will merge as soon as @paddycarver  confirms that the CI server is properly configured.

alex1x:
Awesome, thanks for fixing that too.

rosbo:
@danawillow Are you onboard with the naming convention for Source Repo?

`google_sourcerepo_repository` for a repository.

And whenever they add new feature, we could have something like

`google_sourcerepos_webhook`

rileykarson:
Just to chime in with my two cents - I think we should use `sourcerepo` all the time; preferring `google_sourcerepo_webhook` over `google_sourcerepos_webhook`.

My thoughts behind naming it like that are that it is a `google` provider feature from the `sourcerepo` API, and the resource name is `webhook`.

website/google.erb
```diff
@@ -250,6 +250,15 @@
     </ul>
     </li>
 
+    <li<%= sidebar_current("docs-google-sourcerepo") %>>
+    <a href="#">Google Source Repos Resources</a>
```
danawillow: For docs let's use the full name- `Google Source Repositories Resources`

danawillow:
Yup I'm on board for `google_sourcerepo_{feature}`

rosbo:
@rileykarson the "s" was a typo in my comment... I meant using `google_sourcerepo_*`

alex1x:
Woops, I steamrolled over your fix, rosbo, I'll re-add it.

rosbo:
In case you lost my commit when you steamrolled it, I pushed it to my fork and you can take it from there:
https://github.com/rosbo/terraform-provider-google/commit/d12a694c46d0446e838c969ec7df99f497ddcd89

alex1x:
Thanks, should be all good now.

PR #243, opened 2017-07-25
Remove Compute Operation WaitGlobal. 

rileykarson:
Followup to #191.

Remove every occurrence of `computeOperationWaitGlobal`, `computeOperationWaitGlobalTime`, migrating them to `computeOperationWait` and `computeOperationWaitGlobalTime`.

I tested across several modified resources using some basic and some update tests.

Basic:
```
TF_ACC=1 go test ./google -v -run=TestAccComputeBackendBucket_basic -timeout 120m
=== RUN   TestAccComputeBackendBucket_basic
--- PASS: TestAccComputeBackendBucket_basic (27.96s)
=== RUN   TestAccComputeBackendBucket_basicModified
--- PASS: TestAccComputeBackendBucket_basicModified (43.25s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	71.391s
```

```
TF_ACC=1 go test ./google -v -run=TestAccComputeBackendService_basic -timeout 120m
=== RUN   TestAccComputeBackendService_basic
--- PASS: TestAccComputeBackendService_basic (69.36s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	69.552s
```

```
TF_ACC=1 go test ./google -v -run=TestAccComputeFirewall_basic -timeout 120m
=== RUN   TestAccComputeFirewall_basic
--- PASS: TestAccComputeFirewall_basic (52.35s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	52.515s
```

```
TF_ACC=1 go test ./google -v -run=TestAccComputeInstanceTemplate_basic -timeout 120m
=== RUN   TestAccComputeInstanceTemplate_basic
--- PASS: TestAccComputeInstanceTemplate_basic (23.71s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	23.866s
```

Update:
```
TF_ACC=1 go test ./google -v -run=TestAccComputeUrlMap_update_path_matcher -timeout 120m
=== RUN   TestAccComputeUrlMap_update_path_matcher
--- PASS: TestAccComputeUrlMap_update_path_matcher (79.09s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	79.278s
```

```
TF_ACC=1 go test ./google -v -run=TestAccComputeHttpsHealthCheck_update -timeout 120m
=== RUN   TestAccComputeHttpsHealthCheck_update
--- PASS: TestAccComputeHttpsHealthCheck_update (34.30s)
[...]

PR #257, opened 2017-07-27
Revert "Revert metadata item changes (#225)"

selmanj:
This reverts commit 884158d89163298551e663c53f0adf6805f8351c and revives compute_project_metadata as a resource. 

This PR should be immediately followed up with #226.

### DIFF
--- a/google/resource_compute_project_metadata_item.go
+++ b/google/resource_compute_project_metadata_item.go
@@ -1,0 +1,178 @@
+package google
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/compute/v1"
+)
+
+func resourceComputeProjectMetadataItem() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceComputeProjectMetadataItemCreate,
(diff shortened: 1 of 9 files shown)


google/resource_compute_project_metadata_item.go
```diff
@@ -0,0 +1,178 @@
...
+			},
+		).Do()
+
+		if err != nil {
+			return err
+		}
+
+		log.Printf("[DEBUG] SetCommonInstanceMetadata: %d (%s)", op.Id, op.SelfLink)
+
+		return computeOperationWaitGlobal(config, op, project.Name, "SetCommonInstanceMetadata")
```
blamarvt: This should be `computeOperationWait` as per https://github.com/terraform-providers/terraform-provider-google/pull/243

selmanj: Thanks, will fix

selmanj: Done

selmanj:
Thanks. There was only one minor follow-up commit to update the usage of `computeOperationWait` as it was depending on a function that no longer exists ( @rileykarson improved the API to not depend on the type of operation).

Merging.

PR #261, opened 2017-07-28
Add support for legacyAbac to `google_container_cluster`

### LINKED ISSUE #219: Allow creating RBAC google container clusters
Currently the `google_container_cluster` doesn't allow enabling/disabling legacy authorization which is required for creating RBAC enabled clusters.

https://cloud.google.com/container-engine/docs/role-based-access-control

danawillow:
Fixes #219.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -149,6 +149,21 @@ func resourceContainerCluster() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 
+			"legacy_abac": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+					},
+				},
+			},
+
 			"logging_service": {
 				Type:     schema.TypeString,
 				Optional: true,
@@ -409,6 +424,12 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		cluster.Description = v.(string)
 	}
 
+	if _, ok := d.GetOk("legacy_abac"); ok {
+		cluster.LegacyAbac = &container.LegacyAbac{
+			Enabled: d.Get("legacy_abac.0.enabled").(bool),
+		}
+	}
+
 	if v, ok := d.GetOk("logging_service"); ok {
 		cluster.LoggingService = v.(string)
 	}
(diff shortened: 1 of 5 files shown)


google/resource_container_cluster.go
```diff
@@ -149,6 +149,21 @@ func resourceContainerCluster() *schema.Resource {
...
+			"legacy_abac": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:     schema.TypeBool,
+							Required: true,
```
catsby: I suspect we want a default value here

google/resource_container_cluster.go
```diff
@@ -409,6 +424,12 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		cluster.Description = v.(string)
 	}
 
+	if _, ok := d.GetOk("legacy_abac"); ok {
```
catsby: if the user specifies `false`, this block won't get triggered, right? So we'll never send `false`, if I recall correctly

google/resource_container_cluster.go
```diff
@@ -149,6 +149,21 @@ func resourceContainerCluster() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 
+			"legacy_abac": {
```
catsby: Are there any other parameters here? Should we simplify this for the users into a simple toggle?
`enable_legacy_abac` ?

google/resource_container_cluster.go
catsby: In addition, `LegacyAbac.Enabled` is a boolean that's marshaled with `omitempty`, so even if we do go in here, we won't send `false` unless this field is added to the `LegacyAbac.ForceSendFields` slice

google/resource_container_cluster_test.go
```diff
@@ -291,6 +317,7 @@ func testAccCheckContainerCluster(n string) resource.TestCheckFunc {
 			{"description", cluster.Description},
 			{"endpoint", cluster.Endpoint},
 			{"instance_group_urls", igUrls},
+			{"legacy_abac.0.enabled", strconv.FormatBool(cluster.LegacyAbac.Enabled)},
```
catsby: I don't believe this is accurately testing this as the code stands. We can't send false, so we never actually change the value. The test doesn't seem to verify that the value is now false.

google/resource_container_cluster.go
danawillow: Yeah, I was leaving the option open in case any others get added, but realistically I don't think they will so I went ahead and made the change.

google/resource_container_cluster.go
danawillow: Done

google/resource_container_cluster.go
danawillow: Good point. Fixed.

google/resource_container_cluster_test.go
danawillow: Yup, you're right. Fixed.

PR #266, opened 2017-07-28
Improve docs for `dns_record_set` by adding example for SPF record.

### LINKED ISSUE #232: google_dns_record_set
### Terraform Version
Terraform v0.9.11

### Affected Resource(s)
- google_dns_record_set TXT records

```hcl
resource "google_dns_record_set" "foo-TXT" {
   name = "on-site.com."
   managed_zone = "${google_dns_managed_zone.on-site-com.name}"
   type = "TXT"
   ttl = "300"
   rrdatas = ["v=spf1 blah foo bar ]
}
```

### Expected Behavior
Creates a TXT record as a single string

"v=spf1 blah foo bar"

### Actual Behavior
It is quoting everything in between the spaces which create an invalid SPF record

"v=spf1" "blah" "foo" "bar"

When I add the record via the Google Console it is created properly something in terraform is adding the quotes to everything separated by a space

### Steps to Reproduce

1. Create a TXT type record with a long string
2. Run `terraform apply`
3. Look in the google console or use dig to see that the record contains several strings instead of a single string with qutoes.

### References
This was fixed for another provider 
https://github.com/hashicorp/terraform/pull/13260

rosbo:
Fixes #232 

/cc @paddycarver

### DIFF
--- a/website/docs/r/dns_record_set.markdown
+++ b/website/docs/r/dns_record_set.markdown
@@ -10,11 +10,21 @@ description: |-
 
 Manages a set of DNS records within Google Cloud DNS.
 
-## Example Usage
+## Examples Usage
 
-This example is the common case of binding a DNS name to the ephemeral IP of a new instance:
+### Binding a DNS name to the ephemeral IP of a new instance:
 
 ```hcl
+resource "google_dns_record_set" "frontend" {
+  name = "frontend.${google_dns_managed_zone.prod.dns_name}"
+  type = "A"
+  ttl  = 300
+
+  managed_zone = "${google_dns_managed_zone.prod.name}"
+
+  rrdatas = ["${google_compute_instance.frontend.network_interface.0.access_config.0.assigned_nat_ip}"]
+}
+
 resource "google_compute_instance" "frontend" {
   name         = "frontend"
   machine_type = "g1-small"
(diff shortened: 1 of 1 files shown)


website/docs/r/dns_record_set.markdown
```diff
@@ -10,11 +10,21 @@ description: |-
 
 Manages a set of DNS records within Google Cloud DNS.
 
-## Example Usage
+## Examples Usage
```
paddycarver: I think if we're going to pluralise, it would be "Example Uses", but honestly, I'd prefer this either stay "Example Usage" or be just "Examples".

website/docs/r/dns_record_set.markdown
```diff
@@ -34,15 +44,25 @@ resource "google_dns_managed_zone" "prod" {
   name     = "prod-zone"
   dns_name = "prod.mydomain.com."
 }
+```
 
-resource "google_dns_record_set" "frontend" {
+### Adding a SPF record
+
+`\"` must be added around your `rrdatas` for a SPF record. Otherwise `rrdatas` string gets splitted on spaces.
```
paddycarver: nit: "split" > "splitted"

I think this actually applies to any record that can have spaces in it--TXT is another that comes to mind.

website/docs/r/dns_record_set.markdown
rosbo: Ok, I reverted to "Example Usage"

website/docs/r/dns_record_set.markdown
rosbo: SPF is a TXT field. The other fields like A, CNAME, MX should never have spaces.

I wasn't sure where to put the documentation saying this applies to any TXT record... I added it in the Argument Reference section. I felt that if I put it under the SPF record example, people using TXT field not for SPF probably wouldn't check the documentation in this section...

Another option is to add a `~> **Note:**` block at the top before the examples. Do you think this would be preferable instead of the documentation I put in the Argument Reference?

paddycarver: I think what we have as fine, and if we keep seeing the problem, we can escalate to the higher-visibility **Note:** block :)

PR #270, opened 2017-07-31
Add Google Spanner Support (google_spanner_instance)

nickithewatt:
This PR aims to addresses https://github.com/terraform-providers/terraform-provider-google/issues/67 which looks to add Google Spanner Support, specifically adding a new resource `google_spanner_instance`. CC @rileykarson @danawillow 

In order to make these PR's easier to digest and review, instead of creating one big PR with all Spanner support in it, I have created one for each new resource to be added. This is the first which adds support for the foundational `google_spanner_instance` resource. A separate one (https://github.com/terraform-providers/terraform-provider-google/pull/271) adds support for creating the Database within the Instance, i.e. the `google_spanner_database` resource.

Issue #67, opened 2017-06-13
Support Google Cloud Spanner


hashibot:
_This issue was originally opened by @danielcompton as hashicorp/terraform#12730. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

It would be handy to have support for Google Cloud Spanner. This would be useful in production, but also in development, as it is quite expensive to run, so being able to quickly startup and tear down a Cloud Spanner DB would be helpful.

PR #271, opened 2017-07-31
Add Google Spanner Support (google_spanner_database)

nickithewatt:
This PR aims to addresses #67 which looks to add Google Spanner Support, specifically adding a new resource google_spanner_database. CC @rileykarson @danawillow

In order to make these PR's easier to digest and review, instead of creating one big PR with all Spanner support in it, I have created one for each new resource to be added. This is the 2nd which adds support for the `google_spanner_database` resource. A separate one (PR https://github.com/terraform-providers/terraform-provider-google/pull/270) added the foundational support for creating the Instance, i.e. the `google_spanner_instance` resource.

This PR does build off PR https://github.com/terraform-providers/terraform-provider-google/pull/270 as it uses the the `google_spanner_instance` resource so that should probably be reviewed first, then this one. Alternatively if you want to review all the Spanner support, this PR will have that.


google/resource_spanner_database.go
```diff
@@ -0,0 +1,239 @@
...
+			"project": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"ddl": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
```
rosbo: There is an update method for this. If you decide not to implement it in this PR, please file an issue and add a TODO here link to the issue.

google/resource_spanner_database.go
```diff
@@ -0,0 +1,239 @@
...
+		return err
+	}
+
+	_, err = config.clientSpanner.Projects.Instances.Databases.Get(
+		id.databaseUri()).Do()
+	if err != nil {
+		return handleNotFoundError(err, d, fmt.Sprintf("Spanner database %q", id.databaseUri()))
+	}
+
+	d.SetId(id.terraformId())
```
rosbo: No need to set the id here, it is set either in the create method or the import method.

website/docs/r/spanner_instance.html.markdown
```diff
@@ -0,0 +1,68 @@
...
+
+* `num_nodes` - (Optional, Computed) The number of nodes allocated to this instance.
+   Defaults to `1`. This can be updated after creation.
+
+* `project` - (Optional) The project in which the resource belongs. If it
+    is not provided, the provider project is used.
+
+## Attributes Reference
+
+No additional attributes are computed other than those defined above.
```
rosbo: Add the computed `state` field here once you added it.

google/resource_spanner_database.go
```diff
@@ -0,0 +1,239 @@
...
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"ddl": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
```
rosbo: Please, add the `Computed` state field.

google/resource_spanner_database.go
nickithewatt: I am not sure an update makes sense here? My understanding is that the DDL update here is more additive style statements to apply to the database, rather than the definitive list of statements, i.e. could include update statements as well. Without support for clauses like `DROP IF EXISTS`, this could be quite hard to make idempotent.

nickithewatt:
@rosbo Thanks again for review. Have made one comment re DDL update (see above). Rebased off the spanner-support-instance branch and made other changes as requested. Let me know if there is anything else that needs doing on this :)

google/resource_spanner_database.go
rosbo: You are right. We should leave the ForceNew.

rosbo:
make testacc TESTARGS="-run TestAccSpannerDatabase_" TEST="./google"
TF_ACC=1 go test ./google -v -run TestAccSpannerDatabase_ -timeout 120m
=== RUN   TestAccSpannerDatabase_importInstanceDatabase
--- PASS: TestAccSpannerDatabase_importInstanceDatabase (24.50s)
=== RUN   TestAccSpannerDatabase_importProjectInstanceDatabase
--- PASS: TestAccSpannerDatabase_importProjectInstanceDatabase (24.20s)
=== RUN   TestAccSpannerDatabase_basic
--- PASS: TestAccSpannerDatabase_basic (24.00s)
=== RUN   TestAccSpannerDatabase_basicWithInitialDDL
--- PASS: TestAccSpannerDatabase_basicWithInitialDDL (24.17s)
=== RUN   TestAccSpannerDatabase_duplicateNameError
--- PASS: TestAccSpannerDatabase_duplicateNameError (24.31s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	121.315s

PR #274, opened 2017-08-01
Use a unique key when running project_metadata_item import tests

selmanj:
CI runs tests concurrently and will interfere with each other with the
current global key name.


danawillow:
👍

PR #275, opened 2017-08-01
Fix disk type’Malformed URL’ error

sl1pm4t:
The API expects the disk type to be a SelfLink URL, but the disk type 
name was being used (e.g. “pd-ssd”). This resulted in the following error:

```
Error applying plan:

1 error(s) occurred:

* google_compute_instance.foo: 1 error(s) occurred:

* google_compute_instance.foo: Error creating instance: googleapi: Error 400: Invalid value for field 'resource.disks[0].initializeParams.diskType': 'pd-ssd'. The URL is malformed., invalid
```

### DIFF
--- a/google/resource_compute_instance_test.go
+++ b/google/resource_compute_instance_test.go
@@ -289,6 +289,27 @@ func TestAccComputeInstance_bootDisk_source(t *testing.T) {
 	})
 }
 
+func TestAccComputeInstance_bootDisk_type(t *testing.T) {
+	var instance compute.Instance
+	var instanceName = fmt.Sprintf("instance-test-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeInstanceDestroy,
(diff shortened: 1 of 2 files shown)


sl1pm4t:
Please review @danawillow 
Thanks!

danawillow:
Thanks @sl1pm4t! Wish we had caught this with a test instead :-/ Mind adding (or modifying) a test that sets `type`?

sl1pm4t:
Sure thing @danawillow

google/resource_compute_instance_test.go
```diff
@@ -1370,6 +1409,27 @@ resource "google_compute_instance" "foobar" {
...
 
+func testAccComputeInstance_bootDisk_type(instance string) string {
+	return fmt.Sprintf(`
+resource "google_compute_instance" "foobar" {
+	name         = "%s"
+	machine_type = "n1-standard-1"
+	zone         = "us-central1-a"
+
+	boot_disk {
+    initialize_params {
```
danawillow: Can you fix the indentation here? It should be tabs.

sl1pm4t:
Oops - that test doesn't work.

danawillow:
Thanks, looks good!

PR #276, opened 2017-08-01
compute-component api has been replaced by compute, update tests

danawillow:


danawillow:
Ah yeah that's true. I'm going to go ahead and merge. They're test-only changes anyway so we're not going to cause any harm. I hit some quota limits when trying to run all of them but the ones I was able to run succeeded.

Issue #277, opened 2017-08-01
Typo in the docs  (google provider)


hashibot:
_This issue was originally opened by @roman-cnd as hashicorp/terraform#15681. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

There are some typos on this page: https://www.terraform.io/docs/providers/google/r/sql_database_instance.html#settings

For example, "True iff"

PR #279, opened 2017-08-02
docs: fix typos in sql-database-instance

grubernaut:
Fixes: #277


danawillow:
I assume the original author did actually mean "iff" (if and only if), but I'm happy with these changes for consistency (since we usually assume people will understand what we mean by "if")

catsby:
I deployed this change 👍

PR #280, opened 2017-08-02
Document Beta Features in index.html

rileykarson:


ctavan:
How about replacing the note in https://github.com/terraform-providers/terraform-provider-google/blob/master/website/docs/r/sql_database_instance.html.markdown with a simple link as well now?

catsby:
Hey @rileykarson / @danawillow if it's not too much trouble, it would be awesome if any `beta` attribute/resource had a link to this description. Not a blocker 👍

rileykarson:
@catsby: Done.

PR #282, opened 2017-08-02
Add Beta support & Beta feature deny to google_compute_firewall

### LINKED ISSUE #148: Firewall deny on match
- google_compute_firewall

Is there a way to specify firewall deny on match rule?

Thanks in advance 😄

rileykarson:
- Convert `google_compute_firewall` to a versioned resource with Beta support.
- Add support for the Beta field `deny` to `google_compute_firewall`

`deny` is not able to be updated by the API yet.

I'm not sure what to do here docs-wise where a Beta feature makes a GA field no longer Required; one of `allow` or `deny` is, so I kept `allow` required in docs and said `deny` can replace it. Let me know your thoughts.

Fixes #148.
Related to #93

### DIFF
--- a/google/compute_shared_operation.go
+++ b/google/compute_shared_operation.go
@@ -5,6 +5,25 @@ import (
 	"google.golang.org/api/compute/v1"
 )
 
+func computeSharedOperationWait(config *Config, op interface{}, project string, activity string) error {
+	return computeSharedOperationWaitTime(config, op, project, 4, activity)
+}
+
+func computeSharedOperationWaitTime(config *Config, op interface{}, project string, minutes int, activity string) error {
+	if op == nil {
+		panic("Attempted to wait on an Operation that was nil.")
+	}
+
+	switch op.(type) {
+	case *compute.Operation:
+		return computeOperationWaitTime(config, op.(*compute.Operation), project, activity, minutes)
+	case *computeBeta.Operation:
+		return computeBetaOperationWaitGlobalTime(config, op.(*computeBeta.Operation), project, activity, minutes)
+	default:
+		panic("Attempted to wait on an Operation of unknown type.")
+	}
+}
+
 func computeSharedOperationWaitZone(config *Config, op interface{}, project string, zone, activity string) error {
 	return computeSharedOperationWaitZoneTime(config, op, project, zone, 4, activity)
 }
--- a/google/resource_compute_firewall.go
+++ b/google/resource_compute_firewall.go
@@ -71,12 +71,14 @@ func resourceComputeFirewall() *schema.Resource {
 						"protocol": {
 							Type:     schema.TypeString,
 							Required: true,
+							ForceNew: true,
 						},
 
 						"ports": {
 							Type:     schema.TypeList,
 							Optional: true,
 							Elem:     &schema.Schema{Type: schema.TypeString},
+							ForceNew: true,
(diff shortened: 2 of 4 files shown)


google/resource_compute_firewall.go
```diff
@@ -302,11 +457,12 @@ func resourceFirewall(
 	}
 
 	// Build the firewall parameter
-	return &compute.Firewall{
+	return &computeBeta.Firewall{
 		Name:         d.Get("name").(string),
 		Description:  d.Get("description").(string),
 		Network:      network.SelfLink,
```
danawillow: Does this self link need to be converted to v1? Also, do we actually need to read the network at both versions or can we just read it at v1 since all we need is the self link?

google/resource_compute_firewall_test.go
```diff
@@ -82,6 +82,26 @@ func TestAccComputeFirewall_noSource(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeFirewallDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccComputeFirewall_denied(networkName, firewallName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeFirewallExists("google_compute_firewall.foobar", &firewall),
```
danawillow: Do you want to check that the firewall also has the deny properties we expect?

google/compute_shared_operation.go
```diff
@@ -5,6 +5,25 @@ import (
 	"google.golang.org/api/compute/v1"
 )
 
+func computeSharedOperationWait(config *Config, op interface{}, project string, activity string) error {
```
rileykarson: For sure - what I'd like to do is convert the `Operation` object to `v1` and call the `v1` waits, like in [#187](https://github.com/terraform-providers/terraform-provider-google/pull/187/files#diff-ef92ceb02a5a2c0c742317597b0bc870R104)

Otherwise, we can just copy `compute_operation.go` again.

google/resource_compute_firewall.go
```diff
@@ -226,13 +333,21 @@ func resourceComputeFirewallDelete(d *schema.ResourceData, meta interface{}) err
 	}
 
 	// Delete the firewall
-	op, err := config.clientCompute.Firewalls.Delete(
-		project, d.Id()).Do()
-	if err != nil {
-		return fmt.Errorf("Error deleting firewall: %s", err)
+	var op interface{}
+	switch computeApiVersion {
```
rileykarson: I'm honestly not sure; for now, for completeness for now, I'd like to perform all operations at the resource's API version; we'll be able to bring deletes of versioned resources either down to `v1` or up to `v0beta` every time later, unless we adopt a multiversioned service generated by #240 which makes this mostly a non-issue.

google/resource_compute_firewall.go
rileykarson: This variable never gets stored in state, since we do that in `Read`.

We can just read at `v1`; done.

google/resource_compute_firewall_test.go
rileykarson: Done.

PR #293, opened 2017-08-04
Fix error if container_node_pool deleted out of band

sl1pm4t:
Prior to this change it was possible for Terraform to error during plan / apply with the following message if the Node Pool had been deleted out of band:
```
Error 404: The resource "node pool \"foo\" not found"
```


sl1pm4t:
Please review @danawillow - thanks!

danawillow:
Looks good, thanks!

PR #290, opened 2017-08-03
google_compute_instance can specified the subnetwork using a self_link

rosbo:
Fixes #247 

Currently, you can only specify the `subnetwork` for `google_compute_instance` using the `subnetwork` name. See issue #247 for more context.

This adds the support for both `name` and `self_link` when specifying the `subnetwork` bringing consistency with the way other network/subnetwork fields works.

I also added support for reading the network_interface fields from GCP. 

cc/ @danawillow

PR #289, opened 2017-08-03
Add a test that multiple nics work in google_compute_instance.

rileykarson:
It looks like our code already supports multiple `network_interface` fields on `google_compute_instance`. 
Let's document that the API accepts multiple interfaces, although it is a Beta feature.

Fixes #82

PR #294, opened 2017-08-04
Remove ConflictsWith for network and subnetwork.

rosbo:
We support more than one network_interface. 
ConflictsWith doesn't support this case. 

I added the ConflictsWith an hour ago and I assumed for some reason that network_interface had always a size of 1. See #290. This is wrong, network_interface size is greater than one. @rileykarson added a test that will prevent regression on this in #289.

@danawillow @rileykarson

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -284,7 +284,6 @@ func resourceComputeInstance() *schema.Resource {
 							Computed:         true,
 							ForceNew:         true,
 							DiffSuppressFunc: linkDiffSuppress,
-							ConflictsWith:    []string{"network_interface.0.subnetwork", "network_interface.0.subnetwork_project"},
 						},
 
 						"subnetwork": &schema.Schema{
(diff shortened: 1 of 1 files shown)


rileykarson:
Does `ConflictsWith` compare with the root of the element or is it relative to the current location? I'm curious if we can just mark `subnetwork` and `subnetwork_project` and have it work instead of needing an apply-time check.

rosbo:
It does not support this case. See https://github.com/hashicorp/terraform/issues/11101

It only looks if a value is defined based on the ConflictsWith path you pass in.
https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go?utf8=%E2%9C%93#L1175

if you say `network_interface.subnetwork` in `ConflictsWith`, this field will never be set and therefor, will never conflict. `network_interface.0.subnetwork...network_interface.x.subnetwork` are set, never the non-indexed version.

The schema validation doesn't fail on `network_interface.subnetwork`. This is a bug in the internal validation logic. We use the non-indexed path in one place in our provider. It does nothing and is a lie. 
https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go?utf8=%E2%9C%93#L590

We are misusing the ConflictsWith in one place in the google provider. I filled an issue for this #295

rosbo:
Will do.

PR #300, opened 2017-08-07
Update compute_health_check.html.markdown

wjimenez5271:
Mention limitation of non HTTP health checks with TargetPools to hopefully save other people getting stuck like I did :)

### DIFF
--- a/website/docs/r/compute_health_check.html.markdown
+++ b/website/docs/r/compute_health_check.html.markdown
@@ -10,7 +10,7 @@ description: |-
 
 Manages a health check within GCE. This is used to monitor instances
 behind load balancers. Timeouts or HTTP errors cause the instance to be
-removed from the pool. This cannot be used with GCE `TargetPools` as they only support `HttpHealthChecks`. For more information, see [the official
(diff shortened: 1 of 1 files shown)


website/docs/r/compute_health_check.html.markdown
```diff
@@ -10,7 +10,7 @@ description: |-
 
 Manages a health check within GCE. This is used to monitor instances
 behind load balancers. Timeouts or HTTP errors cause the instance to be
-removed from the pool. For more information, see [the official
+removed from the pool. This cannot be used with GCE `TargetPools` as they only support `HttpHealthChecks`. For more information, see [the official
```
selmanj: I agree that this behavior is confusing.

However it would be nicer to make the documentation match the public API closer. Maybe something like:

```markdown
Be aware that if you're managing a [Network Load Balancer](https://cloud.google.com/compute/docs/load-balancing/network/), you must use one of `google_compute_http_health_check` or `google_compute_https_health_check` instead of this resource. See [Legacy Health Checks](https://cloud.google.com/compute/docs/load-balancing/health-checks#legacy_health_checks) for more information.
```

wjimenez5271: Yes, even better!

selmanj:
Sorry, I know you made the requested changes, but after talking to others it makes more sense to document this on the TargetPool resource documentation itself.

Specifically in `compute_target_pool.html.markdown`, under Argument Reference at the `health_checks` parameter.

rosbo:
I will update the target pool documentation in a separate PR.

Issue #308, opened 2017-08-08
Add ability to change monitoring_service on google_container_cluster without recreation


drzero42:
### Terraform Version
- Core: 0.10.0
- Provider: 0.1.2

### Affected Resource(s)
Please list the resources as a list, for example:
- google_container_cluster

As mentioned during the discussion on #285, it is now possible to change the Monitoring Service for a Container Cluster without recreating it. This means the monitoring_service field on google_container_cluster resources should no longer be ForceNew: true, but should instead have a section in the update function, which either uses [projects.zones.clusters.monitoring](https://cloud.google.com/container-engine/reference/rest/v1/projects.zones.clusters/monitoring) or [projects.zones.clusters.update](https://cloud.google.com/container-engine/reference/rest/v1/projects.zones.clusters/update) with desiredMonitoringService field to change the value.

PR #309, opened 2017-08-08
Add ability to update monitoring_service on google_container_cluster

drzero42:
My proposed fix for #308.
I'm not entirely sure how to write a test for it - I was hoping to see how updating node_version was tested, but alas there does not seem to be a test for that.


drebes:
For a similar test, check either ``TestAccContainerCluster_withAdditionalZones`` or ``TestAccContainerCluster_withLegacyAbac``. Testing updates is basically having multiple steps on the test, one with the initial config, the other with the updated config, and checking that the resource changed accordingly. Using the default ``testAccCheckContainerCluster`` check from the resource should be OK as it generally just confirms that the attributes set in the resource and the parameters read from the API match, and it already tests for ``monitoring_service``.

danawillow:
Test for updating versions: https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_container_cluster_test.go#L142

website/docs/r/container_cluster.html.markdown
```diff
@@ -13,9 +13,20 @@ Creates a GKE cluster. For more information see
 and
 [API](https://cloud.google.com/container-engine/reference/rest/v1/projects.zones.clusters).
 
-!> **Warning:** Due to limitations of the API, all arguments except
-`node_version` are non-updateable. Changing any will cause recreation of the
-whole cluster!
+!> **Warning:** Due to limitations of the API, the following arguments are non-updateable:
```
danawillow: Honestly I'm not sure this warning is particularly helpful, and we don't bother showing a similar one for most resources. Do you think it's ok to remove overall or do you think it's better to leave it in?

danawillow:
@drzero42 are you still working on this? Otherwise I'd like to take it over.

drzero42:
@danawillow Kind of busy with other things, so you go ahead and take it :)

danawillow:
Closing in favor of #598.

PR #310, opened 2017-08-08
Add subnetwork secondary ip ranges beta feature

### LINKED ISSUE #287: Add support for google_compute_subnetwork IP Aliasing Beta feature
https://cloud.google.com/compute/docs/reference/beta/subnetworks#secondaryIpRanges
https://cloud.google.com/compute/docs/alias-ip/

#93

rosbo:
Fixes #287 

peer-programmed w/ @rileykarson on this PR.

cc/ @danawillow

### DIFF
--- a/google/resource_compute_subnetwork.go
+++ b/google/resource_compute_subnetwork.go
@@ -7,8 +7,12 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
+	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
 )
+
+var SubnetworkBaseApiVersion = v1
+var SubnetworkVersionedFeatures = []Feature{}
 
 func resourceComputeSubnetwork() *schema.Resource {
 	return &schema.Resource{
@@ -76,18 +80,8 @@ func resourceComputeSubnetwork() *schema.Resource {
 	}
 }
 
-func createSubnetID(s *compute.Subnetwork) string {
-	return fmt.Sprintf("%s/%s", s.Region, s.Name)
-}
-
-func splitSubnetID(id string) (region string, name string) {
-	parts := strings.Split(id, "/")
-	region = parts[0]
-	name = parts[1]
-	return
-}
-
 func resourceComputeSubnetworkCreate(d *schema.ResourceData, meta interface{}) error {
+	computeApiVersion := getComputeApiVersion(d, SubnetworkBaseApiVersion, SubnetworkVersionedFeatures)
 	config := meta.(*Config)
 
 	region, err := getRegion(d, config)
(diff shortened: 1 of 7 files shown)


google/resource_compute_subnetwork.go
```diff
@@ -68,6 +77,26 @@ func resourceComputeSubnetwork() *schema.Resource {
 				Optional: true,
 			},
 
+			"secondary_ip_range": &schema.Schema{
```
rileykarson: I'm not sure if it will ever make a difference, but this should probably be marked `ForceNew` as well as it's children.

rosbo: Done

rileykarson:
LGTM :+1:, not sure if we should get another reviewer or not here

```
TF_ACC=1 go test ./google -v -run=TestAccComputeSubnetwork -timeout 120m
=== RUN   TestAccComputeSubnetwork_importBasic
--- PASS: TestAccComputeSubnetwork_importBasic (63.93s)
=== RUN   TestAccComputeSubnetwork_basic
--- PASS: TestAccComputeSubnetwork_basic (64.90s)
=== RUN   TestAccComputeSubnetwork_update
--- PASS: TestAccComputeSubnetwork_update (61.32s)
=== RUN   TestAccComputeSubnetwork_secondaryIpRanges
--- PASS: TestAccComputeSubnetwork_secondaryIpRanges (74.19s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	264.553s
```

I'll follow up with another PR cleaning up Beta operations after this.

rosbo:
I added @danawillow to have another pair of eyes looking at this PR.

google/resource_compute_subnetwork.go
```diff
@@ -256,3 +333,41 @@ func resourceComputeSubnetworkImportState(d *schema.ResourceData, meta interface
 
 	return []*schema.ResourceData{d}, nil
 }
+
+func createBetaSubnetID(s *computeBeta.Subnetwork) string {
```
danawillow: This and `createSubnetID` should probably live in the same file

google/self_link_helpers.go
```diff
@@ -28,6 +28,17 @@ func compareSelfLinkRelativePaths(k, old, new string, d *schema.ResourceData) bo
 	return false
 }
 
+// This method should only be used if the self link references a global resource.
+func compareSelfLinkResourceNames(k, old, new string, d *schema.ResourceData) bool {
```
danawillow: Wait I'm confused why `compareSelfLinkRelativePaths` doesn't work on global resources.

Network self links look like https://www.googleapis.com/compute/v1/projects/project/global/networks/network, so comparing the relative one would be comparing `project/global/networks/network` which should work just fine, right?

google/resource_compute_subnetwork.go
rosbo: Moved `createSubnetID` to `resource_compute_subnetwork.go`

google/self_link_helpers.go
rosbo: The `network` field can be a `self_link` or the name of the network.
This method works when the field can be either `name` or `self_link`.

If we were using `compareSelfLinkRelativePaths`, then we would get a diff if we specify the network using its name in the config.

`mynetwork != project/global/networks/mynetwork`

I updated the comment to make it clearer.

danawillow: Should the function also be renamed since it's not necessarily comparing self links?

rosbo: What about `compareGlobalSelfLinkOrResourceName`?

danawillow: sure!

rosbo: Renamed.

Issue #263, opened 2017-07-28
Add InstanceType to Bigtable


danawillow:
Apparently the go client library now supports InstanceType, so we can probably add it to our Bigtable resource.

https://github.com/GoogleCloudPlatform/google-cloud-go/commit/7a4ba9f439fbc50061834a4063b57cf7222ba83f

cc @rileykarson

dasch:
This is critical to setting up developer projects, :+1: 💯

rileykarson:
Closed with #313.

PR #313, opened 2017-08-09
Add support for instance_type to google_bigtable_instance.

rileykarson:
Add support for `instance_type` to `google_bigtable_instance`.

edit: linking #263 

---

Breaking change: `num_nodes` needs to be unset for `DEVELOPMENT` instances, and >3 for `PRODUCTION` ones. I removed the default, which will cause diffs for users that relied on the implicit `3` nodes.

We could avert it by keeping the default - this means `DEVELOPMENT` users will need to explicitly specify `0` instead of not needing to specify `num_nodes` at all.


google/resource_bigtable_instance.go
```diff
@@ -44,11 +44,17 @@ func resourceBigtableInstance() *schema.Resource {
 			},
 
 			"num_nodes": {
-				Type:         schema.TypeInt,
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
```
rosbo: Since the number of nodes is `ForceNew`, we don't want to force our users to delete and recreate their production database.

I would rather have our users specify num_nodes = 1 for DEVELOPMENT.

If you don't send num_nodes to the server, does it set 1 for dev and 3 for production by default? If so, we could use `Computed = True` for num_nodes.

rileykarson: If we send a non-zero value to the server for `num_nodes` for a `DEVELOPMENT` instance, the request fails. There's no server default when you don't send a `num_nodes` for a production one; the request fails.

Users won't actually need to recreate their instances; the `num_nodes` of 3 is preserved in state, and they will just need to update their config when it diffs.

rosbo: Can we leave the `num_nodes` default to 3 and just don't send the property when creating a `DEVELOPMENT` instance? This way, we don't create churn for our users...

rileykarson: Doing things like that breaks the 1:1 mapping from config -> the API, and imo isn't worth it if we ever need to correct later. For example, if we were to gain the ability to read `num_nodes`, we would either need to conditionally read it or would see diffs in the future.

I think it's worth a breaking change here because the number of users is small (and I expect the # of users that relied on the implicit default is smaller), the change is easy to make, and it means we'll be tracking the API as accurately as we can until the next client update.

rileykarson:
```
TF_ACC=1 go test ./google -v -run=TestAccBigtable -timeout 120m
=== RUN   TestAccBigtableInstance_basic
--- PASS: TestAccBigtableInstance_basic (4.91s)
=== RUN   TestAccBigtableInstance_development
--- PASS: TestAccBigtableInstance_development (4.31s)
=== RUN   TestAccBigtableTable_basic
--- PASS: TestAccBigtableTable_basic (6.10s)
=== RUN   TestAccBigtableTable_splitKeys
--- PASS: TestAccBigtableTable_splitKeys (6.73s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	22.203s
```

PR #315, opened 2017-08-09
Add support for Google RuntimeConfig

### LINKED ISSUE #236: Support RuntimeConfig
RuntimeConfig is a GCP API for storing key/value data for a deployment/running application. The most basic use case looks like project metadata, but unlike metadata the API has fine grained IAM roles. It looks like a pretty slick way to inject deployment information into an environment.

### Example Code
```hcl
// psuedo-HCL with non-existent resources!
resource "google_runtime_config" "deployment" {
   name = "my_deployment"
}

resource "google_runtime_config_variable" "load_balancer_ip" {
   config = "${google_runtime_config.deployment.self_link}"
   name = "loadBalancerIp"
   text    = "${google_compute_address.external.address}"
}

resource "google_runtime_config_variable" "ssl_private_key" {
   config = "${google_runtime_config.deployment.self_link}"
   name = "sslPrivateKey"
   value  = "${base64(var.ssl_cert_private_key)}"
}

```

### References
- [rcloadenv](https://github.com/GoogleCloudPlatform/rcloadenv) - Good example app for using RuntimeConfig
- [GCP Docs](https://cloud.google.com/deployment-manager/runtime-configurator/create-and-delete-runtimeconfig-resources)

selmanj:
This allows users to create/manage Google RuntimeConfig resources and
variables. More information here:
https://cloud.google.com/deployment-manager/runtime-configurator/

Closes #236

### DIFF
--- a/google/resource_runtimeconfig_config.go
+++ b/google/resource_runtimeconfig_config.go
@@ -8,11 +8,7 @@ import (
 	"google.golang.org/api/runtimeconfig/v1beta1"
 )
 
-var runtimeConfigFullName *regexp.Regexp
-
-func init() {
-	runtimeConfigFullName = regexp.MustCompile("^projects/([^/]+)/configs/(.+)$")
-}
+var runtimeConfigFullName *regexp.Regexp = regexp.MustCompile("^projects/([^/]+)/configs/(.+)$")
 
 func resourceRuntimeconfigConfig() *schema.Resource {
 	return &schema.Resource{
--- a/google/resource_runtimeconfig_config.go
+++ b/google/resource_runtimeconfig_config.go
@@ -98,11 +98,13 @@ func resourceRuntimeconfigConfigRead(d *schema.ResourceData, meta interface{}) e
 func resourceRuntimeconfigConfigUpdate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
+	// Update works more like an 'overwrite' method - we build a new runtimeconfig.RuntimeConfig struct and it becomes
+	// the new config. This means our Update logic looks an awful lot like Create (and hence, doesn't use
+	// schema.ResourceData.hasChange()).
 	fullName := d.Id()
 	runtimeConfig := runtimeconfig.RuntimeConfig{
 		Name: fullName,
 	}
-	// Update the description (currently its the only thing that could have changed)
 	if v, ok := d.GetOk("description"); ok {
 		runtimeConfig.Description = v.(string)
 	}
--- a/google/resource_runtimeconfig_config.go
+++ b/google/resource_runtimeconfig_config.go
@@ -1,0 +1,148 @@
+package google
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/runtimeconfig/v1beta1"
+)
+
+var runtimeConfigFullName *regexp.Regexp
+
+func init() {
+	runtimeConfigFullName = regexp.MustCompile("^projects/([^/]+)/configs/(.+)$")
+}
+
+func resourceRuntimeconfigConfig() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceRuntimeconfigConfigCreate,
+		Read:   resourceRuntimeconfigConfigRead,
+		Update: resourceRuntimeconfigConfigUpdate,
+		Delete: resourceRuntimeconfigConfigDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateGCPName,
+			},
+
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"project": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceRuntimeconfigConfigCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	name := d.Get("name").(string)
+	fullName := resourceRuntimeconfigFullName(project, name)
+	runtimeConfig := runtimeconfig.RuntimeConfig{
+		Name: fullName,
+	}
+
+	if val, ok := d.GetOk("description"); ok {
+		runtimeConfig.Description = val.(string)
+	}
+
+	_, err = config.clientRuntimeconfig.Projects.Configs.Create("projects/"+project, &runtimeConfig).Do()
+
+	if err != nil {
+		return err
+	}
+	d.SetId(fullName)
+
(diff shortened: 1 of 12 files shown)


website/docs/r/runtimeconfig_config.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+description: |-
+  Manages a RuntimeConfig resource in Google Cloud.
+---
+
+# google\_runtimeconfig\_config
+
+Manages a RuntimeConfig resource in Google Cloud. For more information, see the
+[official documentation](https://cloud.google.com/deployment-manager/runtime-configurator/),
+or the
+[JSON API](hhttps://cloud.google.com/deployment-manager/runtime-configurator/reference/rest/).
```
rosbo: remove extra "h"

website/docs/r/runtimeconfig_variable.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+```hcl
+resource "google_runtimeconfig_config" "my-runtime-config" {
+ 	name = "my-service-runtime-config"
+ 	description = "Runtime configuration values for my service"
+}
+
+resource "google_runtimeconfig_variable" "environment" {
+	parent = "${google_runtimeconfig_config.my-runtime-config.name}"
+	name = "environment"
```
rosbo: I think it would be a more realistic example if you were storing a value in a hierarchy instead.
```hcl
...
name = "prod-variables/hostname"
text = "example.com"
```

website/docs/r/runtimeconfig_variable.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+description: |-
+  Manages a RuntimeConfig variable in Google Cloud.
+---
+
+# google\_runtimeconfig\_variable
+
+Manages a RuntimeConfig variable in Google Cloud. For more information, see the
+[official documentation](https://cloud.google.com/deployment-manager/runtime-configurator/),
+or the
+[JSON API](hhttps://cloud.google.com/deployment-manager/runtime-configurator/reference/rest/).
```
rosbo: same here, remove extra "h"

google/resource_runtimeconfig_config.go
```diff
@@ -0,0 +1,148 @@
...
+	"regexp"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/runtimeconfig/v1beta1"
+)
+
+var runtimeConfigFullName *regexp.Regexp
+
+func init() {
+	runtimeConfigFullName = regexp.MustCompile("^projects/([^/]+)/configs/(.+)$")
```
rosbo: I don't think you need this init block, you can simply do:

```go
var runtimeConfigFullName *regexp.Regexp = regexp.MustCompile("^projects/([^/]+)/configs/(.+)$")
```

google/resource_runtimeconfig_config.go
```diff
@@ -0,0 +1,148 @@
...
+		return err
+	}
+
+	name := d.Get("name").(string)
+	fullName := resourceRuntimeconfigFullName(project, name)
+	runtimeConfig := runtimeconfig.RuntimeConfig{
+		Name: fullName,
+	}
+
+	if val, ok := d.GetOk("description"); ok {
```
rosbo: I think you can simply this by removing the if-check and keeping only:

```go
runtimeConfig.Description = d.Get("description").(string)
```

If description is not set, `d.Get("description").(string)` will return the default value `""` (empty string).

google/resource_runtimeconfig_config.go
```diff
@@ -0,0 +1,148 @@
...
+		runtimeConfig.Description = val.(string)
+	}
+
+	_, err = config.clientRuntimeconfig.Projects.Configs.Create("projects/"+project, &runtimeConfig).Do()
```
rosbo: No waiting on the operation to complete? Is this synchronous?

google/resource_runtimeconfig_config.go
```diff
@@ -0,0 +1,148 @@
...
+
+func resourceRuntimeconfigConfigUpdate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	fullName := d.Id()
+	runtimeConfig := runtimeconfig.RuntimeConfig{
+		Name: fullName,
+	}
+	// Update the description (currently its the only thing that could have changed)
+	if v, ok := d.GetOk("description"); ok {
```
rosbo: use `d.HasChange`. I know right now this is the only field that may change but if we add more fields, it will be easier to extend.

rosbo: I believe this logic is wrong. If you set the description, apply, then unset it, and re-apply, then the description won't be erased. You only change the description if it is set but unsetting it is also something the user might want to do. using hasChange instead like suggested in my previous comment will solve this.

google/resource_runtimeconfig_variable.go
```diff
@@ -0,0 +1,190 @@
...
+
+			"project": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"value": {
+				Type:     schema.TypeString,
+				Optional: true,
```
rosbo: add a ConflictsWith with the "text" field since only one or the other can be set.

website/docs/r/runtimeconfig_variable.html.markdown
selmanj: Oops, done

website/docs/r/runtimeconfig_config.html.markdown
selmanj: Done

website/docs/r/runtimeconfig_variable.html.markdown
selmanj: Fair enough, using a '/' in example now

google/resource_runtimeconfig_config.go
selmanj: Didn't know you could do this; for some reason I thought only primitives could be initialized in this manner (I guess this works because it's a var rather than a constant?)

google/resource_runtimeconfig_config.go
selmanj: I didn't do this because I feel like `GetOk` reads better, although I agree they are equivalent.

google/resource_runtimeconfig_config.go
selmanj: It is!

google/resource_runtimeconfig_config.go
rosbo: correct, a const can't be set at runtime

google/resource_runtimeconfig_config.go
selmanj: Spoke in person; this update API is confusing in that it's more like 'overwrite'. Added comments to help explain this.

selmanj: Same as above; update is like overwrite.  Added comments.

google/resource_runtimeconfig_variable.go
selmanj: Didn't know this one! Added!

PR #320, opened 2017-08-10
Changed zone name to be read correctly in google_compute_disk's detach logic

### LINKED ISSUE #112: Google compute engine instance and disk referencing error
_This issue was originally opened by @nodesocket as hashicorp/terraform#15181. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Think I discovered a bug using `google_compute_disk` and `google_compute_instance` referencing resources.

Create a disk using a resource block like:

```
resource "google_compute_disk" "blog" {
    name  = "blog"
    type  = "pd-ssd"
    zone  = "us-west1-a"
    snapshot = "some-snapshot-here"
    size = 10
}
```

Then create an instance referencing the above disk:

```
resource "google_compute_instance" "blog" {
    name = "blog"
    machine_type = "f1-micro"
    zone = "${google_compute_disk.blog.zone}"

    tags = []

    disk {
        disk = "${google_compute_disk.blog.name}"
        auto_delete = true
    }

    network_interface {
        network = "default"
        access_config {
            # ephemeral external ip address
        }
    }

    metadata {
    }

    service_account {
        scopes = ["https://www.googleapis.com/auth/servicecontrol",
                  "https://www.googleapis.com/auth/service.management.readonly",
                  "https://www.googleapis.com/auth/monitoring.write",
                  "https://www.googleapis.com/auth/trace.append",
                  "https://www.googleapis.com/auth/devstorage.read_only",
                  "https://www.googleapis.com/auth/cloud.useraccounts.readonly"]
    }

    scheduling {
        preemptible = false
        automatic_restart = true
    }
}
```

After the `apply` finishes, run `apply` again. You will see the following error:

```
Error applying plan:

1 error(s) occurred:

* google_compute_disk.blog (destroy): 1 error(s) occurred:

* google_compute_disk.blog: Error detaching disk persistent-disk-0 from instance project-id-here/https://www.googleapis.com/compute/v1/projects/project-id-here/zones/us-west1-a/blog: googleapi: Error 400: Invalid value 'https://www.googleapis.com/compute/v1/projects/project-id-here/zones/us-west1-a'. Values must match the following regular expression: '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?', invalidParameter

```

rileykarson:
We don't support attaching/detaching disks post-create yet in instances, so the instance the disk was previously attached to will no longer be able to be `Read` as a result of this; we may not want to merge for that reason.

Right now, if we end up in the `Delete` for an attached disk, we error out because the `zone` is being set to a zone url and not a zone name in the detach request. 

Fixes #112

### DIFF
--- a/google/resource_compute_disk.go
+++ b/google/resource_compute_disk.go
@@ -301,9 +301,10 @@ func resourceComputeDiskDelete(d *schema.ResourceData, meta interface{}) error {
 			}
 			for _, disk := range i.Disks {
 				if disk.Source == self {
+					zoneParts := strings.Split(i.Zone, "/")
 					detachCalls = append(detachCalls, detachArgs{
 						project:    project,
-						zone:       i.Zone,
+						zone:       zoneParts[len(zoneParts)-1],
 						instance:   i.Name,
 						deviceName: disk.DeviceName,
 					})


danawillow:
This looks fine, but will it alone fix #112?  It'll fix the detach error I assume, but it looks like the bigger issue is the fact the disk is trying to be recreated (which has to do with the snapshot url vs name)

paddycarver:
I think #238 is the fix for snapshot URL v name.

rileykarson:
^Yup, #238 tracks the snapshot url vs name, and #112 is disks not being able to detach themselves. #238 tends to trigger #112, though.

danawillow:
Got it, thanks!

PR #321, opened 2017-08-10
Remove default value for removed value automatic_restart

selmanj:
Import tests for compute_instance_template fail without this change as
they expect a value of true for automatic_restart. As this value was
removed, we're no longer setting it (and therefore it looks like it has
a value of false, which is different from the default).


selmanj:
We can only hope!

PR #325, opened 2017-08-14
Move non-provider related functions into a utils.go file

selmanj:
Small cleanup; there's a lot of miscellaneous functions that ended up in `provider.go` that don't belong anywhere else.


selmanj:
(No changes were made to the functions that were moved)

Issue #90, opened 2017-06-13
Implement Google Apps (G Suite) Google Groups API


hashibot:
_This issue was originally opened by @allandrick as hashicorp/terraform#14415. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,

Is anyone working on implementing the Google Apps (G Suite) APIs in Terraform?

Google Compute Platform IAM relies heavily on Google Apps identities and can also leverage Google Groups as security groups - this is a great way of implementing RBAC in IAM across projects and organizations in GCP.

I'd love to be able to add something like the following to my templates so I can create / update those google groups from within Terraform and then later reference them in my google_project_iam_policy bindings:

```
resource "google_apps_group" "admin-group-one" {
  name = "admingroup1@company.com"
  email = "admingroup1@company.com"
  description = "Administrators of the Aussie GCP Project"
  members = [
# https://developers.google.com/admin-sdk/directory/v1/reference/members/insert
# "MANAGER": This role is only available if the Google Groups for Business is enabled using the Admin console. A MANAGER role can do everything done by an OWNER role except make a member an OWNER or delete the group. A group can have multiple MANAGER members.
# "MEMBER": This role can subscribe to a group, view discussion archives, and view the group's membership list. For more information about member roles, see the administration help center.
# "OWNER": This role can send messages to the group, add or remove members, change member roles, change group's settings, and delete the group. An OWNER must be a member of the group. A group can have more than one OWNER.
    { 
      name = "jimbo@company.com"
      role => "OWNER"
    },
    { 
      name = "neville@company.com"
      role => "MEMBER"
    },
    { 
      name = "matilda@company.com",
      role => "MANAGER"
    },
    { 
      name = "${var.google_project.project.number}-compute@developer.gserviceaccount.com"
      role => "MEMBER"
    },
    { 
      name = "${var.google_service_account.jenkins.email}"
      role => "MEMBER"
    }
  ]
  settings = [
    allowExternalMembers = "false"
    allowGoogleCommunication = "false"
    membersCanPostAsTheGroup = "false"
    showInGroupDirectory = "false"
    whoCanAdd = "ALL_MANAGERS_CAN_ADD"
    whoCanContactOwner = "ALL_IN_DOMAIN_CAN_CONTACT"
    whoCanInvite = "NONE_CAN_INVITE"
[...]

PR #333, opened 2017-08-17
[WIP/HELP WANTED] Add google group

tmshn:
## Summary
As a first step of working on #90 (originally hashicorp/terraform#14415), this PR adds Google Group resource.

## Current Status

I've done rough implementation, but not yet written acctests.
In addition to that, I could not make group because of the following error:

```
* google_admin_group.test: Error creating group: Post https://www.googleapis.com/admin/directory/v1/groups?alt=json: oauth2: cannot fetch token: 401 Unauthorized
Response: {
  "error" : "unauthorized_client",
  "error_description" : "Client is unauthorized to retrieve access tokens using this method."
}
```

Does anyone has idea? (I'm using service account with DwD enabled).

### DIFF
--- a/google/config.go
+++ b/google/config.go
@@ -16,6 +16,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"golang.org/x/oauth2/jwt"
+	adminDirectory "google.golang.org/api/admin/directory/v1"
 	"google.golang.org/api/bigquery/v2"
 	"google.golang.org/api/cloudbilling/v1"
 	"google.golang.org/api/cloudresourcemanager/v1"
@@ -41,6 +42,7 @@ type Config struct {
 	Region      string
 	Impersonate string
 
+	clientAdminDirectory  *adminDirectory.Service
 	clientBilling         *cloudbilling.Service
 	clientCompute         *compute.Service
 	clientComputeBeta     *computeBeta.Service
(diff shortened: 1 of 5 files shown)


paddycarver:
Hey @tmshn,

Thanks for the PR! While I appreciate all the work you've done on this, at this time we're not interested in adding GSuite support to the Google Cloud Platform provider. While GCP and GSuite are tightly integrated, as you've noticed, we feel they'd be better suited as separate providers, given their different target audiences and their separate authentication methods. At the moment, there doesn't seem to be a clean way forward for us to authenticate against GCP and GSuite interchangeably.

This does not mean we would not like to see this functionality in Terraform; I, personally, would love to see a GSuite provider created that could do this. I just feel the technical decision of including it in the GCP provider is going to cause a lot of problems, without much beneft. After all, a separate GSuite provider would still enable all the integration with GCP that including the resource in the GCP provider would, but without the problem of mixing credentials.

Because this is the wrong place for this functionality, I'm going to close this PR. If you'd like to discuss it more, however, please feel free to do so. And I encourage you to keep this functionality in your own custom provider, to enable the integrations you're looking for. If there's enough interest and movement on the provider, it can become an officially supported part of Terraform.

Thanks for all your contributions to Terraform. They're really appreciated.

chicagobuss:
I'm just confused where the work on the new GSuite provider is being captured?

tmshn:
@chicagobuss Unfortunately, I think no one has started that work...

paddycarver:
I don't know of any GitHub repo that has been created for this yet, though I'd encourage anyone that has started one to drop a link to it here so that others may join in, as well!

DeviaVir:
I had a need for such a provider, so I started one here:
https://github.com/DeviaVir/terraform-provider-gsuite

It currently has (pretty basic) support for `users`, `groups` and `group memberships`, hope to see it extended further by the FOSS community but this covers my needs for now.

PR #338, opened 2017-08-17
added block in README.md on upgrading

nodesocket:


paddycarver:
Hey @nodesocket, thanks for the feedback! I'm realising we don't really have `upgrade` covered in our docs anywhere.

I'm hesitant to add it to just the `terraform-provider-google` README, though, because it really applies to all our providers. If it's added to https://www.terraform.io/docs/configuration/providers.html instead, does that solve your concern? Do you think we need to add a pointer to that page to the README?

nodesocket:
@paddycarver sure that makes sense updating the official docs, but I don't think it hurts to add to the `README.md` as well. Your call though.

rosbo:
I agree with @nodesocket, it doesn't hurt to put it there. I like having the most important commands surfaced in our project's readme.

paddycarver:
Agreed, updating this to point to the docs on the website has been on my TODO list for far too long. Apologies for the delay here.

paddycarver:
Does #701 achieve what you were looking for, @nodesocket?

PR #339, opened 2017-08-17
Add support for setting labels to compute_image

selmanj:


google/utils.go
```diff
@@ -229,6 +229,23 @@ func linkDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 	return false
 }
 
+// expandLabels pulls the value of "labels" out of a schema.ResourceData as a map[string]string.
+func expandLabels(d *schema.ResourceData) map[string]string {
```
rosbo: we use the naming convention `expand*` when converting from []interface{} or the ResourceData to the go client object. 

For example:
`func expandAliasIpRanges(ranges []interface{}) []*computeBeta.AliasIpRange`

I find it a bit confusing to reuse the same term for something different.

selmanj: You mean because it's not returning a go client object? Or because the first parameter type should be of type `[]interface{}`? 

This function is indeed converting data from ResourceData but its not directly a go-client object (as the type of the `Labels` attribute is map[string]string). Not using direct go-client objects here lets us reuse this methods across other apis.

rosbo: Gotcha. I assumed that a compute.Labels or something along the line existed. Nevermind my comment then.

PR #341, opened 2017-08-18
Add preemptible as an option to node config

### LINKED ISSUE #334: Feature Request: Allow to use preemptible flag for google container cluster
Now it's not possible, but [API](https://cloud.google.com/container-engine/reference/rest/v1/NodeConfig) has this flag in Node Config.

selmanj:
Closes #334

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -669,6 +669,7 @@ func flattenClusterNodeConfig(c *container.NodeConfig) []map[string]interface{} 
 			"image_type":      c.ImageType,
 			"labels":          c.Labels,
 			"tags":            c.Tags,
+			"preemptible":     c.Preemptible,
 		},
 	}
 
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -471,7 +471,7 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 	d.Set("monitoring_service", cluster.MonitoringService)
 	d.Set("network", d.Get("network").(string))
 	d.Set("subnetwork", cluster.Subnetwork)
-	d.Set("node_config", flattenClusterNodeConfig(cluster.NodeConfig))
+	d.Set("node_config", flattenNodeConfig(cluster.NodeConfig))
 	d.Set("node_pool", flattenClusterNodePools(d, cluster.NodePools))
 
 	if igUrls, err := getInstanceGroupUrlsFromManagerUrls(config, cluster.InstanceGroupUrls); err != nil {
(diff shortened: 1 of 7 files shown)


google/resource_container_cluster.go
```diff
@@ -669,6 +669,7 @@ func flattenClusterNodeConfig(c *container.NodeConfig) []map[string]interface{}
 			"image_type":      c.ImageType,
 			"labels":          c.Labels,
 			"tags":            c.Tags,
+			"preemptible":     c.Preemptible,
```
danawillow: while you're here- it probably makes sense to have this in node_config.go too, right?

google/resource_container_cluster_test.go
```diff
@@ -712,6 +712,7 @@ resource "google_container_cluster" "with_node_config" {
 			foo = "bar"
 		}
 		tags = ["foo", "bar"]
+		preemptible = true
```
danawillow: can you also add a check for this in testAccCheckContainerCluster?

google/resource_container_node_pool_test.go
```diff
@@ -356,6 +356,7 @@ resource "google_container_node_pool" "np_with_node_config" {
 			"https://www.googleapis.com/auth/logging.write",
 			"https://www.googleapis.com/auth/monitoring"
 		]
+		preemptible = true
```
danawillow: likewise for testAccCheckContainerNodePoolMatches

google/resource_container_cluster.go
selmanj: Agreed, I moved it

google/resource_container_cluster_test.go
selmanj: Added!

Interestingly the tests were failing without having to add any changes to check for preemptible (likely because the default is false, and the requested value is true, which creates a perpetual diff unless it's hooked up properly).

BTW, I assumed this method was just checking to see if the resource existed; I didn't realize it did all the checking inside this method. Some other resources look up the remote resource and store the api object in a variable, and then assert on that. Which is preferred for new resources?

google/resource_container_node_pool_test.go
selmanj: Added

google/resource_container_cluster_test.go
danawillow: I'm not really sure- I did this one sort of as an experiment, and it definitely doesn't catch everything since it just checks that state matches remote, but not that those are actually the values we want (I actually had a bug slip through once because of this). Right now I've been using `TestCheckResourceAttrSet` to check that the correct values are in state for what I care about but I don't know if that actually checks that I did it with a .Set.

CGavrila:
@selmanj @danawillow Any idea when this will make it into a release?

danawillow:
👍 , merge at will.

@CGavrila this will be in our next release. I don't have a hard date for it yet but probably within the next week or so.

CGavrila:
Thanks, @danawillow! Quite excited about this one.

PR #345, opened 2017-08-19
Add ability to set priority on compute_firewall

### LINKED ISSUE #342: google_compute_firewall doesn't support priority
### Terraform Version
terraform -v
Terraform v0.10.2

### Affected Resource(s)
google_compute_firewall

In GCE you can put priority on firewall rules so that you can do something like: 
- block tcp:22 to 10.128.0.0/16 priority 5000
- allow tcp:22 ONLY to 10.128.0.5/32 priority 100

This would allow tcp22 to 10.128.0.5 but block tcp:22 to the rest of the subnet.  We  heavily utilize rules in this fashion and need to be able to input priority into terraform configs.  Its not currently documented as being supported at all.  Hopefully this is an easy fix.  

### Here is the gcloud command that would use priority:
gcloud beta compute --project=blah firewall-rules create block --direction=EGRESS --priority=5000 --network=default --deny=all --destination-ranges=10.128.0.0/16

### And the REST call:
POST https://www.googleapis.com/compute/v1/projects/blah/global/firewalls
{
  "name": "block",
  "selfLink": "projects/blah/global/firewalls/block",
  "network": "projects/blah/global/networks/default",
  "direction": "EGRESS",
  "priority": 5000,
  "denied": [
    {
      "IPProtocol": "all"
    }
  ],
  "destinationRanges": [
    "10.128.0.0/16"
  ]
}

selmanj:
Closes #342

### DIFF
--- a/google/resource_compute_firewall.go
+++ b/google/resource_compute_firewall.go
@@ -14,6 +14,8 @@ import (
 	"google.golang.org/api/compute/v1"
 )
 
+const COMPUTE_FIREWALL_PRIORITY_DEFAULT = 1000
+
 var FirewallBaseApiVersion = v1
 var FirewallVersionedFeatures = []Feature{
 	Feature{Version: v0beta, Item: "deny"},
@@ -51,7 +53,7 @@ func resourceComputeFirewall() *schema.Resource {
 				Type:         schema.TypeInt,
 				Optional:     true,
 				ForceNew:     true,
-				Default:      1000,
+				Default:      COMPUTE_FIREWALL_PRIORITY_DEFAULT,
 				ValidateFunc: validation.IntBetween(0, 65535),
 			},
 
@@ -276,6 +278,10 @@ func resourceComputeFirewallRead(d *schema.ResourceData, meta interface{}) error
 		if err != nil {
 			return err
 		}
+		// During firewall conversion from v1 to v0beta, the value for Priority is read as 0 (as it doesn't exist in
+		// v1). Unfortunately this is a valid value, but not the same as the default. To avoid this, we explicitly set
+		// the default value here.
+		firewall.Priority = COMPUTE_FIREWALL_PRIORITY_DEFAULT
 	case v0beta:
 		firewallV0Beta, err := config.clientComputeBeta.Firewalls.Get(project, d.Id()).Do()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


rosbo:
The import test is failing.

You can either update the configuration to set explicitly the priority on the configuration to import or use `ImportStateVerifyIgnore` in the `TestStep`.

make testacc TESTARGS="-run TestAccComputeFirewall_" TEST="./google"
TF_ACC=1 go test ./google -v -run TestAccComputeFirewall_ -timeout 120m
=== RUN   TestAccComputeFirewall_importBasic
--- FAIL: TestAccComputeFirewall_importBasic (58.64s)
	testing.go:435: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

		(map[string]string) (len=1) {
		 (string) (len=8) "priority": (string) (len=1) "0"
		}


		(map[string]string) (len=1) {
		 (string) (len=8) "priority": (string) (len=4) "1000"
		}

=== RUN   TestAccComputeFirewall_basic
--- PASS: TestAccComputeFirewall_basic (66.31s)
=== RUN   TestAccComputeFirewall_update
--- PASS: TestAccComputeFirewall_update (73.46s)
=== RUN   TestAccComputeFirewall_priority
--- PASS: TestAccComputeFirewall_priority (52.01s)
=== RUN   TestAccComputeFirewall_noSource
--- PASS: TestAccComputeFirewall_noSource (52.18s)
=== RUN   TestAccComputeFirewall_denied
--- PASS: TestAccComputeFirewall_denied (59.78s)
=== RUN   TestAccComputeFirewall_egress
--- PASS: TestAccComputeFirewall_egress (59.78s)
FAIL
exit status 1
FAIL	github.com/terraform-providers/terraform-provider-google/google	422.298s
make: *** [testacc] Error 1

selmanj:
Good catch, I forgot to run the import test.

Interesting why this is failing. I believe its due to `resourceComputeFirewallRead` making a choice about api version to use; as this is an import, there's no v0beta features in use, so it chooses v1. Then it converts this resource to a v0beta resource by dumping to json and reparsing. This means it never sees `priority` (and annoyingly, the default value for int64 is 0, which is a valid value for priority).

When a user imports a firewall, which version should we use to read? The code currently decides which version based on what's set in the schema, but when importing should we instead look at the remote api and see if any features are in use?

selmanj:
Spoke to @paddycarver, we decided that the current behavior is correct (when importing, default to non-beta apis). To fix the issue I decided to explicitly set `Priority` to the default when converting the struct.

PR #351, opened 2017-08-23
adding priority setting to firewall

adamenger:
Hi there,

At Etsy we had a use case in our network engineering department to use the `priority` setting for our firewall rules but we found that we were unable to do so using the latest version of the terraform provider. I've extended the schema and beta attributes, and tested successfully using the plan below. 

```
resource "google_compute_firewall" "etsy" {
  name    = "etsy"
  network = "${var.network}"

  target_tags = ["etsy"]
  priority = 1005

  allow {
    protocol = "tcp"
    ports    = ["22"]
  }

  source_ranges = ["1.2.3.4/24"]
}
```

In this commit we've added the support for `priority` and the associated documentation. I looked at the acceptance tests but didn't see a clear place that I could test `priority` in a useful way. 

If you would like some acceptance tests with this patch, could you please provide some direction or suggestions as to how we could include them with this PR?

Best regards,

### DIFF
--- a/google/resource_compute_firewall.go
+++ b/google/resource_compute_firewall.go
@@ -19,6 +19,7 @@ var FirewallVersionedFeatures = []Feature{
 	Feature{Version: v0beta, Item: "deny"},
 	Feature{Version: v0beta, Item: "direction"},
 	Feature{Version: v0beta, Item: "destination_ranges"},
+	Feature{Version: v0beta, Item: "priority"},
 }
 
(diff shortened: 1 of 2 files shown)


selmanj:
Hey Adam

I actually have a PR open right now for this same feature (#345) - hopefully that should be coming along soon.

rosbo:
Closing this PR. The feature has been added in #345 (now merged).

Issue #50, opened 2017-06-13
google_storage_bucket_acl role_entities don't work 


hashibot:
_This issue was originally opened by @mikemcrill as hashicorp/terraform#10612. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
Terraform v0.7.13

### Affected Resource(s)
- google_storage_bucket_acl

### Terraform Configuration Files
```hcl
resource "google_storage_bucket_acl" "scripts-acl" {
  bucket = "${google_storage_bucket.scripts.name}"
  default_acl = "private"

  role_entity = [
    "OWNER:project-owners-123456",
    "READER:user-compute@myproject.iam.gserviceaccount.com",                                                                                          "READER:user-compute-jenk@myproject.iam.gserviceaccount.com",
    "READER:user-compute-salt@myproject.iam.gserviceaccount.com",
  ]
 }
```

### Debug Output
https://gist.github.com/mikemcrill/b43cb01767812985338b9bb890da4a9b

### Expected Behavior
ACL should be applied without errors

### Actual Behavior
ACL tries to delete owner and fails

### Steps to Reproduce
Apply terraform with the above state

### Important Factoids
I tried with and without referencing the OWNER entity, same effect.  Terraform keeps trying to destroy the OWNER permission.

joe-boyce:
Any updates on this one? not being able to use terraform for bucket acl's is making this resource unusable

caseylang:
Having this problem with 0.9.11 as well, including the reappearing work issue described in https://github.com/hashicorp/terraform/issues/10612#issuecomment-303601680 

I differ from the parent bug however. By removing a reference to `OWNER` in my ACL I'm able to create additional permissions, without having terraform attempt to delete the real owner.

It also seems worth mentioning that the permissions Terraform creates are of the older `Storage Legacy Bucket` variety.

paddycarver:
Is this still an ongoing issue? I've got a working config at the moment that adds a user in the OWNER role, and it's working without a hitch for me. Does anyone have a minimal reproduction handy?

lobeck:
This is our case which is affected, the ACL will be updated on every apply

    resource "google_storage_bucket" "blubb_bucket" {
      name     = "${var.bucket_name}"
      location = "${var.location}"
      storage_class = "${var.storage_class}"
    
      website {
[...]

PR #358, opened 2017-08-24
storage: make bucket ACLs control the entire resource.

paddycarver:
Storage bucket ACLs inherited the behaviour of only updating the fields
that were set in the config file. Terraform should track all the fields
in the resource, whether the user has specified a value for them or not,
and correct any drift that may occur.

This has manifested in an issue and unexpected behaviour in #50, and
this PR restores the expected behaviour.

### DIFF
--- a/google/resource_storage_bucket_acl.go
+++ b/google/resource_storage_bucket_acl.go
@@ -30,15 +30,17 @@ func resourceStorageBucketAcl() *schema.Resource {
 			},
 
 			"predefined_acl": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"role_entity"},
 			},
 
 			"role_entity": &schema.Schema{
-				Type:     schema.TypeList,
-				Optional: true,
-				Elem:     &schema.Schema{Type: schema.TypeString},
+				Type:          schema.TypeList,
+				Optional:      true,
+				Elem:          &schema.Schema{Type: schema.TypeString},
+				ConflictsWith: []string{"predefined_acl"},
 			},
 		},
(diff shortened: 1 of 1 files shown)


google/resource_storage_bucket_acl.go
```diff
@@ -138,53 +134,37 @@ func resourceStorageBucketAclCreate(d *schema.ResourceData, meta interface{}) er
...
-			re_local_map[res.Entity] = res.Role
-		}
-
+	// The API offers no way to retrieve predefined ACLs,
+	// and we can't tell which access controls were created
+	// by the predefined roles, so...
+	//
+	// This is, needless to say, a bad state of affairs and
+	// should be fixed.
+	if _, ok := d.GetOk("role_entity"); ok {
```
catsby: It seems we're no long reading `predefined_acl` at all, is that intentional?

paddycarver: Yes, the comment above talks about it briefly. but to expand on it here:

* The resource never actually read `predefined_acl`.
* It _still_ does not read `predefined_acl` because it can't. `predefined_acl` is just a shorthand used when creating or updating a bucket, and isn't actually returned by any API call. It just gets translated into a predefined list of ACLs (which, sadly, aren't predictable--as far as I can tell), and is never seen again. This means `predefined_acl` is probably a bad thing to support as such in terraform, but deprecating it entirely felt out of scope for this PR, especially as I think it would take quite a bit of work to replace it.

catsby: Thanks for explaining!

danawillow:
I think this change broke some tests, can you take a look @paddycarver?
 https://ci.hashicorp.engineering/viewLog.html?buildId=61272&tab=buildResultsDiv&buildTypeId=Terraform_ProviderGoogleCloud

paddycarver:
Investigating this, apologies.

PR #359, opened 2017-08-25
dns: Add special handling for ns records.

### LINKED ISSUE #95: Cannot Manage Google Cloud DNS NS Records
_This issue was originally opened by @j0sh3rs as hashicorp/terraform#14796. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

When Terraform creates a google_dns_managed_zone, the zone automatically gets created with NS records specific to Google Cloud (nameservers they provide that the zone will live on).  Terraform does not import these NS records into state, so all future attempts to add or manipulate NS records in any Google Cloud Zone will fail, citing that the record set already exists:
```
* google_dns_record_set.clouddns_ns: Error creating DNS RecordSet: googleapi: Error 409: The resource 'entity.change.additions[0]' named 'domain.com. (NS)' already exists, alreadyExists
```

### Terraform Version
```
$ terraform -v
Terraform v0.9.5
```
### Affected Resource(s)
- google_dns_managed_zone
- google_dns_record_set

```
resource "google_dns_managed_zone" "public" {
  name        = "domain-name"
  dns_name    = "domain.com."
  description = "domain.com"
}

resource "google_dns_record_set" "clouddns_ns" {
  count = "${var.type == "NS" ? 1 : 0}"
  name = "${var.name == "" ? format("%v.", var.domain) : format("%v.%v.",var.name,var.domain)}"
  type = "${var.type}"
  ttl  = "${var.ttl}"

  managed_zone = "${replace(var.domain, ".", "-")}"

  rrdatas = ["${formatlist("%v.", var.value)}"]
}

module "domain-com-NS" {
  source = "../../modules/pub_dns_record"
  domain = "domain.com"
  name   = ""
  type   = "NS"
  value  = ["ns1.example.com", "ns2.example.com"]
  ttl    = "172800"
}
```

### Expected Behavior
Terraform should either automatically import the NS records into state for management, or provide some method for managing them

### Actual Behavior
NS records are completely unmanageable and must be manually adjusted

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
This generally applies to all GCE zone sets, not just to ones Terraform creates

paddycarver:
Cloud DNS requires every managed zone to have an NS record at all times.
This means if people want to manage their own NS records, we need to add
their new record and remove the old one in the same call. It also means
we can't delete NS records, as we wouldn't know what to replace it with.
So deleting of NS records short-circuits. For the case of `terraform
destroy`, this prevents the error. It does mean if the user explicitly
tries to remove their NS zone from their project, it silently does
nothing, but that's unavoidable unless we want to A) restore a default
value (and it looks like the default values change from zone to zone?
And that is arguably just as unexpected?) or B) let the (arguably more
reasonable) `terraform destroy` case be impossible.

Fixes #95.

### DIFF
--- a/google/resource_dns_record_set.go
+++ b/google/resource_dns_record_set.go
@@ -77,6 +77,25 @@ func resourceDnsRecordSetCreate(d *schema.ResourceData, meta interface{}) error 
 		},
 	}
 
+	if d.Get("type").(string) == "NS" {
+		log.Printf("{DEBUG] DNS record list request for %q", zone)
+		res, err := config.clientDns.ResourceRecordSets.List(project, zone).Do()
+		if err != nil {
+			return fmt.Errorf("Error retrieving record sets for %q: %s", zone, err)
+		}
+		var deletions []*dns.ResourceRecordSet
+
+		for _, record := range res.Rrsets {
+			if record.Type != "NS" {
+				continue
+			}
+			deletions = append(deletions, record)
+		}
+		if len(deletions) > 0 {
+			chg.Deletions = deletions
+		}
+	}
+
 	log.Printf("[DEBUG] DNS Record create request: %#v", chg)
 	chg, err = config.clientDns.Changes.Create(project, zone, chg).Do()
 	if err != nil {
@@ -135,6 +154,13 @@ func resourceDnsRecordSetRead(d *schema.ResourceData, meta interface{}) error {
 }
 
 func resourceDnsRecordSetDelete(d *schema.ResourceData, meta interface{}) error {
+
+	// NS records must always have a value, so we short-circuit delete
+	// this allows terraform delete to work, but may have unexpected
+	// side-effects when deleting just that record set.
+	if d.Get("type").(string) == "NS" {
+		return nil
+	}
 	config := meta.(*Config)
 
 	project, err := getProject(d, config)
--- a/google/resource_dns_record_set.go
+++ b/google/resource_dns_record_set.go
@@ -78,7 +78,7 @@ func resourceDnsRecordSetCreate(d *schema.ResourceData, meta interface{}) error 
 	}
 
 	if d.Get("type").(string) == "NS" {
-		log.Printf("{DEBUG] DNS record list request for %q", zone)
+		log.Printf("[DEBUG] DNS record list request for %q", zone)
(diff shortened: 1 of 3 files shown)


google/resource_dns_record_set.go
```diff
@@ -77,6 +77,25 @@ func resourceDnsRecordSetCreate(d *schema.ResourceData, meta interface{}) error
 		},
 	}
 
+	if d.Get("type").(string) == "NS" {
+		log.Printf("{DEBUG] DNS record list request for %q", zone)
```
catsby: typo `{` instead of `[`

google/resource_dns_record_set.go
```diff
@@ -135,6 +154,13 @@ func resourceDnsRecordSetRead(d *schema.ResourceData, meta interface{}) error {
 }
 
 func resourceDnsRecordSetDelete(d *schema.ResourceData, meta interface{}) error {
+
+	// NS records must always have a value, so we short-circuit delete
+	// this allows terraform delete to work, but may have unexpected
+	// side-effects when deleting just that record set.
+	if d.Get("type").(string) == "NS" {
```
catsby: Let's put a `[DEBUG]` log here; Terraform is doing something "unexpected", or otherwise lying to people, so let's tell them that we aren't doing X because of Y, and link to some documentation. Either Google's or preferably some kind of `NOTE` on https://www.terraform.io/docs/providers/google/r/dns_record_set.html that explains the what and why here.

danawillow:
@paddycarver, what's the status on this?

paddycarver:
Sorry for the delay on this, believe I've addressed all the comments.

google/resource_dns_record_set.go
```diff
@@ -77,6 +77,25 @@ func resourceDnsRecordSetCreate(d *schema.ResourceData, meta interface{}) error
 		},
 	}
 
+	if d.Get("type").(string) == "NS" {
```
danawillow: Can you add a comment explaining why this section needs to happen? (even just a link to the docs is probably fine)

paddycarver:
@danawillow just updated with the comment. Good to merge?

danawillow:
Yup

PR #360, opened 2017-08-25
Don't attempt to delete default users on replica

### LINKED ISSUE #347: New Google Cloud SQL failover replica resource shouldn't try to delete root user
### Terraform Version
0.10.2

### Affected Resource(s)
- google_sql_database_instance

```hcl
// GCP Cloud SQL Primary DB in Primary Region, Primary Zone
resource "google_sql_database_instance" "db-primary" {
  provider           = "google.gcp-${var.region_primary}"
  region             = "${var.sqlregion_primary}"
  database_version   = "MYSQL_5_7"

  settings {
    tier             = "db-n1-standard-1"

    database_flags {
        name         = "character_set_server"
        value        = "utf8mb4"
    }

    backup_configuration {
        binary_log_enabled = "true"
        enabled            = "true"
        start_time         = "18:00"
    }

    ip_configuration {
        require_ssl        = "true"
    }

    location_preference {
        zone               = "${var.region_primary_zone_primary}"
    }
  }
}

// GCP Cloud SQL Secondary DB in Primary Region, Secondary Zone
resource "google_sql_database_instance" "db-failover" {
  provider           = "google.gcp-${var.region_primary}"
  region             = "${var.sqlregion_primary}"
  database_version   = "MYSQL_5_7"
  master_instance_name = "${google_sql_database_instance.db-primary.name}"

  replica_configuration {
    connect_retry_interval = "30"
    failover_target        = "true"
  }

  settings {
    tier             = "db-n1-standard-1"

    database_flags {
        name         = "character_set_server"
        value        = "utf8mb4"
    }

    ip_configuration {
        require_ssl        = "true"
    }

    location_preference {
        zone               = "${var.region_primary_zone_secondary}"
    }
  }
}
```

### Expected Behavior
The resource should be created without issue.

### Actual Behavior
The resource is created, but with error output in red:
```
Error applying plan:

1 error(s) occurred:

* google_sql_database_instance.db-failover: 1 error(s) occurred:

* google_sql_database_instance.db-failover: Error, failed to delete default 'root'@'*' user, but the database was created successfully: googleapi: Error 400: The requested operation is not valid for a read-replica instance., errorReadReplicaInvalidOperation

```

### Steps to Reproduce
1. Create a database instance with resource type google_sql_database_instance
2. Create a failover replica instance with almost all of the same settings, but remove the backup_configuration section and add the replica_configuration section.
3. Watch error output.

### Important Factoids
On page https://www.terraform.io/docs/providers/google/r/sql_database_instance.html, at the top in yellow it specifically says this resource type will delete the default 'root' user that Cloud SQL creates for you, and you need to define your own root user (which I do). However, this isn't a valid action for the failover replica.

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- https://github.com/terraform-providers/terraform-provider-google/issues/100

selmanj:
Fixes #347

### DIFF
--- a/google/resource_sql_database_instance.go
+++ b/google/resource_sql_database_instance.go
@@ -617,20 +617,23 @@ func resourceSqlDatabaseInstanceCreate(d *schema.ResourceData, meta interface{})
 		return err
 	}
 
-	// If a root user exists with a wildcard ('%') hostname, delete it.
-	users, err := config.clientSqlAdmin.Users.List(project, instance.Name).Do()
-	if err != nil {
-		return fmt.Errorf("Error, attempting to list users associated with instance %s: %s", instance.Name, err)
-	}
-	for _, u := range users.Items {
-		if u.Name == "root" && u.Host == "%" {
-			op, err = config.clientSqlAdmin.Users.Delete(project, instance.Name, u.Host, u.Name).Do()
-			if err != nil {
-				return fmt.Errorf("Error, failed to delete default 'root'@'*' user, but the database was created successfully: %s", err)
-			}
-			err = sqladminOperationWait(config, op, "Delete default root User")
-			if err != nil {
-				return err
+	// If a default root user was created with a wildcard ('%') hostname, delete it. Note that if the resource is a
+	// replica, then any users are inherited from the master instance and should be left alone.
+	if !sqlResourceIsReplica(d) {
+		users, err := config.clientSqlAdmin.Users.List(project, instance.Name).Do()
+		if err != nil {
+			return fmt.Errorf("Error, attempting to list users associated with instance %s: %s", instance.Name, err)
+		}
+		for _, u := range users.Items {
+			if u.Name == "root" && u.Host == "%" {
+				op, err = config.clientSqlAdmin.Users.Delete(project, instance.Name, u.Host, u.Name).Do()
+				if err != nil {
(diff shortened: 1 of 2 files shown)


selmanj:
It did not. I was going to write a test but thought that the additional test might not be worth it since it's a failure situation that users shouldn't run into anymore (and each test has a cost, after all; in this case it's in the minutes).

However I think there's something to be said about someone who wants their own `root@'%'` user that's managed within terraform; if they end up adding replicas later, they'll end up hitting this same path. In that case it probably makes sense to have this failure scenario covered.

selmanj:
I added a test ensuring we don't delete a root user found on the replica. It currently fails on master and passes on this branch. 

The test itself is not great to read; would love input on how to make this more readable using the `resource.TestCase`/`resource.TestStep` structs.

ryan-mf:
This was merged into master 12 days ago, but terraform-provider-google hasn't had a release since 1.3 on 8/17/2017:
https://github.com/terraform-providers/terraform-provider-google/releases

Now that Terraform is modular, we have to wait for the module to get its own release. Any idea when there will be a new release for terraform-provider-google?

selmanj:
I agree @ryan-mf; we're focusing on that now (we are working on making our underlying test suite happy).  Should be very soon...

PR #362, opened 2017-08-26
Added support for updating a cloud storage object using source field

rodcloutier:
This PR intends to add support for updating a cloud storage object if the content of the source fields changes.
Currently if the content of the source changes, the object storage does not detect that the state changes.
To do so we use a `StateFunc` which will encode the md5 for the source file. If the file changes, the md5 will also changes which will invalidate the state and trigger an update.

### DIFF
--- a/google/resource_storage_bucket_object.go
+++ b/google/resource_storage_bucket_object.go
@@ -2,8 +2,11 @@ package google
 
 import (
 	"bytes"
+	"crypto/md5"
+	"encoding/base64"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"os"
 
@@ -92,6 +95,9 @@ func resourceStorageBucketObject() *schema.Resource {
 				Optional:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"content"},
+				StateFunc: func(src interface{}) string {
+					return fileNameWithMd5(src.(string))
+				},
 			},
 
 			"storage_class": &schema.Schema{
@@ -106,6 +112,20 @@ func resourceStorageBucketObject() *schema.Resource {
 
 func objectGetId(object *storage.Object) string {
 	return object.Bucket + "-" + object.Name
+}
+
+func fileNameWithMd5(filename string) string {
+	data, err := ioutil.ReadFile(filename)
+	if err != nil {
+		log.Printf("[WARN] Failed to read source file %s. Will not compute md5 to store in state", filename)
+		return filename
+	}
+
+	h := md5.New()
+	h.Write(data)
+	md5 := base64.StdEncoding.EncodeToString(h.Sum(nil))
+
+	return filename + ":" + md5
 }
 
 func resourceStorageBucketObjectCreate(d *schema.ResourceData, meta interface{}) error {
--- a/google/resource_storage_bucket_object_test.go
+++ b/google/resource_storage_bucket_object_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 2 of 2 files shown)


selmanj:
Let me first say that this is a cool change!

I see we already have a `md5hash` Computed value stored in the resource state. Is there any way to use this when checking the file locally? That way if the remote object changes, terraform could detect the diff (rather than only detecting if the source file changes).

rodcloutier:
Good point. Isn't it already supported?

The current state will have the `md5hash` of the current file. In the sync, read will fetch the `md5hash` of the newer file in storage. Since they will differ, the storage will be updated.

Feel free to correct me if I am wrong.

I did some experiments to use the `md5hash`  while preparing this PR. 

We could change the `md5hash` field to be manual (as in not computed) and inject in the terraform configuration manually and it would work. But this is cumbersome and not really user friendly.

I have tried to automatically generate the `md5hash` but sadly, I couldn't find any hook in the terraform plugin system to generate a value before the diff. All diff are directly done on the native types (string, int, bool, etc).

We can change the validity of a diff but we cannot generate the value of a field before diffing. 

This also applies to the source or content. I couldn't find a way to have a custom diff that could use some input to tell terraform that there is a diff. That is why I resorted to the solution proposed.

If you are aware of something I have missed, I would be more than willing to use it!

google/resource_storage_bucket_object.go
```diff
@@ -92,6 +95,7 @@ func resourceStorageBucketObject() *schema.Resource {
 				Optional:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"content"},
+				StateFunc:     createSourceState,
```
selmanj: How about writing an anonymous function here that turns the `interface{}` into a string, and then renaming `createSourceState` to something like `filenameWithMd5` with the signature:
```go
func filenameWithMd5(filename string) string {
   // ...
}
```

That way the function is a bit more general and better named.

google/resource_storage_bucket_object.go
```diff
@@ -108,6 +112,21 @@ func objectGetId(object *storage.Object) string {
 	return object.Bucket + "-" + object.Name
 }
 
+func createSourceState(src interface{}) string {
+
```
selmanj: Minor quibble: this space can be removed

google/resource_storage_bucket_object.go
```diff
@@ -108,6 +112,21 @@ func objectGetId(object *storage.Object) string {
...
+	source := src.(string)
+	data, err := ioutil.ReadFile(source)
+	if err != nil {
```
selmanj: Probably should log a warning here

google/resource_storage_bucket_object.go
```diff
@@ -108,6 +112,21 @@ func objectGetId(object *storage.Object) string {
...
+		return source
+	}
+
+	h := md5.New()
+	h.Write(data)
+	md5 := base64.StdEncoding.EncodeToString(h.Sum(nil))
+
+	return source + "." + md5
```
selmanj: Rather than using a "." as the separator which is commonly found in filenames, how about something like ":"?

google/resource_storage_bucket_object_test.go
```diff
@@ -44,6 +45,45 @@ func TestAccGoogleStorageObject_basic(t *testing.T) {
...
+		ioutil.WriteFile(name, data, 0644)
+		return data_md5
+	}
+	data_md5 := writeFile(tf.Name(), []byte("data data data"))
+	updated_data_md5 := writeFile(tf.Name()+".update", []byte("datum"))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			if err != nil {
+				panic(err)
```
selmanj: I see other code does this, but I'm surprised by the panic (which would end all testing from running, right?)

How about changing it to a `t.Errorf`? There's no point halting everything as it'll just fail anyway.

google/resource_storage_bucket_object_test.go
```diff
@@ -44,6 +45,45 @@ func TestAccGoogleStorageObject_basic(t *testing.T) {
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccGoogleStorageObjectDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testGoogleStorageBucketsObjectBasic(bucketName),
+				Check:  testAccCheckGoogleStorageObject(bucketName, objectName, data_md5),
+			},
+			resource.TestStep{
+				PreConfig: func() {
+					os.Rename(tf.Name()+".update", tf.Name())
```
selmanj: Needs a check for err

google/resource_storage_bucket_object_test.go
```diff
@@ -44,6 +45,45 @@ func TestAccGoogleStorageObject_basic(t *testing.T) {
 	})
 }
 
+func TestAccGoogleStorageObject_update(t *testing.T) {
```
selmanj: I don't think we should use the `_update` suffix here. Those tests in general test the `update` method in the resource module, whereas the parameter we are updating forces a new resource.

selmanj:
Good point, I forgot that `md5sum` itself would trigger this behavior as is. 

I understand your pain points; let's proceed with this for now as it's a pretty useful change

google/resource_storage_bucket_object_test.go
rodcloutier: Seems reasonable. I just copied what was done in the other tests. Should we keep it isync with what the other tests are doing, only change this one or convert the other tests as well?

google/resource_storage_bucket_object_test.go
rodcloutier: You are right. Do you think a `_recreate` suffix would be fine?

google/resource_storage_bucket_object.go
rodcloutier: Done

google/resource_storage_bucket_object.go
rodcloutier: Done

google/resource_storage_bucket_object.go
rodcloutier: Good suggestion. Changed.

google/resource_storage_bucket_object.go
rodcloutier: Done

rodcloutier:
Sorry for the delay. I have addressed all the changes requested.

rodcloutier:
Ping

rosbo:
Sorry Rodrigue, selmanj@ transitioned to a new project and this PR slipped through the cracks.

Thank you for your contribution

rosbo:
I reverted this change. The major issue is that it creates a diff for existing storage object when people upgrade their Google Terraform provider.

Usually, we can solve this with a MigrateFunc. However, this func is called at each `terraform plan` until `terraform apply` is called to save the new state. It means that if you update the file content and the state hasn't been migrated yet, you won't detect any changes because the migrate function will recompute the hash based on the source file and the statefunc will do the same effectively cancelling each other.

We should try to find a more cleaver approach, ideally, relying on the md5 hash sent by the server, this way, we can also detect changes if they happen outside of Terraform.

rosbo:
I created a new issue to track this work #788

PR #378, opened 2017-08-31
Import for compute_address now requires region/name. Changed the ID format.

### LINKED ISSUE #374: It's very unobvious what id to provide to `import google_compute_address`
### Terraform Version
Terraform v.0.10.3

### Affected Resource(s)
- google_compute_address

Doesn't matter.

### Debug Output
Doesn't matter?

### Expected Behavior

Terraform should provide a helpful example of google_compute_address ID that you have to provide for importing - that it's just a name of the address, not for example the long numeric id that you get with `gcloud compute addresses describe <name>`...

~It should also tell you that if the address is in other region than the default for your project, then you have to prepend it to the address region like "<region>.<name>"~

Update: actually you can't use this syntax, you need to temporarily change region in your provider... Thanks @selmanj !

### Actual Behavior
I got only the unfriendly error message: `Values must match the following regular expression: '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?'`

### Steps to Reproduce
`terraform import google_compute_address.<some_name> <some_name>` if the address is in other region than the default for your project

### Important Factoids
Information at https://www.terraform.io/docs/providers/google/r/compute_address.html#import is not helpful at all.

rosbo:
Fixes #374

Also performed the migration to the new ID format for backward compatibility.

```sh
make testacc TESTARGS="-run TestAccComputeAddress_basic" TEST="./google"
TF_ACC=1 go test ./google -v -run TestAccComputeAddress_basic -timeout 120m
--- PASS: TestAccComputeAddress_basic (23.69s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	23.831s

make testacc TESTARGS="-run TestAccComputeAddress_importBasic" TEST="./google"
TF_ACC=1 go test ./google -v -run TestAccComputeAddress_importBasic -timeout 120m
--- PASS: TestAccComputeAddress_importBasic (23.08s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	23.217s
```

@paddycarver We use this PR to discuss also including the project in the import string and the the relative part of the self link (anything right of the version) as the resource id.

### DIFF
--- a/google/resource_compute_address.go
+++ b/google/resource_compute_address.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"google.golang.org/api/compute/v1"
+	"strings"
 )
 
 func resourceComputeAddress() *schema.Resource {
@@ -16,6 +17,10 @@ func resourceComputeAddress() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
+
+		SchemaVersion: 1,
+		MigrateState:  resourceComputeAddressMigrateState,
+
 		Schema: map[string]*schema.Schema{
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
@@ -37,6 +42,7 @@ func resourceComputeAddress() *schema.Resource {
 			"region": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 				ForceNew: true,
 			},
 
@@ -70,7 +76,10 @@ func resourceComputeAddressCreate(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	// It probably maybe worked, so store the ID now
-	d.SetId(addr.Name)
+	d.SetId(computeAddressId{
+		Region: region,
+		Name:   addr.Name,
+	}.terraformId())
 
 	err = computeOperationWait(config, op, project, "Creating Address")
 	if err != nil {
@@ -83,7 +92,7 @@ func resourceComputeAddressCreate(d *schema.ResourceData, meta interface{}) erro
 func resourceComputeAddressRead(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	region, err := getRegion(d, config)
+	addressId, err := parseComputeAddressId(d.Id())
 	if err != nil {
 		return err
 	}
@@ -94,7 +103,7 @@ func resourceComputeAddressRead(d *schema.ResourceData, meta interface{}) error 
 	}
 
 	addr, err := config.clientCompute.Addresses.Get(
-		project, region, d.Id()).Do()
+		project, addressId.Region, addressId.Name).Do()
 	if err != nil {
 		return handleNotFoundError(err, d, fmt.Sprintf("Address %q", d.Get("name").(string)))
 	}
@@ -102,6 +111,7 @@ func resourceComputeAddressRead(d *schema.ResourceData, meta interface{}) error 
 	d.Set("address", addr.Address)
 	d.Set("self_link", addr.SelfLink)
 	d.Set("name", addr.Name)
+	d.Set("region", addr.Region)
 
 	return nil
 }
@@ -109,7 +119,7 @@ func resourceComputeAddressRead(d *schema.ResourceData, meta interface{}) error 
 func resourceComputeAddressDelete(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	region, err := getRegion(d, config)
+	addressId, err := parseComputeAddressId(d.Id())
 	if err != nil {
 		return err
 	}
@@ -122,7 +132,7 @@ func resourceComputeAddressDelete(d *schema.ResourceData, meta interface{}) erro
 	// Delete the address
 	log.Printf("[DEBUG] address delete request")
 	op, err := config.clientCompute.Addresses.Delete(
-		project, region, d.Id()).Do()
+		project, addressId.Region, addressId.Name).Do()
 	if err != nil {
 		return fmt.Errorf("Error deleting address: %s", err)
 	}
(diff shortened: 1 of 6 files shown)


selmanj:
@rosbo What do you think about using the self_link here as the import id instead? Is that more/less confusing?

rosbo:
I think having the user type `projects/MY-PROJECT/region/MY-REGION/address/MY-ADDRESS is very confusing for the user because it is different for the import for all the other resources. It also requires more typing.

selmanj:
Right, but let's say a user wants to import a resource that's not within their provider's default project. Wouldn't they have the same problem as above re: zone? 

I think we all discussed the idea of using self-link as id here but I can't remember where we landed. Pinging @paddycarver for comments

rosbo:
All resources are associated with a project. Right now, we don't support import from a project different than the default project. If we change it to support project. I think we should do it in one swoop to avoid having different resources working in different ways.

We discussed using the relative part of the `self-link` as the resource ID which is a good idea I think. I am not sure if we should use the same for the import ID. i.e.:
- `terraform import my-resource.a-name projects/MY-PROJECT/region/MY-REGION/address/MY-ADDRESS`
Or
- `terraform import my-resource.a-name MY-PROJECT/MY-REGION.MY-ADDRESS`

selmanj:
> We discussed using the relative part of the self-link as the resource ID which is a good idea I think. I am not sure if we should use the same for the import ID. i.e.:

Aren't these the same? Can you distinguish between the two during import?

rosbo:
Yes you can distinguish between the two during import. You just need to write your own import function instead of using a pass through. See:
https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_subnetwork.go#L319

Note: You can call SetId in this function to construct the self-link from the different parts received in the import ID.

selmanj:
Thanks, I didn't know that!

I think that if Id is the self-link, we should allow importing from the self-link as well; naturally the first thing a person will do when checking to see what the id is is look at their state file and copy/paste. 

Given what you've pasted above, maybe we could support importing fragments of the self link? E.g. if the id has no '/' symbols then use the default provider region/project, if it contains region/foo/address/bar then use the provided region, etc...

What do you think?

rosbo:
Good idea.

I will create a generic ID class we can reuse for all the resources. Each resource will instantiate this class by specifying the parts of its id. It will then deal with the various import id format parsing, generating the id and so on. I suggest to do it in a separate PR.

4 import id format can be supported:

* project/name
* name
* relative self-link 
* full self-link

rosbo:
I added the project to the import id for this PR to avoid having to do another id migration.

rosbo:
make testacc TESTARGS="-run TestAccComputeAddress_" TEST="./google"
TF_ACC=1 go test ./google -v -run TestAccComputeAddress_ -timeout 120m
--- PASS: TestAccComputeAddress_importBasic (24.35s)
--- PASS: TestAccComputeAddress_basic (23.96s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	48.438s

selmanj:
I love the import idea.

Just one last thing; the id here is not the self-link (or even a fragment of the self link as it's missing the projects, regions and addresses fragments). Wouldn't it be cleaner to just use the self-link? Customers can already see the self-link in gcloud which would make it easy to copy and paste.

rosbo:
I had planned to support the 4 cases listed above in a different PRs but finally decided to do it in this one to avoid having to do multiple id migration. (this addresses your last comment about supporting self_link).

I will generalize the `computeAddressId` struct and the `parseComputeAddressId` to be able to reuse it for all resources in my next PR.

rosbo:
```sh
TF_ACC=1 go test ./google -v -run TestAccComputeAddress_ -timeout 120m
--- PASS: TestAccComputeAddress_importBasic (23.62s)
--- PASS: TestAccComputeAddress_basic (23.54s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	47.321s

go test -i ./google || exit 1
echo ./google | \
		xargs -t -n4 go test -run TestComputeAddressIdParsing -timeout=30s -parallel=4
go test -run TestComputeAddressIdParsing -timeout=30s -parallel=4 ./google
ok  	github.com/terraform-providers/terraform-provider-google/google	0.122s
/Users/rosbo/go/src/github.com/terraform-providers/terraform-provider-google
```

google/resource_compute_address.go
```diff
@@ -6,6 +6,13 @@ import (
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"google.golang.org/api/compute/v1"
+	"regexp"
+	"strings"
+)
+
+var (
+	computeAddressIdTemplate = "projects/%s/regions/%s/addresses/%s"
+	computeAddressLinkRegex  = regexp.MustCompile("projects/(.*)/regions/(.*)/addresses/(.*)$")
```
selmanj: Minor quibble: using .+ is preferred to .* as it excludes a few invalid compute regexes that match (that probably won't ever occur but oh well)

google/resource_compute_address.go
```diff
@@ -83,55 +99,107 @@ func resourceComputeAddressCreate(d *schema.ResourceData, meta interface{}) erro
...
+	addressId, err := parseComputeAddressId(d.Id(), config)
 	if err != nil {
 		return err
 	}
 
-	project, err := getProject(d, config)
-	if err != nil {
-		return err
-	}
+	log.Println("[rosbo] resourceComputeAddressRead", addressId)
```
selmanj: Oops this guy snuck in

google/resource_compute_address.go
```diff
@@ -83,55 +99,107 @@ func resourceComputeAddressCreate(d *schema.ResourceData, meta interface{}) erro
...
+		if config.Project == "" {
+			return nil, fmt.Errorf("Invalid compute address id. The default project for the provider must be set.")
+		}
+
+		return &computeAddressId{
+			Project: config.Project,
+			Region:  parts[0],
+			Name:    parts[1],
+		}, nil
+	}
```
selmanj: Missing one more case (only specifying `name`, which is the current behavior), or was this not included on purpose?

google/resource_compute_address.go
rosbo: Oupps. Removed.

google/resource_compute_address.go
rosbo: Totally agree, fixed.

google/resource_compute_address.go
rosbo: It was on purpose but now that I think about it, I see no good reason for not supporting this case. Support added.

PR #382, opened 2017-09-02
Fix docs. Node Pool autoscaling attribute names

sl1pm4t:
Documentation fix

Correct attribute names for google_container_node_pool:
* `minNodeCount` -> `min_node_count` as expected by resource config
* `maxNodeCount` -> `max_node_count` as expected by resource config


rosbo:
Thanks for catching and fixing this.

PR #384, opened 2017-09-03
Expose additional properties during cluster creation

brianhealeyRMN:
Allow for sending IssueClientCertificate as part of the cluster creation request
Allow for sending the Management object as part of the cluster creation request
Allow for sending the AutoScale object as part of the cluster creation request

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -66,6 +66,11 @@ func resourceContainerCluster() *schema.Resource {
 							Required: true,
 							ForceNew: true,
 						},
+						"issue_client_certificate": {
+							Type:             schema.TypeBool,
+							Optional:         true,
+							DiffSuppressFunc: linkDiffSuppress,
+						},
 					},
 				},
 			},
@@ -266,6 +271,60 @@ func resourceContainerCluster() *schema.Resource {
 						},
 
 						"node_config": schemaNodeConfig,
+
+						"management": {
+							Type:     schema.TypeList,
+							Optional: true,
+							ForceNew: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+
+									"auto_repair": {
+										Type:     schema.TypeBool,
+										Optional: true,
+										ForceNew: true,
+									},
+
+									"auto_upgrade": {
+										Type:     schema.TypeBool,
+										Optional: true,
+										ForceNew: true,
+									},
+								},
+							},
+						},
+
+						"auto_scaling": {
+							Type:     schema.TypeList,
+							Optional: true,
+							ForceNew: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+
+									"enabled": {
+										Type:     schema.TypeBool,
+										Optional: true,
+										ForceNew: true,
+									},
+
+									"min_node_count": {
+										Type:         schema.TypeInt,
+										Optional:     true,
+										ForceNew:     true,
+										ValidateFunc: validation.IntAtLeast(1),
+									},
+
+									"max_node_count": {
+										Type:         schema.TypeInt,
+										Optional:     true,
+										ForceNew:     true,
+										ValidateFunc: validation.IntAtLeast(1),
+									},
+								},
+							},
+						},
 					},
 				},
 			},
@@ -303,6 +362,12 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		cluster.MasterAuth = &container.MasterAuth{
 			Password: masterAuth["password"].(string),
 			Username: masterAuth["username"].(string),
+		}
+
+		if v, ok := masterAuth["issue_client_certificate"]; ok {
+			cluster.MasterAuth.ClientCertificateConfig = &container.ClientCertificateConfig{
+				IssueClientCertificate: v.(bool),
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


google/resource_container_cluster.go
```diff
@@ -266,6 +271,60 @@ func resourceContainerCluster() *schema.Resource {
...
+									"auto_upgrade": {
+										Type:     schema.TypeBool,
+										Optional: true,
+										ForceNew: true,
+									},
+								},
+							},
+						},
+
+						"auto_scaling": {
```
danawillow: Can this be `autoscaling` to match what's in `google_container_node_pool` and the API docs (which has it as one word)?

google/resource_container_cluster.go
```diff
@@ -476,13 +587,23 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
...
-			"password":               cluster.MasterAuth.Password,
-			"client_certificate":     cluster.MasterAuth.ClientCertificate,
-			"client_key":             cluster.MasterAuth.ClientKey,
-			"cluster_ca_certificate": cluster.MasterAuth.ClusterCaCertificate,
+			"username":                 cluster.MasterAuth.Username,
+			"password":                 cluster.MasterAuth.Password,
+			"client_certificate":       cluster.MasterAuth.ClientCertificate,
+			"client_key":               cluster.MasterAuth.ClientKey,
+			"cluster_ca_certificate":   cluster.MasterAuth.ClusterCaCertificate,
+			"issue_client_certificate": true,
```
danawillow: Instead of setting it to true here and potentially overriding it later, how about just set it later?

google/resource_container_cluster_test.go
```diff
@@ -635,6 +687,7 @@ resource "google_container_cluster" "with_master_auth" {
 	master_auth {
 		username = "mr.yoda"
 		password = "adoy.rm"
+		issue_client_certificate = false
```
danawillow: Does this change anything from what was there before? I think if you want to test `issue_client_certificate` it probably makes more sense to set it to `true`

google/resource_container_cluster.go
```diff
@@ -66,6 +66,11 @@ func resourceContainerCluster() *schema.Resource {
 							Required: true,
 							ForceNew: true,
 						},
+						"issue_client_certificate": {
+							Type:             schema.TypeBool,
+							Optional:         true,
```
danawillow: I think this needs to be `ForceNew` as well.

google/resource_container_cluster.go
```diff
@@ -66,6 +66,11 @@ func resourceContainerCluster() *schema.Resource {
 							Required: true,
 							ForceNew: true,
 						},
+						"issue_client_certificate": {
```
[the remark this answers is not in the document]
brianhealeyRMN: Updated the orders for the schema to be alphabetical

google/resource_container_cluster.go
brianhealeyRMN: Done

google/resource_container_cluster.go
```diff
@@ -66,6 +66,11 @@ func resourceContainerCluster() *schema.Resource {
...
+							Type:             schema.TypeBool,
+							Optional:         true,
+							DiffSuppressFunc: linkDiffSuppress,
```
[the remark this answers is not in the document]
brianhealeyRMN: Removed

google/resource_container_cluster.go
```diff
@@ -266,6 +271,60 @@ func resourceContainerCluster() *schema.Resource {
 						},
 
 						"node_config": schemaNodeConfig,
+
+						"management": {
```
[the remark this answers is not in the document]
brianhealeyRMN: done

google/resource_container_cluster.go
brianhealeyRMN: Modified as suggested

google/resource_container_cluster.go
```diff
@@ -266,6 +271,60 @@ func resourceContainerCluster() *schema.Resource {
...
+
+						"auto_scaling": {
+							Type:     schema.TypeList,
+							Optional: true,
+							ForceNew: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+
+									"enabled": {
```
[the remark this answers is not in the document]
brianhealeyRMN: Modified as suggested

google/resource_container_cluster.go
brianhealeyRMN: Not a problem. Default behavior is "true". This is a flag vs an actual setting so I added code so that it behaves "as a setting" by checking for the existence of the client certitiface

google/resource_container_cluster_test.go
brianhealeyRMN: True is the default behavior, so setting it to false does change, however I think the larger concern is that both cases were not being tested.  I added test specifically to test true/false

brianhealeyRMN:
Thanks @danawillow 
I updated the the PR with your suggestions and fixed the integration tests.  I also added additional tests to provide better coverage.

davidquarles:
Yeah, this is awesome @brianhealeyRMN!  @danawillow and I were discussing version upgrades in #633, which led me here.  Would it be possible to extend the `nodePool.Management` work to target the node pools directly, rather than (or in addition to?) the parent cluster resource, and to handle the update lifecycle as well?  I can totally help, if you'd like to collaborate.  I can also open a new issue and start from scratch, if this is entirely out of scope here.

justinsb:
I would love to see the `issue_client_certificate` support land, that was implemented in this PR.  @brianhealeyRMN do you think you might rebase or split this PR up?  Or would you mind if I (or someone else, if anyone else is keen) added that support?

danawillow:
I'm not @brianhealeyRMN, but I think you should go for it @justinsb :)

paddycarver:
This PR has been inactive for a long time, with no response, so I'm going to close it out. If the author wants to carry on with it, feel free to post in this thread, and we can re-open it. If someone else is interested in picking this up, feel free to open a new PR for it.

PR #395, opened 2017-09-07
Add support for labels to compute_global_forwarding_rule

### LINKED ISSUE #354: Feature request: Labels support for Global Forwarding Rule
Hey folks!

Labels landed in alpha for Global Forwarding Rule (GFW). The API is at https://developers.google.com/apis-explorer/#search/globalforwardingrule/m/compute/alpha/compute.globalForwardingRules.insert

Here's the use-case: I'm building a tool to automate the creation, revocation, and renewal of Let's Encrypt/acme certificates for GFWs. I would love to label global forwarding rules that should have certs, something like:

```
labels: {
  "domain": "a.example.com",
  "expires": "2017-08-24T16:22:42Z"
}
```

This would be super neat because I wouldn't need to store information about which GFWs need certs in a separate database. The GFW List and Get APIs become the database. Cool!

Thanks for considering!

selmanj:
Closes #354

### DIFF
--- a/google/resource_compute_global_forwarding_rule.go
+++ b/google/resource_compute_global_forwarding_rule.go
@@ -156,26 +156,16 @@ func resourceComputeGlobalForwardingRuleCreate(d *schema.ResourceData, meta inte
 		return err
 	}
 
-	// If we have labels to update, try to set those too
+	// If we have labels to set, try to set those too
 	if _, ok := d.GetOk("labels"); ok {
-		if computeApiVersion != v0beta {
-			return fmt.Errorf(
-				"Unable to set labels due to an internal error: can only handle v0beta but compute api logic indicates %d",
-				computeApiVersion)
-		}
-
+		labels := expandLabels(d)
 		// Do a read to get the fingerprint value so we can update
-		createdFrule, err := config.clientComputeBeta.GlobalForwardingRules.Get(project, frule.Name).Do()
+		fingerprint, err := resourceComputeGlobalForwardingRuleReadLabelFingerprint(config, computeApiVersion, project, frule.Name)
 		if err != nil {
-			return fmt.Errorf("Unable to read global forwarding rule to update labels: %s", err)
+			return err
 		}
 
-		setLabels := computeBeta.GlobalSetLabelsRequest{
-			Labels:           expandLabels(d),
-			LabelFingerprint: createdFrule.LabelFingerprint,
-		}
-		op, err := config.clientComputeBeta.GlobalForwardingRules.SetLabels(project, frule.Name, &setLabels).Do()
-		err = computeSharedOperationWait(config, op, project, "Setting labels on Global Fowarding Rule")
+		err = resourceComputeGlobalForwardingRuleSetLabels(config, computeApiVersion, project, frule.Name, labels, fingerprint)
 		if err != nil {
 			return err
 		}
@@ -235,29 +225,15 @@ func resourceComputeGlobalForwardingRuleUpdate(d *schema.ResourceData, meta inte
 		d.SetPartial("target")
 	}
 	if d.HasChange("labels") {
-		// Only able to set labels via the beta API right now. Fail loudly if we're told to use a different api.
-		if computeApiVersion != v0beta {
-			return fmt.Errorf(
-				"Unable to update labels due to an internal error: can only handle v0beta but compute api logic indicates %d",
-				computeApiVersion)
-		}
+		labels := expandLabels(d)
+		fingerprint := d.Get("label_fingerprint").(string)
 
-		gslr := computeBeta.GlobalSetLabelsRequest{
-			Labels:           expandLabels(d),
-			LabelFingerprint: d.Get("label_fingerprint").(string),
-		}
-
-		op, err := config.clientComputeBeta.GlobalForwardingRules.SetLabels(project, d.Id(), &gslr).Do()
+		err = resourceComputeGlobalForwardingRuleSetLabels(config, computeApiVersion, project, d.Get("name").(string), labels, fingerprint)
 		if err != nil {
 			return err
 		}
 
 		d.SetPartial("labels")
-
-		err = computeSharedOperationWait(config, op, project, "Updating labels on Global Forwarding Rule")
-		if err != nil {
-			return err
-		}
 	}
 
 	d.Partial(false)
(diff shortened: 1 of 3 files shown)


google/resource_compute_global_forwarding_rule.go
```diff
@@ -141,6 +156,31 @@ func resourceComputeGlobalForwardingRuleCreate(d *schema.ResourceData, meta inte
 		return err
 	}
 
+	// If we have labels to update, try to set those too
+	if _, ok := d.GetOk("labels"); ok {
+		if computeApiVersion != v0beta {
```
rosbo: Could you extract part of this logic in its own method to avoid the duplication? The only difference I see is reading the fingerprint in create and the SetPartial.

google/resource_compute_global_forwarding_rule.go
```diff
@@ -194,6 +234,31 @@ func resourceComputeGlobalForwardingRuleUpdate(d *schema.ResourceData, meta inte
 
 		d.SetPartial("target")
 	}
+	if d.HasChange("labels") {
+		// Only able to set labels via the beta API right now. Fail loudly if we're told to use a different api.
+		if computeApiVersion != v0beta {
```
rosbo: Is that even possible? The computeApiVersion should prevent this case from happening.

google/resource_compute_global_forwarding_rule_test.go
```diff
@@ -254,6 +321,130 @@ func testAccComputeGlobalForwardingRule_basic1(fr, proxy1, proxy2, backend, hc,
...
 }
 
+func testAccComputeGlobalForwardingRule_labels(fr, proxy1, proxy2, backend, hc, urlmap string) string {
+	return fmt.Sprintf(`
+	resource "google_compute_global_forwarding_rule" "foobar" {
+		description = "Resource created for Terraform acceptance testing"
+		ip_protocol = "TCP"
+		name = "%s"
+		port_range = "80"
+		target = "${google_compute_target_http_proxy.foobar1.self_link}"
```
rosbo: I believe target is optional. Since you are already testing the "setting a target" part in the basic tests, you can speed up this test by removing the target. You can then remove the target http proxies, health checks and backend service.

google/resource_compute_global_forwarding_rule.go
selmanj: I tried, but even though there's duplication the function needs a lot of the surrounding state for context; the method signature looks like:

```go
func resourceComputeGlobalForwardingRuleSetLabels(config *Config, project, name string, labels map[string]string, labelFingerprint string) error {
```

With 5 parameters, I tend to think this logic is clearer duplicated twice... But if you disagree or know of a better way let me know.

google/resource_compute_global_forwarding_rule.go
selmanj: It's not, at least in the current state. This is more to make sure we get a nice error message if someone changes the `Feature` to no longer be in v0beta (for example, if labels is supported in v1).

I could have written it to use the configured client api version but that felt silly since you always have to use the beta api currently.

google/resource_compute_global_forwarding_rule_test.go
selmanj: I got so excited when I read this, but an attempt yields:

```
		Errors: []string{"google_compute_global_forwarding_rule.foobar: \"target\": required field is not set"}
```
😞

rosbo: Oh. I was reading the docs for `compute_forwarding_rule` for which the target field is optional... https://www.terraform.io/docs/providers/google/r/compute_forwarding_rule.html

For `global_forwarding_rule`, we set the attribute as `Required=true`. Is it an artificial constraint we added or is it constrained by the API. It seems weird to me that the behavior of the API for the global vs regional forwarding rule would differ on this.

google/resource_compute_global_forwarding_rule.go
rileykarson: Doesn't guarding `computeApiVersion` like this (and above) just duplicate the information about the required version for a feature that lives in the `Feature` struct? It now lives in two places, and they might drift.

Guarding against incorrect changes to the `Feature` version for some feature is done by the tests already; they would fail if someone incorrectly changed the API version from `v0beta` to `v1`, since they would read at Beta and see that the value is not what is expected.

google/resource_compute_global_forwarding_rule_test.go
rosbo: I just tested, GCP API requires target to be set for global forwarding rule. You can forget this thread.

google/resource_compute_global_forwarding_rule.go
rosbo: +1

google/resource_compute_global_forwarding_rule.go
rosbo: I would rather have this method than duplicate a block of 20 lines of code.

selmanj: I think it's more like 6-7 lines; I'll update the code and you can decide if it's improved or not.

Personally I tend to avoid factoring out functions that can't be named nicely or require a lot of surrounding context unless they're duplicated quite a lot.  Obviously there's a lot of subjectivity here so I'm fine either way.

google/resource_compute_global_forwarding_rule.go
selmanj: The tests would catch this without the explicit guard I wrote as they wouldn't see the label being set anymore upon creation. This would probably be surprising; hence the explicit guard.

If this was a normal feature that was just another beta attribute to be set on the `GlobalForwardingRule` on creation, then everything would just work. However because only the beta api has a `SetLabels` method, we can only write the beta codepath.

Does that make sense?

rosbo: My suggestion was to remove this explicit guard since the Feature already guards it. Like you said, tests would catch a bug if this wasn't true. And leave only the beta codepath and maybe a `// Labels for GlobalForwardingRules are in beta` comment.

selmanj: Actually, I was wrong; without the guard, then we will always use the beta api to call `SetLabels`. All the tests would pass but we'd be using the wrong API.

rosbo: Which is what we want! If labels gets modified (remember, we are in a HasChange("labels") block)  then we need to be using the API to either add, update or delete labels.

selmanj: Yes, that works now. But in the future, when `Labels` is supported on v1 and someone removes it from the beta Features list, we will continue to use the beta API instead of the v1 api and all of the tests will continue to pass. This is undesirable as beta APIs have different SLAs compared to non-beta.

rosbo: Yes, once Labels becomes GA, we will need to update this code to remove it from the future and add the v1 codepath

selmanj: The 'add the v1 codepath' is something I might forget to do, especially as a normal beta feature would not require an explicit 'add the v1 codepath' (even more surprisingly, other resources that use `Labels` in beta don't require an explicit codepath to be added as they are able to set labels upon creation). I prefer the tests to catch this bug rather than relying on someone to explicitly check for this.

Can you quantify what the cost is of leaving the guard in? I'm not seeing why this is controversial.

rosbo: I see the value in your argument that 'at some point it will become GA and this safeguard will make sure we don't forget to update this part of the code'. I am fine with leaving it in.

If we didn't know for sure that we would have to update this code, I would feel strongly for removing an explicit guard guarding nothing because IMO this would be equivalent to leaving dead code.

But again, in this particular situation, we know that GA is coming for sure and I see the value in leaving it in.

Thanks

rileykarson: Good point! This does help us more explicitly manage the version.

What if we added the switch on `computeApiVersion` now, and made the `v1` block throw this error instead of wrapping all this in an if? That way we still fail tests, and it's very explicit where the code will need to go to fix the problem; we'll also make smaller diffs when the feature goes `v1` since we will match the convention for versioned resources more closely.

google/resource_compute_global_forwarding_rule.go
selmanj: This is done, please take a look.

google/resource_compute_global_forwarding_rule.go
selmanj: Done. I'm not sure it will make a smaller diff but it's pretty much isomorphic to the original code.

rosbo:
```sh
TF_ACC=1 go test ./google -v -run TestAccComputeGlobalForwardingRule_labels -timeout 120m
=== RUN   TestAccComputeGlobalForwardingRule_labels
--- PASS: TestAccComputeGlobalForwardingRule_labels (135.37s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	135.505s
```

Issue #372, opened 2017-08-31
Support Shared VPC/XPN in GCE


ebonydad:
### Terraform Version

v0.10.2

### Expected Behavior

To be able to create, provision, deprovision Shared VPC/XPN via Terraform:

### Documentation:

https://cloud.google.com/compute/docs/shared-vpc/provisioning-shared-vpc

### SDK:

https://cloud.google.com/sdk/gcloud/reference/compute/xpn/

### Actual Behavior

No command sets are currently in Terraform.

paddycarver:
I _think_ this may already be supported. Does https://www.terraform.io/docs/providers/google/r/compute_instance.html#subnetwork_project do what you're looking for? Could you help me understand better if it doesn't fit your needs? Sorry, I'm not overly familiar with XPN.

danawillow:
@paddycarver, we have support for using XPN but not for provisioning it. @rosbo and I are working on adding support for that now so I'm going to go ahead and unassign you from this :)

danawillow:
Fixed in #396.

PR #396, opened 2017-09-08
Add new `google_compute_shared_vpc` resource

danawillow:
Part of #372.

(sorry to steal your thunder on this @rosbo, I already had it mostly written before you volunteered for it)


google/resource_compute_shared_vpc_host.go
```diff
@@ -0,0 +1,91 @@
...
+func resourceComputeSharedVpcHost() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceComputeSharedVpcHostCreate,
+		Read:   resourceComputeSharedVpcHostRead,
+		Delete: resourceComputeSharedVpcHostDelete,
+
+		Schema: map[string]*schema.Schema{
+			"project": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
rosbo: This should be required.

website/docs/r/compute_shared_vpc_host.html.markdown
```diff
@@ -0,0 +1,29 @@
...
+resource "google_shared_vpc_host" "host" {
+  project = "your-project-id"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project` - (Optional) The project ID. If not specified, uses the
```
rosbo: I am confused why this field is marked as optional.

google/resource_compute_shared_vpc_host.go
danawillow: Done

website/docs/r/compute_shared_vpc_host.html.markdown
danawillow: Eh, I figured it _could_ be optional since it'll just fall back to the provider project. But it is sort of a weird UX. I'm fine with just making it Required.

google/resource_compute_shared_vpc_host.go
```diff
@@ -0,0 +1,91 @@
...
+	config := meta.(*Config)
+
+	projectID, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	op, err := config.clientCompute.Projects.EnableXpnHost(projectID).Do()
+	if err != nil {
+		return fmt.Errorf("Error disabling XPN Host: %s", err)
```
rosbo: Typo: We are enabling here, not disabling. 

Consider using "Shared VPC" instead of XPN since this is the nomenclature we are using in our docs.

Maybe: "Error enabling shared vpc host for project %s: %s"

google/resource_compute_shared_vpc_host.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+func resourceComputeSharedVpcHost() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceComputeSharedVpcHostCreate,
+		Read:   resourceComputeSharedVpcHostRead,
+		Delete: resourceComputeSharedVpcHostDelete,
+
+		Schema: map[string]*schema.Schema{
+			"project": &schema.Schema{
```
rosbo: For Xpn Resource, we need to specify both the host project and the service project. Should we call this field "host_project" instead? This would add consistency across both resources and potentially be more explicit for users what project is expected here.

This made me think that maybe having both resource merge into one would be more intuitive:

```hcl
resource "google_compute_shared_vpc" "my-shared-vpc" {
  host_project = "my-host-project"
  service_projects = ["my-service-project1", "my-service-project2"]
}

Looking at enableXpnResource method, it can support potentially other xpn resources than project. If we want to add support for new type of xpn resource, we could easily create a new field. Copying the xpn_resource structure would make it less user friendly IMO. 
See: https://cloud.google.com/compute/docs/reference/latest/projects/enableXpnResource

Comment (rosbo) [reply google/resource_compute_shared_vpc_host.go]: The way I see people using this is the following:

Let's say we have 3 projects. `HOST_ PROJECT`, `SERVICE_PROJECT_1` and `SERVICE_PROJECT_2`.

They have 3 terraform directories.

The first one for the `HOST_PROJECT`:
- terraform service credentials must have compute.XpnAdmin role (or the same permissions through another role)
- sets up the network, firewall rules, etc.
- grants compute.networkUser role to both service projects.
- sets up XPN. Enabling the host and adding the 2 service projects as xpn resources. Since enabling the xpn host and adding xpn resource is likely to be done by the same Terraform config/user, IMO, it makes sense to bundle them up together.

Each service directory:
- Adds the necessary compute/container resources to perform the operations the service is responsible for.
- Since it has compute.networkUser role on the host project, it can read from the host network whatever it needs.

File: google/resource_compute_shared_vpc_host.go [hunk 5]
```diff
@@ -0,0 +1,91 @@
...
+
+func resourceComputeSharedVpcHostCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	projectID, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	op, err := config.clientCompute.Projects.EnableXpnHost(projectID).Do()
```
rosbo: What is the behavior of the EnableXpnHost API call if Xpn is already enabled? We should make sure that we don't fail the create operation if xpnHost is already enabled.

google/resource_compute_shared_vpc_host.go
danawillow: Done

google/resource_compute_shared_vpc_host.go
danawillow: Yup, that makes sense. Changed to your proposal, let me know what you think

google/resource_compute_shared_vpc_host.go
danawillow: It succeeds!

google/resource_compute_shared_vpc.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func resourceComputeSharedVpcUpdate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	hostProject := d.Get("host_project").(string)
+
+	if d.HasChange("service_projects") {
+		old, new := d.GetChange("service_projects")
+		// Helper to convert slice to map
+		m := func(vals []interface{}) map[string]struct{} {
```
rosbo: Consider moving this to our `utils.go` file.

danawillow: Done.

PR #401, opened 2017-09-08
Add DSF for address' region field.

paddycarver:
Region used to require the full self_link of a region to be in the
config. Now it accepts the name or the self_link.

Note that this will have weird consequences if referenced from a
field that doesn't support both name and self_link as well, but that
honestly is a bug in the other field, and should be fixed there, in my
opinion.

The other option is a StateFunc that translates everything to a
self_link, at least making things consistent in state, but I wasn't sure
it was worth the effort. IF we like, I can take that route instead.

### DIFF
--- a/google/resource_compute_address.go
+++ b/google/resource_compute_address.go
@@ -4,10 +4,11 @@ import (
 	"fmt"
 	"log"
 
-	"github.com/hashicorp/terraform/helper/schema"
-	"google.golang.org/api/compute/v1"
 	"regexp"
 	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/compute/v1"
 )
 
 var (
@@ -46,10 +47,11 @@ func resourceComputeAddress() *schema.Resource {
 			},
 
 			"region": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
-				ForceNew: true,
+				Type:             schema.TypeString,
+				Optional:         true,
+				Computed:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: linkDiffSuppress,
 			},
 
 			"self_link": &schema.Schema{
--- a/google/resource_compute_address_test.go
+++ b/google/resource_compute_address_test.go
@@ -93,6 +93,25 @@ func TestAccComputeAddress_basic(t *testing.T) {
 	})
 }
 
+func TestAccComputeAddress_region(t *testing.T) {
+	var addr compute.Address
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeAddressDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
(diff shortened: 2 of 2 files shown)


rosbo:
Hey,

I am confused. It works already for name-only. 

`testAccComputeAddress_basic` already test this since it gets the region from the config which is the name only.

Did you meant to support also self_link without perpetual diff?

If so, you should update your test to reflect this and update the description of this PR.

paddycarver:
I'm a little confused too, to be honest. But if you comment out the DiffSuppressFunc line, then run the included test, you get this:

```
make testacc TEST=./google TESTARGS='-run=TestAccComputeAddress'
TF_ACC=1 go test ./google -v -run=TestAccComputeAddress -timeout 120m
--- PASS: TestAccComputeAddress_importBasic (23.46s)
--- PASS: TestAccComputeAddress_basic (23.03s)
=== RUN   TestAccComputeAddress_region
--- FAIL: TestAccComputeAddress_region (23.07s)
        testing.go:434: Step 0 error: After applying this step, the plan was not empty:

                DIFF:

                DESTROY/CREATE: google_compute_address.foobar
                  address:   "104.154.172.214" => "<computed>"
                  name:      "test-1596546341124767833" => "test-1596546341124767833"
                  region:    "https://www.googleapis.com/compute/v1/projects/hc-terraform-testing/regions/us-central1" => "us-central1" (forces new resource)
                  self_link: "https://www.googleapis.com/compute/v1/projects/hc-terraform-testing/regions/us-central1/addresses/test-1596546341124767833" => "<computed>"

                STATE:

                google_compute_address.foobar:
                  ID = projects/hc-terraform-testing/regions/us-central1/addresses/test-1596546341124767833
                  address = 104.154.172.214
                  name = test-1596546341124767833
                  region = https://www.googleapis.com/compute/v1/projects/hc-terraform-testing/regions/us-central1
                  self_link = https://www.googleapis.com/compute/v1/projects/hc-terraform-testing/regions/us-central1/addresses/test-1596546341124767833
FAIL
exit status 1
FAIL    github.com/terraform-providers/terraform-provider-google/google 69.562s
GNUmakefile:15: recipe for target 'testacc' failed
make: *** [testacc] Error 1
```
My understanding is thus:

* There's a `region` field in the config that can be set, which (because it's in the config) Terraform will try to enforce.
* The region configured on the provider is not in the config, so Terraform won't care if there's drift.
* The new test specifies the region in the config, so drift is now detected and Terraform returns a diff.
* The `Region` property of the returned address is in the form of a `self_link`, not a name.
* Because the `Region` property does not match what's in the `region` field of the config, we get a diff.

Because of this, using a name to set the region of the address _in the config_ creates a perpetual diff.

rosbo:
Do we want to support specifying in the region using a self_link too?

If so, use `compareSelfLinkOrResourceName` instead of `linkDiffSuppress`. This way, you can support beta vs v1 links without creating a perpetual diff.

Also, add a test with a region self_link if we want to support this.

If we decide to support only the region name, feel free to merge this PR as-is.

rosbo:
Closing. Fixed in #422. See:
https://github.com/terraform-providers/terraform-provider-google/commit/f83618624cf97dbba05f98fa1a414ceaec483ecf

PR #402, opened 2017-09-09
Cross-reference multi-zone and single-zone documentation

rosbo:
Also fix the API link for `google_compute_instance_group_manager` documentation.

cc/ @selmanj

### DIFF
--- a/website/docs/r/compute_region_instance_group_manager.html.markdown
+++ b/website/docs/r/compute_region_instance_group_manager.html.markdown
@@ -13,6 +13,8 @@ of homogeneous Compute Engine virtual machine instances from a common instance
 template. For more information, see [the official documentation](https://cloud.google.com/compute/docs/instance-groups/distributing-instances-with-regional-instance-groups)
 and [API](https://cloud.google.com/compute/docs/reference/latest/regionInstanceGroupManagers)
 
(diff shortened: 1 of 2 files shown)


website/docs/r/compute_region_instance_group_manager.html.markdown
```diff
@@ -13,6 +13,8 @@ of homogeneous Compute Engine virtual machine instances from a common instance
 template. For more information, see [the official documentation](https://cloud.google.com/compute/docs/instance-groups/distributing-instances-with-regional-instance-groups)
 and [API](https://cloud.google.com/compute/docs/reference/latest/regionInstanceGroupManagers)
 
+~> **Note:** Use [google_compute_instance_group_manager](/docs/providers/google/r/google_compute_instance_group_manager.html) to create a single-zone instance group manager.
```
selmanj: Is '~>' valid markdown? I've never seen that before

selmanj: If it's not I'd prefer to drop it as the ** emphasis should be enough

rosbo: It is an extension of markdown. It creates a block with a light yellow background. You can see an example here:
https://www.terraform.io/docs/providers/google/r/compute_network_peering.html

We uses it already in 9 different pages in our docs.

selmanj: Thanks, it's completely new to me.

PR #410, opened 2017-09-12
Refactor tests to defer to common method when looking up project

selmanj:


google/provider_test.go
```diff
@@ -40,41 +59,16 @@ func testAccPreCheck(t *testing.T) {
...
 	}
 
-	projs := []string{
-		"GOOGLE_PROJECT",
-		"GCLOUD_PROJECT",
-		"CLOUDSDK_CORE_PROJECT",
-	}
-	if v := multiEnvSearch(projs); v == "" {
-		t.Fatalf("One of %s must be set for acceptance tests", strings.Join(projs, ", "))
+	if v := multiEnvSearch(projectEnvVars); v == "" {
```
rosbo: Consider using the `getTestProjectFromEnv` you created below.

rosbo:
Thanks for cleaning this up :)

google/provider_test.go
selmanj: I couldn't decide which was better; they are functionally equivalent but I felt the current implementation was less surprising since the others don't have get functions...

Which do you prefer?

selmanj: OK then I'll leave it as is, not because I think it's better but because I'm lazy

PR #416, opened 2017-09-13
Create, update, move and delete a GCP folders.

### LINKED ISSUE #412: Create, update, move, delete a GCP folder.
Creates a new resource `google_folder`.

It should support top-level folder (folder directly under a GCP organization) and nested folder (folder inside a folder). It should also support moving a folder inside a different folder or a move the a top-level folder.

Implementation note: Use the resource manager v2 folders API.

rosbo:
Fixes #412 

cc/ @danawillow, @srivasta, @ubschmidt2


google/resource_google_folder.go
```diff
@@ -0,0 +1,160 @@
...
+	err = resourceManagerV2Beta1OperationWait(config, op, "creating folder")
+
+	if err != nil {
+		return fmt.Errorf("Error creating folder '%s' in '%s': %s", displayName, parent, err)
+	}
+
+	// Since we waited above, the operation is guaranteed to have been successful by this point.
+	waitOp, err := config.clientResourceManager.Operations.Get(op.Name).Do()
+	if err != nil {
+		return fmt.Errorf("The folder '%s' has been created but we could not retrieve its id. Delete the folder manually and retry or use 'terraform import'.", displayName)
```
selmanj: The err is eaten here; should probably include it in the error message

google/resource_google_folder.go
```diff
@@ -0,0 +1,160 @@
...
+		Update: resourceGoogleFolderUpdate,
+		Delete: resourceGoogleFolderDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: resourceGoogleFolderImportState,
+		},
+
+		Schema: map[string]*schema.Schema{
+			// Format is 'folders/{folder_id}.
+			// The folder id holds the same value.
```
selmanj: I think you mean 'id' or 'folder_id' as in terraform id, right? This comment was confusing to me because I don't see folder_id anywhere (this is exacerbated by the fact that other resources have fields like instance_id (which I am responsible for sorry) that are distinct from id).

google/resource_google_folder.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return fmt.Errorf("Error creating folder '%s' in '%s': %s", displayName, parent, err)
+	}
+
+	// Since we waited above, the operation is guaranteed to have been successful by this point.
+	waitOp, err := config.clientResourceManager.Operations.Get(op.Name).Do()
+	if err != nil {
+		return fmt.Errorf("The folder '%s' has been created but we could not retrieve its id. Delete the folder manually and retry or use 'terraform import'.", displayName)
+	}
+	response := waitOp.Response.(map[string]interface{})
```
selmanj: I'm surprised this works, the documentation seems to indicate that there should be some sort of struct here.

Given we don't control the code that is in Response and the type assertion will cause a big nasty error if it fails, how about using the checked type assertion here? eg:
```go
response, ok := waitOp.Response.(map[string]interface{})
if !ok {
  // ...
}
```

google/resource_google_folder.go
```diff
@@ -0,0 +1,160 @@
...
+	return nil
+}
+
+func resourceGoogleFolderUpdate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	displayName := d.Get("display_name").(string)
+
+	d.Partial(true)
+	if d.HasChange("display_name") {
+		_, err := config.clientResourceManagerV2Beta1.Folders.Patch(d.Id(), &resourceManagerV2Beta1.Folder{
```
selmanj: Any way you could combine the patch code for display_name and parent into one api call? This would result in less code and be less likely to fail.

google/resource_google_folder.go
```diff
@@ -0,0 +1,160 @@
...
+
+func resourceGoogleFolderDelete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	displayName := d.Get("display_name").(string)
+
+	_, err := config.clientResourceManagerV2Beta1.Folders.Delete(d.Id()).Do()
+	if err != nil {
+		return fmt.Errorf("Error deleting folder %s", displayName)
+	}
+
```
selmanj: Missing a `d.SetId("")` call to indicate successful deletion

Although apparently that isn't needed since the tests passed? I'm curious if we should be doing that or not.

google/resource_google_folder_test.go
```diff
@@ -0,0 +1,163 @@
...
+	parent := "organizations/" + org
+	folder := &resourceManagerV2Beta1.Folder{}
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGoogleFolderDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccGoogleFolder_basis(folderDisplayName, parent),
```
selmanj: minor nit
basis -> basic (unless this config is really the 'basis' for the test?)

google/resource_google_folder_test.go
```diff
@@ -0,0 +1,163 @@
...
+)
+
+func TestAccGoogleFolder_rename(t *testing.T) {
+	skipIfEnvNotSet(t, "GOOGLE_ORG")
+
+	folderDisplayName := "tf-test-" + acctest.RandString(10)
+	newFolderDisplayName := "tf-test-renamed-" + acctest.RandString(10)
+	org := os.Getenv("GOOGLE_ORG")
+	parent := "organizations/" + org
+	folder := &resourceManagerV2Beta1.Folder{}
```
selmanj: Another nit; I've seen other tests declare the resource variable as the struct type, not the pointer type (e.g. `resourceManagerV2Beta1.Folder` rather than `*resourceManagerV2Beta1.Folder`. The pointer allocation here feels a bit less explicit. But if this was intentional go ahead and leave it in as this is mainly stylistic.

google/resource_google_folder_test.go
```diff
@@ -0,0 +1,163 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGoogleFolderExists("google_folder.folder1", folder),
+					testAccCheckGoogleFolderParent(folder, parent),
+					testAccCheckGoogleFolderDisplayName(folder, newFolderDisplayName),
+				)},
+		},
+	})
+}
+
+func TestAccGoogleFolder_move(t *testing.T) {
```
selmanj: I got this initially confused with rename (move in hierarchies is often used to rename the element, like in filesystems). Consider adding the suffix `_moveParent` to the test name?

google/resource_google_folder_test.go
```diff
@@ -0,0 +1,163 @@
...
+					testAccCheckGoogleFolderParent(folder, parent),
+					testAccCheckGoogleFolderDisplayName(folder, folder1DisplayName),
+				),
+			},
+			resource.TestStep{
+				Config: testAccGoogleFolder_move(folder1DisplayName, folder2DisplayName, parent),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGoogleFolderExists("google_folder.folder1", folder),
+					testAccCheckGoogleFolderDisplayName(folder, folder1DisplayName),
+					testAccCheckGoogleFolderExists("google_folder.folder2", folder),
```
selmanj: Can you use a separate variable here for folder 2? Overwriting folder with the contents of folder 2 and then making similar assertions got really hard to read for me

website/docs/r/google_folder.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+    Must be of the form `folders/{folder_id}` or `organizations/{org_id}`.
+
+## Attributes Reference
+
+In addition to the arguments listed above, the following computed attributes are
+exported:
+
+* `name` - The resource name of the Folder. Its format is folders/{folder_id}.
+* `lifecycle_state` - The lifecycle state of the folder such as `ACTIVE` or `DELETE_REQUESTED`.
+* `name` - Timestamp when the Folder was created. Assigned by the server.
```
selmanj: name -> create_time

google/resource_google_folder.go
rosbo: Replaced 'folder id' by 'terraform id'

google/resource_google_folder.go
rosbo: Fixed.

google/resource_google_folder.go
rosbo: Good idea. Done

google/resource_google_folder.go
rosbo: No we can't.  The display_name must be updated with `Patch` and the parent must be updated with `Move`

google/resource_google_folder.go
rosbo: In resource.go in terraform core:

```go
// Destroy the resource since it is created
if err := r.Delete(data, meta); err != nil {
  return r.recordCurrentSchemaVersion(data.State()), err
}

// Make sure the ID is gone.
data.SetId("")
```

It does it for us. Calling it again won't cause harm. Should I add it for consistency?

google/resource_google_folder_test.go
rosbo: Done

google/resource_google_folder_test.go
rosbo: I meant the basis because I use it as a basic for two tests but I guess the intention wasn't clear. I renamed it basic.

google/resource_google_folder_test.go
rosbo: renamed to `_moveParent`

google/resource_google_folder_test.go
rosbo: Done

website/docs/r/google_folder.html.markdown
rosbo: Thanks :)

google/resource_google_folder.go
```diff
@@ -0,0 +1,166 @@
...
+		Delete: resourceGoogleFolderDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: resourceGoogleFolderImportState,
+		},
+
+		Schema: map[string]*schema.Schema{
+			// Format is 'folders/{folder_id}.
+			// The terraform id holds the same value.
+			"name": &schema.Schema{
```
danawillow: The convention we've mostly been falling into is to list all Required fields first, then Optional, then Computed. I think that would make it more clear that display_name is what the user sets, rather than name (since for so many other resources it's name)

google/resource_google_folder.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return fmt.Errorf("Error creating folder '%s' in '%s': %s", displayName, parent, err)
+	}
+
+	// Since we waited above, the operation is guaranteed to have been successful by this point.
+	waitOp, err := config.clientResourceManager.Operations.Get(op.Name).Do()
+	if err != nil {
+		return fmt.Errorf("The folder '%s' has been created but we could not retrieve its id. Delete the folder manually and retry or use 'terraform import': %s", displayName, err)
+	}
+	response, ok := waitOp.Response.(map[string]interface{})
```
danawillow: nit: if you felt like saving a line of code, you could also do
`if response, ok := waitOp.Response.(map[string]interface{}); ok {`

website/docs/r/google_folder.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+# google\_folder
+
+Allows management of a Google Cloud Platform folder. For more information see 
+[the official documentation](https://cloud.google.com/resource-manager/docs/creating-managing-folders)
+and 
+[API](https://cloud.google.com/resource-manager/reference/rest/v2/folder).
+
+A folder can contain projects, other folders, or a combination of both. You can use folders to group projects under an organization in a hierarchy. For example, your organization might contain multiple departments, each with its own set of Cloud Platform resources. Folders allows you to group these resources on a per-department basis. Folders are used to group resources that share common IAM policies.
+
+Folders created lives inside an Organization. See the [Organization documentation](https://cloud.google.com/resource-manager/docs/quickstarts) for more details.
```
danawillow: nit: s/lives/live

google/resource_google_folder.go
```diff
@@ -0,0 +1,166 @@
...
+			// Format is 'folders/{folder_id}.
+			// The terraform id holds the same value.
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			// Format is either folders/{folder_id} or organizations/{org_id}.
+			"parent": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
danawillow: Should this be required?

website/docs/r/google_folder.html.markdown
paddycarver: "Created folders live", or some other variation?

google/resource_google_folder.go
selmanj: I missed that, thanks.

google/resource_google_folder.go
selmanj: Good to know, will omit that on delete in the future.

google/resource_google_folder.go
selmanj: I'd prefer if the ok was negated altogether as it reduces nesting. In general if one of the paths leads to an error (or other early return) that's the one that I nest inside if.

It's also a bit more explicit (it requires a bit of careful reading to realize that when `val, ok := response["name"]; ok` evaluates to false, it falls to the error at the bottom.

selmanj:
Sorry, meant to change review status to just "comments" without explicit approval but apparently I can only switch from request changes -> approve.

This is good to merge with me as soon as everyone else's comments are addressed.

google/resource_google_folder.go
rosbo: Done

google/resource_google_folder.go
rosbo: Yes

google/resource_google_folder.go
rosbo: Discussed offline. Added a comment.

website/docs/r/google_folder.html.markdown
rosbo: Done

rosbo: Done

rosbo:
```sh
TF_ACC=1 go test ./google -v -run TestAccGoogleFolder_ -timeout 120m
=== RUN   TestAccGoogleFolder_import
--- PASS: TestAccGoogleFolder_import (13.73s)
=== RUN   TestAccGoogleFolder_rename
--- PASS: TestAccGoogleFolder_rename (14.40s)
=== RUN   TestAccGoogleFolder_moveParent
--- PASS: TestAccGoogleFolder_moveParent (37.29s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	65.553s
```

PR #391, opened 2017-09-06
Container cluster importable

drzero42:
Fixes #165 
I had to change resourceContainerClusterRead to use container.Network rather than d.Get("network"), otherwise it would set network = "" during import. Not sure why network was set this way before?

PR #419, opened 2017-09-13
Add DiffSuppressFunc to GKE cluster networks

danawillow:
`TestAccContainerCluster_network` was failing with a perma-diff after #391 because GKE returns the network as a name only, but we allow users to set it to the URL. Add a DiffSuppressFunc to fix the test:
```
$ make testacc TEST=./google TESTARGS='-run=TestAccContainerCluster_network'
TF_ACC=1 go test ./google -v -run=TestAccContainerCluster_network -timeout 120m
=== RUN   TestAccContainerCluster_network
--- PASS: TestAccContainerCluster_network (348.82s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	348.976s
```

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -176,10 +176,11 @@ func resourceContainerCluster() *schema.Resource {
 			},
 
 			"network": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Default:  "default",
-				ForceNew: true,
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "default",
+				ForceNew:         true,
+				DiffSuppressFunc: linkDiffSuppress,
 			},
 			"subnetwork": {
 				Type:     schema.TypeString,
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -176,11 +176,11 @@ func resourceContainerCluster() *schema.Resource {
 			},
 
 			"network": {
-				Type:             schema.TypeString,
-				Optional:         true,
-				Default:          "default",
-				ForceNew:         true,
-				DiffSuppressFunc: linkDiffSuppress,
+				Type:      schema.TypeString,
+				Optional:  true,
+				Default:   "default",
+				ForceNew:  true,
+				StateFunc: StoreResourceName,
 			},
 			"subnetwork": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


google/resource_container_cluster.go
```diff
@@ -176,10 +176,11 @@ func resourceContainerCluster() *schema.Resource {
...
 			"network": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Default:  "default",
-				ForceNew: true,
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "default",
+				ForceNew:         true,
+				DiffSuppressFunc: linkDiffSuppress,
```
rosbo: You should use `compareSelfLinkOrResourceName`.

google/utils.go
```diff
@@ -251,8 +251,9 @@ func isConflictError(err error) bool {
 }
 
 func linkDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
-	parts := strings.Split(old, "/")
-	if parts[len(parts)-1] == new {
+	oldParts := strings.Split(old, "/")
```
rosbo: The problem with updating this method is that this will cause issue for zonal or regional self-link

If field was "regions/us-east1/someResource/myResource" and now is "regions/us-west1/someResource/myResource" this would suppress the diff. 

I wouldn't update this method and use the other more robust `compareSelfLinkOrResourceName`

google/resource_container_cluster.go
rosbo: Hum. This method assume that the API always return a self-link... It wouldn't work and updating linkDiffSuppress is dangerous because moving a resource to a different zone wouldn't trigger a diff like explained below.

You can either create a new method. Or when you store the network in the Read method, you can generate the relative part of the network and use `compareSelfLinkOrResourceName`.

danawillow:
All right, I solved it in a slightly different way from your suggestion (used a StateFunc instead), let me know what you think.

```
$ make testacc TEST=./google TESTARGS='-run=TestAccContainerCluster_network'
TF_ACC=1 go test ./google -v -run=TestAccContainerCluster_network -timeout 120m
=== RUN   TestAccContainerCluster_network
--- PASS: TestAccContainerCluster_network (402.15s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	402.316sv
```

(if you're happy with this feel free to merge instead of waiting for me since time zones / travel and whatnot)

PR #427, opened 2017-09-13
Expose instance_id as a computed field on compute_instance

selmanj:
Surprisingly, the Id of the compute instance is not currently exposed. Here I added it as `instance_id` to avoid colliding with terraform's internal id.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -250,6 +250,11 @@ func resourceComputeInstance() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"instance_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
 			"zone": &schema.Schema{
 				Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


google/resource_compute_instance.go
```diff
@@ -1175,6 +1180,7 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("cpu_platform", instance.CpuPlatform)
 	d.Set("min_cpu_platform", instance.MinCpuPlatform)
 	d.Set("self_link", ConvertSelfLinkToV1(instance.SelfLink))
+	d.Set("instance_id", fmt.Sprintf("%d", instance.Id))
```
rosbo: Have you considered using https://golang.org/pkg/strconv/#Itoa

selmanj: I almost did that, but `Itoa` takes an `int` and `instance.Id` is of type int64.

PR #432, opened 2017-09-13
Add stackdriver project sink support

selmanj:
This adds support for management of Stackdriver Logging sinks. The use case is to setup logging exports to buckets, pubsub topics or bigtable datasets.

See the example in the `logging_project_sink.html.markdown` file for an example of using terraform to manage logging a VM.

Adding followup support for logging folders, organizations, billingAccounts etc should be easier once this initial sink is checked in.

Documentation: https://cloud.google.com/logging/docs/api/tasks/exporting-logs

### DIFF
--- a/google/logging_utils.go
+++ b/google/logging_utils.go
@@ -5,8 +5,8 @@ import (
 	"regexp"
 )
 
-// loggingSinkTypes contains all the possible Stackdriver Logging resource types. Used to parse ids safely.
-var loggingSinkTypes = []string{
+// loggingSinkResourceTypes contains all the possible Stackdriver Logging resource types. Used to parse ids safely.
+var loggingSinkResourceTypes = []string{
 	"billingAccount",
 	"folders",
 	"organizations",
@@ -15,9 +15,9 @@ var loggingSinkTypes = []string{
 
 // LoggingSinkId represents the parts that make up the canonical id used within terraform for a logging resource.
 type LoggingSinkId struct {
-	typ     string
-	typName string
-	name    string
+	resourceType string
+	resourceId   string
+	name         string
 }
 
 // loggingSinkIdRegex matches valid logging sink canonical ids
@@ -25,12 +25,12 @@ var loggingSinkIdRegex = regexp.MustCompile("(.+)/(.+)/sinks/(.+)")
 
 // canonicalId returns the LoggingSinkId as the canonical id used within terraform.
 func (l LoggingSinkId) canonicalId() string {
-	return fmt.Sprintf("%s/%s/sinks/%s", l.typ, l.typName, l.name)
+	return fmt.Sprintf("%s/%s/sinks/%s", l.resourceType, l.resourceId, l.name)
 }
 
 // parent returns the "parent-level" resource that the sink is in (e.g. `folders/foo` for id `folders/foo/sinks/bar`)
 func (l LoggingSinkId) parent() string {
-	return fmt.Sprintf("%s/%s", l.typ, l.typName)
+	return fmt.Sprintf("%s/%s", l.resourceType, l.resourceId)
 }
 
 // parseLoggingSinkId parses a canonical id into a LoggingSinkId, or returns an error on failure.
@@ -39,21 +39,22 @@ func parseLoggingSinkId(id string) (*LoggingSinkId, error) {
 	if parts == nil {
 		return nil, fmt.Errorf("unable to parse logging sink id %#v", id)
 	}
-	// If our type is not a valid logging sink type, complain loudly
-	validLoggingSinkType := false
-	for _, v := range loggingSinkTypes {
+	// If our resourceType is not a valid logging sink resource type, complain loudly
+	validLoggingSinkResourceType := false
+	for _, v := range loggingSinkResourceTypes {
 		if v == parts[1] {
-			validLoggingSinkType = true
+			validLoggingSinkResourceType = true
 			break
 		}
 	}
 
-	if !validLoggingSinkType {
-		return nil, fmt.Errorf("Logging type %s is not valid. Valid types: %#v", parts[1], loggingSinkTypes)
+	if !validLoggingSinkResourceType {
+		return nil, fmt.Errorf("Logging resource type %s is not valid. Valid resource types: %#v", parts[1],
+			loggingSinkResourceTypes)
 	}
 	return &LoggingSinkId{
-		typ:     parts[1],
-		typName: parts[2],
-		name:    parts[3],
+		resourceType: parts[1],
+		resourceId:   parts[2],
+		name:         parts[3],
 	}, nil
 }
--- a/google/logging_utils.go
+++ b/google/logging_utils.go
@@ -1,0 +1,59 @@
+package google
+
+import (
+	"fmt"
+	"regexp"
+)
+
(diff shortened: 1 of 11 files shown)


website/docs/r/logging_project_sink.html.markdown
```diff
@@ -0,0 +1,120 @@
...
+layout: "google"
+page_title: "Google: google_logging_project_sink"
+sidebar_current: "docs-google-logging-project-sink"
+description: |-
+  Manages a project-level logging sink.
+---
+
+# google\_logging\_project\_sink
+
+Manages a project-level logging sink.. For more information see
```
rosbo: typo ".."

google/resource_logging_project_sink.go
```diff
@@ -0,0 +1,147 @@
...
+		typ:     "projects",
+		typName: project,
+		name:    name,
+	}
+
+	sink := logging.LogSink{
+		Name:        d.Get("name").(string),
+		Destination: d.Get("destination").(string),
+	}
+	if v, ok := d.GetOk("filter"); ok {
```
rosbo: Do you need this if block? d.Get("filter") will return `""` and should work without this extra if block

google/logging_utils.go
```diff
@@ -0,0 +1,59 @@
...
+	"billingAccount",
+	"folders",
+	"organizations",
+	"projects",
+}
+
+// LoggingSinkId represents the parts that make up the canonical id used within terraform for a logging resource.
+type LoggingSinkId struct {
+	typ     string
+	typName string
```
rosbo: In the docs, the `parent` parameter documentation says: 
```md
The resource in which to create the sink:
"projects/[PROJECT_ID]"
"organizations/[ORGANIZATION_ID]"
"billingAccounts/[BILLING_ACCOUNT_ID]"
"folders/[FOLDER_ID]"
```

Instead of type, we could use resource. 

resourceType instead of typ
resourceId instead of typName

rosbo:
Very good. Just some minor typos and suggestions.

website/docs/r/logging_project_sink.html.markdown
selmanj: Oops, fixed

google/resource_logging_project_sink.go
selmanj: That's easier to read. Fixed.

google/logging_utils.go
selmanj: Your names are much better! Fixed.

google/resource_logging_project_sink_test.go
```diff
@@ -0,0 +1,203 @@
...
+		_, err := config.clientLogging.Projects.Sinks.Get(attributes["id"]).Do()
+		if err == nil {
+			return fmt.Errorf("project sink still exists")
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckLoggingProjectSinkExists(n string, sink *logging.LogSink) resource.TestCheckFunc {
```
danawillow: Since everything this function does is also in the next one, is it worth it to keep them separate?

google/resource_logging_project_sink_test.go
```diff
@@ -0,0 +1,203 @@
...
+		if sink.WriterIdentity != attributes["writer_identity"] {
+			return fmt.Errorf("mismatch on writer_identity: api has %s but client has %s", sink.WriterIdentity, attributes["writer_identity"])
+		}
+
+		return nil
+	}
+}
+
+func testAccLoggingProjectSink_basic(name, bucketName string) string {
+	return fmt.Sprintf(`resource "google_logging_project_sink" "basic" {
```
danawillow: can you put a line break after the `? (makes it easier to see what's going on and copy and paste too) (here and elsewhere)

google/resource_logging_project_sink_test.go
```diff
@@ -0,0 +1,203 @@
...
+	name = "%s"
+	destination = "storage.googleapis.com/${google_storage_bucket.log-bucket.name}"
+	filter = "logName=\"projects/%s/logs/compute.googleapis.com%%2Factivity_log\" AND severity>=ERROR"
+	unique_writer_identity = false
+}
+
+resource "google_storage_bucket" "log-bucket" {
+	name     = "%s"
+}`, name, getTestProjectFromEnv(), bucketName)
```
danawillow: likewise, line break before the `

google/resource_logging_project_sink_test.go
```diff
@@ -0,0 +1,203 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckLoggingProjectSink(
+						"google_logging_project_sink.unique_writer"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccLoggingProjectSink_updateUniqueWriter(t *testing.T) {
```
danawillow: Isn't this updating the bucket (/ the destination), not the unique writer?

website/docs/r/logging_project_sink.html.markdown
```diff
@@ -0,0 +1,120 @@
...
+Note that you must have the "Logs Configuration Writer" IAM role (`roles/logging.configWriter`)
+granted to the credentials used with terraform.
+
+## Example Usage
+
+```hcl
+resource "google_logging_project_sink" "my-sink" {
+    name = "my-pubsub-instance-sink"
+
+    // Can export to pubsub, cloud storage, or bigtable
```
danawillow: https://www.terraform.io/docs/configuration/syntax.html
> Single line comments start with #

google/resource_logging_project_sink.go
```diff
@@ -0,0 +1,145 @@
...
+	d.SetId(id.canonicalId())
+
+	return resourceLoggingProjectSinkRead(d, meta)
+}
+
+func resourceLoggingProjectSinkRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	sink, err := config.clientLogging.Projects.Sinks.Get(d.Id()).Do()
+	if err != nil {
```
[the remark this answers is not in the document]
selmanj: Good catch; added

google/resource_logging_project_sink_test.go
selmanj: Yeah, it is. The test is poorly named. I changed it to '_updatePreservesUniqueWriter' (as that's explicitly tested)

google/resource_logging_project_sink_test.go
selmanj: I kept them separate for now since `testAccCheckLoggingProjectSinkExists` assigns the found api object to a provided `*logging.LogSink` parameter and `testAccCheckLoggingProjectSink` checks the remote object against the locally stored attributes

However, since they both look up the api object I changed `testAccCheckLoggingProjectSink` to take an api object as the first parameter; now it does less duplicate work

google/resource_logging_project_sink_test.go
selmanj: Done

google/resource_logging_project_sink_test.go
selmanj: Done

website/docs/r/logging_project_sink.html.markdown
selmanj: Hah apparently it also accepts // as i've been using them in my hcl unknowingly for a while!

Changed to #

PR #434, opened 2017-09-14
Renaming cluster import test to match other acceptance tests naming scheme

rosbo:
Update the name of the import acceptance test to match the naming scheme of the other acceptances for gke cluster:

TestAccGoogleContainerCluster_import -> TestAccContainerCluster_import

TestAccContainerCluster_basic

This way, when we run tests for `TestAccContainerCluster_`, the import tests are running too.

cc/ @danawillow

### DIFF
--- a/google/import_container_cluster_test.go
+++ b/google/import_container_cluster_test.go
@@ -8,7 +8,7 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 )
 
-func TestAccGoogleContainerCluster_import(t *testing.T) {
+func TestAccContainerCluster_import(t *testing.T) {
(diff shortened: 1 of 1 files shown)


selmanj:
Also, again, really wish we would just move the import tests into the main resource test file; I know it differs from other projects but I feel we'll just keep running into this sort of issue otherwise...

danawillow:
Agreed. @paddycarver, do you know why they're generally kept separate?

paddycarver:
I believe it's just historical. A feature was added later, test files were already huge, so they were just started as new files. I couldn't get a compelling reason for them to _need_ to stay in their own files, so if people combine them, I'm not gonna complain. :)

Issue #411, opened 2017-09-12
Assign/Reassign a google project to a folder.


rosbo:
This affects the existing `google_project` resource.

The following use cases should be supported:
* Create a new project inside an existing folder
(Implementation note: I'd suggest to add a "folder_id" attribute to "google_project". This should only require a minor change to the call to v1/projects/create. If "folder_id" is set, then parent ResourceId.type="folder".)
* Move a TF-managed GCP project between folders and between org node and folder.
(Implementation note: This should be just a matter of updating the project with the appropriate parent ResourceId.)

#132 is the parent task tracking all the work necessary to fully support `Folders` in Terraform.

rosbo:
@ubschmidt2 & @srivasta are working on it

srivasta:
I now have a working solution which passes acceptnce tests.

rosbo:
Great. Add me as a reviewer to your PR whenever it is ready. Thank you Manoj!

srivasta:
Hi,


   I don't have permissions to add reviewers, but
https://github.com/terraform-providers/terraform-provider-google/pull/438
is out there to be reviewed.


   Manoj

On Fri, Sep 15, 2017 at 9:49 AM, Vincent Roseberry <notifications@github.com

PR #438, opened 2017-09-15
Folder support: Assign/Reassign a google project to a folder.

srivasta:
This branch addresses terraform-provider-google issue #411

### DIFF
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -51,9 +51,16 @@ func resourceGoogleProject() *schema.Resource {
 				Required: true,
 			},
 			"org_id": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"folder_id"},
+			},
+			"folder_id": &schema.Schema{
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"org_id"},
 			},
 			"policy_data": &schema.Schema{
 				Type:     schema.TypeString,
@@ -95,11 +102,9 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
 	project := &cloudresourcemanager.Project{
 		ProjectId: pid,
 		Name:      d.Get("name").(string),
-		Parent: &cloudresourcemanager.ResourceId{
-			Id:   d.Get("org_id").(string),
-			Type: "organization",
-		},
 	}
+
+	getParentResourceId(d, project)
 
 	if _, ok := d.GetOk("labels"); ok {
 		project.Labels = expandLabels(d)
@@ -155,7 +160,14 @@ func resourceGoogleProjectRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("labels", p.Labels)
 
 	if p.Parent != nil {
-		d.Set("org_id", p.Parent.Id)
+		switch p.Parent.Type {
+		case "organization":
+			d.Set("org_id", p.Parent.Id)
+			d.Set("folder_id", "")
+		case "folder":
+			d.Set("folder_id", p.Parent.Id)
+			d.Set("org_id", "")
+		}
 	}
 
 	// Read the billing account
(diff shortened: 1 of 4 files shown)


srivasta:
@rosbo I don't have enough rights to actually assign reviewrs. I re-rebased, and resolved all new conflicts

google/resource_google_project.go
```diff
@@ -52,8 +52,13 @@ func resourceGoogleProject() *schema.Resource {
...
 				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Optional: true,
+				Computed: true,
+			},
+			"folder_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
```
rosbo: If you specify a folder, you should not be able to specify an org. To make sure it fails at 
 the `terraform plan` step, add `ConflictsWith: []string{"org_id"}` in the schema definition for the `folder_id` field. And vice-versa

google/resource_google_project.go
```diff
@@ -155,7 +172,12 @@ func resourceGoogleProjectRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("labels", p.Labels)
 
 	if p.Parent != nil {
-		d.Set("org_id", p.Parent.Id)
+		switch p.Parent.Type {
+		case "organization":
+			d.Set("org_id", p.Parent.Id)
```
rosbo: add `d.Set("folder_id", "")` here. and the same for the `org_id field in the folder case below. The reason is that if you move a project from a folder to an org outside of Terraform, we want Terraform to clear the other field when reading the state.

google/resource_google_project.go
```diff
@@ -198,30 +221,60 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
...
 		if err != nil {
-			return fmt.Errorf("Error updating project %q: %s", p.Name, err)
+			return fmt.Errorf("Error updating project %q: %s", project_name, err)
 		}
+		d.SetPartial("name")
+	}
+
+	// Project parent has changed
+	if d.HasChange("org_id") || d.HasChange("folder_id") {
+		if v, ok := d.GetOk("org_id"); ok {
```
rosbo: Consider extracting line 239-253 in a `getParentResourceId` (or something along these lines) and reuse it for update and create.

website/docs/r/google_project.html.markdown
```diff
@@ -62,8 +71,16 @@ The following arguments are supported:
     [below](#id-field) for more information about its behaviour.
 
 * `org_id` - (Optional) The numeric ID of the organization this project belongs to.
-    This is required if you are creating a new project.
-    Changing this forces a new project to be created.
+    Changing this forces a new project to be created.  If the `org_id`
+    is specified, but not the `folder_id`, then the project is created
+    at the top level. If both are specified, `org_id` is ignored.
```
rosbo: If you add the ConflictsWith I suggested earlier, you can update this comment to say that folder_id or org_id can be specified but not both.

google/resource_google_project.go
srivasta: Done

srivasta:
Made all the changes requested. I also squashed the commits in this branch, since intermediate points were not interesting, and will give a cleaner history on the mainline.

google/resource_google_project.go
srivasta: Done.

google/resource_google_project.go
srivasta: Done.

website/docs/r/google_project.html.markdown
srivasta: Done

google/resource_google_project_test.go
```diff
@@ -24,15 +24,42 @@ var (
 	originalPolicy *cloudresourcemanager.Policy
 )
 
+// Test that a Project resource can be created without an organization
+func TestAccGoogleProject_createWithoutOrg(t *testing.T) {
```
rosbo: We run our acceptance tests using a service account (not user credentials). Unfortunately, a service account belongs to an org and therefor, cannot create a project without a parent organization.

One solution: we could add a skip if a service account is used. This means we could run it locally but they wouldn't be ran by our CI server.

@paddycarver Any other solutions for this? Running user credentials for the CI server? Not sure it is safe/worth it.

google/resource_google_project_test.go
```diff
@@ -24,15 +24,42 @@ var (
...
+			resource.TestStep{
+				Config: testAccGoogleProject_createWithoutOrg(pid, pname),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGoogleProjectExists("google_project.acceptance", pid),
+				),
+			},
+		},
+	})
+}
+
```
rosbo: Now that you can create folders, you can add acceptance tests for assigning a project to a folder and moving a project from a folder to an org and vice versa. Since you will be on vacation, I can also take care of writing the new tests. LMK what you prefer.

srivasta: That sounds great. I won't have time for this till middle of next month, and so it would be great if you could add the tests with folder creation and movement between folders. I'll circle back in case you don't have time in the next couple of weeks

google/resource_google_project_test.go
rosbo: I added logic to skip the test if a service account is used. @paddycarver if you have a better suggestion, we can fix it in a separate PR.

google/resource_google_project_test.go
rosbo: Oki. I will write these tests today. Thanks!

google/resource_google_project_test.go
paddycarver: I'm late to the party here--sorry, I was out of office on Thursday--but IIRC we were talking about the possibility of getting the service account our CI runs with whitelisted for creating a project without a parent organisation, though I've lost track of where that discussion landed or what its current status is. That's my preferred solution, honestly, but we can follow up on this with a more permanent fix in a separate PR.

PR #439, opened 2017-09-15
Deal with undeletable bucket ACLs in storage.

paddycarver:
When GCS buckets are created, they're created with a set of default
ACLs:

* `OWNER:project-owners-{project_number}`
* `OWNER:project-editors-{project_number}`
* `READER:project-viewers-{project_number}`

Normally, this would be fine, or a minor inconvenience. Terraform could
either delete them itself, or the first apply of a user would overwrite
them.

However, trying to remove the `OWNER:project-owners-{project_number}`
ACL yields an API error that the bucket owner must maintain OWNER access
to the bucket. This breaks things like `terraform destroy`, but also
means any config without that line in it will fail to apply, not just
overwrite the value.

To make matters worse, trying to *add* the
`OWNER:project-owners-{project_number}` ACL to any bucket that already
has it _also_ yields the same error about not being able to remove it.

To get around this, the storage_bucket_acl resource has been updated to
largely ignore _just this_ ACL. It will not try to add it if it already
exists, will not try to remove it at all. This does mean that Terraform
is incapable of removing this ACL from a bucket, but I'm not sure it's
possible to do that with the API, anyways.

Tests were also updated to keep the default ACLs as part of the config,
and to change the email addresses to addresses we actually own. I tried
changing to non-existent hashicorp.com email addresses, but was
rejected; only email addresses that are backed by actual Google accounts
can be used, sadly.

This fixes the broken storage bucket ACL tests.

### DIFF
--- a/google/resource_storage_bucket_acl.go
+++ b/google/resource_storage_bucket_acl.go
@@ -3,6 +3,7 @@ package google
 import (
 	"fmt"
 	"log"
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
@@ -101,9 +102,26 @@ func resourceStorageBucketAclCreate(d *schema.ResourceData, meta interface{}) er
 
 	}
 	if len(role_entity) > 0 {
+		current, err := config.clientStorage.BucketAccessControls.List(bucket).Do()
+		if err != nil {
+			return fmt.Errorf("Error retrieving current ACLs: %s", err)
+		}
 		for _, v := range role_entity {
 			pair, err := getRoleEntityPair(v.(string))
-
+			if err != nil {
+				return err
+			}
+			var alreadyInserted bool
+			for _, cur := range current.Items {
+				if cur.Entity == pair.Entity && cur.Role == pair.Role {
+					alreadyInserted = true
+					break
+				}
+			}
+			if alreadyInserted {
+				log.Printf("[DEBUG]: pair %s-%s already exists, not trying to insert again\n", pair.Role, pair.Entity)
+				continue
+			}
 			bucketAccessControl := &storage.BucketAccessControl{
 				Role:   pair.Role,
 				Entity: pair.Entity,
@@ -172,8 +190,16 @@ func resourceStorageBucketAclUpdate(d *schema.ResourceData, meta interface{}) er
 	config := meta.(*Config)
 
 	bucket := d.Get("bucket").(string)
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
 
 	if d.HasChange("role_entity") {
+		p, err := config.clientResourceManager.Projects.Get(project).Do()
+		if err != nil {
+			return fmt.Errorf("Error retrieving project %q: %s", project, err)
+		}
 		o, n := d.GetChange("role_entity")
 		old_re, new_re := o.([]interface{}), n.([]interface{})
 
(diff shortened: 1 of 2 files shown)


paddycarver:
This is ready for review now, now that I've finished arguing with the test framework :( Test strategy isn't as clean as I'd like it to be, but them's the breaks.

google/resource_storage_bucket_acl.go
```diff
@@ -172,8 +190,16 @@ func resourceStorageBucketAclUpdate(d *schema.ResourceData, meta interface{}) er
 	config := meta.(*Config)
 
 	bucket := d.Get("bucket").(string)
+	project, err := getProject(d, config)
```
selmanj: technically this might not match the project that owns the bucket, right? If someone is managing a bucket in terraform that doesn't belong to their provider's configured project this might cause issues...

Fixing this might be too painful right now so if what I said is true, I'm fine with just acknowledging it with a TODO

google/resource_storage_bucket_acl.go
```diff
@@ -252,15 +278,30 @@ func resourceStorageBucketAclUpdate(d *schema.ResourceData, meta interface{}) er
 func resourceStorageBucketAclDelete(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
+	project, err := getProject(d, config)
```
selmanj: Same comment as above re: which project owns the bucket

google/resource_storage_bucket_acl.go
```diff
@@ -215,7 +237,11 @@ func resourceStorageBucketAclUpdate(d *schema.ResourceData, meta interface{}) er
 			}
 		}
 
-		for entity, _ := range old_re_map {
+		for entity, role := range old_re_map {
+			if entity == "project-owners-"+strconv.FormatInt(p.ProjectNumber, 10) && role == "OWNER" {
```
selmanj: Would it be easier to read to do `fmt.Sprintf("project-owners-%d", p.ProjectNumber)` instead of string concatenating `strconv.FormatInt`?

google/resource_storage_bucket_acl.go
```diff
@@ -252,15 +278,30 @@ func resourceStorageBucketAclUpdate(d *schema.ResourceData, meta interface{}) er
...
+	}
+
 	re_local := d.Get("role_entity").([]interface{})
 	for _, v := range re_local {
 		res, err := getRoleEntityPair(v.(string))
 		if err != nil {
 			return err
 		}
 
+		if res.Entity == "project-owners-"+strconv.FormatInt(p.ProjectNumber, 10) && res.Role == "OWNER" {
```
selmanj: Same as above re: string concat

google/resource_storage_bucket_acl.go
paddycarver: This is a good point, and is actually easier to implement; we can just read the bucket from the API, and pull the project number off it. Great point, and thanks for pointing it out. I like that solution _much_ better.

google/resource_storage_bucket_acl.go
paddycarver: Fair, I'll do that. Sorry, I'm just used to the non-`reflect`-based string construction methods from an old performance-sensitive project, and now they're reflex. Thanks for catching it :)

google/resource_storage_bucket_acl.go
selmanj: I didn't realize the project id is on the bucket resource; that makes this change a lot easier

paddycarver:
Updated, all tests pass, concerns were addressed.

PR #446, opened 2017-09-21
note ipv4 addresses can't be disabled on 2nd gen

ickymettle:
Adds a note that IPv4 addresses can't be disabled on 2nd generation cloudsql instances to save end user confusion.


rosbo:
Thanks for the addition!

PR #447, opened 2017-09-22
Assign IAM policy to folder.

### LINKED ISSUE #413: Configure IAM policy of an existing GCP folders.
Blocked by #412 

Implementation note: This looks very similar to what already exists for "google_project". However, since IAM policies can be set on many different resource types, it might make sense to revisit the current approach, maybe it can be generalized a bit so that it works for other resource types, too.

rosbo:
Fixes #413 

This overrides any existing permissions on the folder. If users have a valid use cases for managing some roles in terraform and some other outside of terraform, we can create a new resource `google_folder_iam_policy_binding` to support it. For now, I believe this will cover most use cases from our users.

### DIFF
--- a/google/resource_google_folder_iam_policy.go
+++ b/google/resource_google_folder_iam_policy.go
@@ -57,7 +57,7 @@ func resourceGoogleFolderIamPolicyRead(d *schema.ResourceData, meta interface{})
 	}
 
 	d.Set("etag", policy.Etag)
-	d.Set("policy_data", encodeV2IamPolicy(policy))
+	d.Set("policy_data", marshalV2IamPolicy(policy))
 
 	return nil
 }
@@ -92,7 +92,7 @@ func resourceGoogleFolderIamPolicyDelete(d *schema.ResourceData, meta interface{
 
 func setFolderIamPolicy(d *schema.ResourceData, config *Config) error {
 	folder := d.Get("folder").(string)
-	policy, err := decodeV2IamPolicy(d.Get("policy_data").(string))
+	policy, err := unmarshalV2IamPolicy(d.Get("policy_data").(string))
 	if err != nil {
 		return fmt.Errorf("'policy_data' is not valid for %s: %s", folder, err)
 	}
@@ -105,14 +105,14 @@ func setFolderIamPolicy(d *schema.ResourceData, config *Config) error {
 	return err
 }
 
-func encodeV2IamPolicy(policy *resourceManagerV2Beta1.Policy) string {
+func marshalV2IamPolicy(policy *resourceManagerV2Beta1.Policy) string {
 	pdBytes, _ := json.Marshal(&resourceManagerV2Beta1.Policy{
 		Bindings: policy.Bindings,
 	})
 	return string(pdBytes)
 }
 
-func decodeV2IamPolicy(policyData string) (*resourceManagerV2Beta1.Policy, error) {
+func unmarshalV2IamPolicy(policyData string) (*resourceManagerV2Beta1.Policy, error) {
 	policy := &resourceManagerV2Beta1.Policy{}
 	if err := json.Unmarshal([]byte(policyData), policy); err != nil {
 		return nil, fmt.Errorf("Could not unmarshal policy data %s:\n%s", policyData, err)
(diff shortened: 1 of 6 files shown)


google/resource_google_folder_iam_policy.go
```diff
@@ -0,0 +1,129 @@
...
+
+	_, err = config.clientResourceManagerV2Beta1.Folders.SetIamPolicy(folder, &resourceManagerV2Beta1.SetIamPolicyRequest{
+		Policy:     policy,
+		UpdateMask: "bindings",
+	}).Do()
+
+	return err
+}
+
+func encodeV2IamPolicy(policy *resourceManagerV2Beta1.Policy) string {
```
selmanj: Why not marhall/unmarshall instead of encode/decode?

google/resource_google_folder_iam_policy_test.go
```diff
@@ -0,0 +1,168 @@
...
+				},
+			},
+		},
+	}
+	policy2 := &resourceManagerV2Beta1.Policy{
+		Bindings: []*resourceManagerV2Beta1.Binding{
+			{
+				Role: "roles/editor",
+				Members: []string{
+					"user:admin@hashicorptest.com",
```
selmanj: Does GCP validate that the domain exists? If not I'd prefer to generate a random email address/domain name here to avoid cases where a test doesn't clean up and leaves permissions open

alternatively if we can set an overriding policy rule that ensures nobody has access...

google/resource_google_folder_iam_policy_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+		p, err := config.clientResourceManagerV2Beta1.Folders.GetIamPolicy(rs.Primary.ID, &resourceManagerV2Beta1.GetIamPolicyRequest{}).Do()
+		if err != nil {
+			return err
+		}
+
+		if !reflect.DeepEqual(p.Bindings, policy.Bindings) {
+			return fmt.Errorf("Incorrect iam policy bindings. Expected '%s', got '%s'", policy.Bindings, p.Bindings)
+		}
+
```
selmanj: Also needs a check to make sure etag was stored

website/docs/r/google_folder.html.markdown
```diff
@@ -1,7 +1,7 @@
 ---
 layout: "google"
 page_title: "Google: google_folder"
-sidebar_current: "docs-google-folder"
+sidebar_current: "docs-google-folder-x"
```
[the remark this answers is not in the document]
rosbo: It does a prefix match when doing highlighting of the current page you are visiting. If you don't put the `-x`, then, when visiting docs-google-folder-iam-policy it will highlight both links in the sidebar which is not desirable.

For instance, the aws provider has this issue for some of their resource:
https://www.terraform.io/docs/providers/aws/d/ami_ids.html

See how `aws_ami` and `aws_ami_ids` links are highlighted in the sidebar. Adding the -x prevents this issue.

google/resource_google_folder_iam_policy_test.go
rosbo: It does check if a user exists.

google/resource_google_folder_iam_policy.go
rosbo: Done

google/resource_google_folder_iam_policy_test.go
rosbo: Done

rosbo:
```sh
TF_ACC=1 go test ./google -v -run TestAccGoogleFolderIamPolicy_ -timeout 120m
=== RUN   TestAccGoogleFolderIamPolicy_basic
--- PASS: TestAccGoogleFolderIamPolicy_basic (16.08s)
=== RUN   TestAccGoogleFolderIamPolicy_update
--- PASS: TestAccGoogleFolderIamPolicy_update (18.30s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	34.522s
```

website/docs/r/google_folder.html.markdown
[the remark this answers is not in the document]
selmanj: I'll remember this, thanks!

google/resource_google_folder_iam_policy_test.go
selmanj: Bummer. For now hashicorptest.com works but at some point we should have a review of these sorts of tests

Issue #455, opened 2017-09-25
Pubsub Subscription should allow updating push config


rosbo:
Remove `ForceNew`, implement an update method leveraging this APIs:
* https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/modifyPushConfig

rosbo:
```sh
TF_ACC=1 go test ./google -v -run TestAccPubsubSubscription_ -timeout 120m
=== RUN   TestAccPubsubSubscription_import
--- PASS: TestAccPubsubSubscription_import (5.86s)
=== RUN   TestAccPubsubSubscription_basic
--- PASS: TestAccPubsubSubscription_basic (6.17s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	12.159s
```

rosbo:
ACK deadline at the subscription level cannot be modified. The `modifyAckDeadline` method allows to update ack deadline only for specific messages. See:
https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/modifyAckDeadline

Updated the issue title and description to reflect that only the push config can be updated.

Issue #393, opened 2017-09-06
google_pubsub_subscription read function does nothing


drzero42:
While working on #386 I discovered that resourcePubsubSubscriptionRead() does not actually do anything. It sends a Get call, but throws the result away instead of updating the ResourceData, which doesn't really matter that much if you only change the resource through Terraform, but if the remote resource has changed, Terraform has no way of discovering this and offering to change it back.

resourcePubsubSubscriptionRead() should be changed to actually set the various data attributes based on the data returned by the Get request.

PR #456, opened 2017-09-25
Pubsub subscription read state from API and import support

### LINKED ISSUE #386: Make pubsub topics and subscriptions importable

rosbo:
Fixes #386, #393 

* Fixed the Read method which was doing nothing.
* Added import support
* `topic` field support either the `path` or the `name`.
* `push_configs` cannot have more than one item. Express this in the schema definition to bring this validation at `terraform plan` time.
* `push_configs.push_endpoint` is required. Update the schema accordingly.
* Renamed acceptance test to follow our naming convention
* Tested manually `push_configs` (see below why this is not automated).

Next steps: I filled an issue (#455) to add update support for `push_configs` and `ack_deadline`.

### DIFF
--- a/google/resource_pubsub_subscription.go
+++ b/google/resource_pubsub_subscription.go
@@ -139,7 +139,7 @@ func resourcePubsubSubscriptionRead(d *schema.ResourceData, meta interface{}) er
 	d.Set("topic", subscription.Topic)
 	d.Set("ack_deadline_seconds", subscription.AckDeadlineSeconds)
 	d.Set("path", subscription.Name)
-	d.Set("push_config", flattenPushConfigs(subscription.PushConfig))
+	d.Set("push_config", flattenPushConfig(subscription.PushConfig))
 
 	return nil
 }
@@ -172,7 +172,7 @@ func resourcePubsubSubscriptionStateImporter(d *schema.ResourceData, meta interf
 	return []*schema.ResourceData{d}, nil
 }
 
-func flattenPushConfigs(pushConfig *pubsub.PushConfig) []map[string]interface{} {
+func flattenPushConfig(pushConfig *pubsub.PushConfig) []map[string]interface{} {
 	configs := make([]map[string]interface{}, 0, 1)
 
 	if pushConfig == nil || len(pushConfig.PushEndpoint) == 0 {
--- a/google/resource_pubsub_subscription.go
+++ b/google/resource_pubsub_subscription.go
@@ -13,6 +13,10 @@ func resourcePubsubSubscription() *schema.Resource {
 		Read:   resourcePubsubSubscriptionRead,
 		Delete: resourcePubsubSubscriptionDelete,
 
+		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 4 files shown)


google/resource_pubsub_subscription_test.go
```diff
@@ -28,6 +30,17 @@ func TestAccPubsubSubscriptionCreate(t *testing.T) {
...
+//
+// Testing push endpoints is tricky for the following reason:
+// - You need a publicly accessible HTTPS server to handle POST requests in order to receive push messages.
+// - The server must present a valid SSL certificate signed by a certificate authority
+// - The server must be routable by DNS.
+// - You also need to validate that you own the domain (or have equivalent access to the endpoint).
+// - Finally, you must register the endpoint domain with the GCP project.
+//
+// An easy way to test this would be to create an App Engine Hello World app. With AppEngine, SSL certificate, DNS and domain registry is handled for us.
+// App Engine is not yet supported by Terraform but once it is, it will provide an easy path to testing push configs.
```
selmanj: Cloud functions is an alternative here that might be quicker to deploy

google/resource_pubsub_subscription.go
```diff
@@ -21,14 +25,16 @@ func resourcePubsubSubscription() *schema.Resource {
...
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: compareSelfLinkOrResourceName,
 			},
 
 			"ack_deadline_seconds": &schema.Schema{
```
selmanj: There's an opportunity to specify the default here instead of in the create codepath below (I see below it's 10; does that match the GCP api?).

google/resource_pubsub_subscription_test.go
rosbo: Good point. Are you still working on adding support for those in Terraform?

selmanj: Not actively; don't wait on me

google/resource_pubsub_subscription.go
rosbo: I prefer to leave it as `Computed` right now to avoid creating diffs for our users (if for instance they updated this field from the UI). A diff right now forces the creation of a new resources. We don't support update. I will fix this once we add update support in #455. I added a comment about this in #455

google/resource_pubsub_subscription_test.go
rosbo: I added a note about Cloud Functions being an option too.

google/resource_pubsub_subscription.go
selmanj: Ah, yeah, that sounds like a much better experience. 👍

PR #457, opened 2017-09-25
Add resource_logging_billing_account_sink resource

selmanj:
This adds support for `resource_logging_billing_account_sink`; the API looks identical to the one used for project-level sinks, so I opted to do some refactoring here to establish a common schema (there is also organizations and folder-level sinks coming soon).

### DIFF
--- a/google/resource_logging_billing_account_sink.go
+++ b/google/resource_logging_billing_account_sink.go
@@ -52,7 +52,7 @@ func resourceLoggingBillingAccountSinkRead(d *schema.ResourceData, meta interfac
 func resourceLoggingBillingAccountSinkUpdate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	sink := expandResourceloggingSinkForUpdate(d)
+	sink := expandResourceLoggingSinkForUpdate(d)
 
 	// The API will reject any requests that don't explicitly set 'uniqueWriterIdentity' to true.
 	_, err := config.clientLogging.BillingAccounts.Sinks.Patch(d.Id(), sink).UniqueWriterIdentity(true).Do()
--- a/google/resource_logging_billing_account_sink.go
+++ b/google/resource_logging_billing_account_sink.go
@@ -1,0 +1,75 @@
+package google
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceLoggingBillingAccountSink() *schema.Resource {
+	schm := &schema.Resource{
+		Create: resourceLoggingBillingAccountSinkCreate,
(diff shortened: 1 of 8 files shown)


google/resource_logging_billing_account_sink.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceLoggingBillingAccountSink() *schema.Resource {
+	schm := &schema.Resource{
+		Create: resourceLoggingBillingAccountSinkCreate,
+		Read:   resourceLoggingBillingAccountSinkRead,
+		Delete: resourceLoggingBillingAccountSinkDelete,
+		Update: resourceLoggingBillingAccountSinkUpdate,
+		Schema: resourceLoggingSinkSchema(),
```
rosbo: Import support? You have all the context you need fresh in your mind. Since this PR is long, I think it is a good idea to do it in a separate PR.

website/docs/r/logging_billing_account_sink.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the logging sink.
+
+* 'billing_account' - (Required) The billing account exported to the sink.
```
rosbo: use the same quote style: `` not `'`

google/resource_logging_sink.go
```diff
@@ -0,0 +1,70 @@
...
+
+func flattenResourceLoggingSink(d *schema.ResourceData, sink *logging.LogSink) {
+	d.Set("name", sink.Name)
+	d.Set("destination", sink.Destination)
+	d.Set("filter", sink.Filter)
+	d.Set("writer_identity", sink.WriterIdentity)
+}
+
+func expandResourceloggingSinkForUpdate(d *schema.ResourceData) *logging.LogSink {
+	// Can only update destination/filter right now. Despite the method below using 'Patch', the API requires both
```
rosbo: Capital `L` for logging.

google/resource_logging_billing_account_sink.go
selmanj: Agreed! I actually started adding that but stopped as it got pretty complicated. My approach was modeled on yours

website/docs/r/logging_billing_account_sink.html.markdown
selmanj: oops, fixed

google/resource_logging_sink.go
selmanj: oops, fixed

Issue #380, opened 2017-09-01
[Feature] Prevent instance destroy on machine_type change


TraGicCode:
### Terraform Version
 v0.10.2

### Plugin Version
google (0.1.3)

### Affected Resource(s)
Please list the resources as a list, for example:
- google_compute_instance

When changing the machine_type of an already created instance terraform thinks it should recreate the instance when a modified should be done.  Also google requires the machine to be stopped before you can change the type.  Not sure how this affects things.

rosbo:
You are listing Oracle Public Cloud resources as the affected resources but mentions google in your comment. Did you mean `google_compute_instance` in the affected resources section?

TraGicCode:
Yes I did sorry I didn't replace that from the pr template

paddycarver:
Looks like this is similar to hashicorp/terraform#11998. I don't see why we couldn't apply the similar logic/workflow here. I'll open a quick PR.

rosbo:
Duplicate. Closing in favor of #728 which contains more details and implementation ideas.

PR #463, opened 2017-09-26
Update example resources in resource template

danawillow:
It feels weird to have opc as an example if the issue is something in the google provider


rosbo:
Good. It will prevent issues like here #380 👍

PR #470, opened 2017-09-27
Add google_logging_folder_sink resource

selmanj:
Also snuck in a minor page title fix relating to `google_logging_billing_account_sink` sorry

### DIFF
--- a/google/resource_logging_folder_sink.go
+++ b/google/resource_logging_folder_sink.go
@@ -33,11 +33,7 @@ func resourceLoggingFolderSink() *schema.Resource {
 func resourceLoggingFolderSinkCreate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	folder := d.Get("folder").(string)
-	if strings.HasPrefix(folder, "folders/") {
-		folder = folder[len("folders/"):]
-	}
-
+	folder := parseFolderId(d.Get("folder"))
 	id, sink := expandResourceLoggingSink(d, "folders", folder)
 	sink.IncludeChildren = d.Get("include_children").(bool)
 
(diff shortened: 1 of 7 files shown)


google/resource_logging_folder_sink_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+		includeChildren := false
+		if attributes["include_children"] != "" {
+			includeChildren, err = strconv.ParseBool(attributes["include_children"])
+			if err != nil {
+				return err
+			}
+		}
+		if sink.IncludeChildren != includeChildren {
+			return fmt.Errorf("mismatch on include_children: api has %s but client has %s", sink.IncludeChildren, attributes["include_children"])
```
rosbo: `go vet` fails because you are coercing a bool to a string. Use `%t`

google/resource_logging_folder_sink.go
```diff
@@ -0,0 +1,95 @@
...
+		Default:  false,
+	}
+
+	return schm
+}
+
+func resourceLoggingFolderSinkCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	folder := d.Get("folder").(string)
```
rosbo: You can use `parseFolderId`

selmanj:
Haha whoops yeah that name was bad, sorry cold medicine is making me terse

google/resource_logging_folder_sink_test.go
selmanj: Fixed, went with %v (and the correct variable this time)

google/resource_logging_folder_sink.go
selmanj: Done

Issue #444, opened 2017-09-21
Feature Request: Support for Cloud Identity Aware Proxy in Compute Engine


migibert:
Hi there,

We would love to see support for Cloud Identity Aware Proxy (https://cloud.google.com/iap/) as it is now available in GA.

Relevant links:
* https://cloud.google.com/iap/docs/reference/compute-engine-apis

Cidan:
I'll go ahead and take a look at this in the next week or two.

homelessnessbo:
Hey guys, as I can see - PR is already merged, right? Is it going to be included in the nearest release, or should I build it from master?

rosbo:
Hi @homelessnessbo,

Yes, this will be included in the next release. I also realized that we had forgot to update our docs so I updated them.

Thank you

Cidan:
@rosbo It's actually already released iirc, no?

rosbo:
@Cidan you are right. This was release in version 1.2.0. However, the documentation was not updated. You can look at #839 for the documentation on how to use it.

homelessnessbo:
Thanks guys!

b4b4r07:
Thank you for your development. This feature is what I want! BTW, can we manage ACL about IAP via terraform? As far as I see docs, it seems that it cannot be manage. https://www.terraform.io/docs/providers/google/r/compute_backend_service.html

Cidan:
@b4b4r07 Can you open a new issue to add ACL support? Thanks!

b4b4r07:
OK, I'll do later. Thanks

nightmare-moon:
this this feature acl and iap support in the product or not... 
I'm seeing conflicting information. 

I need this feature...

PR #471, opened 2017-09-27
Add IAP support for backend services

Cidan:
WIP PR, do not merge yet

Adds IAP support for #444


Cidan:
I think this is more or less good to go, though I might be missing something. @danawillow, whatcha think?

Cidan:
sorry, one last change, wrote a reciprocal function for flatten to keep in line with previous work. now it should be good!

google/resource_compute_backend_service.go
```diff
@@ -37,6 +37,32 @@ func resourceComputeBackendService() *schema.Resource {
 				MaxItems: 1,
 			},
 
+			"iap": &schema.Schema{
+				Type: schema.TypeSet,
```
danawillow: Since it's MaxItems: 1 anyway, I'd recommend just making this a list. Then you can also get rid of the Set fn on line 63

google/resource_compute_backend_service.go
```diff
@@ -37,6 +37,32 @@ func resourceComputeBackendService() *schema.Resource {
...
+				Elem: &schema.Resource{
```
danawillow: Most of our other resources have Elem as the last item in the struct. I'm not sure why this resource does it the other way but it would be awesome if you could move this one and change some of the others in this file too!

google/resource_compute_backend_service.go
```diff
@@ -37,6 +37,32 @@ func resourceComputeBackendService() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"enabled": &schema.Schema{
+							Type:     schema.TypeBool,
+							Optional: true,
```
danawillow: I wonder whether we should make this Required, that way people don't accidentally try to enable it by putting in the block but forgetting this field

google/resource_compute_backend_service.go
```diff
@@ -261,6 +288,27 @@ func resourceComputeBackendServiceDelete(d *schema.ResourceData, meta interface{
 	return nil
 }
 
+func expandIap(configured []interface{}) *compute.BackendServiceIAP {
+	data := configured[0].(map[string]interface{})
+	iap := &compute.BackendServiceIAP{
+		Enabled:            data["enabled"].(bool),
```
danawillow: Without also setting `ForceSendFields`, these three fields won't appear in the request if they're set to empty (or false). It might still work since the outer block will still be there, but if you could check that you can disable iap using this code that would be great.

google/resource_compute_backend_service.go
```diff
@@ -203,6 +229,7 @@ func resourceComputeBackendServiceRead(d *schema.ResourceData, meta interface{})
 	d.Set("self_link", service.SelfLink)
 	d.Set("backend", flattenBackends(service.Backends))
 	d.Set("connection_draining_timeout_sec", service.ConnectionDraining.DrainingTimeoutSec)
+	d.Set("iap", flattenIap(service.Iap))
```
danawillow: will service.Iap ever be nil?

google/resource_compute_backend_service.go
Cidan: I considered this as well. I think about how I work, and commenting out IAP enabled would "disable" it, without having to edit the line and set it as false. What's the standard here?

google/resource_compute_backend_service.go
Cidan: Only in-so-far as the service object doesn't initialize it a few lines up from here. I'll add a check in the flatten function.

google/resource_compute_backend_service.go
Cidan: I will double check and also add tests.

google/resource_compute_backend_service.go
Cidan: Something else to consider here, should we even have the enabled flag? Should it be implicit by providing the keys?

Cidan:
Alright, had to make quite a few changes here. I'll try to explain my reasoning.

There seems to be a bug with the IAP API it self where if you disable IAP and attempt to zero out the secret (ie, remove the entire stanza), only the Enable -> false flag is set and the client secret is not cleared. As a result, the next time an apply is called, the state doesn't match and the apply will happen a second time, correctly clearing the key.

So, what I did was removed the notion of enable as an option. When IAP is added to a backend, you must provide a key and secret to enable it. Dropping the stanza for IAP disables it -- this is somewhat cleaner since the API it self requires the id and secret when enabling anyway. I've also made the service send up a default empty IAP object via ForceSendFields when the IAP stanza is missing, thus disabling the service if the previous state was enabled.

The IAP API never returns the secret once it's send, only its' SHA256 sum, so I've added a function for comparing the SHA256 sum we get from the API vs the sum of the locally defined key.

Tests still to come.

Cidan:
Tests written and pass acceptance tests. @danawillow, how does it look now?

google/resource_compute_backend_service.go
```diff
@@ -261,6 +288,33 @@ func resourceComputeBackendServiceDelete(d *schema.ResourceData, meta interface{
...
+		Oauth2ClientSecret: data["oauth2_client_secret"].(string),
+		ForceSendFields:    []string{"Enabled", "Oauth2ClientId", "Oauth2ClientSecret"},
+	}
+
+	return iap
+}
+
+func flattenIap(iap *compute.BackendServiceIAP) []map[string]interface{} {
+	if iap == nil {
+		return make([]map[string]interface{}, 1, 1)
```
danawillow: just curious, why 1, 1?

google/resource_compute_backend_service.go
```diff
@@ -261,6 +288,33 @@ func resourceComputeBackendServiceDelete(d *schema.ResourceData, meta interface{
...
+	}
+
+	result := make([]map[string]interface{}, 0, 1)
+	iapMap := map[string]interface{}{
+		"enabled":              iap.Enabled,
+		"oauth2_client_id":     iap.Oauth2ClientId,
+		"oauth2_client_secret": iap.Oauth2ClientSecretSha256,
+	}
+	result = append(result, iapMap)
```
danawillow: Totally up to you, but a shorter way to do this would be:
```
iapMap := map[string]interface{}{
  // properties
}
return []map[string]interface{}{iapMap}
```

google/resource_compute_backend_service.go
```diff
@@ -326,6 +379,13 @@ func expandBackendService(d *schema.ResourceData) compute.BackendService {
 	service := compute.BackendService{
 		Name:         d.Get("name").(string),
 		HealthChecks: healthChecks,
+		Iap: &compute.BackendServiceIAP{
+			ForceSendFields: []string{"Enabled", "Oauth2ClientId", "Oauth2ClientSecret"},
```
danawillow: Can you add a comment as to why `Iap` has to be set (and why the three fields are force)?

google/resource_compute_backend_service_test.go
```diff
@@ -114,6 +114,47 @@ func TestAccComputeBackendService_withBackendAndUpdate(t *testing.T) {
...
+					serviceName, igName, itName, checkName, 10),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeBackendServiceExists(
+						"google_compute_backend_service.lipsum", &svc),
+				),
+			},
+		},
+	})
+
+	if svc.TimeoutSec != 10 {
```
danawillow: Eh, I don't necessarily mind having these checks here but they're not _really_ what the test is testing.

google/resource_compute_backend_service_test.go
```diff
@@ -273,6 +314,38 @@ func testAccCheckComputeBackendServiceExists(n string, svc *compute.BackendServi
 	}
 }
 
+func testAccCheckComputeBackendServiceExistsWithIAP(n string, svc *compute.BackendService) resource.TestCheckFunc {
```
danawillow: Instead of a full testcheck here you could just use the existing one and then do checks on the returned svc like you're already doing.

google/resource_compute_backend_service_test.go
```diff
@@ -438,6 +511,60 @@ resource "google_compute_http_health_check" "default" {
...
+	return fmt.Sprintf(`
+resource "google_compute_backend_service" "lipsum" {
+  name        = "%s"
+  description = "Hello World 1234"
+  port_name   = "http"
+  protocol    = "HTTP"
+  timeout_sec = %v
+
+  backend {
+    group = "${google_compute_instance_group_manager.foobar.instance_group}"
```
danawillow: can you fix the formatting here? I think it's a tabs vs spaces issue

google/resource_compute_backend_service.go
Cidan: I hadn't considered that -- I agree, much cleaner.

google/resource_compute_backend_service.go
Cidan: Because this is called directly on the ResourceData object via .Set, which will always have at most one IAP object. I'm not sure why the IAP API has this as a list -- perhaps future expansion of IAP options and credentials?

google/resource_compute_backend_service_test.go
Cidan: Other tests do this too, even though it's not relevant (and I agree, I found it odd), so I decided to leave in for this test as well.

google/resource_compute_backend_service.go
Cidan: Done!

google/resource_compute_backend_service_test.go
Cidan: Done!

google/resource_compute_backend_service_test.go
Cidan: I actually prefer the testcheck personally versus the svc object test, as it feels more complete -- I actually would rather rip the svc checks out since they seem like they aren't really needed.

Cidan:
@danawillow That's interesting on the test failure -- I can't reproduce it. Any guidance on your setup?

```
TF_ACC=1 go test ./google -v -run=TestAccComputeBackendService_withBackendAndIAP -timeout 120m
=== RUN   TestAccComputeBackendService_withBackendAndIAP
--- PASS: TestAccComputeBackendService_withBackendAndIAP (128.16s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	128.190s
```

google/resource_compute_backend_service.go
danawillow: Sorry, my question wasn't "why is this a list?", but "why is it being initialized with 1, 1 instead of 0, 1?"
0, 1 will have the list not have anything in it, whereas 1, 1 will have the list have an empty map in it (see https://play.golang.org/p/SFCt6l5Fsw). The `flatten` functions for all other resources are currently using 0, 1.

google/resource_compute_backend_service_test.go
```diff
@@ -120,6 +120,47 @@ func TestAccComputeBackendService_withBackendAndUpdate(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeBackendServiceExistsWithIAP(
+						"google_compute_backend_service.lipsum", &svc),
+				),
+			},
+			resource.TestStep{
+				Config: testAccComputeBackendService_withBackend(
+					serviceName, igName, itName, checkName, 10),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeBackendServiceExists(
```
danawillow: This checks that the backend service exists, but it doesn't check that IAP actually got disabled (I'd highly recommend just adding that check in below with the other checks on `svc`)

google/resource_compute_backend_service_test.go
danawillow: In what way is it more complete? Both ways would do the exact same thing, except one is duplicating a lot of code between functions and the other is calling an existing function.

google/resource_compute_backend_service.go
Cidan: Ah, my apologies, I misread your question. Will adjust, :)

google/resource_compute_backend_service_test.go
Cidan: I see your point. Will adjust.

Cidan: Actually, I take it back. I started to implement the svc object check when I remembered why I don't like it, particularly for multi-stage tests. The svc object is ambiguous in both timing, and in scope. When is the svc object actually set? After destruction? Before? What about when you have multiple stages, which stage are we checking, IAP being set, or not? The answer may seem obvious, and it is (it's after all tests run), but it's not explicit and self contained. What we end up with is two different ways to validate a test run, and picking which way to test now follows a branching pattern (single stage? test svc. multi stage? test all but last stage with svc, and other stages with a check function).

It makes more sense to me to follow a single pattern -- check each stage with a function and be explicit about intent.

Cidan:
@danawillow Alright, let's try this one more time!

Made the recommended changes, and testing now checks for IAP enabled and disabled properly.

danawillow:
Thanks @Cidan!

PR #204, opened 2017-07-16
[WIP] Add support for google_service_account_key

michaelahern:
Adds support for creating and deleting service account keys. Fixes #116 and https://github.com/hashicorp/terraform/issues/15240.

Still a work in progress -- need to finish up acceptance tests and add docs. In the meantime, wanted to get some guidance on an acceptance test issue. When running a test to create a service account with an associated service account key, it fails with the error below. The service account key was successfully created though, but Google is returning an HTTP 404 for the service account key resource when refreshing state inside the test. It seems like the API takes at least a few seconds for get requests to succeed against the key resource. What is the best practice to handle these this kind of waiting for resource to be created or accessible situations?

```
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccGoogleServiceAccountKey_basic -timeout 120m
?   	github.com/terraform-providers/terraform-provider-google	[no test files]
=== RUN   TestAccGoogleServiceAccountKey_basic
--- FAIL: TestAccGoogleServiceAccountKey_basic (5.45s)
	testing.go:428: Step 0 error: After applying this step and refreshing, the plan was not empty:
		
		DIFF:
		
		CREATE: google_service_account_key.acceptance
		  key_algorithm:           "" => "KEY_ALG_RSA_2048" (forces new resource)
		  name:                    "" => "<computed>"
		  private_key:             "<sensitive>" => "<sensitive>" (attribute changed)
		  private_key_encrypted:   "" => "<computed>"
		  private_key_fingerprint: "" => "<computed>"
		  private_key_type:        "" => "TYPE_GOOGLE_CREDENTIALS_FILE" (forces new resource)
		  service_account_id:      "" => "projects/terraform-test/serviceAccounts/ae4h8z98l0a@terraform-test.iam.gserviceaccount.com" (forces new resource)
		  valid_after:             "" => "<computed>"
		  valid_before:            "" => "<computed>"
		
		STATE:
		
		google_service_account.acceptance:
		  ID = projects/terraform-test/serviceAccounts/ae4h8z98l0a@terraform-test.iam.gserviceaccount.com
		  account_id = ae4h8z98l0a
		  display_name = Terraform Test
		  email = ae4h8z98l0a@terraform-test.iam.gserviceaccount.com
		  name = projects/terraform-test/serviceAccounts/ae4h8z98l0a@terraform-test.iam.gserviceaccount.com
		  unique_id = 108219597091612926163
FAIL
[...]

PR #472, opened 2017-09-27
Add support for google_service_account_key

sebglon:
to correct PR #204


google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		d.Set("private_key_encrypted", encrypted)
+		d.Set("private_key_fingerprint", fingerprint)
+	} else {
+		d.Set("private_key", sak.PrivateKeyData)
+	}
+
+	d.SetId(sak.Name)
+	d.Set("name", sak.Name)
```
danawillow: Instead of setting these values on create, we usually call the read function at the end of the create function. Is it possible for you to do that here? Alternatively, if your goal was to avoid an extra API call you could add a helper function that could take the ServiceAccountKey object and do the d.Set calls

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,143 @@
...
+		Create: resourceGoogleServiceAccountKeyCreate,
+		Read:   resourceGoogleServiceAccountKeyRead,
+		Delete: resourceGoogleServiceAccountKeyDelete,
+		Schema: map[string]*schema.Schema{
+			"service_account_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"name": &schema.Schema{
```
danawillow: We're trying to get to a more clear way of ordering fields in the schema, by doing Requireds, then Optionals, then Computeds. Mind doing that here too?

google/resource_google_service_account_key_test.go
```diff
@@ -0,0 +1,56 @@
...
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No ID is set")
+		}
+
+		return nil
+	}
+}
+
+func testAccGoogleServiceAccountKey(account, name string) string {
+	t := `resource "google_service_account" "acceptance" {
```
danawillow: for consistency with other resources, can you just start this with the fmt.Sprintf instead of making it a variable?

google/resource_google_service_account_key_test.go
```diff
@@ -0,0 +1,56 @@
...
+	account_id = "%v"
+	display_name = "%v"
+}
+
+resource "google_service_account_key" "acceptance" {
+	service_account_id = "${google_service_account.acceptance.id}"
```
danawillow: do you also want to test some of the other fields? There's some logic regarding pgp keys in your create function so at a minimum I'd like to see that tested, but ideally the other fields as well

google/resource_google_service_account_key_test.go
```diff
@@ -0,0 +1,56 @@
...
+	accountId := "a" + acctest.RandString(10)
+	displayName := "Terraform Test"
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccGoogleServiceAccountKey(accountId, displayName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGoogleServiceAccountKeyExists("google_service_account_key.acceptance"),
```
danawillow: You can check values with fields with `resource.TestCheckResourceAttr` to ensure they have the values you want instead of just that the resource exists

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,143 @@
...
+	if v, ok := d.GetOk("private_key_type"); ok {
+		r.PrivateKeyType = v.(string)
+	}
+
+	sak, err := config.clientIAM.Projects.ServiceAccounts.Keys.Create(serviceAccount, r).Do()
+	if err != nil {
+		return fmt.Errorf("Error creating service account key: %s", err)
+	}
+
+	if v, ok := d.GetOk("pgp_key"); ok {
```
danawillow: I'm a bit confused about this part, since it's just reading a value from the config, encrypting it, and then storing it back in state. I'm not totally against it, but it does make me cautious since it's not necessarily matching the API. What's the use case?

sebglon: Use case is to get GPG_key on output for using it with another tools. If it is possible to have them on outpup without storing it, i'm interesting.

sebglon:
@danawillow I have modified all comment except Adding test on all fields.
Is it possible to have field output without storing them? If it's possible, i can remove all added vendor.

sebglon:
A few time, google create key return empty without error.
sho GoogleCloudPlatform/google-cloud-go#787

sebglon:
I have open on issue on Google GoogleCloudPlatform/google-cloud-go

sebglon:
I have added a workaroud for waiting public Key.
Is ti possible to Validate doc?

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,175 @@
...
+
+	if pubKey == "" {
+
+		if v, ok := d.GetOk("private_key_type"); ok {
+			r.PrivateKeyType = v.(string)
+		}
+
+		sak, err := config.clientIAM.Projects.ServiceAccounts.Keys.Create(serviceAccount, r).Do()
+		if err != nil || sak == nil {
+			return fmt.Errorf("Error creating service account key: %s", err)
```
danawillow: it might be nice to differentiate in the message the case where it returned an error vs returning nothing

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,175 @@
...
+		Read:   resourceGoogleServiceAccountKeyRead,
+		Delete: resourceGoogleServiceAccountKeyDelete,
+		Schema: map[string]*schema.Schema{
+			// Required
+			"service_account_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"public_key_type": &schema.Schema{
```
danawillow: Looks like this value has a reasonable default: https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts.keys/get. Any reason why it's `Required` vs `private_key_type` and `key_algorithm` being Optional?

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,175 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
```
danawillow: For enum types, would you mind adding a ValidateFunc that checks the values?
```
ValidateFunc: validation.StringInSlice([]string{"TYPE_NONE", "TYPE_X509_PEM_FILE", "TYPE_RAW_PUBLIC_KEY"}, false),
```

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,175 @@
...
+			"public_key": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
+				ForceNew: true,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
```
danawillow: This is marked `Optional`, but it doesn't look like setting it will have any effect.

google/resource_google_service_account_key_test.go
```diff
@@ -0,0 +1,60 @@
...
+					resource.TestCheckResourceAttrSet(resourceName, "public_key"),
+					resource.TestCheckResourceAttrSet(resourceName, "valid_after"),
+					resource.TestCheckResourceAttrSet(resourceName, "valid_before"),
+				),
+			},
+		},
+	})
+}
+
+func testAccCheckGoogleServiceAccountKeyExists(r string) resource.TestCheckFunc {
```
danawillow: usually our exists tests also read from the API to confirm the resource exists remotely. If there's a good reason why that shouldn't happen in this test, mind writing a comment as to why?

google/resource_google_service_account_key_test.go
```diff
@@ -0,0 +1,60 @@
...
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No ID is set")
+		}
+
+		return nil
+	}
+}
+
+func testAccGoogleServiceAccountKey(account, name string) string {
+	return fmt.Sprintf(`resource "google_service_account" "acceptance" {
```
danawillow: for consistency with the other resources' tests, can you move everything after `fmt.Sprintf(` to the next line? And then at the end, everything starting at the ` can also go on the next line.
(also be careful with formatting)

google/service_account_waiter.go
```diff
@@ -0,0 +1,61 @@
+package google
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"google.golang.org/api/googleapi"
+	iam "google.golang.org/api/iam/v1"
```
danawillow: why does this import need to be named?

google/resource_google_service_account_key_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No ID is set")
+		}
+
+		return nil
+	}
+}
+
+func testAccGoogleServiceAccountKey(account, name string) string {
```
danawillow: It looks like a fair number of the fields in the schema are untested. Mind making sure there's coverage of the rest of the fields?

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+layout: "google"
+page_title: "Google: google_service_accout_key"
+sidebar_current: "docs-google-service-account-key"
+description: |-
+  Allows management of a Google Cloud Platform service account Key Pair
+---
+
+# google\_service\_account\_key
+
+Allows management of a key, and must be created or imported for use with
```
danawillow: It would be nice to have a bit more text on what a key is. Maybe something like:
```
Creates and manages service account key-pairs, which allow the user to establish identity of a service account outside of GCP. For more information, see [the official documentation](https://cloud.google.com/iam/docs/creating-managing-service-account-keys) and [API](https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts.keys).
```

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+  public_key_type = "TYPE_X509_PEM_FILE"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+* `name` - The name used for this key pair (not used on create)
+
+* `service_account_id` - (Required) The Serice account id of the Key Pair.
```
danawillow: typo: service

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `key_algorithm` - (Optional) The output format of the private key. GOOGLE_CREDENTIALS_FILE is the default output format. Valid values [ServiceAccountPrivateKeyType](https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts.keys#ServiceAccountPrivateKeyType) (only used on create)
```
danawillow: just to make a complete sentence, how about `Valid values are listed at [ServiceAccountPrivateKeyType](...)`

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,175 @@
...
+				ForceNew: true,
+			},
+
+			"pgp_key": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+			"private_key": &schema.Schema{
```
danawillow: I think this should be down in the `Computed` section

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+when creating a new key, and when no `pgp_key` is provided
+* `encrypted_private_key` – the private key material, base 64 encoded and
+encrypted with the given `pgp_key`. This is only populated when creating a new
+key and `pgp_key` is supplied
+* `encrypted_fingerprint` - The MD5 public key fingerprint for the encrypted
+private key
+
+## Import
+
+Lightsail Key Pairs cannot be imported, because the private and public key are
```
danawillow: what's lightsail?

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+}
+
+resource "google_service_account_key" "acceptance" {
+  service_account_id = "${google_service_account.acceptance.id}"
+  pgp_key = "keybase:keybaseusername"
+  public_key_type = "TYPE_X509_PEM_FILE"
+}
+```
+
+## Argument Reference
```
danawillow: missing docs for `public_key_type`

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `pgp_key` – (Optional) An optional PGP key to encrypt the resulting private
+key material. Only used when creating or importing a new key pair
+
+~> **NOTE:** a PGP key is not required, however it is strongly encouraged.
+Without a PGP key, the private key material will be stored in state unencrypted.
+
+## Attributes Reference
+
+The following attributes are exported in addition to the arguments listed above:
```
danawillow: also valid_after and valid_before

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `fingerprint` - The MD5 public key fingerprint as specified in section 4 of RFC 4716.
+* `public_key` - the public key, base64 encoded
+* `private_key` - the private key, base64 encoded. This is only populated
+when creating a new key, and when no `pgp_key` is provided
+* `encrypted_private_key` – the private key material, base 64 encoded and
```
danawillow: In the schema these are called `private_key_encrypted` and `private_key_fingerprint`

website/docs/r/google_service_account_key.html.markdown
sebglon: Ok

website/docs/r/google_service_account_key.html.markdown
sebglon: yes

google/resource_google_service_account_key_test.go
sebglon: added

google/resource_google_service_account_key_test.go
sebglon: added

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,169 @@
...
+		KeyAlgorithm: d.Get("key_algorithm").(string),
+	}
+
+	var err error
+
+	if _, ok := d.GetOk("public_key"); !ok {
+
+		r.PrivateKeyType = d.Get("private_key_type").(string)
+
+		sak, err := config.clientIAM.Projects.ServiceAccounts.Keys.Create(serviceAccount, r).Do()
```
danawillow: I'm confused, why do we only call the create function if `public_key` is unset?

arekkas:
Anything I can do to help speed up this PR? :) Creating service account keys is indispensable to configuring Google Cloud SQL on GKE containers as the Cloud SQL Proxy requires those credentials - otherwise it's not possible to connect from containers within GKE to Google Cloud SQL which in turn makes it impossible to automate the process with terraform at the moment.

google/resource_google_service_account_key.go
sebglon: yes of course.

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,166 @@
...
+func resourceGoogleServiceAccountKeyCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	serviceAccount := d.Get("service_account_id").(string)
+
+	r := &iam.CreateServiceAccountKeyRequest{
+		KeyAlgorithm: d.Get("key_algorithm").(string),
+	}
+
+	var err error
```
danawillow: You don't have to declare this here since you initialize it just a few lines down with the `:=`

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,166 @@
...
+
+	r.PrivateKeyType = d.Get("private_key_type").(string)
```
danawillow: You can also add this into the struct on line 94ish like you did with `key_algorithm`

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+
+		d.Set("private_key_encrypted", encrypted)
+		d.Set("private_key_fingerprint", fingerprint)
+	} else {
+		d.Set("private_key", sak.PrivateKeyData)
+	}
+
+	err = serviceAccountKeyWaitTime(config.clientIAM.Projects.ServiceAccounts.Keys, d.Id(), d.Get("public_key_type").(string), "Creating Service account key", 4)
+	if err != nil {
```
danawillow: If there's an error creating it, we probably want to `d.SetId("")` so it doesn't end up in state.

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+  service_account_id = "${google_service_account.acceptance.id}"
+  pgp_key = "keybase:keybaseusername"
+  public_key_type = "TYPE_X509_PEM_FILE"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+* `name` - The name used for this key pair (not used on create)
```
danawillow: Since `name` is computed, it probably belongs in the `Attributes Reference` section (and then you can get rid of the parenthetical)

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+key material. Only used when creating or importing a new key pair
+
+~> **NOTE:** a PGP key is not required, however it is strongly encouraged.
+Without a PGP key, the private key material will be stored in state unencrypted.
+
+## Attributes Reference
+
+The following attributes are exported in addition to the arguments listed above:
+
+* `fingerprint` - The MD5 public key fingerprint as specified in section 4 of RFC 4716.
```
danawillow: nit: can you make sure there's a new line between each bullet point?

google/resource_google_service_account_key_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+// Test that a service account key can be created and destroyed
+func TestAccGoogleServiceAccountKey_basic(t *testing.T) {
```
danawillow: Can you add t.Parallel() (and then a new line) to the beginning of each of these tests?

google/resource_google_service_account_key_test.go
```diff
@@ -0,0 +1,134 @@
...
+	account_id = "%s"
+	display_name = "%s"
+}
+
+resource "google_service_account_key" "acceptance" {
+	service_account_id = "${google_service_account.acceptance.id}"
+	public_key_type = "TYPE_X509_PEM_FILE"
+	pgp_key = <<EOF
+%s
+EOF
```
danawillow: you have an extra space after the `EOF` which is causing the test to fail. If you remove the extra space it passes.

danawillow:
Oh also you'll want to add an entry into `website/google.erb` so the documentation shows up on the sidebar

google/resource_google_service_account_key.go
sebglon: This is actualiy a bug on GCE IAM, the key is is well created but not downloadable just after creation. it is necessary to wait a few second or minut to download it.

The setId is good because key well exist

sebglon:
All change are made.

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,163 @@
...
+		return err
+	}
+	resourceGoogleServiceAccountKeyRead(d, meta)
```
danawillow: this can just be `return resourceGoogleServiceAccountKeyRead(d, meta)` and then you can get rid of the rest of the function

google/resource_google_service_account_key.go
```diff
@@ -0,0 +1,163 @@
...
+				Type:         schema.TypeString,
+				Default:      "TYPE_GOOGLE_CREDENTIALS_FILE",
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringInSlice([]string{"TYPE_UNSPECIFIED", "TYPE_PKCS12_FILE", "TYPE_GOOGLE_CREDENTIALS_FILE"}, false),
+			},
+			"public_key": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
```
danawillow: I think this `Optional` line should be removed, and this whole field should move into the Computed section

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,85 @@
...
+The following attributes are exported in addition to the arguments listed above:
+
+* `name` - The name used for this key pair
+
+* `fingerprint` - The MD5 public key fingerprint as specified in section 4 of RFC 4716.
+
+* `public_key` - The public key, base64 encoded
+
+* `private_key` - The private key, base64 encoded. This is only populated
+
```
danawillow: nit: the newline should be after the next line, not before it.

google/resource_google_service_account_key_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+// Test that a service account key can be created and destroyed
+func TestAccGoogleServiceAccountKey_basic(t *testing.T) {
+	t.Parallel()
```
danawillow: nit: empty line after all `t.Parallel()`s

website/docs/r/google_service_account_key.html.markdown
```diff
@@ -0,0 +1,85 @@
+---
+layout: "google"
+page_title: "Google: google_service_accout_key"
```
danawillow: typo: account

website/google.erb
```diff
@@ -92,6 +92,10 @@
       <li<%= sidebar_current("docs-google-service-account") %>>
         <a href="/docs/providers/google/r/google_service_account.html">google_service_account</a>
       </li>
+      <li<%= sidebar_current("docs-google-service-account") %>>
```
danawillow: this should be `docs-google-service-account-key`

sebglon:
@danawillow  All are good

danawillow:
Thanks @sebglon!
```
$ make testacc TEST=./google TESTARGS='-run=TestAccGoogleServiceAccountKey'
TF_ACC=1 go test ./google -v -run=TestAccGoogleServiceAccountKey -timeout 120m
=== RUN   TestAccGoogleServiceAccountKey_basic
=== RUN   TestAccGoogleServiceAccountKey_pgp
--- PASS: TestAccGoogleServiceAccountKey_pgp (59.03s)
--- PASS: TestAccGoogleServiceAccountKey_basic (60.13s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	60.324s
```

arekkas:
Awesome, thank you both for your work on this! Is there a way to check the ETA of unreleased versions (specifically 1.1.2 :) )?

danawillow:
Our goal is to release every two weeks or so, but we don't give exact dates because sometimes we need the flexibility to release sooner or later.

arekkas:
Thank you for the swift response, I'm looking forward to having this released!

arekkas:
Hope this doesn't come by as annoying, but this issue has actually blocked my team and me since it came up. It would be extremely awesome if it was possible to release some time this week, especially because i have no idea how to add custom built plugins to terraform (installed from binaries).

Issue #431, opened 2017-09-13
All `network` fields should accept a name or self-link


rosbo:
For consistency and also to ensure compatibility when referencing via interpolation the `network` field of a resource, all of them should support either a network name or a network self_link.

These resources have already support properly the network field:
* `google_compute_firewall`
* `google_compute_router`
* `google_compute_vpn_gateway`
* `google_compute_forwarding_rule`
* `google_compute_subnetwork`
* `google_compute_route`

These resources DO NOT support properly the network field

* `google_compute_instance_template`
* `google_compute_instance`

rosbo:
```sh
=== RUN   TestAccComputeRouter_import
--- PASS: TestAccComputeRouter_import (80.44s)
=== RUN   TestAccComputeRouter_basic
--- PASS: TestAccComputeRouter_basic (89.93s)
=== RUN   TestAccComputeRouter_noRegion
--- PASS: TestAccComputeRouter_noRegion (78.67s)
=== RUN   TestAccComputeRouter_networkLink
--- PASS: TestAccComputeRouter_networkLink (86.02s)
=== RUN   TestAccComputeVpnGateway_basic
--- PASS: TestAccComputeVpnGateway_basic (63.16s)
```

PR #477, opened 2017-09-27
Firewall network field now supports self_link in addition of name

### LINKED ISSUE #377: google_compute_firewall should accept network self-link
As of now, `name` is properly supported. If we use a self-link, the resource is created but it creates a perpetual diff.

rosbo:
Fixes #377. 

`network` field for other resources will be able to reuse `NetworkFieldValue`. #431 is listing the other resources with a network field that is not properly supported.

### DIFF
--- a/google/field_helpers.go
+++ b/google/field_helpers.go
@@ -7,7 +7,7 @@ import (
 
 const networkLinkTemplate = "projects/%s/global/networks/%s"
 
-var networkLinkRegex = regexp.MustCompile(fmt.Sprintf(networkLinkTemplate, "(.+)", "(.+)"))
+var networkLinkRegex = regexp.MustCompile("projects/(.+)/global/networks/(.+)")
 
 type NetworkFieldValue struct {
 	Project string
@@ -35,6 +35,6 @@ func ParseNetworkFieldValue(network string, config *Config) *NetworkFieldValue {
 	}
 }
 
-func (f *NetworkFieldValue) RelativeLink() string {
+func (f NetworkFieldValue) RelativeLink() string {
 	return fmt.Sprintf(networkLinkTemplate, f.Project, f.Name)
 }
--- a/google/field_helpers.go
+++ b/google/field_helpers.go
@@ -1,0 +1,40 @@
+package google
+
+import (
+	"fmt"
+	"regexp"
+)
+
+const networkLinkTemplate = "projects/%s/global/networks/%s"
+
+var networkLinkRegex = regexp.MustCompile(fmt.Sprintf(networkLinkTemplate, "(.+)", "(.+)"))
+
+type NetworkFieldValue struct {
+	Project string
+	Name    string
+}
+
+// Parses a `network` supporting 4 different formats:
+// - https://www.googleapis.com/compute/{version}/projects/myproject/global/networks/my-network
+// - projects/myproject/global/networks/my-network
+// - global/networks/my-network (default project is used)
+// - my-network (default project is used)
+func ParseNetworkFieldValue(network string, config *Config) *NetworkFieldValue {
+	if networkLinkRegex.MatchString(network) {
+		parts := networkLinkRegex.FindStringSubmatch(network)
+
+		return &NetworkFieldValue{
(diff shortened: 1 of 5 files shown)


google/resource_compute_firewall_test.go
```diff
@@ -299,7 +331,7 @@ func testAccComputeFirewall_update(network, firewall string) string {
 	resource "google_compute_firewall" "foobar" {
 		name = "firewall-test-%s"
 		description = "Resource created for Terraform acceptance testing"
-		network = "${google_compute_network.foobar.name}"
+		network = "${google_compute_network.foobar.self_link}"
```
rosbo: Note: Other tests are referencing the network by name.

google/field_helpers.go
```diff
@@ -0,0 +1,40 @@
...
+	Project string
+	Name    string
+}
+
+// Parses a `network` supporting 4 different formats:
+// - https://www.googleapis.com/compute/{version}/projects/myproject/global/networks/my-network
+// - projects/myproject/global/networks/my-network
+// - global/networks/my-network (default project is used)
+// - my-network (default project is used)
+func ParseNetworkFieldValue(network string, config *Config) *NetworkFieldValue {
```
selmanj: Brainstorming outloud here; none of these are action items:

I wonder if there's a way we can eventually merge all the url parsing code; unfortunately it doesn't seem obvious how since each has their own path scheme. Right now the code is relatively short and simple so the cost of having it isn't much.

Maybe at some point we should combine them all into a single namespace? urltools or resourcetools or something?

google/field_helpers.go
```diff
@@ -0,0 +1,40 @@
+package google
+
+import (
+	"fmt"
+	"regexp"
+)
+
+const networkLinkTemplate = "projects/%s/global/networks/%s"
+
+var networkLinkRegex = regexp.MustCompile(fmt.Sprintf(networkLinkTemplate, "(.+)", "(.+)"))
```
selmanj: This made me quite uncomfortable; something about printing regex chunks into another regex made me squirm a bit

How do you feel about just inlining the regex here?

google/field_helpers.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+	}
+
+	return &NetworkFieldValue{
+		Project: config.Project,
+		Name:    GetResourceNameFromSelfLink(network),
+	}
+}
+
+func (f *NetworkFieldValue) RelativeLink() string {
```
selmanj: Minor quibble; if the struct is small it's nice to not use a ptr here for the receiving struct; that way its obvious that your method doesn't mutate the struct its called on (there's runtime benefits as well but I don't think we should be concerned with that)

google/resource_compute_firewall.go
```diff
@@ -405,12 +404,6 @@ func resourceComputeFirewallDelete(d *schema.ResourceData, meta interface{}) err
 
 func resourceFirewall(d *schema.ResourceData, meta interface{}, computeApiVersion ComputeApiVersion) (*computeBeta.Firewall, error) {
 	config := meta.(*Config)
-	project, _ := getProject(d, config)
-
-	network, err := config.clientCompute.Networks.Get(project, d.Get("network").(string)).Do()
```
selmanj: nice, you eliminated an API call with this change :)

google/field_helpers.go
rosbo: Sure. Done

google/field_helpers.go
rosbo: Agree. I think once we have created a few of these FieldValue classes, we will see patterns emerging and we can create common utils to make it easy to create these FieldValue classes.

google/resource_compute_firewall.go
rosbo: Yup. And the other good thing is that sometimes, this call was failing if the user didn't have the iam permission for reading the network! It was kind of unexpected since we are setting up a firewall.

google/field_helpers.go
rosbo: Done.

PR #491, opened 2017-09-28
Remove deprecated resources for 1.0.0.

paddycarver:
In advance of 1.0.0, let's take the opportunity to remove the fields on
resources that have been deprecated for a while.

Note: this still needs tests run on it.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -749,6 +749,7 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 	}
 
 	// Build up the list of networkInterfaces
+	networkInterfacesCount := d.Get("network_interface.#").(int)
 	networkInterfaces := make([]*computeBeta.NetworkInterface, 0, networkInterfacesCount)
 	for i := 0; i < networkInterfacesCount; i++ {
 		prefix := fmt.Sprintf("network_interface.%d", i)
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -135,10 +135,10 @@ func resourceComputeInstance() *schema.Resource {
 			},
 
 			"disk": &schema.Schema{
-				Type:       schema.TypeList,
-				Optional:   true,
-				ForceNew:   true,
-				Deprecated: "Use boot_disk, scratch_disk, and attached_disk instead",
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Removed:  "Use boot_disk, scratch_disk, and attached_disk instead",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						// TODO(mitchellh): one of image or disk is required
@@ -200,8 +200,6 @@ func resourceComputeInstance() *schema.Resource {
 				},
 			},
 
-			// Preferred way of adding persistent disks to an instance.
-			// Use this instead of `disk` when possible.
 			"attached_disk": &schema.Schema{
 				Type:     schema.TypeList,
 				Optional: true,
@@ -615,7 +613,6 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 		disks = append(disks, bootDisk)
 	}
 
-	var hasScratchDisk bool
 	if _, hasScratchDisk := d.GetOk("scratch_disk"); hasScratchDisk {
 		scratchDisks, err := expandScratchDisks(d, config, zone, project)
 		if err != nil {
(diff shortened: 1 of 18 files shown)


rosbo:
Tests are failling

paddycarver:
Yup, commit coming for that shortly. :)

google/resource_google_project_iam_policy.go
```diff
@@ -31,8 +31,9 @@ func resourceGoogleProjectIamPolicy() *schema.Resource {
 				DiffSuppressFunc: jsonPolicyDiffSuppress,
 			},
 			"authoritative": &schema.Schema{
-				Type:     schema.TypeBool,
-				Optional: true,
+				Type:       schema.TypeBool,
+				Optional:   true,
+				Deprecated: "Use google_project_iam_policy_binding nad google_project_iam_policy_member instead.",
```
rosbo: typo: s/nad/and

website/docs/r/compute_instance.html.markdown
```diff
@@ -116,10 +116,6 @@ The following arguments are supported:
 * `disk` - (DEPRECATED) Disks to attach to the instance. This can be specified
     multiple times for multiple disks. Structure is documented below.
 
-* `network` - (DEPRECATED) Networks to attach to the instance. This
```
rosbo: Delete the `network` block below too.

google/resource_compute_instance.go
```diff
@@ -135,10 +135,10 @@ func resourceComputeInstance() *schema.Resource {
...
 
 			"disk": &schema.Schema{
-				Type:       schema.TypeList,
-				Optional:   true,
-				ForceNew:   true,
-				Deprecated: "Use boot_disk, scratch_disk, and attached_disk instead",
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Removed:  "Use boot_disk, scratch_disk, and attached_disk instead",
```
danawillow: I'm okay with this but I just want to point it out. It might be good having it at the same time as the migration fn? That way we're saying "hey we're getting rid of this but hypothetically you don't need to do anything to replace it". I'm a bit nervous that we'll find a bug in the migration and people will be stuck though.

paddycarver: My biggest concern is that if we don't remove it in the same release that we run the migration, we run the risk of people continuing to use it (despite it being deprecated) and then winding up with abandoned state (or having to run another migration). Basically, what if people upgrade, run the migration, and then continue to use the `disk` field, despite the deprecation?

danawillow: Yeah, I guess they're stuck in that case too.

website/docs/r/google_project_iam_policy.html.markdown
```diff
@@ -50,16 +50,18 @@ The following arguments are supported:
...
     you out** of your project until an Organization Administrator grants you
     access again, so please exercise caution. If this argument is `true` and you
     want to delete the resource, you must set the `disable_project` argument to
     `true`, acknowledging that the project will be inaccessible to anyone but the
-    Organization Admins, as it will no longer have an IAM policy.
+    Organization Admins, as it will no longer have an IAM policy. Rather than using
+    this, you should use `google_project_iam_policy_binding` and
+    `google_project_iam_policy_member`.
 
-* `disable_project` - (Optional) A boolean value that must be set to `true`
```
rosbo: It is not marked as disable in the schema definition

paddycarver:
I've deprecated `disable_project`. 👍, @rosbo?

rosbo:
Yup. I already gave my conditional 👍  Merge at will :)

google/resource_google_project_iam_policy.go
```diff
@@ -31,8 +31,9 @@ func resourceGoogleProjectIamPolicy() *schema.Resource {
 				DiffSuppressFunc: jsonPolicyDiffSuppress,
 			},
 			"authoritative": &schema.Schema{
-				Type:     schema.TypeBool,
-				Optional: true,
+				Type:       schema.TypeBool,
+				Optional:   true,
+				Deprecated: "Use google_project_iam_policy_binding and google_project_iam_policy_member instead.",
```
chrisboulton: Quick sanity check on this for me if I may. Right now `google_project_iam_policy` defaults attempting to merge policies, and the intent is in the future to make it authoritative for policy management on a project (ie, remove this code, and have the code always call setIamPolicy instead of attempt merging). Today, if I wanted that behaviour, I'd set `authoritative = true`, right? Right now when I do that, I'm faced with this deprecation warning here, and I just want to make sure that's intentional.

paddycarver: That's correct. When 2.0.0 ships, the resource will stick around, and there will be no "authoritative" field; it will just always control the entire IAM policy for the project.

Unfortunately, we have no way to silence the deprecation warning if it's set to "true", and I'm not sure we'd want to if we could, because you'll still need to delete the line from your config when you upgrade to 2.0.0. But I'll see if I can make the deprecation message a bit clearer.

chrisboulton: I figured that was the case - and it makes sense to me.. Just wanted to make sure I wasn't backing myself into a corner. Thanks! 🙏

paddycarver: I opened #1343 to make this a bit clearer.

Issue #481, opened 2017-09-28
Add IAM support for storage bucket


sebglon:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
This is a feature proposal

### Affected Resource(s)
Please list the resources as a list, for example:
- google_storage_bucket
- google_storage_bucket_acl
- google_storage_object_acl
- google_storage_object

### Expected Behavior
Add support for IAM Role on Google storage bucket

### Actual Behavior
use new feature : https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy

sebglon:
proposition:
```
resource "google_storage_bucket" "image-store" {
  name     = "image-store-bucket"
  location = "EU"
}

resource "google_storage_bucket_iam" "image-store-iam" {
  bucket = "${google_storage_bucket.image-store.name}"

  iam_policies = [ {
             role = "custom/mu_role"
             members = [
                  "my_account@fff.com",
                  "ffff@ffff.com"
            ]
}
  ]
}
```

sebglon:
for using SetIamPolicy, we can upgrade "google.golang.org/api/storage/v1" to a more recent version

TiGz:
I want this too! kthxbye :)

danawillow:
@emilymye

greendog99:
Looks like the PR is nearly done (thanks @sebglon!). Any idea which upcoming release this might end up in? Looking forward to this feature!

amoiseiev:
@rosbo @danawillow hey guys. It looks like you are close to the finish line on this one. Do you think it can be merged or checked / commented in the next week or two? GCS is one of the core components and the ability to use IAM bindings is must have.

rosbo:
I haven't started the work for this but with the refactor I did in #744 and #776 it shouldn't take long.

One thing to note, after looking briefly at the API, it seems like we can only attach IAM policies to buckets, not particular objects.

sebglon:
The ACL is no on specific object, it is on object action in the bucket.

benbro:
I suggest adding two resources:
- google_storage_bucket_iam_member
- google_storage_bucket_iam_binding

In my opinion it's much cleaner and intuitive than using iam_policy.
Similar to google_project_iam_binding  and google_project_iam_member.

sebglon:
[...]

PR #493, opened 2017-09-29
Init Storage IAM policy

sebglon:
refer to issue #481


sebglon:
Tested OK for Bucket Iam Policy

sebglon:
```
make testacc TEST=./google TESTARGS='-run=TestAccGoogleStorageObjectIAMPolicy'
TF_ACC=1 go test ./google -v -run=TestAccGoogleStorageObjectIAMPolicy -timeout 120m
=== RUN   TestAccGoogleStorageObjectIAMPolicy_basic
--- SKIP: TestAccGoogleStorageObjectIAMPolicy_basic (0.00s)
	resource_google_project_test.go:432: Environment variable GOOGLE_PROJECT_NUMBER is not set
=== RUN   TestAccGoogleStorageObjectIAMPolicy_upgrade
--- SKIP: TestAccGoogleStorageObjectIAMPolicy_upgrade (0.00s)
	resource_google_project_test.go:432: Environment variable GOOGLE_PROJECT_NUMBER is not set
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	0.009s
```

google/resource_storage_bucket_iampolicy.go
```diff
@@ -0,0 +1,122 @@
+package google
```
paddycarver: This is definitely a nitpick, but I'd love to see a `_` between `iam` and `policy` in the filename here.

google/resource_storage_bucket_iampolicy.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	_, err = config.clientStorage.Buckets.SetIamPolicy(bucket, policy).Do()
+	return err
+}
+
+func resourceStorageBucketIAMPolicyCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	if err := setBucketIamPolicy(d, config); err != nil {
```
paddycarver: We usually like to do a `log.Printf("[DEBUG] Setting IAM policy for bucket %q", bucket)` before we make an API call, and a `log.Printf("[DEBUG] Set IAM policy for bucket %q", bucket)` after the request completes, to help debug.

google/resource_storage_bucket_iampolicy.go
```diff
@@ -0,0 +1,122 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func marshalStorageIamPolicy(policy *storagev1.Policy) string {
+	pdBytes, _ := json.Marshal(&storagev1.Policy{
```
paddycarver: I know it seems impossible, but I'd really be much more comfortable if we checked this error here.

google/resource_storage_bucket_imapolicy_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	storagev1 "google.golang.org/api/storage/v1"
+)
+
+func testBucketIamName() string {
+	return fmt.Sprintf("%s-%d", "tf-test-iam-bucket", acctest.RandInt())
```
paddycarver: You can just use `acctest.RandomWithPrefix("tf-test-iam-bucket")` which is a little less convoluted.

google/resource_storage_object_iampolicy.go
```diff
@@ -0,0 +1,106 @@
...
+		return fmt.Errorf("'policy_data' is not valid for %s: %s", object, err)
+	}
+
+	_, err = config.clientStorage.Objects.SetIamPolicy(bucket, object, policy).Do()
+	return err
+}
+
+func resourceStorageObjectIAMPolicyCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
```
paddycarver: Again, debug logs here would be awesome.

paddycarver: Actually, just before and after every API request would be 💯

google/resource_storage_object_iampolicy.go
```diff
@@ -0,0 +1,106 @@
...
+	if err := setObjectIamPolicy(d, config); err != nil {
+		return err
+	}
+
+	d.SetId(d.Get("object").(string))
```
paddycarver: Shouldn't this be a combination of bucket and object, to ensure uniqueness? Also, for future import, we'd need at least project, bucket, and object, I think. So I think setting it to a combination of the three would be really useful.

google/resource_storage_bucket_iampolicy.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func resourceStorageBucketIAMPolicyCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	if err := setBucketIamPolicy(d, config); err != nil {
+		return err
+	}
+
+	d.SetId(d.Get("bucket").(string))
```
paddycarver: If we set this to a combination of project and bucket, we won't need to change IDs to support import.

vendor/vendor.json
```diff
@@ -852,10 +852,10 @@
 			"revisionTime": "2016-12-14T09:25:55Z"
 		},
 		{
-			"checksumSHA1": "FEzQdhqmb6aqGL1lKnjOcUHIGSY=",
+			"checksumSHA1": "F/N7JzfuHZAk1UXi2zw/62hFPpM=",
 			"path": "google.golang.org/api/bigquery/v2",
```
paddycarver: Seems like we're bringing in a lot of unnecessary vendor changes for this PR. Could we clean this up to just update storage, or is there a reason it needs to bring bigquery along?

paddycarver: (Also, is a storage library update necessary to support this functionality?)

google/resource_storage_bucket_iampolicy.go
sebglon: I don't understand exactly what you want. Actualy i do not have project var.
As seen in resource_storage_bucket_acl, ican set bucket+ "-iam_policy"?

vendor/vendor.json
sebglon: Yes an update is necessary and this update bring bigquery lib with small change type

google/resource_storage_bucket_iampolicy.go
paddycarver: So what I'm suggesting here and why:

* When we make resources importable, we need to be able to identify the resources based on their ID. This means the ID needs to not only be unique, but contain all the information necessary to retrieve the resource.
* For this resource, it looks like we need both the project and the bucket ID to be able to retrieve the bucket.
* So ideally, our ID would be something like `{project}/{bucket}`, which we could then parse into the info we need to be able to retrieve the resource given just its ID.

sebglon: On google Policy, i have etag parameter that is an entity ID (return by API). but this it does not contain project.
I don't see how to make it.

sebglon:
@paddycarver  I have cleaned dependencies with master change

sebglon:
@danawillow and @paddycarver is itpossible to have status on this PR?

google/resource_storage_bucket_iam_policy_test.go
```diff
@@ -0,0 +1,182 @@
...
+	var bindingBuffer bytes.Buffer
+
+	for _, binding := range policy.Bindings {
+		bindingBuffer.WriteString("binding {\n")
+		bindingBuffer.WriteString(fmt.Sprintf("role = \"%s\"\n", binding.Role))
+		bindingBuffer.WriteString(fmt.Sprintf("members = [\n"))
+		for _, member := range binding.Members {
+			bindingBuffer.WriteString(fmt.Sprintf("\"%s\",\n", member))
+		}
+		bindingBuffer.WriteString("]}\n")
```
paddycarver: This part of the code scares me, to be honest.

google/resource_storage_object_iam_policy.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func resourceStorageObjectIAMPolicyCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	if err := setObjectIamPolicy(d, config); err != nil {
+		return err
+	}
+
+	d.SetId(d.Get("bucket").(string) + "-" + d.Get("object").(string) + "-iam-policy")
```
paddycarver: If we could use a separator here that is an invalid character for a bucket name, that'd be awesome. (`/` is a good choice, I believe.)

The idea is that we can then parse the ID to get the bucket and object back, which helps in a few places. As it stands, there's no real way to reliably turn this ID back into its constituent parts. In `my-bucket-object-id-iam-policy`, for example, we'd have no idea what was the bucket and what was the object.

google/resource_storage_object_iam_policy_test.go
```diff
@@ -0,0 +1,189 @@
...
+	bucket = "google_storage_bucket.bucket.name"
+}
+resource "google_storage_bucket" "bucket" {
+  display_name = "%s"
+}
+data "google_iam_policy" "test" {
+  %s
+}
+resource "google_storage_Object_iam_policy" "test" {
+  Object = "${google_storage_Object.Object.name}"
```
paddycarver: Looks like a find and replace is doing some weird capitalising here.

google/resource_storage_object_iam_policy_test.go
```diff
@@ -0,0 +1,189 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	storagev1 "google.golang.org/api/storage/v1"
+)
+
+func testObjectIamName() string {
+	return fmt.Sprintf("%s-%d", "tf-test-iam-Object", acctest.RandInt())
```
paddycarver: Looks like a find and replace is doing some weird capitalising here.

google/resource_storage_object_iam_policy_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+		return nil
+	}
+}
+
+func testAccGoogleStorageObjectIAMPolicyDestroy(s *terraform.State) error {
+	config := testAccProvider.Meta().(*Config)
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "google_storage_Object_iam_policy" {
```
paddycarver: Looks like a find and replace is doing some weird capitalising here.

website/docs/r/storage_bucket_iam_policy.html.markdown
```diff
@@ -0,0 +1,58 @@
+---
```
paddycarver: Looks like you've duplicated the bucket docs here?

website/docs/r/storage_object_iam_policy.html.markdown
```diff
@@ -0,0 +1,63 @@
+---
+layout: "google"
+page_title: "Google: google_storage_object_iam_policy"
+sidebar_current: "docs-google--storage-object-iam-policy"
+description: |-
+ Allows management of an IAM policy for a Google Storage Bucket Object.
+---
+
+# google\_storage\-_object\_iam\_policy
```
paddycarver: Looks like you have an extra `-` in here that's messing things up.

website/docs/r/storage_object_iam_policy.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+## Example Usage
+
+```hcl
+resource "google_storage_bucket" "bucket" {
+  display_name = "%s"
+}
+
+resource "google_storage_object" "object" {
+  display_name = "%s"
+	bucket = "${google_storage_bucket.bucket.name}"
```
paddycarver: Tabs/spaces problem. Generally we ask that documentation examples be run through `terraform fmt`.

website/docs/r/storage_object_iam_policy.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+
+* `object` - (Required) The Object ID.
+
+* `policy_data` - (Required) The `google_iam_policy` data source that represents
+    the IAM policy that will be applied to the project. The policy will be
+    merged with any existing policy applied to the project.
+
+    Changing this updates the policy.
+
+    Deleting this removes the policy, but leaves the original object policy
```
paddycarver: Is this true? Looking up at the code, it looks like the policy is just set to an empty policy on delete.

google/resource_storage_bucket_iam_policy_test.go
sebglon: I don't understand Why?

website/docs/r/storage_bucket_iam_policy.html.markdown
sebglon: No; Why?

greendog99: Looks like you have both a `google_storage_bucket_iam_policy.html.markdown` and a nearly identical `storage_bucket_iam_policy.html.markdown` file (without the "google" prefix). I assume that's what @paddycarver is referring to.

paddycarver: Yup, sorry. I'm seeing two different files here with almost the same name and can't tell what happened.

greendog99: FYI, @sebglon deleted the duplicate file. I think the only outstanding issue left is the "this part of the code scares me" section, if you can expand on that?

google/resource_storage_bucket_iam_policy_test.go
sebglon: this is the same than resource_google_folder_iam_policy_test.go;

I have added Comment on source code

paddycarver: Apologies, missed this response. It largely scares me because we're dynamically building the config string in a buffer, which is more complicated than tests usually require. I think this may be more generalized than is strictly necessary, but it's really not worth holding a merge over.

paddycarver:
It looks like after 580b153, the code no longer builds.

sebglon:
@paddycarver Compile is corrected.

sebglon:
@danawillow , @paddycarver  Can i have status on validation; this lock my project?

tragiclifestories:
Cc @rosbo who has been working on the IAM system

rosbo:
In #744 we refactored the IAM code to make it easier to write new IAM resources. #822 adds IAM support for storage bucket using the new architecture.

Closing this one in favor #822.

PR #498, opened 2017-09-29
Network peering gets network project from network self_link

### LINKED ISSUE #496: `google_compute_network_peering` project field is required but cannot be set
When creating a `google_compute_network_peering` resource with a provider with no default project, the following error is received:

```
* google_compute_network_peering.my_peering: "project": required field is not set
```

This is the expected behaviour. However, when adding the `project` field as requested on the resource, the following error is received:

```
* google_compute_network_peering.my_peering: : invalid or unknown key: project
```

The resource should support the `project` field for Terraform configurations without a default project on the provider.

rosbo:
Fixes #496 

Network peering requires two network `self_link`. One for each of the network in the peering.
We were reading the project from `config.Project`. This means that if `config.Project` is not set, we could not set the peering. 

Since we already require a network `self_link`, we should extract the project from `self_link`.

cc/ @rochdev

### DIFF
--- a/google/resource_compute_network_peering.go
+++ b/google/resource_compute_network_peering.go
@@ -3,7 +3,6 @@ package google
 import (
 	"fmt"
 	"log"
-	"regexp"
 	"sort"
 
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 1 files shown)


rochdev:
I tend to use `network.name` most of the time. Would it be possible to also support an explicit project? Or should I maybe switch to `self_link` instead?

rosbo:
This resource requires a `self_link`, otherwise the ValidateFunc will fail. Since having a `self_link` in the `network` field is enforced at plan time, I liked the idea of having only one way to specify a project to prevent the situation where the `network` field contains a `self_link` with project `A` and the `project` field contains `B`. In that case, should we fail and display an error message? Should we silently pick `A`? Or `B`?

Additionally, the `peer_network` field can reference a network in the same or a different project. This means, for consistency, we would need to support specifying only the network name in `peer_network`. However, since the peer network may not be in the same project, then we would need a separate project field (we could call it `peer_project`). This would create the same problem as above if someone use a self_link and a project with a project that doesn't match.

The use of self_link and its enforcement at **plan time** remove all this confusion for the users.

PR #500, opened 2017-09-29
Firewall uses v1 API if the priority is unset or has the default value.

### LINKED ISSUE #429: google_compute_firewall always uses the beta API even when not using any beta fields.
In #345, we added a new beta feature. Rule priority.

The `priority` field has a default value of 1000. Even if the field is not set, we have a value of 1000 for the field. The `getComputeApiVersion` considers the field as set and therefor activates the beta API.

rosbo:
Fixes #429 

Right now, the `google_compute_firewall` always uses the beta API even if no beta fields are set.

The is because the beta `priority` field has a default value set to 1000 in the schema. Even if you don't specify a value for the `priority` field, the field still has a value (the default value) and `getComputeApiVersion` or `getComputeApiVersionUpdate` consider that the field is used and returns that the `v0beta` should be used.

### DIFF
--- a/google/api_versions.go
+++ b/google/api_versions.go
@@ -45,7 +45,7 @@ type TerraformResourceData interface {
 func getComputeApiVersion(d TerraformResourceData, resourceVersion ComputeApiVersion, features []Feature) ComputeApiVersion {
 	versions := map[ComputeApiVersion]struct{}{resourceVersion: struct{}{}}
 	for _, feature := range features {
-		if feature.InUseBy(d) {
+		if feature.InUseByDefault(d) {
 			versions[feature.Version] = struct{}{}
 		}
 	}
@@ -58,8 +58,12 @@ func getComputeApiVersion(d TerraformResourceData, resourceVersion ComputeApiVer
 // version, to determine what version of the API is required in order to update the resource.
 func getComputeApiVersionUpdate(d TerraformResourceData, resourceVersion ComputeApiVersion, features, updateOnlyFields []Feature) ComputeApiVersion {
 	versions := map[ComputeApiVersion]struct{}{resourceVersion: struct{}{}}
-	schemaVersion := getComputeApiVersion(d, resourceVersion, features)
-	versions[schemaVersion] = struct{}{}
+
+	for _, feature := range features {
+		if feature.InUseByUpdate(d) {
+			versions[feature.Version] = struct{}{}
+		}
+	}
 
 	for _, feature := range updateOnlyFields {
 		if feature.HasChangeBy(d) {
@@ -83,6 +87,10 @@ type Feature struct {
 	//
 	// Note: beta field nested inside a SET are NOT supported at the moment.
 	Item string
+
+	// Optional, only set if your field has a default value.
+	// If the value for the field is equal to the DefaultValue, we assume the beta feature is not activated.
+	DefaultValue interface{}
 }
 
 // Returns true when a feature has been modified.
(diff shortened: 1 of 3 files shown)


google/api_versions.go
```diff
@@ -93,16 +101,37 @@ func (s Feature) HasChangeBy(d TerraformResourceData) bool {
...
 }
 
-// Return true when a feature appears in schema or has been modified.
-func (s Feature) InUseBy(d TerraformResourceData) bool {
-	return inUseBy(d, s.Item)
+type InUseFunc func(d TerraformResourceData, path string, defaultValue interface{}) bool
+
+func defaultInUseFunc(d TerraformResourceData, path string, defaultValue interface{}) bool {
+	// At read and delete time, there is no change.
+	// At create time, all fields are marked has changed. We should only considered the feature active if the field has
```
danawillow: nit: consider

google/api_versions.go
```diff
@@ -93,16 +101,37 @@ func (s Feature) HasChangeBy(d TerraformResourceData) bool {
...
+	// a value set and that this value is not the default value.
+	value, ok := d.GetOk(path)
+	return ok && value != defaultValue
+}
+
+func updateInUseFunc(d TerraformResourceData, path string, defaultValue interface{}) bool {
+	// During a resource update, if the beta field has changes, the feature is considered active even if the new value
+	// is the default value. This is because the beta API must be call to change the value of the field back to the
```
danawillow: nit: called

google/api_versions_test.go
```diff
@@ -29,7 +29,7 @@ func TestComputeApiVersion(t *testing.T) {
 				Update:     baseVersion,
 			},
 		},
-		"beta fields no set": {
+		"beta field no set": {
```
danawillow: nit: not

google/api_versions_test.go
```diff
@@ -104,6 +104,60 @@ func TestComputeApiVersion(t *testing.T) {
...
+				"normal_field": "foo",
+				"beta_field":   "bar",
+			},
+			ExpectedApiVersions: ExpectedApiVersions{
+				Create:     baseVersion,
+				ReadDelete: baseVersion,
+				Update:     baseVersion,
+			},
+		},
+		"beta fields is being updated to default value": {
```
danawillow: nit: field

google/api_versions_test.go
```diff
@@ -104,6 +104,60 @@ func TestComputeApiVersion(t *testing.T) {
...
+				"list_field.1.normal_field":      "bar",
+				"list_field.1.beta_nested_field": "baz",
+			},
+			ExpectedApiVersions: ExpectedApiVersions{
+				Create:     baseVersion,
+				ReadDelete: baseVersion,
+				Update:     baseVersion,
+			},
+		},
+		"nested beta field is being updated default value": {
```
danawillow: nit: updated to

google/api_versions.go
rosbo: Done

google/api_versions.go
rosbo: Done

google/api_versions_test.go
rosbo: done

google/api_versions_test.go
rosbo: done

google/api_versions_test.go
rosbo: done

Issue #433, opened 2017-09-14
google_container_cluster should allow disabling of KubernetesDashboard addon


calind:
Currently the `google_container_cluster` allows enabling/disabling of `HorizontalPodAutoscaling` and `HttpLoadBalancing` addons. Since https://cloud.google.com/container-engine/release-notes#july_25_2017 it's allowed to enable/disable the KubernetesDashboard addon as well.

sebglon:
I have created a PR for this #505

selmanj:
@sebglon Thanks for your PR!  Unfortunately I beat you to it with #464. Please keep them coming though!

PR #464, opened 2017-09-26
Add kubernetes dashboard addon

selmanj:
To support this I had to:
1. Update the container api
1. Modify the testing logic to check addons; right now if the api returned nil it would ignore whatever was in the terraform state

PR #505, opened 2017-10-01
allow disabled KubernetesDashboard

sebglon:
#433 
Adding possibility to disable KubernetesDashboard


selmanj:
Already merged via #464. Thanks for the PR though, looking forward to seeing more!

PR #512, opened 2017-10-02
Add update support for pubsub subscription push config field

### LINKED ISSUE #455: Pubsub Subscription should allow updating push config

rosbo:
Fixes #455.

As discussed in #456, testing in not very practical for the `push_config` field. You can learn more also here: https://github.com/rosbo/terraform-provider-google/blob/f79044a57c7ed52d13144daa09899dc553e0ec2b/google/resource_pubsub_subscription_test.go#L33. I tested it manually.

### DIFF
--- a/google/resource_pubsub_subscription.go
+++ b/google/resource_pubsub_subscription.go
@@ -11,6 +11,7 @@ func resourcePubsubSubscription() *schema.Resource {
 	return &schema.Resource{
 		Create: resourcePubsubSubscriptionCreate,
 		Read:   resourcePubsubSubscriptionRead,
+		Update: resourcePubsubSubscriptionUpdate,
 		Delete: resourcePubsubSubscriptionDelete,
 
 		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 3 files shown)


google/resource_pubsub_subscription.go
```diff
@@ -186,3 +188,17 @@ func flattenPushConfig(pushConfig *pubsub.PushConfig) []map[string]interface{} {
 
 	return configs
 }
+
+func expandPubsubSubscriptionPushConfig(configured []interface{}) *pubsub.PushConfig {
+	if len(configured) != 1 {
```
selmanj: Minor: `len(configured) == 0` would be more explicit and match the comment

google/utils.go
```diff
@@ -318,6 +318,14 @@ func convertStringSet(set *schema.Set) []string {
 	return s
 }
 
+func convertStringMap(v map[string]interface{}) map[string]string {
```
selmanj: expandStringMap does the same as this i believe

rosbo: Not quite. `expandStringMap` works with a `schema.ResourceData`. This one works with a `map[string]interface{}`.

google/resource_pubsub_subscription.go
rosbo: Agree, easier to read. Done.

google/utils.go
selmanj: Can you rewrite one in terms of the other? `expandStringMap` can defer to this method

rosbo: Done

Issue #441, opened 2017-09-16
Feature Request: Support for KMS KeyRings / CryptoKeys


mrparkers:
Hi everyone,

I was looking into using Terraform to create my next GCP project and I noticed that there are no resources for Google's KMS, which I use in most of my projects.

I was wondering if there was already a plan to add support for these resources.  If not, I'd like to take a stab at submitting a PR to implement them (if there are no objections).

Relevant links:
- [KMS REST reference](https://cloud.google.com/kms/docs/reference/rest/)
- [cloudkms Go package reference](https://godoc.org/google.golang.org/api/cloudkms/v1)

mayakacz:
Hi,
I'm the PM on Cloud KMS. That would be a great addition. If you have any implementation questions please let us know!

danawillow:
Hey @Mrparkers, my team was planning on taking this on in Q4 but if you'd like to do it go for it! Feel free to reach out if you have any questions, and keep me updated on your progress 😃

Mrparkers:
Hi @mayakacz and @danawillow, thanks for the replies!

I did have one question about the implementation, and I wasn't able to find an example in the codebase that answered my question.  Since it doesn't look like you can delete a KeyRing or a CryptoKey using the KMS API, what is the expected behavior for the `Destroy` function for these resources?  My guess is that I would just throw an error that explains that the resource cannot be destroyed, but I wasn't able to find an example to affirm this.

Thanks!

danawillow:
Oh, that's a good question. I don't like the idea of throwing an error since that means you can never do a `terraform destroy` on your full configuration. I think we could just remove it from state on destroy (i.e. do nothing), and make sure that if you try to create something that already exists it doesn't throw an error. I wonder what happens if you just don't implement a delete function at all?
 I don't know whether that would be allowed, but you could try! And we should also be really clear in the docs that this is different behavior from other resources because KeyRings and CryptoKeys can never be deleted.

Just to confirm @mayakacz, is this intentional based on how KMS works that these can't be deleted or is it just that it can't be done via the API (looks like gcloud doesn't have delete functionality either)

Mrparkers:
Okay, I'll go with your suggestion of removing it from state and allowing it to be "re-created" to add it back.

[...]

PR #518, opened 2017-10-03
Adds support for creating KMS KeyRing resources

Mrparkers:
- [x] Create resource with simple acceptance test
- [x] Update acceptance test to create a new GCP project to create the resource in
- [x] Add docs

Discussion in #441 

I'm mostly looking for advice on what I could be doing better as I continue to work on this in my spare time.


google/resource_kms_key_ring.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return err
+	}
+
+	name := d.Get("name").(string)
+	location := d.Get("location").(string)
+
+	parent := createKmsResourceParentString(project, location)
+
+	_, err = config.clientKms.Projects.Locations.KeyRings.
```
danawillow: style nit: for consistency with other resources, I'd recommend either not breaking at all or breaking after `Create(`

google/resource_kms_key_ring.go
```diff
@@ -0,0 +1,90 @@
...
+			"location": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func createKmsResourceParentString(project, location string) string {
```
danawillow: readability nit: I'm not a huge fan of `create` being used in functions that don't actually create something, especially in a code base like this where it realistically could be making API calls. How about just `kmsResourceParentString`?

google/resource_kms_key_ring.go
```diff
@@ -0,0 +1,90 @@
...
+
+	keyRing, err := config.clientKms.Projects.Locations.KeyRings.
+		Get(keyRingName).
+		Do()
+
+	if err != nil {
+		return fmt.Errorf("Error reading KeyRing: %s", err)
+	}
+
+	d.SetId(keyRing.Name)
```
danawillow: I don't think we usually set the id again on read- any reason why you needed it here?

google/resource_kms_key_ring_test.go
```diff
@@ -0,0 +1,78 @@
...
+					testAccCheckGoogleKmsKeyRingExists("google_kms_key_ring.key_ring"),
+				),
+			},
+		},
+	})
+}
+
+func testGoogleKmsKeyRing_basic(name string) string {
+	return fmt.Sprintf(`
+	resource "google_kms_key_ring" "key_ring" {
```
danawillow: nit- can you align this all the way to the left?

google/resource_kms_key_ring_test.go
```diff
@@ -0,0 +1,78 @@
...
+				Config: testGoogleKmsKeyRing_basic(name),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGoogleKmsKeyRingExists("google_kms_key_ring.key_ring"),
+				),
+			},
+		},
+	})
+}
+
+func testGoogleKmsKeyRing_basic(name string) string {
```
danawillow: I don't have a good reason behind this, but we tend to keep these functions at the end of the file

google/resource_kms_key_ring_test.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+		}
+
+		return fmt.Errorf("KeyRing not found: %s", keyRingName)
+	}
+}
+
+// TODO
+// KMS KeyRings cannot be deleted. This will test if the resource can be added back to state after being removed
+func testGoogleKmsKeyRing_recreate(name string) resource.TestCheckFunc {
```
danawillow: If the resource already exists server-side, we can add import functionality instead of trying to get it to essentially import on create.

Mrparkers:
Thanks for the comments @danawillow.  I was planning on setting aside some time tonight to work on this some more, and I can address your comments then.

google/resource_kms_key_ring.go
Mrparkers: I thought I saw another resource do this, but I can't remember where I saw it.  I'll remove this.

google/resource_kms_key_ring_test.go
Mrparkers: What is the expected behavior of attempting to create a KeyRing that already exists?  Should it fail because the resource already exists on the server?  Or should it handle it similar to how an import would?

I think most resources would fail here because you'd expect to have to use `terraform import` for resources not managed by terraform.  This seems like a special situation though, because for all other resources, you can do `terraform destroy` and `terraform apply` to re-create your infrastructure without having to modify the names of any of your resources.  If calling `Create` for a KMS Keyring fails for KeyRings that already exist on the server, you wouldn't be able to do this.

In either case, my intention was for this test to capture that logic, where I either expect a failure when re-creating, or for it to be handled gracefully.

google/resource_kms_key_ring.go
Mrparkers: Sorry, I'm not sure what you mean by this.  Could you give me another resource to look at as an example?

Mrparkers:
That sounds good.  I'm guessing that means `import` will just work as it normally does with other resources then?

google/resource_kms_key_ring_test.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		name := rs.Primary.Attributes["name"]
+		location := rs.Primary.Attributes["location"]
+		project := rs.Primary.Attributes["project"]
+
+		parent := kmsResourceParentString(project, location)
+		keyRingName := kmsResourceParentKeyRingName(project, location, name)
+
+		listKeyRingsResponse, err := config.clientKms.Projects.Locations.KeyRings.List(parent).Do()
```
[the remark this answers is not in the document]
Mrparkers: I used list because I don't like it when my tests look similar to the production code.  I thought this would be a good way to still test the intended functionality without making it too similar to `Read`.

I am open to changing it if you would like me to.

google/resource_kms_key_ring.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return err
+	}
+
+	name := d.Get("name").(string)
+	location := d.Get("location").(string)
+
+	parent := kmsResourceParentString(project, location)
+
+	keyRing, err := config.clientKms.Projects.Locations.KeyRings.
```
danawillow: sorry to point this out again, but mind fixing the style here and elsewhere? https://github.com/golang/go/wiki/CodeReviewComments#line-length
I think one line would be fine, two is also fine (we usually break after the left paren)

google/resource_kms_key_ring.go
```diff
@@ -0,0 +1,104 @@
...
+
+/*
+	Because KMS KeyRing resources cannot be deleted on GCP, we are only going to remove it from state.
+	Re-creation of this resource through Terraform will produce an error.
+ */
+
+func resourceKmsKeyRingDelete(d *schema.ResourceData, meta interface{}) error {
+	keyRingName := d.Id()
+
+	log.Printf("[INFO] KMS KeyRing resources cannot be deleted from GCP. This KeyRing %s will be removed from Terraform state, but will still be present on the server.", keyRingName)
```
danawillow: Let's do WARNING juuuuuuust in case

google/resource_kms_key_ring_test.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/*
+	This test runs in its own project, otherwise the test project would start to get filled
+	with undeletable resources
+ */
+func testGoogleKmsKeyRing_basic(projectId, projectOrg, projectBillingAccount, keyRingName string) string {
+	return fmt.Sprintf(`
+resource "google_project" "acceptance" {
+    name            = "%s"
```
danawillow: can you fix the formatting here? I think you just have a mix of tabs and spaces

google/resource_kms_key_ring.go
Mrparkers: Ah, I don't think I understood before but I see what you're saying now.  I'll change it to a single line.

Mrparkers:
@danawillow Thanks for your comments.  I went ahead and added import functionality, so if you could review that, I would appreciate it.  I'll leave some comments myself where I had questions.

Also, I just exceeded my project creation quota for the GCP free trial, so I won't be able to run my acceptance test locally anymore 😅 

I will work on adding the docs tomorrow so this can get wrapped up.  Thanks for your patience!

google/import_kms_key_ring_test.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func testGoogleKmsKeyRing_import(keyRingName string) string {
+	return fmt.Sprintf(`
+resource "google_kms_key_ring" "key_ring" {
+	name     = "%s"
+	location = "us-central1"
+}
+	`, keyRingName)
+}
```
Mrparkers: Should I follow the same pattern as I did with my other test and run it in a separate project?  The only reason I didn't do this was because I used up my GCP free trial project quota, so I wouldn't have been able to test it myself.

google/import_kms_key_ring_test.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"testing"
+
+	"fmt"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccGoogleKmsKeyRing_importBasic(t *testing.T) {
+	t.Parallel()
```
Mrparkers: I'm unsure if I will leave this here if I go down the route of running this test in its own GCP project.

google/resource_kms_key_ring.go
```diff
@@ -0,0 +1,142 @@
...
+	d.SetId("")
+
+	return nil
+}
+
+func parseKmsKeyRingId(id string) (*kmsKeyRingId, error) {
+	keyRingIdRegex := regexp.MustCompile("projects/(.+)/locations/(.+)/keyRings/(.+)$")
+
+	if !keyRingIdRegex.MatchString(id) {
+		return nil, fmt.Errorf("Invalid KeyRing id format, expecting projects/{projectId}/locations/{locationId}/keyRings/{keyRingName}")
```
Mrparkers: What is the normal pattern for IDs when importing?  I figured I would make it as similar to the REST API as possible, which puts the literal strings `project`, `location`, and `keyRing` in the ID.  Are these normally included in other resources?

danawillow: eh not usually. See https://www.terraform.io/docs/providers/google/r/compute_instance_group.html#import for an example that uses multiple elements in the id.
There's some pretty good discussion on what the best import format is at https://github.com/terraform-providers/terraform-provider-google/pull/378. They ended up with a solution that allows for multiple different ways of importing it, which is pretty neat but not strictly necessary. I'll let you figure out which approach you think works best for KeyRings.

Mrparkers: Okay, I switched it to `{projectId}/{location}/{keyRing}`

google/import_kms_key_ring_test.go
Mrparkers: Okay, I updated that test.  Feel free to run it whenever you can.

Mrparkers:
@danawillow I addressed your comments and added the documentation page.  Please let me know if there's anything you'd like me to change.

google/resource_kms_key_ring_test.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/*
+	This test runs in its own project, otherwise the test project would start to get filled
+	with undeletable resources
+*/
+func testGoogleKmsKeyRing_basic(projectId, projectOrg, projectBillingAccount, keyRingName string) string {
+	return fmt.Sprintf(`
+resource "google_project" "acceptance" {
+    name            = "%s"
```
danawillow: can you fix the formatting here? I think you've got a spaces vs tabs issue (we usually use tabs)

google/resource_kms_key_ring_test.go
```diff
@@ -0,0 +1,196 @@
...
+	name     = "%s"
+	location = "us-central1"
+}
+	`, projectId, projectId, projectOrg, projectBillingAccount, keyRingName)
+}
+
+func testGoogleKmsKeyRing_removed(projectId, projectOrg, projectBillingAccount string) string {
+	return fmt.Sprintf(`
+resource "google_project" "acceptance" {
+    name            = "%s"
```
danawillow: here too

website/docs/r/google_kms_key_ring.html.markdown
```diff
@@ -0,0 +1,61 @@
+---
+layout: "google"
+page_title: "Google: google_kms_key_ring"
+sidebar_current: "docs-google-kms-key-ring-x"
```
danawillow: why x?

google/resource_kms_key_ring_test.go
Mrparkers: Oops, my editor automatically converts tabs to spaces for me.  I can fix this.

website/docs/r/google_kms_key_ring.html.markdown
Mrparkers: The file I was using as a reference had it as well.  I'll remove it.

Mrparkers:
I'm not sure I understand this failure.  Is it complaining because it tried to import the project as well?

I ran this test last night without setting up the extra project (I just commented out the `google_project` and the `google_project_services` resources) and it passed.

danawillow:
It's checking that the resource that got imported has exactly the same configuration as the one in the test. Since the one in the test had `project` set explicitly, it wants `project` to be set in state. (I can explain this with examples if it's confusing.) Right now, nothing sets "project" in state so that's where the mismatch was. There are a few different ways to solve this- the one I just tested out adds this block into the `ImportState` function:
```
	if config.Project != keyRingId.Project {
		d.Set("project", keyRingId.Project)
	}
```
The other ways to solve this would be to add the set in the Read fn or just always do it in Import, but I kind of like the idea that you only need to set "project" on your resource if the project doesn't match the one in the config.

Mrparkers:
That makes perfect sense, thank you for explaining it to me.

I must be remembering incorrectly, because there's no way that test would have passed just by commenting out the `google_project` and the `google_project_services` resources.

I'll use the block of code you posted and add it to `ImportState` like you suggested.

danawillow:
```
$ make testacc TEST=./google TESTARGS='-run=TestAccGoogleKmsKeyRing' GOOGLE_USE_DEFAULT_CREDENTIALS=true
TF_ACC=1 go test ./google -v -run=TestAccGoogleKmsKeyRing -timeout 120m
=== RUN   TestAccGoogleKmsKeyRing_importBasic
--- PASS: TestAccGoogleKmsKeyRing_importBasic (53.64s)
=== RUN   TestAccGoogleKmsKeyRing_basic
--- PASS: TestAccGoogleKmsKeyRing_basic (51.80s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	105.617s
```
Looks good, thanks @Mrparkers!

jasonjho:
@danawillow  @mrparkers  Sorry if this has already been addressed, but I am confirming behavior on v0.11.1 that due to the immutability of key rings, a destroy and subsequent create will fail with the following error:

```Error creating KeyRing: googleapi: Error 409: KeyRing projects/<proj>/locations/global/keyRings/<some_key> already exists., alreadyExists```

What's the proper way to handle this kind of scenario? Any help appreciated.

Thanks!

danawillow:
Hey @jasonjho, is your goal to get the key ring back into your terraform state? If so, you can import it: https://www.terraform.io/docs/providers/google/r/google_kms_key_ring.html#import. It sounds like a data source would be a good fit for this type of scenario too- feel free to file a feature request for one.

PR #520, opened 2017-10-03
Fix: Updates sidebar text in docs for google_client_config resource

Mrparkers:
I noticed that the sidebar in the docs had `google_compute_network` listed twice:

image

It looks like the anchor tag was copy/pasted without changing the text content.


rosbo:
Thank you!

Issue #449, opened 2017-09-22
Infinite apply loop when sql database instance region is us-central instead of us-central1


danielcompton:
### Terraform Version
Terraform v0.10.6
GCP provider v0.1.3

### Affected Resource(s)

- google_compute_instance_template
- google_sql_database_instance

### Terraform Configuration Files
```hcl
resource "google_sql_database_instance" "master" {
  region           = "us-central"
  database_version = "POSTGRES_9_6"

  settings {
    tier = "${terraform.env == "default" ? "db-custom-1-4096" : "db-f1-micro"}"

    backup_configuration {
      enabled = true
    }
  }
}

resource "google_sql_database" "app" {
  name     = "deps_prod"
  instance = "${google_sql_database_instance.master.name}"
}
```

### Expected Behavior
After running an apply, terraform should reach a stable configuration

### Actual Behavior
Every time I run plan after an apply, the database instance wants to be recreated.

### Steps to Reproduce
1. `terraform apply`
2. `terraform plan`

### Important Factoids
When writing up this issue, I found that the region was changing from us-central1 back to us-central. The [docs](https://www.terraform.io/docs/providers/google/r/sql_database_instance.html#region) say that you need to use us-central, but it seems like now us-central1 is ok?

This might be good to put in some upgrade notes for people upgrading to 0.10 Terraform. I assumed something was broken with the provider when I first saw this, and almost didn't bother upgrading and trying to debug it.

danawillow:
I'm going to go ahead and close this based on #521, but if we see it come up more I can look into a more robust solution.

PR #521, opened 2017-10-03
Update region in sql db docs

danawillow:
Since the available regions might change over time, link to the GCP documentation instead of trying to provide our own list.

@danielcompton would you say this is enough for #449?

### DIFF
--- a/website/docs/r/sql_database_instance.html.markdown
+++ b/website/docs/r/sql_database_instance.html.markdown
@@ -36,8 +36,8 @@ resource "google_sql_database_instance" "master" {
 The following arguments are supported:
 
 * `region` - (Required) The region the instance will sit in. Note, this does
-    not line up with the Google Compute Engine (GCE) regions - your options are
-    `us-central`, `asia-west1`, `europe-west1`, and `us-east1`.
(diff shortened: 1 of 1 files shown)


danielcompton:
It looks like for second generation instances all of the names line up with [actual regions](https://cloud.google.com/compute/images/zones_diagram.svg). The only oddball in the SQL list is `us-central` for first-gen SQL instances. For second-gen I don't think there is any distinction to make. I think it would be less confusing to say:

> region - (Required) The region the instance will sit in. Note, first-generation Cloud SQL instance regions do not line up with the Google Compute Engine (GCE) regions - choose from one of the options listed here.

danawillow:
True, but not all GCE regions can be used for Cloud SQL (asia-southeast1 is missing)

danielcompton:
How about:

> region - (Required) The region the instance will sit in. Note, first-generation Cloud SQL instance regions do not line up with the Google Compute Engine (GCE) regions, and Cloud SQL is not available in all regions - choose from one of the options listed here.

Otherwise what you've got is fine too.

danawillow:
I like it, thanks!

PR #523, opened 2017-10-04
Add support for org policies at the organization level

### LINKED ISSUE #507: Add support for org policies at the organization level
Organization policies provide the following benefits:
- Policies can be set per project or per organization.
- Policies are inherited down the resource hierarchy, and can be overridden at any level on which an org policy can be set.
- Policies are not necessarily managed by the owner of the resource. Instead, policies are managed by your organization's policy administrator (IAM role: roles/orgpolicy.policyAdmin). This means that individual users and project owners cannot override policies.

https://cloud.google.com/resource-manager/docs/organization-policy/overview

rosbo:
Fixes #507 

cc/ @ubschmidt2

### DIFF
--- a/google/resource_google_organization_policy.go
+++ b/google/resource_google_organization_policy.go
@@ -1,0 +1,327 @@
+package google
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/cloudresourcemanager/v1"
+	"strings"
+)
+
+func resourceGoogleOrganizationPolicy() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGoogleOrganizationPolicyCreate,
+		Read:   resourceGoogleOrganizationPolicyRead,
+		Update: resourceGoogleOrganizationPolicyUpdate,
+		Delete: resourceGoogleOrganizationPolicyDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: resourceGoogleOrganizationPolicyImportState,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"org_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"constraint": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: linkDiffSuppress,
+			},
+			"boolean_policy": {
+				Type:          schema.TypeList,
+				Optional:      true,
+				MaxItems:      1,
+				ConflictsWith: []string{"list_policy"},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enforced": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+					},
+				},
+			},
+			"list_policy": {
+				Type:          schema.TypeList,
+				Optional:      true,
+				MaxItems:      1,
+				ConflictsWith: []string{"boolean_policy"},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"allow": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"all": {
+										Type:     schema.TypeBool,
+										Optional: true,
+										Default:  false,
+									},
+									"values": {
+										Type:     schema.TypeSet,
+										Optional: true,
+										Elem:     &schema.Schema{Type: schema.TypeString},
+										Set:      schema.HashString,
+									},
+								},
+							},
+						},
+						"deny": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"all": {
+										Type:     schema.TypeBool,
+										Optional: true,
+										Default:  false,
+									},
+									"values": {
+										Type:     schema.TypeSet,
+										Optional: true,
+										Elem:     &schema.Schema{Type: schema.TypeString},
+										Set:      schema.HashString,
+									},
+								},
+							},
+						},
+						"suggested_value": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+						},
+					},
+				},
+			},
+			"version": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+			},
+			"etag": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"update_time": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceGoogleOrganizationPolicyCreate(d *schema.ResourceData, meta interface{}) error {
+	if err := setOrganizationPolicy(d, meta); err != nil {
+		return err
+	}
+
+	d.SetId(fmt.Sprintf("%s:%s", d.Get("org_id"), d.Get("constraint").(string)))
+
+	return resourceGoogleOrganizationPolicyRead(d, meta)
+}
+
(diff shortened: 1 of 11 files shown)


rosbo:
```sh
TF_ACC=1 go test ./google -v -run TestAccGoogleOrganizationPolicy_ -timeout 120m
=== RUN   TestAccGoogleOrganizationPolicy_import
--- PASS: TestAccGoogleOrganizationPolicy_import (2.38s)
=== RUN   TestAccGoogleOrganizationPolicy_boolean_enforced
--- PASS: TestAccGoogleOrganizationPolicy_boolean_enforced (1.70s)
=== RUN   TestAccGoogleOrganizationPolicy_boolean_notEnforced
--- PASS: TestAccGoogleOrganizationPolicy_boolean_notEnforced (1.80s)
=== RUN   TestAccGoogleOrganizationPolicy_boolean_update
--- PASS: TestAccGoogleOrganizationPolicy_boolean_update (4.30s)
=== RUN   TestAccGoogleOrganizationPolicy_list_allowAll
--- PASS: TestAccGoogleOrganizationPolicy_list_allowAll (2.20s)
=== RUN   TestAccGoogleOrganizationPolicy_list_allowSome
--- PASS: TestAccGoogleOrganizationPolicy_list_allowSome (1.80s)
=== RUN   TestAccGoogleOrganizationPolicy_list_denyAll
--- PASS: TestAccGoogleOrganizationPolicy_list_denyAll (1.80s)
=== RUN   TestAccGoogleOrganizationPolicy_list_denySome
--- PASS: TestAccGoogleOrganizationPolicy_list_denySome (1.90s)
=== RUN   TestAccGoogleOrganizationPolicy_list_update
--- PASS: TestAccGoogleOrganizationPolicy_list_update (4.84s)
PASS

TF_ACC=1 go test ./google -v -run TestAccGoogleFolder_rename -timeout 120m
=== RUN   TestAccGoogleFolder_rename
--- PASS: TestAccGoogleFolder_rename (16.88s)
PASS
```

rosbo:
I wrote the preliminary documentation but wanted to first get an agreement on the schema structure before spending more time finishing it.

Specifically for List Policy. There is many ways we could structure the schema. The API structure is not user-friendly so I would rather not follow it.

Option A (the one I implemented)

```hcl
... nested
 list_policy {
  allow {
    values = [
      "cloudresourcemanager.googleapis.com",
      "compute.googleapis.com",
    ]
  }
}

list_policy {
  deny {
    all = true
  }
}
...
```

Option B
```hcl
... nested
list_policy {
  allow_values = [
      "cloudresourcemanager.googleapis.com",
      "compute.googleapis.com",
    ]
  }
}

list_policy {
  deny_all = true
}
...
```

google/resource_google_organization_policy_test.go
```diff
@@ -0,0 +1,311 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGoogleOrganizationPolicy_listAll(org, "allow"),
+				Check:  testAccCheckGoogleOrganizationListPolicyAll("listAll", "ALLOW"),
+			},
+		},
+	})
+}
+
+func TestAccGoogleOrganizationPolicy_list_allowSome(t *testing.T) {
```
rosbo: this test and the deny tests could have side effects with other tests running in parallel

google/resource_google_organization_policy.go
```diff
@@ -0,0 +1,327 @@
...
+	listPolicy := configured[0].(map[string]interface{})
+	return &cloudresourcemanager.ListPolicy{
+		AllValues:      allValues,
+		AllowedValues:  allowedValues,
+		DeniedValues:   deniedValues,
+		SuggestedValue: listPolicy["suggested_value"].(string),
+	}, nil
+}
+
+func canonicalOrgPolicyConstraint(constraint string) string {
```
danawillow: if we allow people to set a non-canonical version, wouldn't that lead to diffs next plan once the constraint comes back from the api with a canonical version? Should be an easy enough fix with a diffsuppress

google/resource_google_organization_policy_test.go
```diff
@@ -0,0 +1,311 @@
...
+			continue
+		}
+
+		org := "organizations/" + rs.Primary.Attributes["org_id"]
+		constraint := rs.Primary.Attributes["constraint"]
+		_, err := config.clientResourceManager.Organizations.GetOrgPolicy(org, &cloudresourcemanager.GetOrgPolicyRequest{
+			Constraint: constraint,
+		}).Do()
+
+		if err != nil {
```
danawillow: wait, shouldn't this be the opposite? I'd expect a nil error to mean it exists.

google/resource_google_organization_policy.go
```diff
@@ -0,0 +1,327 @@
...
+				ConflictsWith: []string{"boolean_policy"},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"allow": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"all": {
```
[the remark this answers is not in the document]
rosbo: Right, I thought ConflictsWith didn't work with list but when there is only one item in the list, it does. I will fix this.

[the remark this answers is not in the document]
rosbo: Done

google/resource_google_organization_policy.go
```diff
@@ -0,0 +1,327 @@
...
+	if len(configured) == 0 {
+		return nil, nil
+	}
+
+	listPolicyMap := configured[0].(map[string]interface{})
+
+	allow := listPolicyMap["allow"].([]interface{})
+	deny := listPolicyMap["deny"].([]interface{})
+
+	if len(allow) > 0 && len(deny) > 0 {
```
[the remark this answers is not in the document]
rosbo: Done

google/resource_google_organization_policy.go
```diff
@@ -0,0 +1,327 @@
...
+		return nil, fmt.Errorf("You cannot specified an `allow` and a `deny` block")
```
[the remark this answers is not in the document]
rosbo: Code has been removed

google/resource_google_organization_policy.go
rosbo: There is already a `DiffSuppress` func set in the schema.

google/resource_google_organization_policy_test.go
rosbo: Good catch. I fixed it. It basically always returns an org policy. If it has been cleared, both the boolean and list policy field are `nil`.

google/resource_google_organization_policy_test.go
rosbo: I have been able to do 2). However, I haven't found a way to test the denyAll without side-effects but knowing the implementation, if the allowAll works, it is very likely that the denyAll works also, it is not ideal but good enough.

Also, the same methods (`expandListOrganizationPolicy` and `flattenListOrganizationPolicy`) will be used for org policies on folders. Since I can create a new folder easily, I will be able to test the denyAll case without side effects on other tests running in parallel.

website/docs/r/google_organization_policy.html.markdown
```diff
@@ -0,0 +1,82 @@
...
+
+The following arguments are supported:
+
+* `org_id` - (Required) The numeric ID of the organization to set the policy for.
+
+* `constraint` - (Required) The name of the Constraint the Policy is configuring, for example, `serviceuser.services`. Check out the [complete list of available constraints](https://cloud.google.com/resource-manager/docs/organization-policy/understanding-constraints#available_constraints).
+
+* `version` - (Optional) Version of the Policy. Default version is 0.
+
+TODO(rosbo): Add boolean and list policy block once I get agreement on the schema format.
```
danawillow: ^

rosbo: Done

PR #488, opened 2017-09-28
Give addresses beta support, add internal addresses.

paddycarver:
Update the `google_compute_address` resource to take advantage of
differing API versions. Use this new functionality to support internal
reserved IP addresses. Fixes #366.

PR #525, opened 2017-10-04
Adding functionality for static internal addresses to be configured

### LINKED ISSUE #366: reserve static internal ip address
I would like the ability to reserve a static internal ip address as described here 
https://cloud.google.com/compute/docs/ip-addresses/reserve-static-internal-ip-address

elliottsam:
This adds the ability to reserve internal addresses and resolves #366

### DIFF
--- a/google/compute_operation.go
+++ b/google/compute_operation.go
@@ -93,6 +93,10 @@ func computeOperationWaitTime(config *Config, op *compute.Operation, project, ac
 	return nil
 }
 
+func computeBetaOperationWait(config *Config, op *computeBeta.Operation, project, activity string) error {
+	return computeBetaOperationWaitTime(config, op, project, activity, 4)
+}
+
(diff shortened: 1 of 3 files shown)


paddycarver:
This looks like a duplicate of #488.

elliottsam:
Oh my mistake would it be possible to add specifying the IP address to that one? As it seems to not allow that at the moment

paddycarver:
Ah, you're correct. I'd prefer to add that functionality to #488 (which I believe should only take a couple of modifications to the code--I'll do that) rather than go forward with this one, just because #488 properly supports the beta API rather than forcing all users of the `compute_address` resource to use the beta API even if they don't need to. How do you feel about this path forward?

elliottsam:
Yeah that seems like a good way to go with it. Thanks

PR #528, opened 2017-10-04
Add support for target_tcp_proxy

### LINKED ISSUE #363: Add google_compute_target_tcp_proxy resource
Add support for TCP Proxies to create global TCP load balancers.

API Docs: https://cloud.google.com/compute/docs/reference/latest/targetTcpProxies#resource

The `google_compute_global_forwarding_rule` may also need to be updated to accept the TCP proxy in addition to the HTTP and HTTPS proxies.

enxebre:
Fixes #363 

Introduces support google_compute_target_tcp_proxy

### DIFF
--- a/google/resource_compute_target_tcp_proxy.go
+++ b/google/resource_compute_target_tcp_proxy.go
@@ -1,0 +1,171 @@
+package google
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/compute/v1"
+)
+
+func resourceComputeTargetTcpProxy() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceComputeTargetTcpProxyCreate,
+		Read:   resourceComputeTargetTcpProxyRead,
+		Delete: resourceComputeTargetTcpProxyDelete,
+		Update: resourceComputeTargetTcpProxyUpdate,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"backend_service": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"proxy_header": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "NONE",
+			},
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"project": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"self_link": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceComputeTargetTcpProxyCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	proxy := &compute.TargetTcpProxy{
+		Name:    d.Get("name").(string),
+		Service: d.Get("backend_service").(string),
+	}
+
+	if v, ok := d.GetOk("description"); ok {
+		proxy.Description = v.(string)
+	}
+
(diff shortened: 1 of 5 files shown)


google/resource_compute_target_tcp_proxy.go
```diff
@@ -0,0 +1,171 @@
...
+				Optional: true,
+				Default:  "NONE",
+			},
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"id": &schema.Schema{
```
rosbo: Could you rename it to `proxy_id`? We are moving away from "id" field. See #399 for more details.

google/resource_compute_target_tcp_proxy.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return fmt.Errorf("Error creating TargetTcpProxy: %s", err)
+	}
+
+	err = computeOperationWait(config, op, project, "Creating Target Tcp Proxy")
+	if err != nil {
+		return err
+	}
+
+	d.SetId(proxy.Name)
```
rosbo: Should probably set this to proxy.Id. 

Can two separate proxies have the same name?

google/resource_compute_target_tcp_proxy.go
```diff
@@ -0,0 +1,171 @@
...
+		return err
+	}
+
+	proxy := &compute.TargetTcpProxy{
+		Name:    d.Get("name").(string),
+		Service: d.Get("backend_service").(string),
+	}
+
+	if v, ok := d.GetOk("description"); ok {
+		proxy.Description = v.(string)
```
rosbo: No need for the if statement around it. Move it inside `&compute.TargetTcpProxy{...}` above. `d.Get("description")` will return an empty string if not set which will yield the same behavior.

google/resource_compute_target_tcp_proxy.go
```diff
@@ -0,0 +1,171 @@
...
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	proxy := &compute.TargetTcpProxy{
+		Name:    d.Get("name").(string),
+		Service: d.Get("backend_service").(string),
```
rosbo: Have you forgot to set the `ProxyHeader` field?

google/resource_compute_target_tcp_proxy_test.go
```diff
@@ -0,0 +1,119 @@
...
+	target := fmt.Sprintf("ttcp-test-%s", acctest.RandString(10))
+	backend := fmt.Sprintf("ttcp-test-%s", acctest.RandString(10))
+	hc := fmt.Sprintf("ttcp-test-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeTargetTcpProxyDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
```
rosbo: How is this testing the update feature? You need at least two `TestStep`. Make sure to test the update with the proxyHeader field (only updateable field) otherwise it will do a Delete/Create operation which won't test the update method.

website/docs/r/compute_target_tcp_proxy.html.markdown
```diff
@@ -0,0 +1,72 @@
...
+
+* `project` - (Optional) The project in which the resource belongs. If it
+    is not provided, the provider project is used.
+
+## Attributes Reference
+
+In addition to the arguments listed above, the following computed attributes are
+exported:
+
+* `id` - A unique ID assigned by GCE.
```
rosbo: Rename to `proxy_id` like suggested in a comment above.

google/resource_compute_target_tcp_proxy.go
enxebre: At least through the UI you can not have more than one with the same name. I'd think I'd leave it with name which is also in line with the others target resources

enxebre:
Hey @rosbo thanks for your comments! proxy_id and tests should be fine now

rosbo:
Thanks Alberto for adding this new resource. Merging in!

```sh 
TF_ACC=1 go test ./google -v -run TestAccComputeTargetTcpProxy_ -timeout 120m
=== RUN   TestAccComputeTargetTcpProxy_basic
--- PASS: TestAccComputeTargetTcpProxy_basic (67.52s)
=== RUN   TestAccComputeTargetTcpProxy_update
--- PASS: TestAccComputeTargetTcpProxy_update (79.31s)
PASS
```

Issue #535, opened 2017-10-05
google_compute_autoscaler does not support google_compute_region_instance_group_manager


zbikmarc:
It looks like that newly added `google_compute_region_instance_group_manager` can not be used with auto scaling because `google_compute_autoscaler` has `zone` [parameter as required](https://www.terraform.io/docs/providers/google/r/compute_autoscaler.html#zone).

### Terraform Version
`Terraform-0.10.6`
`terraform-provider-google-1.0.1`

### Affected Resource(s)
- google_compute_autoscaler
- google_compute_region_instance_group_manager

Those two resources are connected with eachother

### Expected Behavior
It should be able to create `google_compute_region_instance_group_manager` with auto scaling enabled.

### Important Factoids
This feature is supported from Google Cloud console. Snippet from Google documentation:

> Regional managed instance groups also support autoscaling, internal load balancing, and external load balancing.

Example API call is provided under [this location](https://cloud.google.com/compute/docs/instance-groups/distributing-instances-with-regional-instance-groups#enabling_autoscaling)

### References
- https://www.terraform.io/docs/providers/google/r/compute_autoscaler.html#zone
- https://cloud.google.com/compute/docs/instance-groups/distributing-instances-with-regional-instance-groups
- https://cloud.google.com/compute/docs/instance-groups/distributing-instances-with-regional-instance-groups#enabling_autoscaling

rosbo:
We will need to create a new resource called `google_compute_region_autoscaler`. It will be very similar to `google_compute_autoscaler` but will take a region instead of a zone and call the region autoscaler API using `config.clientCompute.RegionAutoscalers.*`

PR #544, opened 2017-10-06
Add resource_compute_region_autoscaler

usererror:
# Why
Attempts to fix https://github.com/terraform-providers/terraform-provider-google/issues/535.

It seemed easy enough based on @rosbo 's comment and I think this works but please let me know if I've missed anything! I tried to copy the regular resource_compute_autoscaler as much as possible and luckily they both use the same underlying type.

# Tests
I ran acceptance tests pointing at one of my google projects and they all passed.

```shell
➜  google git:(region-autoscaler) ✗ TF_ACC=1 go test -v -run TestAccComputeRegionAutoscaler
=== RUN   TestAccComputeRegionAutoscaler_basic
--- PASS: TestAccComputeRegionAutoscaler_basic (70.90s)
=== RUN   TestAccComputeRegionAutoscaler_update
--- PASS: TestAccComputeRegionAutoscaler_update (83.31s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google    154.249s
```

### DIFF
--- a/google/resource_compute_region_autoscaler.go
+++ b/google/resource_compute_region_autoscaler.go
@@ -1,0 +1,372 @@
+package google
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/compute/v1"
+)
+
+func resourceComputeRegionAutoscaler() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceComputeRegionAutoscalerCreate,
+		Read:   resourceComputeRegionAutoscalerRead,
+		Update: resourceComputeRegionAutoscalerUpdate,
+		Delete: resourceComputeRegionAutoscalerDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+
+			"target": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"autoscaling_policy": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"min_replicas": &schema.Schema{
+							Type:     schema.TypeInt,
+							Required: true,
+						},
+
+						"max_replicas": &schema.Schema{
+							Type:     schema.TypeInt,
+							Required: true,
+						},
+
+						"cooldown_period": &schema.Schema{
+							Type:     schema.TypeInt,
+							Optional: true,
+							Default:  60,
+						},
+
+						"cpu_utilization": &schema.Schema{
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"target": &schema.Schema{
+										Type:     schema.TypeFloat,
+										Required: true,
+									},
+								},
+							},
+						},
+
+						"metric": &schema.Schema{
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"name": &schema.Schema{
+										Type:     schema.TypeString,
+										Required: true,
+									},
+									"target": &schema.Schema{
+										Type:     schema.TypeFloat,
+										Required: true,
+									},
+
+									"type": &schema.Schema{
+										Type:     schema.TypeString,
+										Required: true,
+									},
+								},
+							},
+						},
+
+						"load_balancing_utilization": &schema.Schema{
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"target": &schema.Schema{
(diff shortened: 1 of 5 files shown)


google/resource_compute_region_autoscaler.go
```diff
@@ -0,0 +1,372 @@
...
+
+	err = computeOperationWait(config, op, project, "Creating Autoscaler")
+	if err != nil {
+		return err
+	}
+
+	return resourceComputeRegionAutoscalerRead(d, meta)
+}
+
+func flattenRegionAutoscalingPolicy(policy *compute.AutoscalingPolicy) []map[string]interface{} {
```
usererror: I'm not sure this needs to be redefined for the regional autoscaler since they both use compute.Autoscaler... anyway. I didn't touch this function actually, just renamed it.

rosbo: You can delete this method and reuse the one in the non-regional version

google/resource_compute_region_autoscaler.go
```diff
@@ -0,0 +1,372 @@
...
+				Required: true,
+			},
+
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"autoscaling_policy": &schema.Schema{
```
rosbo: You can extract this &schema.Schema definition and reuse it for both autoscaler

google/resource_compute_region_autoscaler.go
```diff
@@ -0,0 +1,372 @@
...
+
+			"self_link": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func buildRegionAutoscaler(d *schema.ResourceData) (*compute.Autoscaler, error) {
```
rosbo: You can remove this method and use the `buildAutoscaler`. In fact, I think you are not even using it and already using the `buildAutoscaler`.

google/resource_compute_region_autoscaler.go
```diff
@@ -0,0 +1,372 @@
...
+		log.Printf("[WARN] Removing Autoscaler %q because it's gone", d.Get("name").(string))
+		d.SetId("")
+		return nil
+	}
+
+	regionUrl := strings.Split(scaler.Region, "/")
+	d.Set("self_link", scaler.SelfLink)
+	d.Set("name", scaler.Name)
+	d.Set("target", scaler.Target)
+	d.Set("region", regionUrl[len(regionUrl)-1])
```
rosbo: You can use the method `GetResourceNameFromSelfLink`

google/resource_compute_region_autoscaler.go
```diff
@@ -0,0 +1,372 @@
...
+
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"autoscaling_policy": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
```
rosbo: Add `MaxItems: 1` since we can only specify at most on `autoscaling_policy`. It is also not Optional since you must have at least one `autoscaling_policy`. Please add `Required: true`

google/resource_compute_region_autoscaler.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	// Optional fields
+	if v, ok := d.GetOk("description"); ok {
+		scaler.Description = v.(string)
+	}
+
+	aspCount := d.Get("autoscaling_policy.#").(int)
+	if aspCount != 1 {
+		return nil, fmt.Errorf("The autoscaler must have exactly one autoscaling_policy, found %d.", aspCount)
```
rosbo: You can remove this line once this requirement is enforced by the schema like suggested above. The advantage of enforcing this in the schema is that it fails at `terraform plan` instead of `terraform apply`.

google/resource_compute_region_autoscaler_test.go
```diff
@@ -0,0 +1,262 @@
+package google
```
rosbo: Can you also add import acceptance tests?

By convention, we put them in a separate file. See `import_compute_autoscaler_test.go"

usererror:
Thanks for the review Vincent! I fixed up the zonal autoscaler on your recommendation and got rid of the duplicate regional functions and both are much more readable now. I also added the import acceptance tests for the regional autoscaler and ran all the acceptance tests again. This was a pleasure to work on, very easy to test and work with!

# Acceptance Tests
```shell
➜  google git:(region-autoscaler) ✗ TF_ACC=1 go test -v -run TestAccComputeAutoscaler
=== RUN   TestAccComputeAutoscaler_importBasic
--- PASS: TestAccComputeAutoscaler_importBasic (69.92s)
=== RUN   TestAccComputeAutoscaler_basic
--- PASS: TestAccComputeAutoscaler_basic (68.78s)
=== RUN   TestAccComputeAutoscaler_update
--- PASS: TestAccComputeAutoscaler_update (81.43s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	220.148s
```

```shell
➜  google git:(region-autoscaler) ✗ TF_ACC=1 go test -v -run TestAccComputeRegionAutoscaler
=== RUN   TestAccComputeRegionAutoscaler_importBasic
--- PASS: TestAccComputeRegionAutoscaler_importBasic (70.69s)
=== RUN   TestAccComputeRegionAutoscaler_basic
--- PASS: TestAccComputeRegionAutoscaler_basic (70.94s)
=== RUN   TestAccComputeRegionAutoscaler_update
--- PASS: TestAccComputeRegionAutoscaler_update (83.86s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	225.508s
```

rosbo:
Awesome work! Thanks for adding support for this new resource.

PR #548, opened 2017-10-06
fix default for connection_draining_timeout_sec

mioi:
The default value for the connection_draining_timeout_sec arg was changed in commit 5cd3e1ec99398698b5caabe4e39a9306253d4496 from 0 to 300. Fixing the doc to reflect this change.


rosbo:
Thank you Mioi for fixing the docs :)

PR #565, opened 2017-10-10
Fix import for compute_route

rosbo:
Fixes one resource for #431 

* Reorder fields in schema for style consistency
* Add reusable ZonalFieldValue
* Import tests were not testing the right thing. Import was broken. I fixed the import.
* Read state from the API
* Generate network link without calling the API

### DIFF
--- a/google/field_helpers.go
+++ b/google/field_helpers.go
@@ -6,8 +6,11 @@ import (
 )
 
 const (
-	globalLinkTemplate    = "projects/%s/global/%s/%s"
-	globalLinkBasePattern = "projects/(.+)/global/%s/(.+)"
+	globalLinkTemplate          = "projects/%s/global/%s/%s"
+	globalLinkBasePattern       = "projects/(.+)/global/%s/(.+)"
+	zonalLinkTemplate           = "projects/%s/zones/%s/%s/%s"
+	zonalLinkBasePattern        = "projects/(.+)/zones/(.+)/%s/(.+)"
+	zonalPartialLinkBasePattern = "zones/(.+)/%s/(.+)"
 )
 
 // ------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


google/field_helpers.go
```diff
@@ -77,6 +82,86 @@ func parseGlobalFieldValue(resourceType, fieldValue, projectSchemaField string,
...
+// Parses a zonal field supporting 5 different formats:
+// - https://www.googleapis.com/compute/ANY_VERSION/projects/{my_project}/zones/{zone}/{resource_type}/{resource_name}
+// - projects/{my_project}/zones/{zone}/{resource_type}/{resource_name}
+// - zones/{zone}/{resource_type}/{resource_name}
+// - resource_name
+// - "" (empty string). RelativeLink() returns empty if isEmptyValid is true.
+//
+// If the project is not specified, it first tries to get the project from the `projectSchemaField` and then fallback on the default project.
+// If the zone is not specified, it takes the value of `zoneSchemaField`.
+
```
selmanj: empty space

google/field_helpers.go
```diff
@@ -77,6 +82,86 @@ func parseGlobalFieldValue(resourceType, fieldValue, projectSchemaField string,
...
+func parseZonalFieldValue(resourceType, fieldValue, projectSchemaField, zoneSchemaField string, d TerraformResourceData, config *Config, isEmptyValid bool) (*ZonalFieldValue, error) {
```
selmanj: This function has 7 arguments. Is it doing too much? You might consider breaking it down into smaller functions or somehow grouping related attributes together.  Up to you if you think that's simpler

google/field_helpers.go
```diff
@@ -6,8 +6,11 @@ import (
 )
 
 const (
-	globalLinkTemplate    = "projects/%s/global/%s/%s"
-	globalLinkBasePattern = "projects/(.+)/global/%s/(.+)"
+	globalLinkTemplate          = "projects/%s/global/%s/%s"
+	globalLinkBasePattern       = "projects/(.+)/global/%s/(.+)"
```
selmanj: I still don't enjoy seeing a mixture of regexps and formatting patterns (when reading this it's not obvious to me whether the regexp is matched before or after the print). However if this ends up reducing the amount of code needed (which I'm guessing it does given the large number of constants) then I'll defer to your judgement

Alternatively, consider inlining these (especially if they're only used once). Then the formatting/regexp matching is much more explicit

google/field_helpers.go
```diff
@@ -77,6 +82,86 @@ func parseGlobalFieldValue(resourceType, fieldValue, projectSchemaField string,
...
+		}, nil
+	}
+
+	project, err := getProjectFromSchema(projectSchemaField, d, config)
+	if err != nil {
+		return nil, err
+	}
+
+	r = regexp.MustCompile(fmt.Sprintf(zonalPartialLinkBasePattern, resourceType))
+	if r.MatchString(fieldValue) {
```
selmanj: You might be able to collapse this and the following line via:
```go
if parts := r.FindStringSubmatch(fieldValue); parts != nil {
   // stuff happens
}
```

google/field_helpers.go
rosbo: `*Template` is used in more than one place.
`*BasePattern` is used in only one place but I like to have both next to each other. It is easy to catch a mistake.

google/field_helpers.go
rosbo: Removed

google/field_helpers.go
rosbo: Then entry point would still require all the parameters. The logic using those parameters is extracted away in method like `getProjectFromSchema`.

google/field_helpers.go
rosbo: Done

google/field_helpers.go
selmanj: Well, not exactly. Some of these parameters aren't "request-level" (e.g. they don't change inside the specific resource's code) and could be extracted into a more general constructor. You could imagine something like

```go
// Defined at top of resource
var zoneParser := ZoneParser{ResourceType: "networks", IsEmptyValid: false}

func myResourceCreate(d *schema.ResourceData, meta interface{}) {
    // a bunch of stuff happens here
    // [...]

    zoneParser.parse(myFieldValue, "foo_field", "project", d, config)
}
```

Similar to how a regex's pattern is defined globally ahead of where it's actually used. But it's up to you; it may be this representation is uglier.

PR #569, opened 2017-10-10
Add new resource google_compute_target_ssl_proxy 

### LINKED ISSUE #62: Missing Google resource: Target SSL Proxy
_This issue was originally opened by @maikzumstrull as hashicorp/terraform#12422. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

There is a resource on Google Cloud Platform called a [Target SSL Proxy](https://cloud.google.com/compute/docs/reference/latest/targetSslProxies/insert). It seems Terraform currently cannot create/manage these proxies.

Please add an appropriate resource, probably `google_compute_target_ssl_proxy`.

rosbo:
Fixes #62 

```sh
=== RUN   TestAccComputeTargetSslProxy_import
--- PASS: TestAccComputeTargetSslProxy_import (66.96s)
=== RUN   TestAccComputeTargetSslProxy_basic
--- PASS: TestAccComputeTargetSslProxy_basic (66.86s)
=== RUN   TestAccComputeTargetSslProxy_update
=== RUN   TestAccComputeTargetSslProxy_update
--- PASS: TestAccComputeTargetSslProxy_update (110.85s)
```

### DIFF
--- a/google/resource_compute_target_ssl_proxy.go
+++ b/google/resource_compute_target_ssl_proxy.go
@@ -1,0 +1,247 @@
+package google
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/compute/v1"
+)
+
+func resourceComputeTargetSslProxy() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceComputeTargetSslProxyCreate,
+		Read:   resourceComputeTargetSslProxyRead,
+		Delete: resourceComputeTargetSslProxyDelete,
+		Update: resourceComputeTargetSslProxyUpdate,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"backend_service": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"ssl_certificates": &schema.Schema{
+				Type:     schema.TypeList,
+				Required: true,
+				Elem: &schema.Schema{
+					Type:             schema.TypeString,
+					DiffSuppressFunc: compareSelfLinkOrResourceName,
+				},
+			},
+
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"proxy_header": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "NONE",
+			},
+
+			"project": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"proxy_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"self_link": &schema.Schema{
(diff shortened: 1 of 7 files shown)


google/resource_compute_target_ssl_proxy.go
```diff
@@ -0,0 +1,247 @@
...
+	err = computeOperationWait(config, op, project, "Deleting Target SSL Proxy")
+	if err != nil {
+		return err
+	}
+
+	d.SetId("")
+	return nil
+}
+
+func expandSslCertificates(configured []interface{}, d *schema.ResourceData, config *Config) ([]string, error) {
```
selmanj: This function would be a bit simpler if it didn't take the `configured` parameter (which is of a very confusing type). Instead you can use the `d` parameter to extract out the `ssl_certificates` field

google/resource_compute_target_ssl_proxy.go
```diff
@@ -0,0 +1,247 @@
...
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	d.Partial(true)
+
+	if d.HasChange("proxy_header") {
+		proxy_header := d.Get("proxy_header").(string)
```
selmanj: I hate to be that guy but this is using snake case rather than camel case.

It's totally readable to me but I imagine others might find it bad style

google/resource_compute_target_ssl_proxy.go
```diff
@@ -0,0 +1,247 @@
...
+		proxy_header_payload := &compute.TargetSslProxiesSetProxyHeaderRequest{
```
selmanj: Same here

google/resource_compute_target_ssl_proxy_test.go
```diff
@@ -0,0 +1,189 @@
...
+		found, err := config.clientCompute.TargetSslProxies.Get(
+			config.Project, rs.Primary.ID).Do()
+		if err != nil {
+			return err
+		}
+
+		if found.Name != rs.Primary.ID {
+			return fmt.Errorf("TargetSslProxy not found")
+		}
+
```
selmanj: Do we want to do more testing here (e.g. that the stored parameters match the set parameters)?

rosbo: Done

google/resource_compute_target_ssl_proxy_test.go
```diff
@@ -103,6 +104,19 @@ func testAccCheckComputeTargetSslProxyExists(n string) resource.TestCheckFunc {
...
+		if found.ProxyHeader != proxyHeader {
+			return fmt.Errorf("Wrong proxy header. Expected '%s', got '%s'", proxyHeader, found.ProxyHeader)
+		}
+
+		foundCertsName := make([]string, 0, len(found.SslCertificates))
+		for _, foundCert := range found.SslCertificates {
+			foundCertsName = append(foundCertsName, GetResourceNameFromSelfLink(foundCert))
+		}
+
+		if !reflect.DeepEqual(foundCertsName, sslCerts) {
```
selmanj: Does order matter? If it doesn't, you might consider sorting or something before doing the deepEqual

rosbo: Actually, after waiting for the full test run to finish and reading the API docs, one and only one certificate must be specified at this time (they might support more in the future). So I added a `MaxItems:1` on the schema for now and simplified this check.

From the docs:
"Currently exactly one SslCertificate resource must be specified."

PR #570, opened 2017-10-10
Add label support to google_compute_snapshot

selmanj:
Add labels for some resource in #17.

### DIFF
--- a/google/resource_compute_snapshot.go
+++ b/google/resource_compute_snapshot.go
@@ -125,7 +125,7 @@ func resourceComputeSnapshotCreate(d *schema.ResourceData, meta interface{}) err
 	// It probably maybe worked, so store the ID now
 	d.SetId(snapshot.Name)
 
-	err = computeOperationWait(config, op, project, "Creating Snapshot")
+	err = computeOperationWait(config.clientCompute, op, project, "Creating Snapshot")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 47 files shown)


google/resource_compute_snapshot.go
```diff
@@ -116,6 +129,29 @@ func resourceComputeSnapshotCreate(d *schema.ResourceData, meta interface{}) err
...
 		return err
 	}
+
+	// Now if labels are set, go ahead and apply them
+	if labels := expandLabels(d); len(labels) > 0 {
+		// First, read the remote resource in order to find the fingerprint
+		apiSnapshot, err := config.clientCompute.Snapshots.Get(project, d.Id()).Do()
+		if err != nil {
+			return fmt.Errorf("Eror when reading snapshot for label update: %s", err)
+		}
```
rosbo: 1. Call `d.Set("label_fingerprint", apiSnapshot.LabelFingerprint)` here
1. Extract line 141-153 into a private method call `updateSnapshotLabel` or something along those lines.
1. Reuse the `updateSnapshotLabel` for the create and update method.

selmanj: 1. I don't know if it makes sense to set the fingerprint here, as this fingerprint is immediately out of date once the following set occurs. A read will occur later that puts the correct fingerprint value.
2. Done.
3. Done.

google/resource_compute_snapshot.go
```diff
@@ -112,10 +125,25 @@ func resourceComputeSnapshotCreate(d *schema.ResourceData, meta interface{}) err
...
+
+		// HERE
```
rosbo: Not sure why there is this `HERE` comment

selmanj: oops!

PR #522, opened 2017-10-03
Fix BackendService group hash when instance groups use beta features

danawillow:
Fixes #462.

The issue in #462 was that `backend.group` takes a self link of an instance group. Instance Group Managers have a beta field, `auto_healing_policies`. If that is set, then its self link will be the beta one since we have to read from the beta API in order to read that field. In that case, the example config ended up such that state had the v1 self link (because it reads from the v1 BackendServices API) but plan wanted to change it to a beta self link (because it reads from the beta InstanceGroupManagers API).

The link of the instance group is one of the factors that makes up the hash for the backend. In order to make sure changes in api version don't impact the backend hash, I've changed it to use the relative path instead of the full self link. Since this changes the value of the hash, I added a state migration.

While I was in the neighborhood, I also changed `group` to Required since it'll fail server-side if it's not set (and our documentation actually says it's Required) and added a self link DiffSuppress for it.

```
$ make testacc TEST=./google TESTARGS='-run=TestComputeBackendServiceMigrateState'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./google -v -run=TestComputeBackendServiceMigrateState -timeout 120m
=== RUN   TestComputeBackendServiceMigrateState
2017/10/03 16:46:42 [INFO] Found Compute Backend Service State v0; migrating to v1
2017/10/03 16:46:42 [DEBUG] Attributes before migration: map[string]string{"backend.242332812.balancing_mode":"UTILIZATION", "backend.242332812.max_utilization":"0.8", "backend.#":"1", "backend.242332812.group":"https://www.googleapis.com/compute/v1/projects/project_name/zones/zone_name/instances/instanceGroups/igName"}
2017/10/03 16:46:42 [DEBUG] Attributes after migration: map[string]string{"backend.2573491210.balancing_mode":"UTILIZATION", "backend.2573491210.max_utilization":"0.8", "backend.2573491210.group":"https://www.googleapis.com/compute/v1/projects/project_name/zones/zone_name/instances/instanceGroups/igName", "backend.#":"1"}
--- PASS: TestComputeBackendServiceMigrateState (0.00s)
=== RUN   TestComputeBackendServiceMigrateState_empty
2017/10/03 16:46:42 [DEBUG] Empty InstanceState; nothing to migrate.
2017/10/03 16:46:42 [DEBUG] Empty InstanceState; nothing to migrate.
--- PASS: TestComputeBackendServiceMigrateState_empty (0.00s)
PASS
[...]

PR #576, opened 2017-10-12
Change group back to Optional in `google_compute_backend_service`

danawillow:
#522 changed group to Required, since it would fail at apply-time anyway if it isn't set. However, this actually breaks our published modules (https://github.com/hashicorp/terraform/issues/16306). To avoid that breakage for the next release, I changed it back to Optional and while I was there, added some extra checks to ensure it's actually set so it fails with an actual error message instead of a panic.

```
$ make testacc TEST=./google TESTARGS='-run=TestAccComputeBackendService'
TF_ACC=1 go test ./google -v -run=TestAccComputeBackendService -timeout 120m
=== RUN   TestAccComputeBackendService_importBasic
--- PASS: TestAccComputeBackendService_importBasic (44.82s)
=== RUN   TestAccComputeBackendService_basic
--- PASS: TestAccComputeBackendService_basic (66.33s)
=== RUN   TestAccComputeBackendService_withBackend
--- PASS: TestAccComputeBackendService_withBackend (91.32s)
=== RUN   TestAccComputeBackendService_withBackendAndUpdate
--- PASS: TestAccComputeBackendService_withBackendAndUpdate (113.24s)
=== RUN   TestAccComputeBackendService_updatePreservesOptionalParameters
--- PASS: TestAccComputeBackendService_updatePreservesOptionalParameters (54.94s)
=== RUN   TestAccComputeBackendService_withConnectionDraining
--- PASS: TestAccComputeBackendService_withConnectionDraining (43.62s)
=== RUN   TestAccComputeBackendService_withConnectionDrainingAndUpdate
--- PASS: TestAccComputeBackendService_withConnectionDrainingAndUpdate (55.09s)
=== RUN   TestAccComputeBackendService_withHttpsHealthCheck
--- PASS: TestAccComputeBackendService_withHttpsHealthCheck (43.17s)
=== RUN   TestAccComputeBackendService_withCDNEnabled
--- PASS: TestAccComputeBackendService_withCDNEnabled (43.51s)
=== RUN   TestAccComputeBackendService_withSessionAffinity
--- PASS: TestAccComputeBackendService_withSessionAffinity (55.01s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	611.229s
```

### DIFF
--- a/google/resource_compute_backend_service.go
+++ b/google/resource_compute_backend_service.go
@@ -342,6 +342,7 @@ func expandBackendService(d *schema.ResourceData) (*compute.BackendService, erro
 		HealthChecks: healthChecks,
 	}
 
+	var err error
 	if v, ok := d.GetOk("backend"); ok {
 		service.Backends, err = expandBackends(v.(*schema.Set).List())
 		if err != nil {
--- a/google/resource_compute_backend_service.go
+++ b/google/resource_compute_backend_service.go
@@ -2,6 +2,7 @@ package google
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"log"
 
@@ -44,7 +45,7 @@ func resourceComputeBackendService() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"group": &schema.Schema{
 							Type:             schema.TypeString,
-							Required:         true,
+							Optional:         true,
 							DiffSuppressFunc: compareSelfLinkRelativePaths,
 						},
 						"balancing_mode": &schema.Schema{
@@ -150,7 +151,10 @@ func resourceComputeBackendService() *schema.Resource {
 func resourceComputeBackendServiceCreate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	service := expandBackendService(d)
+	service, err := expandBackendService(d)
+	if err != nil {
+		return err
+	}
 
 	project, err := getProject(d, config)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


danawillow:
(not actually looking for 3 reviews, just whichever of you get to it first)

google/resource_compute_backend_service.go
```diff
@@ -263,14 +270,19 @@ func resourceComputeBackendServiceDelete(d *schema.ResourceData, meta interface{
...
-func expandBackends(configured []interface{}) []*compute.Backend {
+func expandBackends(configured []interface{}) ([]*compute.Backend, error) {
 	backends := make([]*compute.Backend, 0, len(configured))
 
 	for _, raw := range configured {
 		data := raw.(map[string]interface{})
 
+		g, ok := data["group"]
+		if !ok {
+			return nil, errors.New("google_compute_backend_service.backend.group must be set")
```
paddycarver: This resource address may not actually be right, if the user names their backend something other than "backend". I'd prefer something along the lines of `fmt.Errorf("group field on backend %q must be set", backend.SelfLink)`, but that's just me.

danawillow: the address wouldn't contain the name of the backend though, that would be the value. the address would technically be something like `google_compute_backend_service.backend.<hash>.group` but I figured omitting the hash was fine in this case. We can't use the self link because there isn't a `compute.Backend` object to take the self link of

danawillow: I could use the index, I guess. Would that work better?

paddycarver: Ah, no, I'm just confused. Ignore me, nitpick withdrawn. :)

PR #577, opened 2017-10-12
Fix cluster version upgrades

danawillow:
This PR does two things (each in a separate commit):
1. Adds a retry when reading the cluster to make sure it's in the RUNNING state. This (sort of) solves the issue of trying to upgrade a cluster while it's already upgrading. Though it would make sense to have the retry be on the upgrade, I don't want users to get confused when they run plan, see a certain version, and then run upgrade and find out they were at a different version. Let me know your thoughts on whether this should actually be in read or in update (or in both).

2. Adds a new field, `min_master_version` and sets `master_version` to computed. As per discussions with @rosbo and @paddycarver, having just one `master_version` field doesn't work when the master gets upgraded automatically. This allows the users to set the minimum master version while also using the current master version for interpolation in other resources.

@paddycarver is in charge of running tests and adding documentation because I have to go catch my bus :)

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
 	"google.golang.org/api/container/v1"
@@ -435,8 +436,18 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 
 	zoneName := d.Get("zone").(string)
 
-	cluster, err := config.clientContainer.Projects.Zones.Clusters.Get(
-		project, zoneName, d.Get("name").(string)).Do()
+	var cluster *container.Cluster
+	err = resource.Retry(1*time.Minute, func() *resource.RetryError {
+		cluster, err = config.clientContainer.Projects.Zones.Clusters.Get(
+			project, zoneName, d.Get("name").(string)).Do()
+		if err != nil {
+			return resource.NonRetryableError(err)
+		}
+		if cluster.Status != "RUNNING" {
+			return resource.RetryableError(fmt.Errorf("Cluster %q has status %q with message %q", d.Get("name"), cluster.Status, cluster.StatusMessage))
+		}
+		return nil
+	})
 	if err != nil {
 		return handleNotFoundError(err, d, fmt.Sprintf("Container Cluster %q", d.Get("name").(string)))
 	}
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	version "github.com/hashicorp/go-version"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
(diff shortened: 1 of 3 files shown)


paddycarver:
I'm seeing this still fail the updateVersion test. I'm going to look into why/what it takes to fix it.

paddycarver:
This appears to be happening because `get-server-config` is returning `1.7.6-gke.1` as the most recent available version for _nodes_, but then claiming it's not a valid version?

paddycarver:
Switching from the latest and 2nd-latest master/node versions in the updateVersion to the 2nd-latest and 3rd-latest resolves the issue, which makes me suspect it really is a problem with the (mistakenly?) reported latest node version. @danawillow do we want to:

* merge this as-is and release even though the test fails?
* wait for the API to be fixed?
* update the test to use the 2nd/3rd latest?

I'm updating the docs for this now, but I don't think I can push to the branch this is on. I'll push to the paddy_577 branch on this repo.

google/resource_container_cluster.go
```diff
@@ -508,8 +528,8 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 	d.Partial(true)
 
 	// The master must be updated before the nodes
-	if d.HasChange("master_version") {
-		desiredMasterVersion := d.Get("master_version").(string)
+	if d.HasChange("min_master_version") {
+		desiredMasterVersion := d.Get("min_master_version").(string)
```
paddycarver: This doesn't take into consideration if the current `master_version` is already greater than `min_master_version`. I think we want to ignore the version in that case, instead of downgrading?

paddycarver:
Sorry for the lots-of-comments on this one. :(

> Let me know your thoughts on whether this should actually be in read or in update (or in both).

I'm in favor of doing it in read. That also ensures that other things that may depend on the cluster being in a RUNNING state _also_ get the benefit of this.

google/resource_container_cluster.go
```diff
@@ -302,12 +307,17 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
...
-	if v, ok := d.GetOk("master_version"); ok {
+	if v, ok := d.GetOk("min_master_version"); ok {
 		cluster.InitialClusterVersion = v.(string)
 	}
 
-	if _, ok := d.GetOk("node_version"); ok {
-		return fmt.Errorf("cannot set node_version on create, use master_version instead")
+	if v, ok := d.GetOk("node_version"); ok {
+		// ignore -gke.X suffix for now. if it becomes a problem later, we can fix it.
+		mv := strings.Split(cluster.InitialClusterVersion, "-")[0]
```
rosbo: If `min_master_version` is not specified, then `cluster.InitialClusterVersion` will be empty right? This means `mv` will be an empty string. Is that what we want? Does that mean that if we don't set `min_master_version`, we should also leave `node_version` empty?

danawillow:
@paddycarver yeah, that's the bug I was talking about where I spoke to the person and it'll be fixed soon. In the meantime I just went ahead and changed it to the 2nd/third latest.

google/resource_container_cluster.go
danawillow: Good point, fixed.

google/resource_container_cluster.go
danawillow: That's what @paddycarver and I ended up deciding on yesterday but now I can't remember for sure why. I went and changed it because it seems reasonable to just specify the `node_version` (since that's what the nodes will be set to anyway) and let GKE handle the master.

danawillow:
Added docs based on @paddycarver's commit- I changed the wording around a little but it's the same basic idea.

```
$ make testacc TEST=./google TESTARGS='-run=TestAccContainerCluster_basic'
TF_ACC=1 go test ./google -v -run=TestAccContainerCluster_basic -timeout 120m
--- PASS: TestAccContainerCluster_basic (366.49s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	366.677s
```
```
$ make testacc TEST=./google TESTARGS='-run=TestAccContainerCluster_withVersion'
TF_ACC=1 go test ./google -v -run=TestAccContainerCluster_withVersion -timeout 120m
=== RUN   TestAccContainerCluster_withVersion
--- PASS: TestAccContainerCluster_withVersion (367.59s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	367.756s
```
```
$ make testacc TEST=./google TESTARGS='-run=TestAccContainerCluster_updateVersion'
TF_ACC=1 go test ./google -v -run=TestAccContainerCluster_updateVersion -timeout 120m
=== RUN   TestAccContainerCluster_updateVersion
--- PASS: TestAccContainerCluster_updateVersion (690.49s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	690.673s
```

danawillow:
Ok, I remembered why we weren't going to allow setting the `node_version` on create. `InitialNodeVersion` only accepts valid master versions. Changed it back.

PR #584, opened 2017-10-13
Fix autoscaler and org policies docs

rosbo:
* Add region autoscaler docs
* Change link to API docs in autoscaler docs
* Add autoscaler and org policies documentation to the side menu

### DIFF
--- a/website/docs/r/compute_region_autoscaler.html.markdown
+++ b/website/docs/r/compute_region_autoscaler.html.markdown
@@ -1,0 +1,155 @@
+---
+layout: "google"
+page_title: "Google: google_compute_region_autoscaler"
+sidebar_current: "docs-google-compute-region-autoscaler"
+description: |-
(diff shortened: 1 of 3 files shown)


website/docs/r/compute_region_autoscaler.html.markdown
```diff
@@ -0,0 +1,155 @@
...
+    scopes = ["userinfo-email", "compute-ro", "storage-ro"]
+  }
+}
+
+resource "google_compute_target_pool" "foobar" {
+  name = "foobar"
+}
+
+resource "google_compute_region_instance_group_manager" "foobar" {
+  name = "foobar"
```
danawillow: Can you line up the `=`s here and below?

rosbo: Done

PR #591, opened 2017-10-16
feat: allow subnet link lookup via field, project and region

gsdevme:
`resource_compute_forwarding_rule` doesn't support a string configuration, its not really ideal as Id like to have the network layer of Terraform separate.

I've functionally tested this and its working. I feel like long term `getSubnetworkLink` might need refactoring because it assumes Zonal configuration to get the region rather than just having region as an argument.

It seems like `getSubnetworkLink` should actually be called something else and more specific to its use case.. for now though Ive done this however

Without this change you get the following error.
```bash
google_compute_forwarding_rule.default: Error creating ForwardingRule: googleapi: Error 400: Invalid value for field 'resource.subnetwork': 'xxxx'. The URL is malformed., invalid
```

### DIFF
--- a/google/resource_compute_forwarding_rule.go
+++ b/google/resource_compute_forwarding_rule.go
@@ -146,6 +146,12 @@ func resourceComputeForwardingRuleCreate(d *schema.ResourceData, meta interface{
 		ports = append(ports, v.(string))
 	}
 
+	subnetwork, err := getSubnetworkLinkWithRegionAndProject(d, config, region, project)
+
+	if err != nil {
+		return err
+	}
+
 	frule := &compute.ForwardingRule{
 		BackendService:      d.Get("backend_service").(string),
 		IPAddress:           d.Get("ip_address").(string),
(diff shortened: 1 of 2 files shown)


gsdevme:
Hi, is there anything I can do to speed up the merging? any suggestions?

danawillow:
Hey @gsdevme, we try to get to all PRs within a week and we're a bit shortstaffed this week, so I just ask for your patience. That being said, code reviews are #2 on my to-do list for today so I suspect I'll review it this afternoon :)

gsdevme:
Ok, sorry to bother. Just wanted to know if I could be any use. Thanks when you get the time.

danawillow:
Hey @gsdevme, thanks for the PR! We actually are just starting to standardize on how we allow fields to have either a name or a self_link value. Take a look at https://github.com/terraform-providers/terraform-provider-google/blob/master/google/field_helpers.go and it's usage (`ParseNetworkFieldValue` is already being used for `network` within `resource_compute_forwarding_rule.go`). You'd need to add a new one for regionFieldValues, but it would be a very welcome addition to the code base!

If that seems like it'll be too much work, we can check this in as-is, but it'll probably get overridden pretty soon. Either way I'd like to see a test (we actually don't have any that set a subnetwork right now, so it would be great to have one for setting by name and one for setting by URL)

ndmckinley:
I've just checked, and as @danawillow said, this no longer needs to be merged.  Thanks for the PR, though!

PR #592, opened 2017-10-16
Migrate boot disk initialize params

### LINKED ISSUE #555: Error in disk migration
Hi, we found an issue while migrating from terraform 9.6 to 10.7 with 1.0.1 google provider. What happens is that when we run `terraform refresh` or `terraform plan`  after` terraform init` the state isn't migrated. 

Instead this error occurs for every compute instance:

```
2017/10/09 12:35:04 [ERROR] root: eval: *terraform.EvalValidateResource, err: Warnings: []. Errors: ["disk": [REMOVED] Use boot_disk, scratch_disk, and attached_disk instead]
2017/10/09 12:35:04 [ERROR] root: eval: *terraform.EvalSequence, err: Warnings: []. Errors: ["disk": [REMOVED] Use boot_disk, scratch_disk, and attached_disk instead]
```

### Affected Resource(s)
- google_compute_instance

```
resource "google_compute_instance" "appserver" {
  name = "appserver"
  zone = "${var.region_zone}"
  disk {
    image = "${var.appserver_image}"
  }
  network_interface {
    network = "${var.network_name}"
    access_config = {
    }
  }
  tags = [
    "https",
    "http"]
}
```
We also tried to change disk property to 
``` 
boot_disk {
    initialize_params {
      image = "${var.appserver_image}"
    }
  }
```

but this forces new resource to be built. 
For us the most important thing is not to have to rebuild our VM's , how can we migrate to the new sintax without running into this problem ? 

Thanks !

### LINKED ISSUE #581: disk->boot_disk initialize_params not migrated for created instances.
### Terraform Version
`Terraform v0.10.7`

### Affected Resource(s)
- google_compute_instance

A "cluster instance" module file:

```
resource "google_compute_disk" "data-pd" {
  count = "${var.machine_count}"
  name = "${format("${var.disk_name_format}", count.index + 1)}"
  type = "${var.disk_type}"
  size = "${var.disk_size}"
  zone = "${var.fixed_zone ? var.fixed_zone_name : element(var.zones, count.index)}"
}

resource "google_compute_instance" "box" {
  depends_on = ["google_compute_disk.data-pd"]
  count = "${var.machine_count}"
  name = "${format("${var.machine_name_format}", count.index + 1)}"
  zone = "${element(var.zones, count.index)}"
  tags = "${var.tags}"

  boot_disk {
    // boot disk is always named exactly the same as the instance
    device_name = "${format("${var.machine_name_format}", count.index + 1)}"
    auto_delete = true
    initialize_params {
      size = 50
      image = "${var.image_name}"
    }
  }

  attached_disk {
    device_name = "fs-data-disk"
    source = "${element(google_compute_disk.data-pd.*.self_link, count.index)}"
  }

  network_interface {
    network = "default"
    access_config {
      // ephemeral IP
    }
  }

  service_account {
    scopes = "${var.oauth_scopes}"
  }

}

output "public_ips" {
  value = ["${google_compute_instance.box.*.network_interface.0.access_config.0.assigned_nat_ip}"]
}
```

### Expected Behavior
When instances previously created with custom image and size of the boot device that have both boot and attached disk defined with `disk{}` and `disk{}` are migrated for `boot_disk{}` and `attached_disk{}` those instances that have been previously created would not be expected to be flagged as needing to be recreated with all of the relevant information migrated over. 

### Actual Behavior
Instances that have been created from the module are seen as having to be recreated because the boot disk initialize_params have changed :

```
      boot_disk.0.initialize_params.#:                     "0" => "1" (forces new resource)
      boot_disk.0.initialize_params.0.image:               "" => "https://www.googleapis.com/compute/v1/projects/redacted/global/images/custom-image-4" (forces new resource)
```

### Important Factoids
Maybe not a Factoid, but more of a use case.

We have clusters of hosts (staging >5 <18 .. prod >32 <64) for services that are created from this base module. c01, c02 etc. etc. The size of these clusters can be adjusted as necessary by incrementing and decrementing the counter value in the .tf file that invokes the module for whichever cluster needs to be expanded or contracted.

If the custom image and size is left in the module with the boot_disk{} definition all the currently running instances are wanting to be recreated. If the values are removed this module can no longer be used to manage these service clusters and new modules must be created along with all of `count.index` offset to get instance names correct from last known state. 

Individual instances that have been created can have the `initialize_params{}` removed (even though it it nice to be able to look there for the config information), but the multi machine clusters built from a module pose an issue.

danawillow:
Even though initialize params don't matter after creation, if the config is being used for repeatable deployments then users want to be able to keep them in their configs.

Fixes #581 
Fixes #555


danawillow:
No, neither attached_disk nor scratch_disk have initialize_params.

PR #594, opened 2017-10-17
Give addresses beta support, add internal addresses

negz:
This PR (hopefully) ties together the existing work in #488. Specifically it:

* Adds support for specifying the internal IP address you wish to reserve per #525 
* Adds documentation for all the new knobs
* Addresses some PR feedback that came in after #488 was approved.
* Resolves some merge conflicts

I haven't contributed to Terraform before, but I'm pretty sure this works as expected:
```bash
$ GOOGLE_XPN_HOST_PROJECT=REDACTED \
   GOOGLE_PROJECT=REDACTED \
   GOOGLE_REGION=us-central1 \
   GOOGLE_USE_DEFAULT_CREDENTIALS=1 \
   TF_ACC=1 \
   go test -v . -run "^TestAccComputeAddress_internal"
=== RUN   TestAccComputeAddress_internal
--- PASS: TestAccComputeAddress_internal (114.13s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 114.153s
```

### DIFF
--- a/google/resource_compute_address.go
+++ b/google/resource_compute_address.go
@@ -17,6 +17,7 @@ var (
 	computeAddressLinkRegex  = regexp.MustCompile("projects/(.+)/regions/(.+)/addresses/(.+)$")
 	AddressBaseApiVersion    = v1
 	AddressVersionedFeatures = []Feature{
+		{Version: v0beta, Item: "address"},
 		{Version: v0beta, Item: "address_type"},
 		{Version: v0beta, Item: "subnetwork"},
 	}
@@ -60,8 +61,12 @@ func resourceComputeAddress() *schema.Resource {
 				DiffSuppressFunc: linkDiffSuppress,
 			},
 
+			// address will be computed unless it is specified explicitly.
+			// address may only be specified for the INTERNAL address_type.
 			"address": &schema.Schema{
 				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
 				Computed: true,
 			},
 
@@ -105,6 +110,10 @@ func resourceComputeAddressCreate(d *schema.ResourceData, meta interface{}) erro
 		Name:        d.Get("name").(string),
 		AddressType: d.Get("address_type").(string),
 		Subnetwork:  d.Get("subnetwork").(string),
+	}
+
+	if desired, ok := d.GetOk("address"); ok {
+		v0BetaAddress.Address = desired.(string)
 	}
 
 	var op interface{}
--- a/google/resource_compute_address.go
+++ b/google/resource_compute_address.go
@@ -12,13 +12,18 @@ import (
 	"google.golang.org/api/compute/v1"
 )
 
+const (
+	addressTypeExternal = "EXTERNAL"
+	addressTypeInternal = "INTERNAL"
+)
+
 var (
 	computeAddressIdTemplate = "projects/%s/regions/%s/addresses/%s"
 	computeAddressLinkRegex  = regexp.MustCompile("projects/(.+)/regions/(.+)/addresses/(.+)$")
 	AddressBaseApiVersion    = v1
 	AddressVersionedFeatures = []Feature{
 		{Version: v0beta, Item: "address"},
-		{Version: v0beta, Item: "address_type"},
+		{Version: v0beta, Item: "address_type", DefaultValue: addressTypeExternal},
 		{Version: v0beta, Item: "subnetwork"},
 	}
 )
@@ -48,9 +53,9 @@ func resourceComputeAddress() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
-				Default:  "EXTERNAL",
+				Default:  addressTypeExternal,
 				ValidateFunc: validation.StringInSlice(
-					[]string{"INTERNAL", "EXTERNAL"}, false),
+					[]string{addressTypeInternal, addressTypeExternal}, false),
 			},
 
 			"subnetwork": &schema.Schema{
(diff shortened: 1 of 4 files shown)


negz:
Ping. Is there anything I can do to help get this merged? :)

google/resource_compute_address.go
```diff
@@ -74,9 +105,30 @@ func resourceComputeAddressCreate(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	// Build the address parameter
-	addr := &compute.Address{Name: d.Get("name").(string)}
-	op, err := config.clientCompute.Addresses.Insert(
-		project, region, addr).Do()
+	v0BetaAddress := &computeBeta.Address{
+		Name:        d.Get("name").(string),
+		AddressType: d.Get("address_type").(string),
```
paddycarver: Echoing @rosbo's comment: we should explicitly document (or use `Default` to set) the default value here.

Source: https://github.com/terraform-providers/terraform-provider-google/pull/488/files#r142192254

google/resource_compute_address.go
```diff
@@ -97,17 +149,38 @@ func resourceComputeAddressCreate(d *schema.ResourceData, meta interface{}) erro
...
+			return err
+		}
+	case v0beta:
+		v0BetaAddr, err := config.clientComputeBeta.Addresses.Get(
+			addressId.Project, addressId.Region, addressId.Name).Do()
+		if err != nil {
+			return handleNotFoundError(err, d, fmt.Sprintf("Address %q", d.Get("name").(string)))
+		}
+
+		err = Convert(v0BetaAddr, addr)
```
paddycarver: @rosbo mentioned we don't actually need to convert here: https://github.com/terraform-providers/terraform-provider-google/pull/488/files#r142192681

negz:
@paddycarver Thanks for the review! Please take another look - comments should be addressed.

negz:
@paddycarver Whoops, my mistake! Should have run all the tests. I believe I've corrected everything you mentioned, but I added a new test `TestAccComputeAddress_importInternal` and can't seem to get it to work:

```
$ TF_ACC=1 go test -v -timeout=120m -run "^TestAccComputeAddress_.+" .
=== RUN   TestAccComputeAddress_importInternal
=== RUN   TestAccComputeAddress_internal
--- PASS: TestAccComputeAddress_internal (306.37s)
--- PASS: TestAccComputeAddress_importBasic (23.33s)
--- PASS: TestAccComputeAddress_basic (23.44s)
--- FAIL: TestAccComputeAddress_importInternal (317.17s)
        testing.go:434: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

                (map[string]string) (len=2) {
                 (string) (len=12) "address_type": (string) (len=8) "EXTERNAL",
                 (string) (len=10) "subnetwork": (string) ""
                }


                (map[string]string) (len=2) {
                 (string) (len=12) "address_type": (string) (len=8) "INTERNAL",
                 (string) (len=10) "subnetwork": (string) (len=121) "https://www.googleapis.com/compute/v1/projects/planet-k8s-staging/regions/us-east1/subnetworks/subnetwork-test-j1sb4wplcw"
                }

FAIL
exit status 1
FAIL    github.com/terraform-providers/terraform-provider-google/google 623.560s
```
It seems as if we're trying to use the `v1` API when importing `google_compute_address.internal_with_subnet` despite that resource having schema fields set that require `v0beta` - i.e. `subnetwork` and `address_type`.

Any idea how I can get this test to pass?

rosbo:
Update `AddressVersionedFeatures` to set the default value for the `address_type` field. 

See an example here:
https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_firewall.go#L23

negz:
@rosbo Your suggested change makes sense, thanks. 

I found that the test still failed after setting the default value for `address_type`, but passed after I also explicitly called `resourceComputeAddressRead` from within `resourceComputeAddressParseState`.

Without having dug into the mechanics of how Terraform imports work I had assumed something under the hood was triggering a read subsequent to the import function being called to determine the resource's ID, but it seems that's not the case?

I believe all integration tests matching `^TestAccComputeAddress_.+` are now passing.

paddycarver:
I'm :/ about explicitly calling Read, as that should be unnecessary, and is a bit of a weird situation--it should be automatically Read after calling Import, anyways, I believe? I looked into this some, and can't explain the behaviour--without your fix, I can't find anything that would set `EXTERNAL` in state the way the test is saying it is. @rosbo, any thoughts? I'll also ask a few other people on the Terraform team if they have any thoughts here, because this is decidedly odd.

negz:
@paddycarver Thanks for following up - glad I wasn't just missing something obvious.

negz:
@rosbo I've removed `address` from the beta fields, and removed the double-read bit.

Given that we've discovered imports are broken for any beta resources, how would you feel merging this PR (perhaps without the internal import acceptance tests?) and addressing the beta import issue separately?

rosbo:
This sounds reasonable to me. Please comment out the import acceptance tests and a quick comment explaining why the test is commented-out.

Thanks

negz:
@rosbo will do. Raised https://github.com/terraform-providers/terraform-provider-google/issues/694 to track the separate issue for now.

rosbo:
Thanks for filling the new issue :)

negz:
@rosbo I've commented out the offending tests:
```
$ go test -timeout=1200s -v . -run "^TestAccComputeAddress_.+$"
=== RUN   TestAccComputeAddress_internal
--- PASS: TestAccComputeAddress_internal (308.59s)
--- PASS: TestAccComputeAddress_importBasic (23.31s)
--- PASS: TestAccComputeAddress_basic (23.35s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 331.967s
```

PR #605, opened 2017-10-18
Allow setting boot and attached disk sources by name or self link

### LINKED ISSUE #460: Boot disk and attached disk should both take self link or name

### LINKED ISSUE #514: attached_disks only work if source is a self_link
When using `attached_disk` on a resource, if a `name` is used instead of a full `self_link`, the API response can't match it to the `attached_disk` field it belongs to, and tries to make it a `disk`. But disks don't exist anymore, so it just ignores it in the state. This manifests as a constant diff showing the need to update the attached disk.

A quick solution to this would be to update the matching that happens in `compute_instance`'s Read function to accept either a `self_link` _or_ a `name` and treat them as equivalent. A _better_ solution to this would probably involve ripping out the code that distinguished between `attached_disk` and `disk` entirely, and reworking the logic of that section, as we no longer need to match against `disk`.

In the meantime, users can workaround the problem by using the `self_link` instead of the `name` in their configs.

danawillow:
Branched off of #600 to make my life a bit easier, so review that one first :)

Fixes #460.
Fixes #514.

```
$ make testacc TEST=./google TESTARGS='-run=TestAccComputeInstance_attachedDisk -parallel 8'
TF_ACC=1 go test ./google -v -run=TestAccComputeInstance_attachedDisk -parallel 8 -timeout 120m
=== RUN   TestAccComputeInstance_attachedDisk
=== RUN   TestAccComputeInstance_attachedDisk_sourceUrl
--- PASS: TestAccComputeInstance_attachedDisk_sourceUrl (71.96s)
--- PASS: TestAccComputeInstance_attachedDisk (81.82s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	82.055s
```


danawillow:
Fixed a quick bug that I found when looking this over, I'm going to go ahead and assume the fix is ok and merge this :)

PR #612, opened 2017-10-21
Expand docs for auto_healing_policies for Instance Group Managers

danielcompton:
This should make it clearer what is required for setting up `auto_healing_policies`, especially that the `health_check` URL is a link to a resource, not to the health check URL directly.


danielcompton:
Fixed!

PR #616, opened 2017-10-21
Warn about specifying ip_version and ip_address together

danielcompton:


danielcompton:
I can't see any docs for it either, but I got an error when I tried to add `ip_version`:

```console
$ tfapply
google_compute_global_forwarding_rule.new-ipv4: Creating...
  ip_address:        "" => "<snip>"
  ip_protocol:       "" => "<computed>"
  ip_version:        "" => "IPV4"
  label_fingerprint: "" => "<computed>"
  name:              "" => "new-forwarding-rule"
  port_range:        "" => "80"
  self_link:         "" => "<computed>"
  target:            "" => "https://www.googleapis.com/compute/v1/projects/<snip>/global/targetHttpProxies/<snip>"
Error applying plan:

1 error(s) occurred:

* google_compute_global_forwarding_rule.new-ipv4: 1 error(s) occurred:

* google_compute_global_forwarding_rule.new-ipv4: Error creating Global Forwarding Rule: googleapi: Error 400: Invalid value for field 'resource.ipVersion': 'IPV4'. Both IP Version and IP Address cannot be specified., invalid

```

paddycarver:
How very bizarre. Let me try a few different types of forwarding rules out (I know there are a few configurations supported that have different requirements) and make sure it's not valid for any of them, then we can go ahead and merge it. In lieu of documentation anywhere else, though, I just want to be sure that this actually holds in _all_ cases.

danielcompton:
My assumption was that it was to be used to create an ephemeral IP address only, but would be good to see what the intended behaviour is.

rosbo:
I agree with your assumption @danielcompton and like the clarification in the docs.

PR #624, opened 2017-10-24
Update container/v1 API

davidquarles:
This is a prerequisite for a few things I'm working on, and I'm guessing it would be preferable to test/merge this change independently.


danawillow:
Thanks @davidquarles!

Issue #583, opened 2017-10-13
Network Policy addon for container cluster is not implemented


LudovicTOURMAN:
### Terraform Version
terraform : v0.10.7
terraform-provider-google : v1.0.1

### Affected Resource(s)
- google_container_cluster

### Terraform Configuration Files
```hcl
resource "google_container_cluster" "my-cluster" {
 ...
  addons_config {
    network_policy { 
      enabled = true
    }
  }
}
```

### Expected Behavior
Should enable [GPE addon for Network Policy](https://cloud.google.com/container-engine/docs/network-policy)

### Actual Behavior
Throw this error
```
1 error(s) occurred:

* module.k8s-cluster.google_container_cluster.my-cluster: addons_config.0: invalid or unknown key: network_policy
``` 

### Steps to Reproduce
1. Add block below to configuration file
2. Set `terraform apply`

sebglon:
Has described on documentation, Network policy is on BETA on gkE kubernetes 1.7.6
https://cloud.google.com/container-engine/docs/network-policy

But this is released on Kubernetes project.

paddycarver:
This looks like it was merged as part of #630, so I'm going to close it out.

PR #630, opened 2017-10-25
[WIP] Add container cluster network policy addon

sebglon:
Add #583

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -128,6 +128,22 @@ func resourceContainerCluster() *schema.Resource {
 								},
 							},
 						},
+						"network_policy": {
+							Type:     schema.TypeList,
+							Optional: true,
+							ForceNew: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"disabled": {
+										Type:     schema.TypeBool,
+										Default:  true,
+										Optional: true,
+										ForceNew: true,
+									},
+								},
+							},
+						},
 					},
 				},
 			},
@@ -859,6 +875,14 @@ func expandClusterAddonsConfig(configured interface{}) *container.AddonsConfig {
 	if v, ok := config["kubernetes_dashboard"]; ok && len(v.([]interface{})) > 0 {
 		addon := v.([]interface{})[0].(map[string]interface{})
 		ac.KubernetesDashboard = &container.KubernetesDashboard{
+			Disabled:        addon["disabled"].(bool),
+			ForceSendFields: []string{"Disabled"},
+		}
+	}
+
+	if v, ok := config["network_policy"]; ok && len(v.([]interface{})) > 0 {
+		addon := v.([]interface{})[0].(map[string]interface{})
+		ac.NetworkPolicyConfig = &container.NetworkPolicyConfig{
 			Disabled:        addon["disabled"].(bool),
 			ForceSendFields: []string{"Disabled"},
 		}
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -128,22 +128,6 @@ func resourceContainerCluster() *schema.Resource {
 								},
 							},
 						},
-						"network_policy": {
-							Type:     schema.TypeList,
-							Optional: true,
-							ForceNew: true,
-							MaxItems: 1,
-							Elem: &schema.Resource{
-								Schema: map[string]*schema.Schema{
-									"disabled": {
-										Type:     schema.TypeBool,
-										Default:  true,
-										Optional: true,
-										ForceNew: true,
-									},
-								},
-							},
-						},
 					},
 				},
 			},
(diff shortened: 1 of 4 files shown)


sebglon:
Actualy i open a case on google because with api, on create cluster param networkPolicyConfig disabled=false is not set on created cluster.

danawillow:
Hey @sebglon, I see the `WIP` in the PR description. Do you have anything you're looking to get feedback on with things as they are now?

sebglon:
Actualy, when i set disable NetworkPolicy tu fase, the network policy is not enabled on create.
For that i have open a case on google to understand why.

wmuizelaar:
I just tried to use this, and when I looked at the webinterface of GKE, you it looked like you first need to enable it on the master-node, and then as well on the nodes. So probably there are two places in the API to look for?

sebglon:
After many exchange with google team, i have changed network_policy on addons by the global network_policy.
Now the code only support global network_policy. 
If the value change, it recreate the cluster.

sebglon:
After many exchange with google team, i have changed network_policy on addons by the global network_policy.
Now the code only support global network_policy. 
If the value change, it recreate the cluster.

@danawillow all is good for me. can you review it?

google/resource_container_cluster.go
```diff
@@ -260,6 +260,31 @@ func resourceContainerCluster() *schema.Resource {
...
+				Computed: true,
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:     schema.TypeBool,
+							ForceNew: true,
+							Optional: true,
+							Default:  true,
```
danawillow: Is this actually enabled by default?

google/resource_container_cluster.go
```diff
@@ -260,6 +260,31 @@ func resourceContainerCluster() *schema.Resource {
 				StateFunc: StoreResourceName,
 			},
 
+			"network_policy": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
```
danawillow: Since you added update functionality, these doesn't need to be `ForceNew`

google/resource_container_cluster.go
```diff
@@ -886,6 +941,30 @@ func expandMasterAuthorizedNetworksConfig(configured interface{}) *container.Mas
...
+				result.Provider = provider.(string)
+			}
+		}
+	}
+	return result
+}
+
+func flattenNetworkPolicy(c *container.NetworkPolicy) []map[string]interface{} {
+	if c == nil {
+		c = &container.NetworkPolicy{}
```
danawillow: I don't like reassigning pointers this way. Instead, how about something like:
```
result := []map[string]interface{}{}
if c != nil {
  result = append(result, map[string]interface{}{
    "enabled": c.Enabled,
    "provider": c.Provider,
  })
}
return result
```

google/resource_container_cluster_test.go
```diff
@@ -921,6 +961,30 @@ resource "google_container_cluster" "with_master_auth" {
...
+		initial_node_count = 1
+	
+		network_policy {
+			enabled = true
+			provider = "CALICO"
+		}	
+	}`, acctest.RandString(10))
+
+var testAccContainerCluster_updateNetworkPolicyEnabled = fmt.Sprintf(`
+		resource "google_container_cluster" "with_network_policy_enabled" {
```
danawillow: nit: can you indent this block one more to the left to match the previous var?

google/resource_container_cluster_test.go
```diff
@@ -921,6 +961,30 @@ resource "google_container_cluster" "with_master_auth" {
...
+			name = "cluster-test-%s"
```
danawillow: Since this config and the previous one both use random strings as part of the cluster name, this will destroy+recreate the cluster so it's not actually testing update.

google/resource_container_cluster_test.go
```diff
@@ -921,6 +961,30 @@ resource "google_container_cluster" "with_master_auth" {
...
+			zone = "us-central1-a"
+			initial_node_count = 1
+		
+			// remove network_policy is equal than enabled=false
```
danawillow: Instead of this comment, what about renaming the test var to `testAccContainerCluster_removeNetworkPolicy`? Right now it's a bit hard to understand.

sebglon:
@danawillow thanks for your review. All change are made

google/resource_container_cluster.go
```diff
@@ -260,6 +260,30 @@ func resourceContainerCluster() *schema.Resource {
...
+			"network_policy": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:     schema.TypeBool,
+							ForceNew: true,
```
danawillow: I'm confused why you've decided to allow adding/removing the network policy but not changing the values within it. Was that intentional? If so, can you add a comment explaining why?

google/resource_container_cluster_test.go
```diff
@@ -911,7 +949,7 @@ resource "google_container_cluster" "primary" {
 
 var testAccContainerCluster_withMasterAuth = fmt.Sprintf(`
 resource "google_container_cluster" "with_master_auth" {
-	name = "cluster-test-%s"
+	name = "%s"
```
danawillow: I think you changed the wrong test here

google/resource_container_cluster.go
sebglon: sorry @danawillow i don't understand exactly what do you say about 'changing the value".
if you do not put network_policy block, it's the same than disable it.
if you put network_policy block, you can choice to enable or diable it.

sebglon:
@danawillow Can i have a feedback

danawillow:
Just added a few small things, but looks good overall, thanks!

PR #636, opened 2017-10-26
Allow attaching and detaching disks from instances

### LINKED ISSUE #33: 'google_compute_instance' adding 'google_compute_disk' forces new resource
_This issue was originally opened by @advanderveer as hashicorp/terraform#6678. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hey, i've encountered unexpected behaviour while updating a google_compute_instance resource by adding a new disk {}.
### Terraform Version

Terraform v0.6.16
### Affected Resource(s)
- google_compute_instance
- google_compute_disk

Consider starting out with the following configuration:

``` hcl
variable "gce_ssh_user" {}
variable "gce_ssh_pub_key_file" {}
variable "gce_service_account" {}
variable "gce_project_id" {}
variable "gce_image" {}
variable "gce_region" {
  default = "europe-west1"
}

variable "gce_network" {}
variable "gce_subnetwork" {}

provider "google" {
  credentials = "${file(var.gce_service_account)}"
  project     = "${var.gce_project_id}"
  region      = "${var.gce_region}"
}

resource "google_compute_disk" "default" {
  name  = "test-disk"
  type  = "pd-ssd"
  zone = "${var.gce_region}-b"
  size = 10
}

resource "google_compute_instance" "remove-me" {
  name       = "disk-test"
  can_ip_forward       = false
  zone = "${var.gce_region}-b"

  disk {
    image = "ubuntu-1604-xenial-v20160429"
  }

  network_interface {
    subnetwork = "${var.gce_subnetwork}"
    access_config {
      // Ephemeral IP
    }
  }

  service_account {
    scopes = ["compute-rw"] 
  }

  metadata {
    sshKeys = "${var.gce_ssh_user}:${file(var.gce_ssh_pub_key_file)}"
  }
}
```

When using the gce interface I would now be able to add the disk 'test-disk' to the instance 'remove-me' without recreating it (or even rebooting). I would expect terraform to behave the same when changing the resource configuration to the following:

``` hcl

variable "gce_ssh_user" {}
variable "gce_ssh_pub_key_file" {}
variable "gce_service_account" {}
variable "gce_project_id" {}
variable "gce_image" {}
variable "gce_region" {
  default = "europe-west1"
}

variable "gce_network" {}
variable "gce_subnetwork" {}

provider "google" {
  credentials = "${file(var.gce_service_account)}"
  project     = "${var.gce_project_id}"
  region      = "${var.gce_region}"
}

resource "google_compute_disk" "default" {
  name  = "test-disk"
  type  = "pd-ssd"
  zone = "${var.gce_region}-b"
  size = 10
}

resource "google_compute_instance" "remove-me" {
  name       = "disk-test"
  can_ip_forward       = false
  zone = "${var.gce_region}-b"

  disk {
    image = "ubuntu-1604-xenial-v20160429"
  }

  #DISK ADDED HERE (using .name attribute instead of self_link yields the same result)
  disk {
    disk = "${google_compute_disk.default.self_link}"
  }

  network_interface {
    subnetwork = "${var.gce_subnetwork}"
    access_config {
      // Ephemeral IP
    }
  }

  service_account {
    scopes = ["compute-rw"] //this grands our filesystemd to modify itself
  }

  metadata {
    sshKeys = "${var.gce_ssh_user}:${file(var.gce_ssh_pub_key_file)}"
  }
}
```
### Debug Output

[https://gist.github.com/advanderveer/fc9e8ef73f1f0b0b8578fedf9140307e](https://gist.github.com/advanderveer/fc9e8ef73f1f0b0b8578fedf9140307e)
### Panic Output

No panic is produced
### Expected Behavior

I would expect the the 'google_compute_instance' not to be recreated
### Actual Behavior

It get destroyed, i.e terraform plan shows multiple "forces new resource":

```
-/+ google_compute_instance.remove-me
    can_ip_forward:                                      "false" => "false"
    disk.#:                                              "1" => "2" **(forces new resource)**
    disk.0.auto_delete:                                  "true" => "true"
    disk.0.image:                                        "ubuntu-1604-xenial-v20160429" => "ubuntu-1604-xenial-v20160429"
    disk.1.auto_delete:                                  "" => "true" **(forces new resource)**
    disk.1.disk:                                         "" => "https://www.googleapis.com/compute/v1/projects/microfactory-test/zones/europe-west1-b/disks/test-disk" (forces new resource)
    machine_type:                                        "n1-standard-1" => "n1-standard-1"
    metadata.#:                                          "1" => "1"
    metadata.sshKeys:                                    "<my public ssh key>"
    metadata_fingerprint:                                "Mq9VPCUvh-E=" => "<computed>"
    name:                                                "disk-test" => "disk-test"
    network_interface.#:                                 "1" => "1"
    network_interface.0.access_config.#:                 "1" => "1"
    network_interface.0.access_config.0.assigned_nat_ip: "146.148.15.98" => "<computed>"
    network_interface.0.address:                         "10.0.0.3" => "<computed>"
    network_interface.0.name:                            "nic0" => "<computed>"
    network_interface.0.subnetwork:                      "microfactory-subnetwork" => "microfactory-subnetwork"
    self_link:                                           "https://www.googleapis.com/compute/v1/projects/microfactory-test/zones/europe-west1-b/instances/disk-test" => "<computed>"
    service_account.#:                                   "1" => "1"
    service_account.0.email:                             "1010578936991-compute@developer.gserviceaccount.com" => "<computed>"
    service_account.0.scopes.#:                          "1" => "1"
    service_account.0.scopes.299962681:                  "https://www.googleapis.com/auth/compute" => "https://www.googleapis.com/auth/compute"
    tags_fingerprint:                                    "42WmSpB8rSM=" => "<computed>"
    zone:                                                "europe-west1-b" => "europe-west1-b"
```
### Steps to Reproduce

1. `terraform plan`
1. `terraform apply`
### Important Factoids

Not that i'm aware of
### References

I couldn't find any maybe [https://github.com/hashicorp/terraform/pull/5241](https://github.com/hashicorp/terraform/pull/5241) is related

danawillow:
Third time's the charm? Fixes #33.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -1161,12 +1161,12 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
 		// were at the time we ran terraform plan.
 		currDisks := map[string]struct{}{}
 		for _, disk := range instance.Disks {
-			if !disk.Boot {
+			if !disk.Boot && disk.Type != "SCRATCH" {
 				currDisks[disk.DeviceName] = struct{}{}
 			}
 		}
 
-		// Keep track of previous config's disks.
+		// Keep track of disks currently in state.
 		// Since changing any field within the disk needs to detach+reattach it,
 		// keep track of the hash of the full disk.
 		oDisks := map[uint64]string{}
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -9,8 +9,10 @@ import (
 
 	"regexp"
 
+	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/mitchellh/hashstructure"
 	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
 	"google.golang.org/api/googleapi"
@@ -204,7 +206,6 @@ func resourceComputeInstance() *schema.Resource {
 			"attached_disk": &schema.Schema{
 				Type:     schema.TypeList,
 				Optional: true,
-				ForceNew: true, // TODO(danawillow): Remove this, support attaching/detaching
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"source": &schema.Schema{
@@ -223,7 +224,6 @@ func resourceComputeInstance() *schema.Resource {
 							Type:      schema.TypeString,
 							Optional:  true,
 							Sensitive: true,
-							ForceNew:  true,
 						},
 
 						"disk_encryption_key_sha256": &schema.Schema{
@@ -623,27 +623,13 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 	attachedDisksCount := d.Get("attached_disk.#").(int)
 
 	for i := 0; i < attachedDisksCount; i++ {
-		prefix := fmt.Sprintf("attached_disk.%d", i)
-		source, err := ParseDiskFieldValue(d.Get(prefix+".source").(string), d, config)
+		diskConfig := d.Get(fmt.Sprintf("attached_disk.%d", i)).(map[string]interface{})
+		disk, err := expandAttachedDisk(diskConfig, d, config)
 		if err != nil {
 			return err
 		}
-		disk := computeBeta.AttachedDisk{
-			Source:     source.RelativeLink(),
-			AutoDelete: false, // Don't allow autodelete; let terraform handle disk deletion
-		}
 
-		if v, ok := d.GetOk(prefix + ".device_name"); ok {
-			disk.DeviceName = v.(string)
-		}
-
-		if v, ok := d.GetOk(prefix + ".disk_encryption_key_raw"); ok {
-			disk.DiskEncryptionKey = &computeBeta.CustomerEncryptionKey{
-				RawKey: v.(string),
-			}
-		}
-
-		disks = append(disks, &disk)
+		disks = append(disks, disk)
 	}
 
 	// Build up the list of networkInterfaces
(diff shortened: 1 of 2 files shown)


danawillow:
```
$ make testacc TEST=./google TESTARGS='-run=TestAccComputeInstance_attachedDisk -parallel 8'
TF_ACC=1 go test ./google -v -run=TestAccComputeInstance_attachedDisk -parallel 8 -timeout 120m
=== RUN   TestAccComputeInstance_attachedDisk
=== RUN   TestAccComputeInstance_attachedDisk_sourceUrl
=== RUN   TestAccComputeInstance_attachedDiskUpdate
--- PASS: TestAccComputeInstance_attachedDisk_sourceUrl (71.67s)
--- PASS: TestAccComputeInstance_attachedDisk (72.10s)
--- PASS: TestAccComputeInstance_attachedDiskUpdate (152.40s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	152.576s
```

google/resource_compute_instance.go
```diff
@@ -1164,6 +1150,101 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
...
+
+	if d.HasChange("attached_disk") {
+		o, n := d.GetChange("attached_disk")
+
+		// Keep track of disks currently in the instance. Because the google_compute_disk resource
+		// can detach disks, it's possible that there are fewer disks currently attached than there
+		// were at the time we ran terraform plan.
+		currDisks := map[string]struct{}{}
+		for _, disk := range instance.Disks {
+			if !disk.Boot {
```
paddycarver: Silly question: does this take scratch disks into account?

google/resource_compute_instance.go
```diff
@@ -1164,6 +1150,101 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
...
+				currDisks[disk.DeviceName] = struct{}{}
+			}
+		}
+
+		// Keep track of previous config's disks.
```
paddycarver: This isn't the previous config, it's the state. Terraform does not version configs itself, and so cannot tell you what was in the previous config.

paddycarver:
> I see in the `READ` method (understandably not being changed here) there seems to be a notion of updating existing disks in config and also adding extra disks and storing them in `attached_disk`. It _seems_ there is an issue here, in that we’re not completely overwriting `attached_disk` with what we find in the API results. Example, if there is a change and a disk Terraform _thinks_ is attached (it’s in the config) is removed outside of Terraform, does Terraform notice on refresh? There doesn’t seem to be any part in the `READ` that’s removing disks not found. If we simply built up the list from the API and wrote those results, regardless of what we had in the config, we could detect that drift. Again, I realize this is outside of the current PR, but I noticed it. 

From what I can see, the disks [are being properly set](https://github.com/terraform-providers/terraform-provider-google/blob/b6af3c2b6dcbdaf6c1fd9ce4564590d3d45d77d2/google/resource_compute_instance.go#L934-L963) based on the API response. That code should detect drift of:

* Disks set in the config but not set on the API
* Disks set in the API but not set in the config

I think the confusion may be around the `extraAttachedDisks` variable, which--as I understand it--exists to work around `attached_disk` being a List and not a Set, but I could be wrong on that. Essentially, the code looks to be ordering disks in the order they're in the config, with any disks not in the config appended to the end.

danawillow:
Yeah, having attached_disks as a Set would definitely simplify things, especially with that Difference method. I'll add that in a new PR and then come back to this one.

google/resource_compute_instance.go
danawillow: I don't think it would have done anything unexpected since we're only using it to check whether a disk in state is still attached, but may as well do the check just in case. Done.

google/resource_compute_instance.go
danawillow: Fixed.

danawillow:
Paddy and I talked this over, and we're not going to try to make `attached_disks` into a set before this change, since it's a fair amount more work than expected, a breaking change, and not currently breaking anyone. See #656 for details.

catsby:
@danawillow 👍

PR #529, opened 2017-10-04
Subnetwork generates network link without calling API

rosbo:
Helps for #431

PR #527, opened 2017-10-04
Improve VPN gateway and router to generate network link without calling network API

rosbo:
Helps for #431 

Not calling the network API for generating the network `self_link` has 2 main advantages:
* Faster, avoid an extraneous network call
* Requires less permission: you may have the permission to create a router without the permission to actually read the network info.

PR #639, opened 2017-10-27
Add Alias IP and Guest Accelerator support to Instance Templates

negz:
...by refactoring a whole bunch of instance and instance template code.

Apologies in advance for the size of this PR. I set out to add Alias IP support to instance templates. I decided to implement this by deduplicating a bunch of code that was replicated between `resource_compute_instance` and `resource_compute_instance_template` and got a little carried away refactoring.

The changes worth note in this PR:
* Per the title instance templates support now Alias IP ranges and Guest Accelerators.
* `resource_compute_instance` and `resource_compute_instance_template` now use the same code to expand and flatten service accounts and network interfaces.
* I switched everything in `resource_compute_instance` over to the v1 API. It was using the beta API for Min CPU Platform and Guest Accelerators, which are both now GA.
* I've deprecated the `network_interface.0.network_ip` key on `resource_compute_instance_template` in favour of `address`, in order to be more consistent with `resource_compute_instance`.

I've run the following tests:
```
$ GOOGLE_PROJECT=REDACTED \
  GOOGLE_XPN_HOST_PROJECT=REDACTED \
  GOOGLE_REGION=us-central1 \
  GOOGLE_USE_DEFAULT_CREDENTIALS=1 \
  TF_ACC=1 \
  go test -v -timeout=120m -run "^TestAccComputeInstance(Template|)_.+$" .
...
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 1356.099s
```


negz:
@rosbo Let me know if there's anything I can do to help you review this - I feel bad about how unwieldy the PR is. Happy to schedule some time to walk through it.

website/docs/r/compute_instance_template.html.markdown
```diff
@@ -255,6 +274,15 @@ The `scheduling` block supports:
...
 
+---
+
+* `guest_accelerator` - (Optional) List of the type and count of accelerator cards attached to the instance. Structure documented below.
+
+The `guest_accelerator` block supports:
+
+* `type` (Required) - The accelerator type resource to expose to this instance. E.g. `nvidia-tesla-k80`.
+
+* `count` (Required) - The number of the guest accelerator cards exposed to this instance.
```
rosbo: Nit: Add a new line after

google/resource_compute_instance_template.go
```diff
@@ -221,6 +242,33 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 										Computed: true,
 										Optional: true,
 									},
+
+									"assigned_nat_ip": &schema.Schema{
```
rosbo: Since an IP is only assigned when an instance is created, it doesn't make sense to have this field in the instance template.

google/resource_compute_instance_template.go
```diff
@@ -184,23 +187,41 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 				ForceNew: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"network": &schema.Schema{
+						"name": &schema.Schema{
```
rosbo: Why did you add a name field here in the `network_interface` block?

google/utils.go
```diff
@@ -124,73 +124,50 @@ func getZonalBetaResourceFromRegion(getResource func(string) (interface{}, error
...
-		if !strings.HasPrefix(network, "https://www.googleapis.com/compute/") {
-			// Network value provided is just the name, lookup the network SelfLink
-			networkData, err := config.clientCompute.Networks.Get(
-				project, network).Do()
-			if err != nil {
-				return "", fmt.Errorf("Error reading network: %s", err)
-			}
-			network = networkData.SelfLink
-		}
+func getNetworkLink(config *Config, project, network string) (string, error) {
```
rosbo: In #527 and #529 I created a helper function to replace this method. I started the work of moving off the `getNetworkLink` method, my next step was to finish off with instance and instance template. Consider deleting this method and use the method [here](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/field_helpers.go#L20). The problem with this method is that it makes an extra network call to generate the self link for the network. 

The other benefit of using this new method is that you can easily extract the project from the link. No need for the extra methods you added in utils.go to do this.

google/utils.go
```diff
@@ -387,3 +363,274 @@ func retry(retryFunc func() error) error {
...
+	for i, accelerator := range accelerators {
+		acceleratorsSchema[i] = map[string]interface{}{
+			"count": accelerator.AcceleratorCount,
+			"type":  accelerator.AcceleratorType,
+		}
+	}
+	return acceleratorsSchema
+}
+
+// Instances want a partial URL, but instance templates want the bare
```
rosbo: You can use the pattern we are using here: https://github.com/terraform-providers/terraform-provider-google/blob/master/google/field_helpers.go#L20

This way both instance and instance template can accept either the name or the URL. And then when calling the api, you can pass whichever the API requires.

google/resource_compute_instance.go
```diff
@@ -336,11 +327,13 @@ func resourceComputeInstance() *schema.Resource {
...
 										Optional: true,
+										Computed: true,
 									},
 
 									"assigned_nat_ip": &schema.Schema{
-										Type:     schema.TypeString,
-										Computed: true,
+										Type:       schema.TypeString,
+										Computed:   true,
+										Deprecated: "Please use nat_ip",
```
rosbo: What is the benefit of changing the way people use the nat ip? This PR is already big, I would leave this change out of it for now and open an issue (explaining why it is desirable and how the benefits outweigh having to force users to change their config) if you feel strongly about it.

Note: I don't know why they where split in two fields in the first place, I wasn't there when the decision was taken, maybe @danawillow has more insights.

google/resource_compute_instance.go
```diff
@@ -328,6 +311,14 @@ func resourceComputeInstance() *schema.Resource {
 							Computed: true,
 						},
 
+						"network_ip": &schema.Schema{
```
rosbo: @danawillow what was the plan with `address` and `network_ip`?

google/resource_compute_instance.go
danawillow: Nah, I wasn't there either. Maybe it can change due to reasons other than setting the field via the API? There's nothing in the docs that gives me that impression though, it seems like if you set it it'll be what you set it to, and if you don't set it it'll assign something. @evandbrown do you know?

google/resource_compute_instance.go
danawillow: Looks like @negz put some info about this in the PR description

rosbo: My bad, it makes total sense to mark it as deprecated.

google/resource_compute_instance.go
negz: I guessed that perhaps the original implementor was unaware that a field could be both optional and computed - i.e. could be set by a human or determined by GCE - and thus deduplicated them.

I don't feel strongly about making this change though - happy to leave it out of this PR.

google/resource_compute_instance_template.go
negz: I added it because `resource_compute_instance.go` has a name field in its `network_interface` block. I was operating under the assumption the schema had to exactly match in those two resources in order for them to share a flatten/expand function.

I'm happy to remove name if that's not the case, since an instance template pretty obviously won't have an instance name.

google/resource_compute_instance_template.go
negz: I can't think of a good way to remove this from instance templates while also sharing interface flatten code with instances. I think that might be what sent me down the path of deprecating `assigned_nat_ip` in favour of just using `nat_ip` for both explicit and computed NAT IPs.

google/resource_compute_instance.go
rosbo: I would remove the `Deprecated` notice from `assigned_nat_ip` but you can leave the `nat_ip` field as `Computed` and set the value from GCE. This way, we don't force user to change their config for no clear benefit and the logic for flatten/expand is simpler.

negz: SGTM - Will do.

website/docs/r/compute_instance_template.html.markdown
negz: Done.

google/resource_compute_instance.go
negz: Done.

google/resource_compute_instance_template.go
negz: I managed to work around this - I only try to set the name field if the returned API object has a Name field, which instance template NICs will not.

google/resource_compute_instance_template.go
negz: Part of me prefers your suggested way, but using it would require passing all of `d` down to `flattenAccessConfigs`, and I feel a little better about that function currently simply acting on a slice of *compute.AccessConfig (at the expense of having to keep a pointless `assigned_nat_ip` in the compute instance schema). I've added a comment.

google/utils.go
negz: I replaced `getNetworkLink` with the helpers you suggested, thanks!

I added some code to parse regional fields, with the intention of using the same helpers for subnetworks. Unfortunately in the context of an instance/instance template the subnetwork's project isn't necessarily the same as the instance's project due to XPN, so I don't think the existing helpers will work.

negz:
@rosbo I believe I've addressed all your comments as best I can. All tests are passing.

google/field_helpers.go
```diff
@@ -220,3 +231,78 @@ func parseOrganizationFieldValue(resourceType, fieldValue string, isEmptyValid b
...
+
+// Parses a regional field supporting 5 different formats:
+// - https://www.googleapis.com/compute/ANY_VERSION/projects/{my_project}/regions/{region}/{resource_type}/{resource_name}
+// - projects/{my_project}/regions/{region}/{resource_type}/{resource_name}
+// - regions/{region}/{resource_type}/{resource_name}
+// - resource_name
+// - "" (empty string). RelativeLink() returns empty if isEmptyValid is true.
+//
+// If the project is not specified, it first tries to get the project from the `projectSchemaField` and then fallback on the default project.
+// If the region is not specified, it takes the value of `regionSchemaField`.
```
rosbo: The Regional case is slightly different from the Zonal case. You can set the default region on the provider. The logic should fallback on the default region if it is not specified in the field value or the schema field.

You can refactor the `getRegion` method to have a structure similar to `getProject`. This will introduce a `getRegionFromSchema` method that you can use here.

negz:
@rosbo Done!

website/docs/r/compute_instance.html.markdown
```diff
@@ -259,7 +259,7 @@ exported:
 
 * `network_interface.0.address` - The internal ip address of the instance, either manually or dynamically assigned.
 
-* `network_interface.0.access_config.0.assigned_nat_ip` - If the instance has an access config, either the given external ip (in the `nat_ip` field) or the ephemeral (generated) ip (if you didn't provide one).
+* `network_interface.0.access_config.0.nat_ip` - If the instance has an access config, either the given external ip (in the `nat_ip` field) or the ephemeral (generated) ip (if you didn't provide one).
```
rosbo: Revert to `network_interface.0.access_config.0.assigned_nat_ip`

negz: Done.

google/resource_compute_instance.go
```diff
@@ -542,41 +537,20 @@ func resourceComputeInstance() *schema.Resource {
...
-			return nil, err
-		}
-	case v0beta:
-		var err error
-		instance, err = config.clientComputeBeta.Instances.Get(
-			project, zone, d.Id()).Do()
-		if err != nil {
-			return nil, handleNotFoundError(err, d, fmt.Sprintf("Instance %s", d.Get("name").(string)))
-		}
+	instance, err := config.clientCompute.Instances.Get(project, zone, d.Id()).Do()
```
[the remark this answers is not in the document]
negz: I want to think through what that would look like for a resource that supported the beta API but that didn't currently use any beta features.

If I understand correctly, using this resource as an example:
* There would be an empty `InstanceVersionedFeatures`, i.e. `var InstanceVersionedFeatures = []Feature{}`
* We'd keep all the `switch computeApiVersion` blocks and keep up-converting `compute` structs to their `computeBeta` counterparts.
* All flatten/expand helper functions (which are potentially shared by multiple resources) would take `computeBeta` arguments, regardless of whether they contained any beta functionality.

I think that could work. It would be a little confusing that everything took `computeBeta` arguments regardless of whether it actually contained any beta functionality, and would be especially unfortunate if we needed to keep maintaining two implementations of each helper function.

I'm on the fence and happy to defer to the provider's maintainers. I can put the scaffolding for beta resources back in. I think the code is easier to reason about without the beta resources, but empathise that someone will likely have to add all the boilerplate back next time we want to support a beta feature on this resource.

google/resource_compute_instance.go
```diff
@@ -948,10 +798,11 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 	}
 	attachedDisks = append(attachedDisks, extraAttachedDisks...)
 
+	d.Set("service_account", flattenServiceAccounts(instance.ServiceAccounts))
```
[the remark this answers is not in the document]
negz: @danawillow I don't follow why that would be.

Previously we set the service accounts inline: https://github.com/terraform-providers/terraform-provider-google/pull/639/files/eb36a7c615c78b220388b693b779a19216a28c09#diff-6ebc1a840e42235a51a46d537f17ddaaL823

This PR breaks that same code out into a `flattenServiceAccounts` function: https://github.com/terraform-providers/terraform-provider-google/pull/639/files/eb36a7c615c78b220388b693b779a19216a28c09#diff-d9f99248626432c3fc0b32182bf3a6faR571

However I don't believe there are any changes to what actually happens - i.e. the code does the same thing.

negz:
> In the future, it would be great if things like this were split across smaller PRs- it might extend the amount of time before the code is fully checked in, but it would definitely make our lives easier to only have to review smaller pieces at a time.

@danawillow Understood - I should know better. I really just wanted Alias IP support for instance templates, but got carried away refactoring. I felt like I'd be contributing to a problem (i.e. two divergent implementations of the same thing) if I were to simply go add Alias IP support to the instance template resource.

google/resource_compute_instance.go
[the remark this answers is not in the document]
negz: I was thinking about this more this morning. I'm increasingly feeling that it's more readable to leave out beta support for resources that don't make use of beta features, assuming that re-adding beta support is as simple as:
1. Re-inserting well-established boilerplate (`InstanceVersionedFeatures`, `switch computeApiVersion`)
2. `s/compute/computeBeta/g`.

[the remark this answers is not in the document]
danawillow: I think it's more readable, but also less maintainable, if that makes sense. Right now, the barrier to entry of using a versioned feature is adding it to the `InstanceVersionedFeatures` list. If we remove all the current beta scaffolding, then the barrier to entry becomes adding it all back. If we get to a point where we don't have any beta features in any resources, then we would need to go back into the files' histories to find examples of how to convert back and forth. Or, a new contributor might not even realize that we have examples already, and then they'd have to try to figure it all out again. For a user adding something unrelated to beta features, the extra scaffolding is easy to work around, but if the code isn't there it could potentially become very difficult for someone to figure out how to add it back. So I stand by my original opinion that we should leave it in.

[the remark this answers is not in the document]
negz: @danawillow Fair enough. How do you think I should proceed here?

I believe that if I re-add the beta scaffolding to `resource_compute_instance` I'll need to switch all of its helper functions to take/produce `computeBeta` structs.

A big part of this PR's intent was to reuse code between `resource_compute_instance` and `resource_compute_instance_template` in order to make the two resources more consistent from a Terraform user's perspective. If I reintroduce the beta scaffolding for `resource_compute_instance` I'll need to also add said scaffolding to `resource_compute_instance_template` in order for the two instances to use the same code.

I'm happy to add beta scaffolding to both resources, but I'm also open to abandoning this PR and sending a new one that adds what I really want (Alias IP support for instance templates) if you think this is all getting too unwieldy.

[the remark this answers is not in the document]
danawillow: Sure, but if you switch it to non-beta and then a beta feature comes along, you still would need to figure out what to do about beta-ifying instance template.

I'm ok with adding the Alias IP support in a new PR so we can keep having the refactoring conversations in this one while that gets in (it'll probably be a quick review since it'll have mostly already been reviewed)

negz:
@danawillow I've re-added the beta scaffolding to compute instance and added it to compute instance template. All tests are passing.

google/resource_compute_instance_template_test.go
```diff
@@ -718,3 +870,104 @@ resource "google_compute_instance_template" "foobar" {
...
+
+  	network_interface {
+   		subnetwork = "${google_compute_subnetwork.inst-test-subnetwork.self_link}"
+
+		// Note that unlike compute instances, instance templates seem to be
+		// only able to specify the netmask here. Trying a full CIDR string
+		// results in: 
+		// Invalid value for field 'resource.properties.networkInterfaces[0].aliasIpRanges[0].ipCidrRange':
+		// '172.16.0.0/24'. Alias IP CIDR range must be a valid netmask starting with '/' (e.g. '/24')
+    	alias_ip_range {
```
danawillow: nit: indentation

google/utils.go
```diff
@@ -408,3 +365,252 @@ func extractFirstMapConfig(m []interface{}) map[string]interface{} {
...
+			return nil, fmt.Errorf("exactly one of network or subnetwork must be provided")
+		}
+
+		nf, err := ParseNetworkFieldValue(network, d, config)
+		if err != nil {
+			return nil, fmt.Errorf("cannot determine selflink for subnetwork '%s': %s", subnetwork, err)
+		}
+
+		subnetworkProject := data["subnetwork_project"].(string)
+		subnetLink, err := getSubnetworkLink(config, project, region, subnetworkProject, subnetwork)
```
danawillow: Were you intending on having this use ParseSubnetworkFieldValue? That got added in this PR but isn't used anywhere.

google/utils.go
```diff
@@ -408,3 +365,252 @@ func extractFirstMapConfig(m []interface{}) map[string]interface{} {
 
 	return m[0].(map[string]interface{})
 }
+
+func expandAliasIpRanges(ranges []interface{}) []*computeBeta.AliasIpRange {
```
danawillow: Since these are only specific to instances/instance templates, I'd rather them be in resource_compute_instance.go (or a new file) instead of the provider-wide utils

google/resource_compute_project_metadata.go
```diff
@@ -87,6 +90,18 @@ func resourceComputeProjectMetadataCreate(d *schema.ResourceData, meta interface
 	return resourceComputeProjectMetadataRead(d, meta)
 }
 
+// This function differs from flattenMetadata only in that it takes
+// compute.metadata rather than computeBeta.metadata as an argument. It should
+// be removed in favour of flattenMetadata if/when this resource gets beta
+// support.
+func flattenCommonInstanceMetadata(metadata *compute.Metadata) map[string]string {
```
danawillow: There are a whole bunch of metadata expanders/flatteners floating around (here, resource_compute_instance, utils.go, metadata.go)- is it possible to have just one of each that live in the same file? (my vote is for metadata.go)

google/utils.go
negz: SGTM. I think I'd prefer a new file. `_helpers.go` and `_utils.go` patterns both exist, so maybe `compute_instance_helpers.go`.

google/resource_compute_project_metadata.go
negz: SGTM, will change.

google/utils.go
negz: https://github.com/terraform-providers/terraform-provider-google/pull/639#discussion_r151016103

> I added some code to parse regional fields, with the intention of using the same helpers for subnetworks. Unfortunately in the context of an instance/instance template the subnetwork's project isn't necessarily the same as the instance's project due to XPN, so I don't think the existing helpers will work.

I did originally, but then it was too much effort per the above comment. :) I think it might be worth doing, but it seems like it might be a good candidate for a separate PR given that `ParseSubnetworkFieldValue` would not work as currently implemented. I'm happy to drop `ParseSubnetworkFieldValue` and the regional field parsing code from this PR since it's unused if that's preferred.

danawillow: Ah ok. At this point my goal is to just get this PR in, so don't worry about it

negz:
@danawillow Think I've got all your comments addressed. It was mostly just shuffling functions around to more sensible places, but I confirmed the tests are still passing too.

danawillow:
Thansk @negz, and sorry for the drawn-out review cycle! Merging now.

negz:
No problem @danawillow ! Thank you and @rosbo for all the help getting this merged.

PR #640, opened 2017-10-27
Subscription with topic in different project

### LINKED ISSUE #649: Create a pubsub subscription with a topic in a different project.
As of now, the topic must be in the same project as the subscription.

The `topic` field should support a `name` or a `self_link`. If a `self_link` is used, it should use the project in that self_link for the topic.

sebasrobert:
Fixes #649 
Right now we can't create subscription on a topic in a different gcp
project since it assume the project from the subscription. The provider
always create the full topic name string
projects/{project}/topics/{topic} with the received topic property.
Using a regexp we validate if the string is already in
the format projects/{project}/topics/{topic} and if it's the case
we don't wrap it again and take it directly. The original functionality
is maintained but it's possible to specify a different project for the
topic.

### DIFF
--- a/google/resource_pubsub_subscription.go
+++ b/google/resource_pubsub_subscription.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"google.golang.org/api/pubsub/v1"
+	"regexp"
 )
 
 func resourcePubsubSubscription() *schema.Resource {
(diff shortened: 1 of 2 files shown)


rosbo:
Thanks @sebasrobert for catching this issue and coming up with a fix. 

I will add a `self_link` field in the topic resource in a separate PR to make it even easier to create Terraform config when the topic is in a different project than the subscription.

Merging this part!

PR #642, opened 2017-10-28
fixing example of google_storage_object_acl

mioi:
some of the resource names seem to be incorrect. fixing to reflect the actual resource names in the example.

### DIFF
--- a/website/docs/r/storage_object_acl.html.markdown
+++ b/website/docs/r/storage_object_acl.html.markdown
@@ -25,13 +25,13 @@ resource "google_storage_bucket" "image-store" {
 
 resource "google_storage_bucket_object" "image" {
   name   = "image1"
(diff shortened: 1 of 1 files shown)


website/docs/r/storage_object_acl.html.markdown
```diff
@@ -25,13 +25,13 @@ resource "google_storage_bucket" "image-store" {
 
 resource "google_storage_bucket_object" "image" {
   name   = "image1"
-  bucket = "${google_storage_bucket.name}"
+  bucket = "${google_storage_bucket.image-store}"
```
danawillow: shouldn't this end in `.name`?

mioi: aha, good catch.

danawillow:
Thanks so much @mioi!

PR #652, opened 2017-10-30
Add labels to `google_storage_bucket`

### LINKED ISSUE #599: Add label creation to storage bucket resource type
Hi,

Was hoping that we could get label creation on storage buckets added to that resource type. Not sure if this is planned or not, but I didn't see it in the last release of the provider.

Was hoping this could follow the pattern of labeling compute vms like:

```
resource "google_storage_bucket" "bucket" {
  name     = "my-bucket"
  labels   = {
      test  = "label"
  }
}
```

danawillow:
Fixes #599.

```
$ make testacc TEST=./google TESTARGS='-run=TestAccStorageBucket_labels'
TF_ACC=1 go test ./google -v -run=TestAccStorageBucket_labels -timeout 120m
=== RUN   TestAccStorageBucket_labels
--- PASS: TestAccStorageBucket_labels (6.83s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	6.993s
```

The extra vendor changes are in here because of some cascading dependencies (storage and bigquery both depend on gensupport).

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -964,10 +964,10 @@
 			"revisionTime": "2016-12-14T09:25:55Z"
 		},
 		{
-			"checksumSHA1": "FEzQdhqmb6aqGL1lKnjOcUHIGSY=",
+			"checksumSHA1": "jhyv7qysp0NZbWzAaUCsKATDGuk=",
 			"path": "google.golang.org/api/bigquery/v2",
-			"revision": "16ab375f94503bfa0d19db78e96bffbe1a34354f",
-			"revisionTime": "2017-03-20T22:51:23Z"
+			"revision": "672d215daf0631fcae4c08c2a4324a763aaaf789",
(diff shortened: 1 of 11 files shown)


vendor/vendor.json
```diff
@@ -964,10 +964,10 @@
 			"revisionTime": "2016-12-14T09:25:55Z"
 		},
 		{
-			"checksumSHA1": "FEzQdhqmb6aqGL1lKnjOcUHIGSY=",
+			"checksumSHA1": "jhyv7qysp0NZbWzAaUCsKATDGuk=",
 			"path": "google.golang.org/api/bigquery/v2",
```
rosbo: How is updating bigquery client to the latest version related to this change?

danawillow: When I updated storage, it didn't build because it relied on a version of gensupport later than what we currently had vendored. When I updated gensupport, it still didn't build because bigquery was relying on an older version of gensupport.

PR #654, opened 2017-10-30
add support for ip aliasing in `google_container_cluster`

### LINKED ISSUE #618: Add support for IP Aliases to `google_container_cluster`
This is now part of the [v1 API](https://godoc.org/google.golang.org/api/container/v1#IPAllocationPolicy).

The two relevant (mutually exclusive) workflows:
- Pre-allocate subnet with secondary ranges for pods/services, then pass in the subnet name and secondary range names
- Dynamically create the subnet as part of cluster creation, and specify pod/service/node CIDRs

IIRC there are slight limitations in the latter under gcloud, but I'm not sure if that holds when talking to the upstream API.  I'll try tackling this today if I can get to it.

davidquarles:
Fixes #618.

Note that there are two mutually exclusive paths for GKE clusters with IP aliasing enabled:
1) pre-allocated subnets
2) dynamically created subnets, which are a bit more magical and strictly tied to cluster lifecycle

I chose to implement only (1), for now, because:
- It seemed like the most flexible path
- There are several tiers of conflicting and dependent attributes.  `ConflictsWith` halfway solves the problem, but AFAIK there's no good abstraction on the other side for dependencies/requirements, making it difficult to implement both strategies and detect failures at plan-time.  There are cases where I wasn't able to do so even with the single strategy.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -341,40 +341,18 @@ func resourceContainerCluster() *schema.Resource {
 			"ip_allocation_policy": {
 				Type:     schema.TypeList,
 				Optional: true,
-				Computed: true,
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"use_ip_aliases": {
-							Type:     schema.TypeBool,
-							Optional: true,
-							Computed: true,
-							Default:  nil,
-							ForceNew: true,
-						},
 						"cluster_secondary_range_name": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Computed: true,
 							ForceNew: true,
-							DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-								if new == "" {
-									return true
-								}
-								return ipCidrRangeDiffSuppress(k, old, new, d)
-							},
 						},
 						"services_secondary_range_name": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Computed: true,
 							ForceNew: true,
-							DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-								if new == "" {
-									return true
-								}
-								return ipCidrRangeDiffSuppress(k, old, new, d)
-							},
 						},
 					},
 				},
@@ -515,47 +493,6 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		cluster.NodePools = nodePools
 	}
 
-	if v, ok := d.GetOk("ip_allocation_policy"); ok {
-		cluster.IpAllocationPolicy, err = expandIPAllocationPolicy(v)
-		if err != nil {
-			return err
-		}
-
-		region := getRegionFromZone(zoneName)
-
-		subnetwork, err := config.clientCompute.Subnetworks.Get(project, region, cluster.Subnetwork).Do()
-		if err != nil {
-			return handleNotFoundError(err, d, fmt.Sprintf("Subnetwork %q", cluster.Subnetwork))
-		}
-
-		clusterRangeExists := false
-		servicesRangeExists := false
-
-		for _, secondaryRange := range subnetwork.SecondaryIpRanges {
-			switch secondaryRange.RangeName {
-			case cluster.IpAllocationPolicy.ClusterSecondaryRangeName:
-				clusterRangeExists = true
-			case cluster.IpAllocationPolicy.ServicesSecondaryRangeName:
-				servicesRangeExists = true
-			}
-		}
-
-		missingRanges := []string{}
-		if !clusterRangeExists {
-			missingRanges = append(missingRanges, fmt.Sprintf("cluster_secondary_range_name=%s",
-				cluster.IpAllocationPolicy.ClusterSecondaryRangeName))
-		}
-		if !servicesRangeExists {
-			missingRanges = append(missingRanges, fmt.Sprintf("services_secondary_range_name=%s",
-				cluster.IpAllocationPolicy.ServicesSecondaryRangeName))
-		}
-
-		if len(missingRanges) > 0 {
-			return fmt.Errorf("The specified secondary ranges are missing on the cluster's subnetwork: [%s]",
-				strings.Join(missingRanges, ", "))
-		}
-	}
-
 	req := &container.CreateClusterRequest{
 		Cluster: cluster,
 	}
@@ -672,7 +609,7 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 	}
 	d.Set("node_pool", nps)
 
-	if cluster.IpAllocationPolicy != nil {
+	if cluster.IpAllocationPolicy != nil && cluster.IpAllocationPolicy.UseIpAliases {
 		d.Set("ip_allocation_policy", flattenIPAllocationPolicy(cluster.IpAllocationPolicy))
 	}
 
@@ -1099,7 +1036,6 @@ func flattenClusterNodePools(d *schema.ResourceData, config *Config, c []*contai
 func flattenIPAllocationPolicy(c *container.IPAllocationPolicy) []map[string]interface{} {
 	return []map[string]interface{}{
 		{
-			"use_ip_aliases":                c.UseIpAliases,
 			"cluster_secondary_range_name":  c.ClusterSecondaryRangeName,
 			"services_secondary_range_name": c.ServicesSecondaryRangeName,
 		},
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -341,6 +341,7 @@ func resourceContainerCluster() *schema.Resource {
 			"ip_allocation_policy": {
 				Type:     schema.TypeList,
 				Optional: true,
+				ForceNew: true,
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
@@ -493,6 +494,13 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		cluster.NodePools = nodePools
 	}
 
+	if v, ok := d.GetOk("ip_allocation_policy"); ok {
+		cluster.IpAllocationPolicy, err = expandIPAllocationPolicy(v)
+		if err != nil {
+			return err
+		}
+	}
+
 	req := &container.CreateClusterRequest{
 		Cluster: cluster,
 	}
@@ -609,8 +617,8 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 	}
 	d.Set("node_pool", nps)
 
-	if cluster.IpAllocationPolicy != nil && cluster.IpAllocationPolicy.UseIpAliases {
-		d.Set("ip_allocation_policy", flattenIPAllocationPolicy(cluster.IpAllocationPolicy))
+	if err := d.Set("ip_allocation_policy", flattenIPAllocationPolicy(cluster.IpAllocationPolicy)); err != nil {
+		return err
 	}
 
 	if igUrls, err := getInstanceGroupUrlsFromManagerUrls(config, cluster.InstanceGroupUrls); err != nil {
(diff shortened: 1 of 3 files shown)


davidquarles:
I can run the full acceptance suite later tonight, but for now: 
``` 
$ make testacc TEST=./google TESTARGS='-run=TestAccContainerCluster_withIPAllocationPolicy'
TF_ACC=1 go test ./google -v -run=TestAccContainerCluster_withIPAllocationPolicy -timeout 120m
=== RUN   TestAccContainerCluster_withIPAllocationPolicy
--- PASS: TestAccContainerCluster_withIPAllocationPolicy (643.20s)
PASS
ok  github.com/terraform-providers/terraform-provider-google/google643.235s
```

google/resource_container_cluster.go
```diff
@@ -279,6 +279,48 @@ func resourceContainerCluster() *schema.Resource {
...
+						"cluster_secondary_range_name": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+							ForceNew: true,
+							DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+								if new == "" {
+									return true
+								}
+								return ipCidrRangeDiffSuppress(k, old, new, d)
```
danawillow: Isn't this field the name, not the range itself?

google/resource_container_cluster.go
```diff
@@ -394,6 +436,47 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		cluster.NodePools = nodePools
 	}
 
+	if v, ok := d.GetOk("ip_allocation_policy"); ok {
+		cluster.IpAllocationPolicy, err = expandIPAllocationPolicy(v)
+		if err != nil {
+			return err
+		}
+
+		region := getRegionFromZone(zoneName)
```
danawillow: I'm not sure we actually need these checks inside of Terraform. If I get rid of them I get this error from the API: `Google Compute Engine: services secondary range "pods" not found in subnet "container-net-asdfafw"` which serves the same purpose

google/resource_container_cluster.go
```diff
@@ -844,6 +956,26 @@ func flattenClusterNodePools(d *schema.ResourceData, config *Config, c []*contai
...
 }
 
+func flattenIPAllocationPolicy(c *container.IPAllocationPolicy) []map[string]interface{} {
+	result := map[string]interface{}{
+		"use_ip_aliases":                c.UseIpAliases,
+		"cluster_secondary_range_name":  c.ClusterSecondaryRangeName,
+		"services_secondary_range_name": c.ServicesSecondaryRangeName,
+	}
+
+	//if c.UseIpAliases {
```
danawillow: These can be removed :)

google/resource_container_cluster_test.go
```diff
@@ -1340,3 +1395,49 @@ resource "google_container_cluster" "with_node_pool_node_config" {
...
+	}
+
+	return fmt.Sprintf(`
+resource "google_compute_network" "container_network" {
+	name = "container-net-%s"
+	auto_create_subnetworks = false
+}
+
+resource "google_compute_subnetwork" "container_subnetwork" {
+    name          = "${google_compute_network.container_network.name}"
```
danawillow: can you fix the indentation here?

google/resource_container_cluster_test.go
```diff
@@ -1340,3 +1395,49 @@ resource "google_container_cluster" "with_node_pool_node_config" {
...
+	var secondaryRanges bytes.Buffer
+	for rangeName, cidr := range ranges {
+		secondaryRanges.WriteString(fmt.Sprintf(`
+	secondary_ip_range {
+	    range_name    = "%s"
+	    ip_cidr_range = "%s"
+	}`, rangeName, cidr))
+	}
+
+	var ipAllocationPolicy bytes.Buffer
```
danawillow: I'm fine leaving this as is, but a potential alternative would be just passing in the configuration block as a string that looked like:
```
  ip_allocation_policy {
    use_ip_aliases                = "true"
    cluster_secondary_range_name  = "pods"
    services_secondary_range_name = "services"
  }
```

google/resource_container_cluster.go
```diff
@@ -279,6 +279,48 @@ func resourceContainerCluster() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"ip_allocation_policy": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
danawillow: Any reason why this is (and the others are) computed?

davidquarles:
@danawillow My apologies for the shoddy / incomplete initial PR.  I was a bit sleep-deprived that week; all of your comments were right on.  I made all the changes you asked for, and I removed `use_ip_aliases` param in favor of simply checking existence of the outer config block (as you suggested with `master_authorized_networks_config`).  Let me know what you think.  Thanks!

```
$ make testacc TEST=./google TESTARGS='-run=TestAccContainerCluster_withIPAllocationPolicy'
TF_ACC=1 go test ./google -v -run=TestAccContainerCluster_withIPAllocationPolicy -timeout 120m
=== RUN   TestAccContainerCluster_withIPAllocationPolicy
--- PASS: TestAccContainerCluster_withIPAllocationPolicy (1336.11s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	1336.136s
```

google/resource_container_cluster.go
davidquarles: 🤦‍♂️

google/resource_container_cluster.go
```diff
@@ -279,6 +279,48 @@ func resourceContainerCluster() *schema.Resource {
...
+							ForceNew: true,
+						},
+						"cluster_secondary_range_name": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+							ForceNew: true,
+							DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+								if new == "" {
+									return true
```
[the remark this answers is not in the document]
davidquarles: 🤦‍♂️

google/resource_container_cluster.go
```diff
@@ -279,6 +279,48 @@ func resourceContainerCluster() *schema.Resource {
...
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"use_ip_aliases": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Computed: true,
+							Default:  nil,
```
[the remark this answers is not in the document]
davidquarles: 🤦‍♂️

google/resource_container_cluster.go
davidquarles: 🤦‍♂️

google/resource_container_cluster.go
davidquarles: Sounds good to me.

website/docs/r/container_cluster.html.markdown
```diff
@@ -184,6 +186,18 @@ maintenance_policy {
 }
 ```
 
+The `ip_allocation_policy` block supports:
+
+* `cluster_secondary_range_name` - (Optional) The name of the secondary range to be
```
danawillow: The last sentence of this (and the next) are no longer applicable.

google/resource_container_cluster_test.go
```diff
@@ -1472,11 +1522,57 @@ resource "google_container_cluster" "with_maintenance_window" {
...
+	name          = "${google_compute_network.container_network.name}"
+	network       = "${google_compute_network.container_network.name}"
+	ip_cidr_range = "10.0.0.0/24"
+	region        = "us-central1"
+
+	%s
+}
+
+resource "google_container_cluster" "with_ip_allocation_policy" {
+	name = "with-ip-allocation-policy"
```
danawillow: Can you give this a name with some randomness? We usually do that so if the test fails and the resource is lingering, we can still run the test before cleaning it up again.

google/resource_container_cluster.go
```diff
@@ -337,6 +337,26 @@ func resourceContainerCluster() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"ip_allocation_policy": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
```
danawillow: Not sure it really makes a difference, but let's make this `ForceNew` as well

google/resource_container_cluster.go
```diff
@@ -589,6 +609,10 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 	}
 	d.Set("node_pool", nps)
 
+	if cluster.IpAllocationPolicy != nil && cluster.IpAllocationPolicy.UseIpAliases {
```
danawillow: Actually I think we should set it no matter what (and this might be a bug elsewhere in the file). If I create a cluster with IP Aliasing, then remove IP Aliasing outside of Terraform, I would want my state to get updated. It's sort of a moot point since IPAllocationPolicy doesn't support update right now, but this will help future-proof it.

Here's what I think `flattenIPAllocationPolicy` should look like:
```
func flattenIPAllocationPolicy(c *container.IPAllocationPolicy) []map[string]interface{} {
	result := []map[string]interface{}{}
	if c != nil && c.UseIpAliases {
		result = append(result, map[string]interface{}{
			"cluster_secondary_range_name":  c.ClusterSecondaryRangeName,
			"services_secondary_range_name": c.ServicesSecondaryRangeName,
		})
	}
	return result
}
```

and then the call to set it would be:
```
	if err := d.Set("ip_allocation_policy", flattenIPAllocationPolicy(cluster.IpAllocationPolicy)); err != nil {
		return err
	}
```

google/resource_container_cluster.go
```diff
@@ -925,6 +949,30 @@ func expandClusterAddonsConfig(configured interface{}) *container.AddonsConfig {
 	return ac
 }
 
+func expandIPAllocationPolicy(configured interface{}) (*container.IPAllocationPolicy, error) {
```
danawillow: I think you accidentally got rid of the call to this function

google/resource_container_cluster_test.go
```diff
@@ -595,6 +595,56 @@ func TestAccContainerCluster_withMaintenanceWindow(t *testing.T) {
...
+						"pods":     "10.1.0.0/16",
+						"services": "10.2.0.0/20",
+					},
+					map[string]string{
+						"cluster_secondary_range_name":  "pods",
+						"services_secondary_range_name": "services",
+					},
+				),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckContainerCluster(
```
danawillow: Can you add checks for `ip_allocation_policy` into `testAccCheckContainerCluster`? I also wouldn't mind seeing  some `resource.TestCheckResourceAttr`s but I won't block on those

google/resource_container_cluster_test.go
davidquarles: Done.

website/docs/r/container_cluster.html.markdown
davidquarles: Done.

google/resource_container_cluster.go
danawillow: You still need to do the nil check, otherwise the code will panic when trying to access the struct's fields (you can prove it by running `TestAccContainerCluster_basic`)

davidquarles: I just realized this comment slipped through the cracks and fixed it.

```  gke_nordstrom-k8s-nonprod_us-west1-a_test-lark/default  make testacc TEST=./google TESTARGS='-run=TestAccContainerCluster_basic'
TF_ACC=1 go test ./google -v -run=TestAccContainerCluster_basic -timeout 120m
--- PASS: TestAccContainerCluster_basic (356.83s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	356.853s
```

google/resource_container_cluster_test.go
davidquarles: Done and done

danawillow:
Thanks @davidquarles!

Issue #46, opened 2017-06-13
google_compute_target_pool is not idempotent when using URLs as values from google_compute_instance


hashibot:
_This issue was originally opened by @jsierles as hashicorp/terraform#9121. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,

Terraform v0.7.4
google_compute_target_pool

``` hcl
resource "google_compute_target_pool" "loadbalancer" {
  name = "target-pool-loadbalancer-${var.env}"
  instances = ["${google_compute_instance.loadbalancer.*.self_link}"]
  health_checks = ["${google_compute_http_health_check.loadbalancer.name}"]
}
```
### Plan output

```
~ google_compute_target_pool.loadbalancer
    instances.0: "europe-west1-d/loadbalancer-0-staging" => "https://www.googleapis.com/compute/v1/projects/myproject-com/zones/europe-west1-d/instances/loadbalancer-0-staging"
    instances.1: "europe-west1-d/loadbalancer-1-staging" => "https://www.googleapis.com/compute/v1/projects/myproject-com/zones/europe-west1-d/instances/loadbalancer-1-staging"
```
### Expected Behavior

This value should not be changed on every plan/apply run.
### Actual Behavior

On every `plan` and `apply`, this resource is modified without makes changes to the configuration.

PR #659, opened 2017-10-31
two-tier example is idempotent

gobengo:
Moving this from old repo (https://github.com/hashicorp/terraform/pull/12258)

This PR makes the two-tier example apply more cleanly and idempotently.

* `google_compute_forwarding_rule.default.port_range = "80"` would save okay, but then would save to state file as "80-80", so it would always think a change was needed
* `google_compute_target_pool.default.instances` items would resolve to full URLs, but saved in state as just `{zone}/{instance.name}`.
  * This PR uses a [workaround](#9121 (comment)) mentioned in #9121

### DIFF
--- a/examples/two-tier/main.tf
+++ b/examples/two-tier/main.tf
@@ -17,7 +17,7 @@ resource "google_compute_http_health_check" "default" {
 
 resource "google_compute_target_pool" "default" {
   name          = "tf-www-target-pool"
-  instances     = ["${google_compute_instance.www.*.self_link}"]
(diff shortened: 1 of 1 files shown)


danawillow:
Instead of this change, I'm going to go ahead and actually fix the issue making this change necessary (#46)

danawillow:
I'm going to go ahead and close this, since I fixed the issue in PR #666 (it'll be in the next release). Thanks though, @gobengo!

gobengo:
@danawillow Way to go. Thanks for making it even better!

Issue #658, opened 2017-10-31
The zonal field for resource disks cannot be empty


allandrick:
Hi there,

I'm upgrading a project from 0.8.2 to 0.10.7 and having problems with an existing google_compute_instance resource and its associated disk(s).

I became aware that previous references to disks using the

```hcl
  disk {
    name = "xxx"
    device_name = "yyy"
    auto_delete = "false"
  }
```

method to link in an existing google_compute_disk resource needed to be changed to:

```hcl
        boot_disk {
                source = "${google_compute_disk.whatever.self_link}"
                initialize_params {
                        image = ""
                }
                device_name = "disk1"
                auto_delete = "false"
        }
```

and this has worked for several VMs across different projects coming from various terraform versions up to and including 0.10.7.  The initialize_params shouldn't have needed to be there, especially given the presence of the source parameter, but terraform plan kept saying it wanted to force new resource on the instance because initialize_params went from 1 => 0 - this work-around was the closest I could get to a clean "no change moving forward" plan.

However, in one particularly old project, I am now getting a strange error on terraform refresh:

```bash
Error refreshing state: 1 error(s) occurred:

* google_compute_instance.xxxxxxxxxxxxxxxx: 1 error(s) occurred:

* google_compute_instance.xxxxxxxxxxxxxxxx: google_compute_instance.xxxxxxxxxxxxxxxx: The zonal field for resource disks cannot be empty.
```

with no other additional info.

### Terraform Version

Terraform v0.10.7

The release notes for 0.10.8 do not indicate any related fixes to this issue and this appears to be a Google provider specific error message.

### Affected Resource(s)
- google_compute_instance
- google_compute_disk

### Terraform Configuration Files
```hcl
resource "google_compute_disk" "os-disk" {
        name = "os-disk"
        type = "pd-standard"
        zone = "us-east1-b"
        size = "100"
        image = "windows-server-2012-r2-dc-v20161012"

        lifecycle {
                prevent_destroy = "true"
        }
}

resource "google_compute_disk" "data-disk" {
        name = "data-disk"
        type = "pd-standard"
        zone = "us-east1-b"
        size = "500"

        lifecycle {
                prevent_destroy = "true"
        }
}

resource "google_compute_instance" "server" {
        name = "server"
        machine_type = "n1-standard-2"
        zone = "us-east1-b"

        lifecycle {
                ignore_changes = [
                                "create_timeout"
                ]
        }

        boot_disk {
[...]

PR #664, opened 2017-11-01
Fix bug with empty `initialize_params` block

danawillow:
In old versions of Terraform, `disk.size` was being set to "0" in state for disks that did not specify a size parameter. Check for this case when checking that the field is empty, and fix states that accidentally added an `initialize_params` block to the `boot_disk` during the migration in the above scenario.

Part of #658.

### DIFF
--- a/google/resource_compute_instance_migrate.go
+++ b/google/resource_compute_instance_migrate.go
@@ -56,7 +56,14 @@ func resourceComputeInstanceMigrateState(
 		if err != nil {
 			return is, err
 		}
-		// when adding case 5, make sure to turn this into a fallthrough
+		fallthrough
+	case 5:
+		log.Println("[INFO] Found Compute Instance State v5; migrating to v6")
+		is, err = migrateStateV5toV6(is)
+		if err != nil {
+			return is, err
+		}
+		// when adding case 6, make sure to turn this into a fallthrough
 		return is, err
 	default:
 		return is, fmt.Errorf("Unexpected schema version: %d", v)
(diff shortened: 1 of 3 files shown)


google/resource_compute_instance_migrate.go
```diff
@@ -489,3 +496,17 @@ func getDiskFromAutoDeleteAndImage(config *Config, instance *compute.Instance, a
...
 }
+
+func migrateStateV5toV6(is *terraform.InstanceState) (*terraform.InstanceState, error) {
+	log.Printf("[DEBUG] Attributes before migration: %#v", is.Attributes)
+	if is.Attributes["boot_disk.0.initialize_params.#"] == "1" {
+		if (is.Attributes["boot_disk.0.initialize_params.0.size"] == "0" ||
+			is.Attributes["boot_disk.0.initialize_params.0.size"] == "") &&
+			is.Attributes["boot_disk.0.initialize_params.0.type"] == "" &&
+			is.Attributes["boot_disk.0.initialize_params.0.image"] == "" {
+			is.Attributes["boot_disk.0.initialize_params.#"] = "0"
```
paddycarver: Is it worth calling `delete(is.Attributes, "boot_disk.0.initialize_params.INSERT_EVERY_ATTRIBUTE_HERE")` to get a properly clean state?

danawillow: Sure, can't hurt. Done.

paddycarver:
👍 LGTM

PR #665, opened 2017-11-01
Make `google_container_cluster.node_pool` mutable (ForceNew=false)

davidquarles:
I'm guessing this is too naive, but I thought it made sense to pose the question before tackling some of the larger problems around node pool mutation / lifecycle – is this all that's required (beyond additional tests) to be able to add/remove node pools from the cluster resource?

/cc @danawillow (as there's a `TODO` here with your name on it)

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -295,10 +295,10 @@ func resourceContainerCluster() *schema.Resource {
 			"node_config": schemaNodeConfig,
 
 			"node_pool": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Optional: true,
 				Computed: true,
-				ForceNew: false,
+				ForceNew: true,
 				Elem: &schema.Resource{
 					Schema: schemaNodePool,
 				},
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -298,7 +298,7 @@ func resourceContainerCluster() *schema.Resource {
 				Type:     schema.TypeList,
 				Optional: true,
 				Computed: true,
-				ForceNew: true, // TODO(danawillow): Add ability to add/remove nodePools
+				ForceNew: false,
 				Elem: &schema.Resource{
 					Schema: schemaNodePool,
 				},


danawillow:
Nope- that'll cause `terraform plan` to say it's an update instead of a destroy/create, but when `resourceContainerClusterUpdate` gets called nothing will happen since there isn't any logic there to handle adding/removing.

davidquarles:
Ah, true.  I figured as much.  I can take a real stab at this soon, then (unless someone else wants to, of course).

birdayz:
Please correct me if i am wrong, but there is some code that allows us to resize node pools - function nodePoolUpdate. However, without the change in this PR (disabling forceUpdate in node_pool), it will never trigger anyway, because any change will cause a full destroy/create cycle. Am i correct here?

If yes, i suppose we need to fully implement the update for node_pool, i.e. new node pools are added and deleted node pools are removed in case of changes. Then we could remove the forceNew. Is this correct?

danawillow:
@birdayz not quite- for a List, a diff means (roughly) that the number of elements in the list has changed. If a specific field in the list has changed, that's considered a change to that field rather than to the whole list. So for an example, I created this cluster:
```
resource "google_container_cluster" "is-665" {
  name       = "is-665"
  zone       = "us-central1-a"

  node_pool {
    name       = "is-665-np"
    node_count = 1
  }
}
```
then when I change `node_count` to 2 and run `terraform plan`, I get:
```

  ~ google_container_cluster.is-665
      node_pool.0.node_count: "1" => "2"
```

Make sense?

davidquarles:
@danawillow What do you think is a sane path forward here?  It feels like `node_pool` should be a set (with id/name used for state/diff, perhaps?), and on cluster update we'd want to diff against previous state and handle deletions / additions / updates in three separate buckets, but you'd know better than I.  Are there any other existing resource in the provider that do this?  I looked around and found similar codepaths, but couldn't find any where the set elements are actually nested resources.  This is obviously a bit complex, and I see you handling similar use cases in your own, past PRs (e.g. disk attach/detach), so I totally get if you want to take this one, but I can also take a stab it with a bit of direction.  Thanks, either way!

danawillow:
Yeah, the best examples I have are the ones that I've done myself :)

This is on my to-do list for next week so how about we circle back in two weeks to see if I've actually done it? And then if not I can give you some guidance.

danawillow:
I'm going to go ahead and close this in favor of the discussion happening in #780.

PR #666, opened 2017-11-01
Update `google_compute_target_pool` to no longer have a plan/apply loop with instance URLs

### LINKED ISSUE #46: google_compute_target_pool is not idempotent when using URLs as values from google_compute_instance

danawillow:
Fixes #46 

```
$ make testacc TEST=./google TESTARGS='-run=TestAccComputeTargetPool_basic'
TF_ACC=1 go test ./google -v -run=TestAccComputeTargetPool_basic -timeout 120m
=== RUN   TestAccComputeTargetPool_basic
--- PASS: TestAccComputeTargetPool_basic (85.47s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	85.689s
```

### DIFF
--- a/google/resource_compute_target_pool.go
+++ b/google/resource_compute_target_pool.go
@@ -10,6 +10,8 @@ import (
 	"google.golang.org/api/compute/v1"
 )
 
+var instancesSelfLinkPattern = regexp.MustCompile(fmt.Sprintf(zonalLinkBasePattern, "instances"))
+
 func resourceComputeTargetPool() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceComputeTargetPoolCreate,
@@ -63,9 +65,15 @@ func resourceComputeTargetPool() *schema.Resource {
 					oldParts := strings.Split(old, "/")
 
 					// instances can also be specified in the config as a URL
-					r := regexp.MustCompile(fmt.Sprintf(zonalLinkBasePattern, "instances"))
-					if parts := r.FindStringSubmatch(new); parts != nil {
-						if parts[2] == oldParts[0] && parts[3] == oldParts[1] {
+					if parts := instancesSelfLinkPattern.FindStringSubmatch(new); len(oldParts) == 2 && len(parts) == 4 {
+						// parts[0] = full match
+						// parts[1] = project
+						// parts[2] = zone
+						// parts[3] = instance name
+
+						oZone, oName := oldParts[0], oldParts[1]
+						nZone, nName := parts[2], parts[3]
+						if oZone == nZone && oName == nName {
 							return true
 						}
 					}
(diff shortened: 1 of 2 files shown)


google/resource_compute_target_pool.go
```diff
@@ -57,6 +58,19 @@ func resourceComputeTargetPool() *schema.Resource {
 				Computed: true,
 				ForceNew: false,
 				Elem:     &schema.Schema{Type: schema.TypeString},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					// instances are stored in state as "zone/name"
+					oldParts := strings.Split(old, "/")
+
+					// instances can also be specified in the config as a URL
+					r := regexp.MustCompile(fmt.Sprintf(zonalLinkBasePattern, "instances"))
```
paddycarver: Ideally we'd compile this to a global variable, not on every DiffSuppressFunc call.

google/resource_compute_target_pool.go
```diff
@@ -57,6 +58,19 @@ func resourceComputeTargetPool() *schema.Resource {
...
+					if parts := r.FindStringSubmatch(new); parts != nil {
```
paddycarver: Let's check that parts has enough entries to avoid panics on invalid inputs.

google/resource_compute_target_pool.go
danawillow: Done

google/resource_compute_target_pool.go
danawillow: Done, also made some variables hopefully to make it a bit clearer what's going on

Issue #596, opened 2017-10-17
google_project_services inconsistent behaviour


hashibot:
_This issue was originally opened by @jp as hashicorp/terraform#16341. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,

I'm having some issues with an inconsistent behaviour of the google_project_services module.
I want to restrict a GCP project to only three services, but terraform will not comply.

### Terraform Version
```
Terraform v0.10.7 
```
### Terraform Configuration Files
```hcl

variable "gcp_services" {
  default = [
    "container.googleapis.com",
    "compute.googleapis.com",
    "servicemanagement.googleapis.com",
  ]
}

resource "google_project_services" "project" {
  project = "terraform-test-project"
  services   = "${var.gcp_services}"
}

```

### Expected Behavior
After apply the TF file above I should find the following result:
```
$ gcloud service-management list --enabled
NAME                               TITLE
compute.googleapis.com             Google Compute Engine API
container.googleapis.com           Google Container Engine API
servicemanagement.googleapis.com   Google Service Management API
```

### Actual Behavior
After running terraform apply, I'm either having those services enabled:
```
$ gcloud service-management list --enabled
NAME                               TITLE
deploymentmanager.googleapis.com   Google Cloud Deployment Manager V2 API
containerregistry.googleapis.com   Google Container Registry API
compute.googleapis.com             Google Compute Engine API
pubsub.googleapis.com              Google Cloud Pub/Sub API
container.googleapis.com           Google Container Engine API
servicemanagement.googleapis.com   Google Service Management API
storage-api.googleapis.com         Google Cloud Storage JSON API
resourceviews.googleapis.com       Google Compute Engine Instance Groups API
replicapool.googleapis.com         Google Compute Engine Instance Group Manager API
replicapoolupdater.googleapis.com  Google Compute Engine Instance Group Updater API
```

Or those services:
```
$ gcloud service-management list --enabled
NAME                              TITLE
servicemanagement.googleapis.com  Google Service Management API
compute.googleapis.com            Google Compute Engine API
```

### Steps to Reproduce
Please list the full steps required to reproduce the issue, for example:
1. `terraform init`
2. `terraform apply`

danawillow:
[...]

PR #668, opened 2017-11-01
Add new `google_project_service` resource for fine-grained service control.

### LINKED ISSUE #482: Make `google_project_services` fine-grained: add resource to enable/disable a single API
Terraform automatically adds services in some cases, and navigating in the UI enables some APIs also which conflicts with the list in Terraform.

There is already an `authoritative` parameter on `google_project_iam_policy`. Could the same be applied to `google_project_services` so that services added outside of Terraform are not removed on update?

danawillow:
Fixes #482.
Potential fix for #5 and #596.

Also adds retry logic for enabling/disabling because the tests were being too flaky without it.

### DIFF
--- a/google/resource_google_project_service.go
+++ b/google/resource_google_project_service.go
@@ -2,6 +2,7 @@ package google
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
 )
@@ -41,7 +42,7 @@ func resourceGoogleProjectServiceCreate(d *schema.ResourceData, meta interface{}
 		return fmt.Errorf("Error enabling service: %s", err)
 	}
 
-	d.SetId(srv)
+	d.SetId(projectServiceId{project, srv}.terraformId())
 	return resourceGoogleProjectServiceRead(d, meta)
 }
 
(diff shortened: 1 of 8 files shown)


google/resource_google_project_service.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+
+	srv := d.Get("service").(string)
+
+	if err = enableService(srv, project, config); err != nil {
+		return fmt.Errorf("Error enabling service: %s", err)
+	}
+
+	d.SetId(srv)
```
paddycarver: Should we include the project in this, so multiple projects can enable the same service within the same Terraform state?

danawillow: Sure! Done.

loffelmacher:
Just to confirm my understanding...I want to use appengine flex in my project, so in order to enable that API I need to "do a little dance" and rely on `google_project_service` instead of `google_project_services` and then enable AppEngine Flex and it's dependencies, in the proper order? Is that about right?

rosbo:
`google_project_service` makes sure that the API specified in that resource is enabled.

`google_project_services` makes sure ONLY the APIs specified in that resource are enabled. If some APIs are enabled and not listed in the `google_project_services` resource, they will be disabled.

For your use case, it is up to you, do you want to make sure only the APIs specified in your config are enabled? If so, use `google_project_services`. If you want to let user enabled API outside of Terraform (for instance: using gcloud or the UI), then go with `google_project_service`.

I hope this helps :)

Issue #633, opened 2017-10-26
Add support for version upgrades in `google_container_node_pool`


davidquarles:
Right now `node_version` is only accessible as a cluster-level field.  What I'd like (aka need) is the ability to trigger upgrades at the node pool level.

Furthermore, according to the [current go docs](https://godoc.org/google.golang.org/api/container/v1#ClusterUpdate), `ClusterUpdate.DesiredNodePoolId` is...

```
    // mandatory if
    // "desired_node_version", "desired_image_family"
    // or
    // "desired_node_pool_autoscaling" is specified and there is more than
    // one
    // node pool on the cluster.
```

Given the existing logic, I think this is not just missing granularity but _broken_ functionality for clusters with multiple node pools:
 ```
if d.HasChange("node_version") {
    desiredNodeVersion := d.Get("node_version").(string)

    req := &container.UpdateClusterRequest{
        Update: &container.ClusterUpdate{
            DesiredNodeVersion: desiredNodeVersion,
        },
    }
```

I _think_ this would subsume #397.

davidquarles:
@danawillow Is this an accurate assessment? Do you have any thoughts / insight / design considerations, if so? I'm still working through IP Aliasing, but I'm happy to help if this is in limbo when I'm finished.

danawillow:
Yup, that sounds accurate. Unfortunately, the way this resource was first designed seems to have tried to oversimplify the API a bit (or maybe the API's changed a bunch since then, who knows).

My initial guess would be that the `node_version` within a cluster should apply to every node pool in the cluster, and that a version attribute should be added to node pool, but I haven't thought much about it yet.

davidquarles:
The API is definitely evolving pretty rapidly, which is rad!  What you're suggesting sounds good to me.  My main question, then -- at both layers -- is how strict that attribute should be.

More verbosely:
- Should we trigger downgrades if/when the node pools have been externally upgraded
- What happens when the world moves on and the given `node_version` is no longer supported?

I do really like the existing `min_master_version` mechanism, if that abstraction makes sense here.

danawillow:
Generally, we try to make Terraform config files the source of truth if it can be. Things become really complicated if we want Terraform to only be the source of truth sometimes. This generally ends up meaning that if an attribute can be changed outside of Terraform without the user's involvement, we need a way to work around that, but if a user decided to change something that's already managed in Terraform they should understand that Terraform will try to clobber those changes.

[...]

Issue #79, opened 2017-06-13
providers/google: Support for Node Auto-Upgrade in google_container_cluster (GKE)


hashibot:
_This issue was originally opened by @cblecker as hashicorp/terraform#13391. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Relates to terraform-providers/terraform-provider-google#44

Requesting support for node auto-upgrade on GKE clusters: https://cloud.google.com/container-engine/docs/node-auto-upgrade

cc: @danawillow @paddyforan

stonith:
Also add to `google_container_node_pool` resource.

danawillow:
Thanks! Linking to https://github.com/terraform-providers/terraform-provider-google/issues/93 so we can easily keep track of all the beta FRs.

danawillow:
Fixed in #669

danawillow:
Hi @petervandenabeele, the `node_pool` block in the `google_container_cluster` resource and the `google_container_node_pool` resource share all of the same properties, so you should be able to do what you're trying to do already.

However, I would recommend using `google_container_node_pool` resources anyway, unless you need the ability to specify the exact set of node pools that you want and guarantee that there are no others (see #1712 and some other issues in this repo for why I make that recommendation). The website hasn't been updated recently, but once it is, https://github.com/terraform-providers/terraform-provider-google/pull/2474 will show a warning about using node pools in the cluster resource.

PR #669, opened 2017-11-02
Add support for `google_container_node_pool` management

### LINKED ISSUE #655: Add support for auto-{upgrade,healing} in `google_container_node_pool`
These features are all in-progress at the _cluster_ level over in #384, but I'd love to see them exposed on the node pool resource (alongside `google_container_cluster_node_pool.version`, discussed elsewhere) with support for full lifecycle management.

davidquarles:
Fixes #655.

See also: #79, #384, #633.

### DIFF
--- a/google/resource_container_node_pool.go
+++ b/google/resource_container_node_pool.go
@@ -85,6 +85,29 @@ var schemaNodePool = map[string]*schema.Schema{
 		Deprecated: "Use node_count instead",
 	},
 
+	"management": {
+		Type:     schema.TypeList,
+		Optional: true,
+		Computed: true,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+
+				"auto_repair": {
+					Type:     schema.TypeBool,
+					Optional: true,
+					Computed: true,
+				},
+
+				"auto_upgrade": {
+					Type:     schema.TypeBool,
+					Optional: true,
+					Computed: true,
+				},
+			},
+		},
+	},
+
 	"name": &schema.Schema{
 		Type:     schema.TypeString,
 		Optional: true,
@@ -301,6 +324,19 @@ func expandNodePool(d *schema.ResourceData, prefix string) (*container.NodePool,
 		}
 	}
 
+	if v, ok := d.GetOk(prefix + "management"); ok {
+		managementConfig := v.([]interface{})[0].(map[string]interface{})
+		np.Management = &container.NodeManagement{}
+
+		if v, ok := managementConfig["auto_repair"]; ok {
+			np.Management.AutoRepair = v.(bool)
+		}
+
+		if v, ok := managementConfig["auto_upgrade"]; ok {
+			np.Management.AutoUpgrade = v.(bool)
+		}
+	}
+
 	return np, nil
 }
 
@@ -331,6 +367,15 @@ func flattenNodePool(d *schema.ResourceData, config *Config, np *container.NodeP
 			map[string]interface{}{
 				"min_node_count": np.Autoscaling.MinNodeCount,
 				"max_node_count": np.Autoscaling.MaxNodeCount,
+			},
+		}
+	}
+
+	if np.Management != nil {
+		nodePool["management"] = []map[string]interface{}{
+			{
+				"auto_repair":  np.Management.AutoRepair,
+				"auto_upgrade": np.Management.AutoUpgrade,
 			},
 		}
 	}
@@ -409,6 +454,37 @@ func nodePoolUpdate(d *schema.ResourceData, meta interface{}, clusterName, prefi
 
 		if prefix == "" {
 			d.SetPartial("node_count")
+		}
+	}
+
+	if d.HasChange(prefix + "management") {
+
+		management := &container.NodeManagement{}
+		if v, ok := d.GetOk(prefix + "management"); ok {
+			managementConfig := v.([]interface{})[0].(map[string]interface{})
+			management.AutoRepair = managementConfig["auto_repair"].(bool)
+			management.AutoUpgrade = managementConfig["auto_upgrade"].(bool)
+		}
+		req := &container.SetNodePoolManagementRequest{
+			Management: management,
+		}
+		op, err := config.clientContainer.Projects.Zones.Clusters.NodePools.SetManagement(
+			project, zone, clusterName, npName, req).Do()
+
+		if err != nil {
+			return err
+		}
+
+		// Wait until it's updated
+		waitErr := containerOperationWait(config, op, project, zone, "updating GKE node pool management", timeoutInMinutes, 2)
+		if waitErr != nil {
+			return waitErr
+		}
+
+		log.Printf("[INFO] Updated management in Node Pool %s", d.Id())
+
+		if prefix == "" {
+			d.SetPartial("management")
 		}
 	}
 
--- a/google/resource_container_node_pool.go
+++ b/google/resource_container_node_pool.go
@@ -92,7 +92,6 @@ var schemaNodePool = map[string]*schema.Schema{
 		MaxItems: 1,
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
-
 				"auto_repair": {
 					Type:     schema.TypeBool,
 					Optional: true,
@@ -371,13 +370,11 @@ func flattenNodePool(d *schema.ResourceData, config *Config, np *container.NodeP
 		}
 	}
 
-	if np.Management != nil {
-		nodePool["management"] = []map[string]interface{}{
-			{
-				"auto_repair":  np.Management.AutoRepair,
-				"auto_upgrade": np.Management.AutoUpgrade,
-			},
-		}
+	nodePool["management"] = []map[string]interface{}{
+		{
+			"auto_repair":  np.Management.AutoRepair,
+			"auto_upgrade": np.Management.AutoUpgrade,
+		},
 	}
 
 	return nodePool, nil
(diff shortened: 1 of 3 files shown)


danawillow:
Hey @davidquarles, are you looking for any sort of feedback right now or just claiming your spot in fixing this? Wasn't sure what to make of the [WIP].

google/resource_container_node_pool.go
```diff
@@ -85,6 +85,29 @@ var schemaNodePool = map[string]*schema.Schema{
...
 	},
 
+	"management": {
+		Type:     schema.TypeList,
+		Optional: true,
+		Computed: true,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+
```
danawillow: nit: extra newline

google/resource_container_node_pool.go
```diff
@@ -85,6 +85,29 @@ var schemaNodePool = map[string]*schema.Schema{
 		Deprecated: "Use node_count instead",
 	},
 
+	"management": {
+		Type:     schema.TypeList,
+		Optional: true,
+		Computed: true,
```
danawillow: Is this actually computed? (note: it might be! it's just sometimes hard to tell from looking at the api, so I usually ask)

google/resource_container_node_pool.go
```diff
@@ -415,6 +460,37 @@ func nodePoolUpdate(d *schema.ResourceData, meta interface{}, clusterName, prefi
 		}
 	}
 
+	if d.HasChange(prefix + "management") {
+
```
danawillow: nit: extra newline

google/resource_container_node_pool.go
```diff
@@ -415,6 +460,37 @@ func nodePoolUpdate(d *schema.ResourceData, meta interface{}, clusterName, prefi
...
+		management := &container.NodeManagement{}
+		if v, ok := d.GetOk(prefix + "management"); ok {
+			managementConfig := v.([]interface{})[0].(map[string]interface{})
+			management.AutoRepair = managementConfig["auto_repair"].(bool)
+			management.AutoUpgrade = managementConfig["auto_upgrade"].(bool)
```
danawillow: since this api is `json:omitempty`, you'll also need to add `management.ForceSendFields = []string{"AutoRepair", "AutoUpgrade"}` so you can change values to false

google/resource_container_node_pool.go
```diff
@@ -415,6 +460,37 @@ func nodePoolUpdate(d *schema.ResourceData, meta interface{}, clusterName, prefi
...
+			return err
+		}
+
+		// Wait until it's updated
+		waitErr := containerOperationWait(config, op, project, zone, "updating GKE node pool management", timeoutInMinutes, 2)
+		if waitErr != nil {
+			return waitErr
+		}
+
+		log.Printf("[INFO] Updated management in Node Pool %s", d.Id())
```
danawillow: I'd use `npName` instead of `d.Id()` since this function gets called for `google_container_node_pool` resources as well as `google_container_cluster` ones.

google/resource_container_node_pool.go
davidquarles: I'm obviously somewhat fuzzy on the exact meaning of `computed` and when exactly to use it / not use it (especially in the context of new attributes with `ForceNew: true` being added to resources that are already out in the wild, though this is not that case), if you don't mind giving me a one-time clarification.  In this context, the field is always passed back from the API and both values are false if unspecified.  Based on my current understanding, I'd think that means the field is `optional` + `computed`, but that could be totally incorrect.  I can do a bit of testing on my own to get a better grasp.

google/resource_container_node_pool.go
davidquarles: Done.

google/resource_container_node_pool.go
davidquarles: Done.

google/resource_container_node_pool.go
davidquarles: Done.

google/resource_container_node_pool.go
davidquarles: Done.

davidquarles:
```
$ make testacc TEST=./google TESTARGS='-run=_withManagement'
TF_ACC=1 go test ./google -v -run=_withManagement -timeout 120m
=== RUN   TestAccContainerNodePool_withManagement
--- PASS: TestAccContainerNodePool_withManagement (1114.72s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	1114.747s
```

google/resource_container_node_pool.go
danawillow: The way I like to think about `Computed` is that it doesn't cause a diff if the value is set in state but not in the config. For a simple case, take `name`. We allow the user to set this field to empty, which means that terraform will generate a name for you, which we then store in state. Since we don't want the next run of `terraform plan` to show a diff (since you aren't _actually_ trying to set the name to empty), we have `name` as `Computed` so that when it's empty, nothing happens.

So in this case, if management is always returned from the api, then it will always get set in state via the flatten fn, so `Computed` is indeed the correct choice. A good example of the opposite is autoscaling- if it isn't enabled, then it doesn't get returned from the API at all (so `np.Autoscaling` gives nil), meaning we don't want it to get set in state.

The reason I like to double-check that a field needs to be`Computed` is that this empty-means-no-diff-behavior means that removing the field from your config doesn't set it to the default value, as might be expected. So for example, if we completely remove the `autoscaling` block, terraform will notice there's a diff between what's in state (autoscaling enabled) and what's in the config (no autoscaling) and will call update appropriately. For management, if (say) `auto_repair` is set to true in state, and then I remove the entire `management` block in the hopes of "removing" management, terraform won't see that as a diff and so nothing will be updated.

The way I usually handle this when developing is to just set nothing to Computed and see what fails in testing. In this particular case, since `Management` is returned from the API, I think you should leave it as `Computed` but set the two properties to be `Default: false`. That way, if they are set to true in the schema and then removed, terraform will see that it needs to set it back to false. It'll just work for those individual fields, not the full management block, but I think it's still better than nothing.

I hope that makes sense! It's taken me a long time to get to my current level of understanding so feel free to ask any more questions you might have :)

google/resource_container_node_pool_test.go
```diff
@@ -550,3 +592,25 @@ resource "google_container_node_pool" "np_with_node_config_scope_alias" {
 		oauth_scopes = ["compute-rw", "storage-ro", "logging-write", "monitoring"]
 	}
 }`, acctest.RandString(10), acctest.RandString(10))
+
+var testAccContainerNodePool_withManagement = fmt.Sprintf(`
+resource "google_container_cluster" "cluster" {
+	name = "tf-cluster-nodepool-test-%s"
+	zone = "us-central1-a"
```
danawillow: nit: mind running `terraform fmt` on this file to align all the `=`s?

google/resource_container_node_pool_test.go
```diff
@@ -550,3 +592,25 @@ resource "google_container_node_pool" "np_with_node_config_scope_alias" {
 		oauth_scopes = ["compute-rw", "storage-ro", "logging-write", "monitoring"]
 	}
 }`, acctest.RandString(10), acctest.RandString(10))
+
+var testAccContainerNodePool_withManagement = fmt.Sprintf(`
+resource "google_container_cluster" "cluster" {
+	name = "tf-cluster-nodepool-test-%s"
```
danawillow: I think you intended on this test checking update, but since the name is changing from the basic test, it'll actually delete and recreate the cluster/np

google/resource_container_node_pool.go
davidquarles: Rad!  Thanks so much for the great, thorough explanation.  That makes complete sense.  I'll make that change and re-test now.

google/resource_container_node_pool_test.go
davidquarles: I didn't actually intend to test updates, but it totally _should_.  I'll change it, re-test, and circle back shortly.

google/resource_container_node_pool_test.go
davidquarles: Unfortunately `vim-terraform` doesn't autodetect nested strings within go files.  😛 
Note that If i visually select each block and pipe out to `terraform fmt` (i.e. `:'<,'>!terraform fmt -`), I end up with two leading spaces, rather than tabs, along with various other format changes for pretty much every single nested block within the file, which is actually what I'd expect from `terraform fmt`.  Should we do this for all the tests uniformly in a separate PR?  I'll push a separate commit doing so for the individual file here, just to illustrate, which I can discard (or not).

davidquarles: `:set expandtab`|`:set noexpandtab` and `:GoFmt` both appear to have no effect, either

danawillow: Ah yeah, we don't really have a consistent story on whether we should use tabs or spaces for configs within tests. I don't really have an opinion either way, I just want the nice alignment :)
I do think though that we should either use spaces for all tests (in a given file) or tabs for all, so for now it probably makes sense to discard the commits and align by hand

davidquarles: Yeah, that seems sane.  I'm generally not a huge fan of owning hundreds of lines of code I don't necessarily understand for the sake of formatting.  😛

davidquarles:
```
make testacc TEST=./google TESTARGS='-run=_withManagement'
TF_ACC=1 go test ./google -v -run=_withManagement -timeout 120m
=== RUN   TestAccContainerNodePool_withManagement
--- PASS: TestAccContainerNodePool_withManagement (671.00s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	671.029s
```

danawillow:
Looks good, thanks @davidquarles!

PR #672, opened 2017-11-02
make it clear that the attribute represents the project id

topliceanu:
Projects in GCP have an id, a name and a project_number. All seem to uniquely identify your project. In this case it is the `id` that is required.


eilgin:
Wouldn't it be better to say "The project id [...]" ?

topliceanu:
I see, that is a good point. I'll scan the doc and see where I can find `project` being used instead of 
"project id". Thanks for getting back to me!

topliceanu:
Hi guys! Sorry about being late with this. I went through the docs and updated the resources.

I went with `the ID of the project ...`. Event though it sounds a bit forced, I think it makes more sense than `Project ID ... ` for most of the sentences. That being said, not a native speaker here.

Also, I haven't actually checked whether all these resource actually use the value of the `project` attribute as the Project ID, but I assume so, if they didn't it would be terribly inconsistent. Do I need this?

Let me know how I can improve this PR! Thanks!

PR #673, opened 2017-11-02
Fix link format

tonglil:
Currently malformed on https://www.terraform.io/docs/providers/google/r/logging_folder_sink.html.


danawillow:
Whoops, thanks @tonglil!

Issue #634, opened 2017-10-26
Validate that DNS record names end in a .


danielcompton:
It could be handy to have an additional validation on DNS names to fail early if they are missing a trailing `.`, rather than letting it go to GCP. I've had this catch me out a few times.

### Affected Resource(s)

- `google_dns_record_set`

### Expected Behavior

When the user tries to enter a domain value without a trailing `.`, Terraform raises an error locally to tell them it is not allowed.

### Actual Behavior

You will only get an error when you actually talk to GCP to try and add the record.

```
* google_dns_record_set.dmarc_txt: Error creating DNS RecordSet: googleapi: Error 400: Invalid value for 'entity.change.additions[0].name': 'text.example.com', invalid
```

danawillow:
@srivasta

srivasta:
On the face of it the requested change looks fine. The documentation: 
    https://cloud.google.com/dns/api/v1/resourceRecordSets#resource
specifies www.example.com. , and points us to RFC 1035 (section 5) and RFC 1034 (section 3.6.1) http://www.ietf.org/rfc/rfc1034.txt, which agree.

The next step is to try to reproduce this, preferably as an acceptance test, which will validate the solution.

srivasta:
(?=^.{1,254}$)(^(?:(?!\d+\.)[a-zA-Z0-9_\-]{1,63}\.?)+(?:[a-zA-Z]{1,})\.$)
Seems like a decent validation using a regexp (after appending the final period)

- http://dns-sd.org./TrailingDotsInDomainNames.html
- https://jdebp.eu./FGA/web-fully-qualified-domain-name.html

srivasta:
(please bear with me, I'm learning this as I go along)

So the acceptance test does not actually test recordset creation in a zone (dug(. It can't easi;y, since anyone should be able to run the test, and we fall into the trap of letting anyone on the internet being able to create records in a managed zone ...

So. The quickstart https://cloud.google.com/dns/quickstart is pretty explicit about adding the final . Also, I tested with the attached file, with the final dot it works, without it I do get the reported error. I am now adding tests to validate the google_dns_record_set.name configuration.

   Manoj

    resource "google_dns_managed_zone" "internal_public" {
       name     = "internal-public"
       # Note the trailing dot
       dns_name = "${vars.bare_managed_domain_name}."
    }

    resource "google_dns_record_set" "tiamat" {
[...]

PR #676, opened 2017-11-02
[validate-dns-rrset-name]: Add checks for DNS RRSet names.

srivasta:
This branch addresses "Validate that DNS record names end in a ."
Issue  #634

The documentation:
https://cloud.google.com/dns/api/v1/resourceRecordSets#resource
specifies www.example.com. , and points us to RFC 1035 (section 5) and
RFC 1034 (section 3.6.1) http://www.ietf.org/rfc/rfc1034.txt, which
agree.  The quickstart https://cloud.google.com/dns/quickstart is
pretty explicit about adding the final .

This catches most invalid DNS names, though it might not catch some
invlaid non-ASCII names it does not implement the TOASCII
functionality defined in https://tools.ietf.org/html/rfc3454

Signed-off-by: Manoj Srivastava <srivasta@google.com>


google/resource_dns_record_set.go
```diff
@@ -3,9 +3,11 @@ package google
 import (
 	"fmt"
 	"log"
+	"regexp"
+
+	dns "google.golang.org/api/dns/v1"
```
danawillow: Can you revert this change?

google/resource_dns_record_set.go
```diff
@@ -55,6 +57,101 @@ func resourceDnsRecordSet() *schema.Resource {
 	}
 }
 
+// This catches most invalid DNS names, though it might not catch some invlaid non-ASCII names
+// it does not implement the TOASCII functionality defined in https://tools.ietf.org/html/rfc3454
+// See:
+// https://tools.ietf.org/html/rfc1034 DOMAIN NAMES - CONCEPTS AND FACILITIES
+// https://tools.ietf.org/html/rfc3490 Internationalizing Domain Names in Applications (IDNA)
+// https://tools.ietf.org/html/rfc3454 Preparation of Internationalized Strings ('stringprep')
+func validate_dns_rrset_name(name string) error {
```
danawillow: golang style is to use mixedCaps for pretty much everything: https://golang.org/doc/effective_go.html#mixed-caps

google/resource_dns_record_set.go
```diff
@@ -65,6 +162,12 @@ func resourceDnsRecordSetCreate(d *schema.ResourceData, meta interface{}) error
 
 	zone := d.Get("managed_zone").(string)
 
+	// Validate the Recordset name to be added
+	err = validate_dns_rrset_name(d.Get("name").(string))
```
danawillow: If you make this a `ValidateFunc` on the `name` attribute instead of doing the check here, then it'll fail during `terraform plan` instead of during `terraform apply`

google/resource_dns_record_set_test.go
```diff
@@ -140,6 +140,42 @@ func testAccCheckDnsRecordSetExists(resourceType, resourceName string) resource.
 	}
 }
 
+func testAccCheckDnsRecordSetValid(resourceType, resourceName string) resource.TestCheckFunc {
```
danawillow: this function doesn't seem to get called from anywhere, and the inputs don't seem to get used. I think you meant to make this its own unit test- take a look at https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_dataproc_cluster_test.go#L24 for an example of how to do this.

google/resource_dns_record_set.go
```diff
@@ -55,6 +57,101 @@ func resourceDnsRecordSet() *schema.Resource {
...
+		if index+1 == len(labels) && len(label) < 1 {
+			return nil
+		}
+
+		if len(label) < 1 {
+			return fmt.Errorf("subdomains must not be empty")
+		}
+		s := []rune(label)
+		if len(s) > 63 {
+			return fmt.Errorf("subdomainss must not have more than 63 character: %d.", len(s))
```
danawillow: here and elsewhere- s/subdomainss/subdomains

danawillow: nit: characters

google/resource_dns_record_set.go
```diff
@@ -55,6 +57,101 @@ func resourceDnsRecordSet() *schema.Resource {
...
+// See:
+// https://tools.ietf.org/html/rfc1034 DOMAIN NAMES - CONCEPTS AND FACILITIES
+// https://tools.ietf.org/html/rfc3490 Internationalizing Domain Names in Applications (IDNA)
+// https://tools.ietf.org/html/rfc3454 Preparation of Internationalized Strings ('stringprep')
+func validate_dns_rrset_name(name string) error {
+	if len(name) > 255 {
+		return fmt.Errorf("DNS name length %d greater than 254", len(name))
+	}
+
+	re, err := regexp.Compile("[\u002e\u3002\uFF0E\uFF61]$")
```
danawillow: Can you add comments as to what each of these unicode codes is and why they need to be specified this way rather than just doing something like:
```
if !strings.HasSuffix(name, ".") {
  return fmt.Errorf(...)
}
```

srivasta: Added. Specifically, these are required for internationalized domain names. 

Whenever dots are used as label separators, the following characters MUST be recognized as dots: U+002E (full stop), U+3002 (ideographic full stop), U+FF0E (fullwidth full stop), U+FF61 (halfwidth ideographic full stop).

google/resource_dns_record_set.go
srivasta: I do need regular expressions in the validator. Am I missing something?

google/resource_dns_record_set.go
srivasta: Will change

google/resource_dns_record_set.go
srivasta: Thanks. Fixed

google/resource_dns_record_set.go
srivasta: Done. Had to refactor to return as full a list of errors as possible rather than the first one

google/resource_dns_record_set.go
srivasta: Also, the dns import seems to be automatically included by gofmt (and since we do have a dns.change() call there, it seems legit.  Color me confused

danawillow: Oh sorry, I just meant that we don't need dns as a named import

srivasta:
This update should ahve fixed requested changes

google/resource_dns_record_set.go
srivasta: Done

google/resource_dns_record_set.go
srivasta: Done.

google/resource_dns_record_set.go
srivasta: Done.

google/resource_dns_record_set.go
srivasta: Done.

srivasta: Done.

google/resource_dns_record_set.go
```diff
@@ -55,6 +57,147 @@ func resourceDnsRecordSet() *schema.Resource {
...
+		errors = append(errors, fmt.Errorf("DNS name length %d greater than 254",
+			len(name)))
+	}
+
+	// https://tools.ietf.org/html/rfc3490#section-3.1 Requirements
+	// Whenever dots are used as label separators, the following
+	// characters MUST be recognized as dots: U+002E (full stop),
+	// U+3002 (ideographic full stop), U+FF0E (fullwidth full
+	// stop), U+FF61 (halfwidth ideographic full stop).
+	re, err := regexp.Compile("[\u002e\u3002\uFF0E\uFF61]$")
```
danawillow: Totally up to you, but here's a way you could make this a bit more concise if you wanted:
```
if re, err := regexp.Compile("[\u002e\u3002\uFF0E\uFF61]$"); err != nil {
  errors = append(errors, fmt.Errorf("Internal error compiling regexp: %s", err))
} else if !re.MatchString(name) {
  errors = append(errors, fmt.Errorf("DNS name must be fully qualified with a trailing dot"))
}

File: google/resource_dns_record_set.go [hunk 8]
```diff
@@ -55,6 +57,147 @@ func resourceDnsRecordSet() *schema.Resource {
...
+		errors = append(errors, fmt.Errorf("Internal error compiling regexp: %s",
+			err))
+		re = nil
+	}
+	if re != nil && !re.MatchString(name) {
+		errors = append(errors,
+			fmt.Errorf("DNS name must be fully qualified with a trailing dot"))
+	}
+
+	dot_re, err := regexp.Compile("[\u002e\u3002\uFF0E\uFF61]")
```
danawillow: same comment with regards to underscores

google/resource_dns_record_set.go
```diff
@@ -55,6 +57,147 @@ func resourceDnsRecordSet() *schema.Resource {
...
+	if err != nil {
+		errors = append(errors, fmt.Errorf("Internal error compiling regexp: %s",
+			err))
+		dot_re = nil
+	}
+
+	verboten_re, err := regexp.Compile("[\x00-\x2C\x2E-\x2F\x3A-\x40\x5B-\x60\x7B-\x7F]+")
```
danawillow: Can you add a comment as to what these characters are?

google/resource_dns_record_set.go
```diff
@@ -55,6 +57,147 @@ func resourceDnsRecordSet() *schema.Resource {
...
+			errors = append(errors,
+				fmt.Errorf("DNS name must not be a top level domain."))
+			labels = nil
+		}
+	}
+
+	for index, label := range labels {
+		// The final label is empty
+		if index+1 == len(labels) && len(label) < 1 {
+			errors = append(errors, nil)
```
danawillow: I'm a bit confused about why nil is being added to the errors list

google/resource_dns_record_set.go
```diff
@@ -55,6 +57,147 @@ func resourceDnsRecordSet() *schema.Resource {
...
+			errors = append(errors,
+				fmt.Errorf("Subdomain labels cannot all be digits: %s",
+					label))
+		}
+
+		if verboten_re != nil {
+			bad_chars := verboten_re.FindAllStringSubmatch(label, -1)
+			if bad_chars != nil {
+				errors = append(errors,
+					fmt.Errorf("Illegal ASCII character(s) % q  in domain name.",
```
danawillow: looks like you accidentally got a space between the % and the q

google/resource_dns_record_set.go
srivasta: This is by design. It probably does not matter in most cases, but I'd like to keep the more expensive regular expression compilation outside the loop. I'll add a comment section that emphasizes this.

google/resource_dns_record_set.go
srivasta: Done

google/resource_dns_record_set.go
srivasta: Err. Thinko. It made sense before this became a catch-all-errors compile time check. Fixed.

google/resource_dns_record_set.go
srivasta: fixed

danawillow:
@rosbo fyi since I won't get a chance to review this before I leave

srivasta:
Is separating the regexp compilation from use a no go?

danawillow:
That part of things is fine. Have you looked at the other changes I suggested?

srivasta:
I think all other issues mentioned have been addressed

danawillow:
https://github.com/terraform-providers/terraform-provider-google/pull/676#discussion_r149513326
https://github.com/terraform-providers/terraform-provider-google/pull/676#discussion_r152412841

google/resource_dns_record_set_test.go
srivasta: forgot to mark this as done

google/resource_dns_record_set.go
srivasta: as above, thew compilation is poutside the loop

srivasta:
explicitly marked those two as done; the code had been changed.

google/resource_dns_record_set.go
```diff
@@ -56,6 +58,156 @@ func resourceDnsRecordSet() *schema.Resource {
...
+		errors = append(errors, fmt.Errorf("Internal error compiling regexp: %s",
+			err))
+		re = nil
+	}
+	if re != nil && !re.MatchString(name) {
+		errors = append(errors,
+			fmt.Errorf("DNS name must be fully qualified with a trailing dot"))
+	}
+
+	dot_re, err := regexp.Compile("[\u002e\u3002\uFF0E\uFF61]")
```
danawillow: Can you use mixedCaps for your variables please? https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

google/resource_dns_record_set_test.go
```diff
@@ -176,6 +176,40 @@ func testAccCheckDnsRecordSetExists(resourceType, resourceName string) resource.
 	}
 }
 
+func testAccCheckDnsRecordSetValid(t *testing.T) {
```
danawillow: capital T

google/resource_dns_record_set.go
srivasta: Ah, _that's_ what you meant. Thank you for your patience.

srivasta: Updated with a new set variable names

google/resource_dns_record_set_test.go
srivasta: Thank you. Updated, and it does work now

srivasta:
Rebased to fix conflict.

google/resource_dns_record_set.go
```diff
@@ -60,6 +64,163 @@ func resourceDnsRecordSet() *schema.Resource {
...
+	}
+
+	for index, label := range labels {
+		// First, the simpler checks
+
+		// Only the final label may be empty
+		if index < len(labels)-1 && len(label) < 1 {
+			errors = append(errors, fmt.Errorf("subdomains must not be empty"))
+		}
+		// errors = append(errors, fmt.Errorf("DEBUG: %d label is %s", index, label))
```
danawillow: Did you mean to include this comment?

google/resource_dns_record_set_test.go
```diff
@@ -192,6 +192,39 @@ func testAccCheckDnsRecordSetExists(resourceType, resourceName string) resource.
 	}
 }
 
+func TestAccCheckDnsRecordSetValid(t *testing.T) {
+	t.Parallel()
+	valid_names := []string{
```
danawillow: validNames (etc)

ndmckinley:
Ping: are you planning to come back around to this?  I'm happy to finish it up and get it merged if you're not going to have time for it in the near future.

sethvargo:
I have opinions on this 😄. IMO Terraform shouldn't be enforcing this detailed level of validation. This level of validation should likely be left up to the API to return a proper error. 

I would be in favor of something very simple like:

```go
if !strings.HasSuffix(name, ".") {
  name = name + "."
}
```

I think the validation of the DNS name should be left up to the API itself. If the API were to change to be more or less inclusive in the future, we wouldn't want Terraform to be a limiting factor for a customer. To put it another way: I'd rather have a user that's slightly frustrated because of a silly typo that results in a delayed error message than a customer that is completely blocked because of client-side validation of their (otherwise fine) DNS name. It makes me reminisce of websites that don't let me use a `+` in my email address and makes me very sad.

PR #692, opened 2017-11-06
Adds support for creating KMS CryptoKeys resources

amfarrell:


amfarrell:
Still need to add docs.

Pushing now so that I can see the result of running the acceptance test because I've not yet figured out access to a setup with permissions to run acceptance tests.

danawillow:
Hey @amfarrell, looks like this doesn't build. Happy to run the tests for you once that's all squared away.

tragiclifestories:
@danawillow This should now build.

Interface looks like
```hcl
resource "google_kms_key_ring" "mykeyring" {
  location = "global"
  name     = "mykeyring"
  project  = "myproject"
}

resource "google_kms_crypto_key" "mykey" {
  name     = "mykey"
  key_ring = "${google_kms_key_ring.mykeyring.id}"
}

resource "google_kms_crypto_key" "myrotatingkey" {
  name            = "myrotatingkey"
  key_ring        = "${google_kms_key_ring.mykeyring.id}"
  # Next rotation time will be set automatically to now + rotation_period
  rotation_period = "100000s"                             
}
```

tragiclifestories:
cc @mrparkers

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,265 @@
...
+
+func resourceKmsCryptoKeyDelete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	cryptoKeyId, err := parseKmsCryptoKeyId(d.Id(), config)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[WARNING] KMS CryptoKey resources cannot be deleted from GCP. This CryptoKey %s will be removed from Terraform state, but will still be present on the server.", cryptoKeyId.cryptoKeyId())
```
mrparkers: It might be helpful to add the bit about destroying all of the CryptoKey's versions to this warning message.

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,265 @@
...
+				Location: parts[1],
+				Name:     parts[2],
+			},
+			Name: parts[3],
+		}, nil
+	}
+
+	if cryptoKeyIdWithoutProjectRegex.MatchString(id) {
+		if config.Project == "" {
+			return nil, fmt.Errorf("The default project for the provider must be set when using the `{location}/{cryptoKeyName}` id format.")
```
mrparkers: You're missing the `keyRingName` in the example format

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,265 @@
...
+			KeyRingId: kmsKeyRingId{
+				Project:  config.Project,
+				Location: parts[0],
+				Name:     parts[1],
+			},
+			Name: parts[2],
+		}, nil
+	}
+
+	return nil, fmt.Errorf("Invalid CryptoKey id format, expecting `{projectId}/{locationId}/{KeyringName}/{cryptoKeyName}` or `{locationId}/{KeyringName}/{cryptoKeyName}.`")
```
mrparkers: Could you use `keyRingName` instead of `KeyringName` here?

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	d.Set("key_ring", cryptoKeyId.KeyRingId.terraformId())
+	d.Set("name", cryptoKeyId.Name)
+
+	if d.Get("purpose") == "" {
+		d.Set("purpose", "ENCRYPT_DECRYPT")
+	}
```
mrparkers: Is this needed?  `purpose` isn't defined in your resource schema.

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,373 @@
...
+		}
+
+		getCryptoKeyResponse, err := config.clientKms.Projects.Locations.KeyRings.CryptoKeys.Get(cryptoKeyId.cryptoKeyId()).Do()
+
+		if err != nil {
+			return err
+		}
+
+		_, err = time.Parse(time.RFC3339Nano, getCryptoKeyResponse.NextRotationTime)
+
```
mrparkers: I think you should also make sure `getCryptoKeyResponse.RotationPeriod` exists.  CryptoKeys are allowed to have a `NextRotationTime` without having a `RotationPeriod`, but that would be bad in this case, since you'd be creating it with both fields.

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,373 @@
...
+/*
+	KMS KeyRings cannot be deleted. This ensures that the CryptoKey resource's CryptoKeyVersion
+    sub-resources were scheduled to be destroyed, rendering the key itself inoperable.
+*/
+
+func testAccCheckGoogleKmsCryptoKeyVersionsDestroyed(projectId, location, keyRingName, cryptoKeyId string) resource.TestCheckFunc {
+	return func(_ *terraform.State) error {
+		config := testAccProvider.Meta().(*Config)
+		gcpResourceUri := fmt.Sprintf("projects/%s/locations/%s/keyRings/%s/cryptoKeys/%s", projectId, location, keyRingName, cryptoKeyId)
```
mrparkers: You could use your `&kmsCryptoKeyId.cryptoKeyId` method here.

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,373 @@
...
+
+		response, _ := config.clientKms.Projects.Locations.KeyRings.CryptoKeys.CryptoKeyVersions.List(gcpResourceUri).Do()
```
mrparkers: Could you handle the potential error here?

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,373 @@
...
+resource "google_kms_key_ring" "key_ring" {
+	project  = "${google_project.acceptance.project_id}"
+	name     = "%s"
+	location = "us-central1"
+}
+
+resource "google_kms_crypto_key" "crypto_key" {
+	name     = "%s"
+    key_ring = "${google_kms_key_ring.key_ring.id}"
+    rotation_period = "100000s"
```
mrparkers: Could you fix the formatting here?

website/docs/r/google_kms_crypto_key.html.markdown
```diff
@@ -0,0 +1,63 @@
+---
+layout: "google"
+page_title: "Google: google_kms_key_ring"
```
mrparkers: Could you fix this?

website/docs/r/google_kms_crypto_key.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+from state and delete all CryptoKeyVersions, rendering the key unusable, but **will not delete the resource on the server**.
+
+## Example Usage
+
+```hcl
+resource "google_kms_crypto_key" "my_crypto_key" {
+  name            = "my-crypto-key"
+  key_ring        = "my_project/us-central1/my-key-ring"
+  rotation_period = "100000s"
+}
```
mrparkers: It might be nice to include a `google_kms_key_ring` resource here as well, so you could demonstrate using string interpolation for the `key_ring` field.

website/docs/r/google_kms_crypto_key.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+* `id` - The ID of the created CryptoKey. Its format is `{projectId}/{location}/{keyRingName}/{cryptoKeyName}`.
+
+## Import
+
+CryptoKeys can be imported using the CryptoKey autogenerated `id`, e.g.
+
+```
+$ terraform import google_kms_key_ring.my_key_ring my-gcp-project/us-central1/my-key-ring/my-crypto-key
+
+$ terraform import google_kms_key_ring.my_key_ring us-central1/my-key-ring/my-crypto-key
```
mrparkers: The resource name should be `google_kms_crypto_key.my_crypto_key`

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,373 @@
...
+			ExpectedError: true,
+		},
+		"id is in location/keyRingName/CyptoKeyID format": {
+			ImportId:            "us-central1/test-key-ring/test-key-id",
+			ExpectedError:       false,
+			ExpectedTerraformId: "test-project/us-central1/test-key-ring/test-key-id",
+			ExpectedCryptoKeyId: "projects/test-project/locations/us-central1/keyRings/test-key-ring/cryptoKeys/test-key-id",
+			Config:              &Config{Project: "test-project"},
+		},
+		"id is in location/keyRingName/CyptoKeyID format without project in config": {
```
mrparkers: For each of these test descriptions, could you use `cryptoKeyName` instead of `CryptoKeyID`?  This way, the format will be consistent with the error messages in `parseKmsCryptoKeyId`

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,265 @@
...
+			},
+			"key_ring": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"rotation_period": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
mrparkers: I believe a CryptoKey's rotation period can be changed, or at least [there's an API for it](https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys/patch).

If this changed, `Update` would have to be implemented as well.  This change probably doesn't need to hold up the PR, but @danawillow can make that call.

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,265 @@
...
+	d.Set("name", cryptoKeyId.Name)
+
+	if d.Get("purpose") == "" {
+		d.Set("purpose", "ENCRYPT_DECRYPT")
+	}
+
+	d.SetId(cryptoKeyId.terraformId())
+
+	return []*schema.ResourceData{d}, nil
+}
```
mrparkers: It might be a good idea to set `rotation_period` if it exists.

google/resource_kms_crypto_key.go
tragiclifestories: It's one of those weird ones you get in GCP (and AWS as well) where there is a parameter that accepts an enum with one possible value ([see the REST API docs for CryptoKey](https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys#CryptoKeyPurpose)). My feeling is that it is not necessary to support this parameter as an argument to the resource, and instead just set it 'on the quiet', until there is another purpose value defined, at which point it will in any case be necessary to add more validation. But I'm new to the Terraform codebases, so perhaps it is better to be explicit but have a default?

google/resource_kms_crypto_key.go
tragiclifestories: Done - further up so comment isn't hidden ...

google/resource_kms_crypto_key.go
tragiclifestories: I did bump into this, actually, but decided to defer it.

On reflection, it might be worth doing given that you can't destroy the damn things properly...

google/resource_kms_crypto_key.go
mrparkers: It looks like its only being set during import though, not creation. I think we should either set it in both places, or neither. Like you mentioned, there's only one possible value for this enum, so I think we should just get rid of it until more are added.

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,378 @@
...
+
+resource "google_kms_key_ring" "key_ring" {
+	project  = "${google_project_services.acceptance.project}"
+	name     = "%s"
+	location = "us-central1"
+}
+
+resource "google_kms_crypto_key" "crypto_key" {
+	name     = "%s"
+    key_ring = "${google_kms_key_ring.key_ring.id}"
```
mrparkers: Could you fix this indentation?

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,378 @@
...
+resource "google_kms_key_ring" "key_ring" {
+	project  = "${google_project.acceptance.project_id}"
+	name     = "%s"
+	location = "us-central1"
+}
+
+resource "google_kms_crypto_key" "crypto_key" {
+	name     = "%s"
+    key_ring = "${google_kms_key_ring.key_ring.id}"
+    rotation_period = "%s"
```
mrparkers: Here as well

mrparkers:
@tragiclifestories sorry for the confusion about the indentation.  Most of it was fine, there were just a couple of lines where spaces were used instead of tabs.  It looks like they were all changed to use spaces, which will probably need to be changed.

The only reason I'm asking is because [I received similar comments on my PR](https://github.com/terraform-providers/terraform-provider-google/pull/518#discussion_r147258198) :wink:

tragiclifestories:
And thanks for a very thorough review - I'm quite new to mucking about with terraform internals so it's all helpful!

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,262 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	d.SetId("")
+	return nil
+}
+
+func validateKmsCryptoKeyRotationPeriod(period string) error {
```
danawillow: This could actually probably become a ValidateFunc, which would allow doing these checks during `terraform plan` instead of during `terraform apply`. Here's an example of what that would look like (this example has it inlined but it can certainly be a separate fn): https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_dataproc_cluster.go#L39

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,262 @@
...
+				Name:     parts[1],
+			},
+			Name: parts[2],
+		}, nil
+	}
+
+	return nil, fmt.Errorf("Invalid CryptoKey id format, expecting `{projectId}/{locationId}/{KeyringName}/{cryptoKeyName}` or `{locationId}/{keyRingName}/{cryptoKeyName}.`")
+}
+
+func resourceKmsCryptoKeyImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
danawillow: I think this is actually similar enough to the Read fn that you don't need a separate Import fn, but instead could use the passthrough one:
```
Importer: &schema.ResourceImporter{
	State: schema.ImportStatePassthrough,
},
```

Then inside the read function, you would do the same `d.Set` calls that you're doing here, plus an additional one for the `rotation_period`.

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,381 @@
...
+
+func TestCryptoKeyIdParsing(t *testing.T) {
+	cases := map[string]struct {
+		ImportId            string
+		ExpectedError       bool
+		ExpectedTerraformId string
+		ExpectedCryptoKeyId string
+		Config              *Config
+	}{
+		"id is in project/location/keyRingName/CryptoKeyName format": {
```
danawillow: nit: just for consistency, let's say cryptoKeyName instead of CryptoKeyName

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,381 @@
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGoogleKmsCryptoKeyWasRemovedFromState("google_kms_crypto_key.crypto_key"),
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testGoogleKmsCryptoKey_basic(projectId, projectOrg, projectBillingAccount, keyRingName, cryptoKeyName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGoogleKmsCryptoKeyExists("google_kms_crypto_key.crypto_key"),
+				),
+			},
+			resource.TestStep{
```
danawillow: Can you add a comment that you need this as a separate test step so the project sticks around? Similar (or identical) to https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_shared_vpc_test.go#L31

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,381 @@
...
+	projectOrg := os.Getenv("GOOGLE_ORG")
+	location := os.Getenv("GOOGLE_REGION")
+	projectBillingAccount := os.Getenv("GOOGLE_BILLING_ACCOUNT")
+	keyRingName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	cryptoKeyName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGoogleKmsCryptoKeyWasRemovedFromState("google_kms_crypto_key.crypto_key"),
```
danawillow: Since your second `TestStep` calls this already, this is just duplicating that check- not wrong, but also not necessary.

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,381 @@
...
+			return err
+		}
+
+		if rotationPeriod != getCryptoKeyResponse.RotationPeriod {
+			return fmt.Errorf("Expected rotation period %s to match input %s", getCryptoKeyResponse.RotationPeriod, rotationPeriod)
+		}
+
+		_, err = time.Parse(time.RFC3339Nano, getCryptoKeyResponse.NextRotationTime)
+
+		return err
```
danawillow: It might be worth it to wrap this error, something like `return fmt.Errorf("Error parsing NextRotationTime: %s", err)`

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,381 @@
...
+	KMS KeyRings cannot be deleted. This ensures that the CryptoKey resource's CryptoKeyVersion
+	sub-resources were scheduled to be destroyed, rendering the key itself inoperable.
+*/
+
+func testAccCheckGoogleKmsCryptoKeyVersionsDestroyed(projectId, location, keyRingName, cryptoKeyName string) resource.TestCheckFunc {
+	return func(_ *terraform.State) error {
+		config := testAccProvider.Meta().(*Config)
+		gcpResourceUri := fmt.Sprintf("projects/%s/locations/%s/keyRings/%s/cryptoKeys/%s", projectId, location, keyRingName, cryptoKeyName)
+
+		response, _ := config.clientKms.Projects.Locations.KeyRings.CryptoKeys.CryptoKeyVersions.List(gcpResourceUri).Do()
```
danawillow: I'd feel more comfortable if this error were handled: if we got a mysterious server error, then `response.CryptoKeyVersions` would (I believe) be empty, which means this test would succeed, even though we don't _actually_ know that the versions were destroyed.

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,381 @@
...
+resource "google_project_services" "acceptance" {
+	project = "${google_project.acceptance.project_id}"
+
+	services = [
+	  "cloudkms.googleapis.com",
+	]
+}
+
+resource "google_kms_key_ring" "key_ring" {
+	project  = "${google_project.acceptance.project_id}"
```
danawillow: Let's also use `${google_project_services.acceptance.project}` like in the basic test, so that we guarantee the API is enabled before trying to use it.

google/resource_kms_crypto_key_test.go
```diff
@@ -0,0 +1,381 @@
...
```
danawillow: here too

danawillow:
Whoops sorry @tragiclifestories, I should be thanking you!

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	log.Printf("[DEBUG] Executing read for KMS CryptoKey %s", cryptoKeyId.cryptoKeyId())
+
+	_, err = config.clientKms.Projects.Locations.KeyRings.CryptoKeys.Get(cryptoKeyId.cryptoKeyId()).Do()
+
+	if err != nil {
+		return fmt.Errorf("Error reading CryptoKey: %s", err)
+	}
+
```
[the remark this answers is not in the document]
mrparkers: Just curious, are you just looking for `rotation_period` here?  Or would you also set `name` and `key_ring` by getting it from `cryptoKeyId`?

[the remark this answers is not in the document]
tragiclifestories: I went for `rotation_period` only, since I *believe* the other fields are immutable anyway. although the docs are unclear about this.

tragiclifestories:
@danawillow I think that's everything ...

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,250 @@
...
+		result = now.UTC().Add(duration).Format(time.RFC3339Nano)
+	}
+
+	return
+}
+
+func parseKmsCryptoKeyId(id string, config *Config) (*kmsCryptoKeyId, error) {
+	parts := strings.Split(id, "/")
+
+	cryptoKeyIdRegex := regexp.MustCompile("^([a-z0-9-]+)/([a-z0-9-])+/([a-zA-Z0-9_-]{1,63})+/([a-zA-Z0-9_-]{1,63})$")
```
danawillow: Sorry I didn't catch this earlier- shouldn't the third plus not be there since it already has the {1, 63}? I think that would end up essentially overriding the {1, 63} by allowing it to repeat some number of times

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,250 @@
...
+	cryptoKeyIdWithoutProjectRegex := regexp.MustCompile("^([a-z0-9-])+/([a-zA-Z0-9_-]{1,63})+/([a-zA-Z0-9_-]{1,63})$")
```
danawillow: likewise

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,250 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+/*
+	Because KMS CryptoKey resources cannot be deleted on GCP, we are only going to remove it from state
```
danawillow: nit: this comment has mismatched tabs/spaces for indentation, mind choosing one or the other?

google/resource_kms_crypto_key.go
```diff
@@ -0,0 +1,250 @@
...
+		},
+	}
+}
+
+type kmsCryptoKeyId struct {
+	KeyRingId kmsKeyRingId
+	Name      string
+}
+
+// TODO: Add the info about rotation frequency and start time.
```
danawillow: nit: can you be a bit more specific here as to what you're adding it to? I don't think it belongs in the id, which is the section of code this comment is living in.

google/resource_kms_crypto_key.go
tragiclifestories: Fixed

danawillow:
```
$ make testacc TEST=./google TESTARGS='-run=TestCryptoKey'
TF_ACC=1 go test ./google -v -run=TestCryptoKey -timeout 120m
=== RUN   TestCryptoKeyIdParsing
=== RUN   TestCryptoKeyNextRotationCalculation
=== RUN   TestCryptoKeyNextRotationCalculation_validation
--- PASS: TestCryptoKeyNextRotationCalculation_validation (0.00s)
--- PASS: TestCryptoKeyNextRotationCalculation (0.00s)
--- PASS: TestCryptoKeyIdParsing (0.01s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	0.167s

$ make testacc TEST=./google TESTARGS='-run=TestAccGoogleKmsCryptoKey'
TF_ACC=1 go test ./google -v -run=TestAccGoogleKmsCryptoKey -timeout 120m
=== RUN   TestAccGoogleKmsCryptoKey_importBasic
=== RUN   TestAccGoogleKmsCryptoKey_basic
=== RUN   TestAccGoogleKmsCryptoKey_rotation
--- PASS: TestAccGoogleKmsCryptoKey_importBasic (58.22s)
--- PASS: TestAccGoogleKmsCryptoKey_rotation (58.62s)
--- PASS: TestAccGoogleKmsCryptoKey_basic (61.67s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	61.851s
```

Looks great, thanks so much @tragiclifestories!

tragiclifestories:
You're welcome, and thanks for the thorough review!

PR #698, opened 2017-11-07
add test for resource_google_project_iam_policy

enxebre:
Better coverage for resource_google_project_iam_policy when using default provider project

### DIFF
--- a/google/resource_google_project_iam_policy_test.go
+++ b/google/resource_google_project_iam_policy_test.go
@@ -260,6 +260,7 @@ func TestAccGoogleProjectIamPolicy_basic(t *testing.T) {
 func TestAccGoogleProjectIamPolicy_defaultProject(t *testing.T) {
 	t.Parallel()
 
+	config := testAccProvider.Meta().(*Config)
 	resource.Test(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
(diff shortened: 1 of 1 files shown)


google/resource_google_project_iam_policy_test.go
```diff
@@ -353,6 +361,15 @@ func testAccCheckGoogleProjectIamPolicyExists(projectRes, policyRes, pid string)
 	}
 }
 
+func testAccCheckGoogleDefaultProjectIamPolicyIsMerged(projectRes, policyRes string) resource.TestCheckFunc {
```
rosbo: Instead of creating a new method, you can call the other method like this:

```go
config := testAccProvider.Meta().(*Config)
testAccCheckGoogleProjectIamPolicyIsMerged("google_project_iam_policy.acceptance", "data.google_iam_policy.admin", config.Project)
```

rosbo:
Thank you Alberto, just a minor comment and we are good to go :)

enxebre:
thanks for pointing that out! should be good now

rosbo:
My suggestion was crashing, I forgot that `testAccProviders.Meta()` is nil before the `Configure` method is called by `resource.Test()`.

I pushed a quick fix to your branch.

Thank you for improving the tests!

PR #701, opened 2017-11-07
Add docs about upgrading to README

paddycarver:
@nodesocket opened #338 to document upgrading to the README. This just moves that around a bit to put it below the section on using the provider, and points to the Terraform website as the source of truth.


danawillow:
LGTM 👍

PR #702, opened 2017-11-08
Add support to google_compute_target_pool for health checks self_link

rosbo:
Instead of requiring the health check name. It now accepts the self_link or the name.

I added:
- Accepts self_link in addition of health check name
- Removes the need for an API call to generate the self link
- Improves the documentation to mention that only the legacy `google_compute_http_health_check` is supported. This will prevent our user from being stuck like mentioned here: #300.
- Adds a `MaxItems:1` in the schema. You can't have more than one. The API will fail. The official docs also says so.
- Adds a check to the acceptance test to ensure the health checks are properly setup.

### DIFF
--- a/google/resource_compute_target_pool_test.go
+++ b/google/resource_compute_target_pool_test.go
@@ -91,7 +91,7 @@ func testAccCheckComputeTargetPoolHealthCheck(targetPool, healthCheck string) re
 
 		hcLink := healthCheckRes.Primary.Attributes["self_link"]
 		if targetPoolRes.Primary.Attributes["health_checks.0"] != hcLink {
-			return fmt.Errorf("Health check not set up. Expected '%s'", hcLink)
+			return fmt.Errorf("Health check not set up. Expected %q", hcLink)
 		}
 
 		return nil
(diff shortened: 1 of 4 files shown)


rosbo:
```sh
TF_ACC=1 go test ./google -v -run TestAccComputeTargetPool_basic -timeout 120m
=== RUN   TestAccComputeTargetPool_basic
--- PASS: TestAccComputeTargetPool_basic (72.02s)
PASS
```

google/resource_compute_target_pool_test.go
```diff
@@ -73,6 +77,27 @@ func testAccCheckComputeTargetPoolExists(n string) resource.TestCheckFunc {
...
+		}
+
+		healthCheckRes, ok := s.RootModule().Resources[healthCheck]
+		if !ok {
+			return fmt.Errorf("Not found: %s", healthCheck)
+		}
+
+		hcLink := healthCheckRes.Primary.Attributes["self_link"]
+		if targetPoolRes.Primary.Attributes["health_checks.0"] != hcLink {
+			return fmt.Errorf("Health check not set up. Expected '%s'", hcLink)
```
danawillow: fyi: '%s' is equivalent to %q (up to you if you want to change it)

rosbo: Will do

PR #713, opened 2017-11-09
Fix the description of the compute instance group

eilgin:
Since this resource is targeting unmanaged instances i've changed the description to better fit its purpose 😄


rosbo:
Thanks!

PR #721, opened 2017-11-10
Correction and clarification of service_account_key docs

tragiclifestories:
Great to have this resource available; playing around with it, I came across a couple of issues with the documentation.

- `key_algorithm` description was erroneous, and actually had information about `private_key_type`
- The `pgp_key` argument also accepts literal key bytes, so I thought that was worth mentioning for people who aren't using vault.


rosbo:
Thank you @tragiclifestories!

PR #724, opened 2017-11-10
Set UseLegacySql to true for compatibility with the BigQuery API

jimmy-btn:
This is required for compatibility with the default value of this field in BigQuery. Without this set, Terraform will always detect that there has been a change to resources that do not explicitly set this field.


danawillow:
Thanks @jimmy-btn!

PR #737, opened 2017-11-13
Fix timeout issue with project services

### LINKED ISSUE #722: Timeout when enabling the compute API
### Terraform Version
```
terraform -v
Terraform v0.10.8
```

### Affected Resource(s)
- google_project_service

```hcl
resource "google_project_service" "compute" {
  project = "${google_project.default.id}"
  service = "compute.googleapis.com"
}
```

### Expected Behavior
```
google_project_service.services: Creation complete after XXs (ID: foo)
```

### Actual Behavior
Activating the "compute" API is sometimes taking a lot of time (more than the max allowed time which seems to be 1 minute):
```
* google_project_service.services: Error enabling service: Error enabling service "compute.googleapis.com" for project "XXX": timeout while waiting for state to become 'success' (timeout: 1m0s)
```

We should allow this resource to set the [`timeout` parameter](https://www.terraform.io/docs/configuration/resources.html#timeouts)

### Steps to Reproduce
1. `terraform plan -out=tfplan`
1. `terraform apply tfplan`

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- [GH-718](https://github.com/terraform-providers/terraform-provider-google/issues/718)

danawillow:
Apparently the retry function (which was set to retry for one minute) was conflicting with our operation wait (which was set to retry for ten minutes) and winning. I looked through the code and it really seems like that the first iteration through the retry it should wait for the function to return, which would take the full ten minutes (or whenever the api is done getting enabled) but this at least fixes it.

I definitely agree that we should have configurable timeouts, but it's a bit complicated with project_services because it makes as many API calls as there are APIs to enable or disable in a given run, and we usually do a timeout per create/read/update/delete function and I don't want it to be misleading. We could pretty easily add it to the fine-grained project_service resource, but it's not as urgent now that the timeout is actually getting set to the value we had set it to originally.

I'll go ahead and say this fixes #722.

### DIFF
--- a/google/resource_google_project_services.go
+++ b/google/resource_google_project_services.go
@@ -205,7 +205,7 @@ func enableService(s, pid string, config *Config) error {
 
 func disableService(s, pid string, config *Config) error {
 	dsr := newDisableServiceRequest(pid)
-	err := retry(func() error {
+	err := retryTime(func() error {
 		sop, err := config.clientServiceMan.Services.Disable(s, dsr).Do()
 		if err != nil {
 			return err
--- a/google/resource_google_project_services.go
+++ b/google/resource_google_project_services.go
@@ -186,7 +186,7 @@ func getApiServices(pid string, config *Config, ignore map[string]struct{}) ([]s
 
 func enableService(s, pid string, config *Config) error {
 	esr := newEnableServiceRequest(pid)
-	err := retry(func() error {
+	err := retryTime(func() error {
 		sop, err := config.clientServiceMan.Services.Enable(s, esr).Do()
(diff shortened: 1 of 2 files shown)


rosbo:
Can't build the test...

arekkas:
https://github.com/terraform-providers/terraform-provider-google/pull/737/files#diff-0dd957e3bf9b308fcd63fcd3659900faR208 should be `retryTime` not `retry`

danawillow:
whoops, fixed.

Issue #132, opened 2017-06-16
Add support for folders


ubschmidt2:
### Terraform Version
Terraform v0.10.0-dev (9706042ddd51b85a46e23356536a1d812efa5a38+CHANGES)

### Affected Resource(s)
- google_project

### References
- https://cloud.google.com/resource-manager/docs/creating-managing-folders
- https://cloud.google.com/resource-manager/reference/rest/v2beta1/folders

danawillow:
This is blocked on https://github.com/terraform-providers/terraform-provider-google/issues/93, but is absolutely on our radar.

rileykarson:
Looking at https://cloud.google.com/resource-manager/docs/creating-managing-folders, it looks like this has entered GA.

ubschmidt2:
Identified use cases:

- affecting the existing resource type "google_project":
  - create a new GCP project in a given existing GCP folder.
    (Implementation note: I'd suggest to add a "folder_id" attribute to "google_project". This should only require a minor change to the call to v1/projects/create. If "folder_id" is set, then parent ResourceId.type="folder".)
  - move a TF-managed GCP project between folders and between org node and folder.
    (Implementation note: This should be just a matter of updating the project with the appropriate parent ResourceId.)
- for configuring GCP folders:
  - create a new GCP folder (+destroy)
    (Implementation note: uses the resource-manager v2/folders API. New TF resource type "google_folder".)
  - move a TF-managed GCP folder into a given existing GCP folder or to the org node.
    (Implementation note: see the v2/folders/move API.)
  - configure the IAM policy of a GCP folder.
    (Implementation note: This looks very similar to what already exists for "google_project". However, since IAM policies can be set on many different resource types, it might make sense to revisit the current approach, maybe it can be generalized a bit so that it works for other resource types, too.)
  - configure constraints on a GCP folder.
    (Implementation note: Constraints can be set on the org node, folder(s), and the project, thus need a solution that works (or could be extended to work) for all levels.)

rosbo:
I created 4 new issues to track the different use cases we identified to make it easier to track progress.

This task will be kept open until all children issues are resolved.

#411 is being tackled by @ubschmidt2 & @srivasta
#412 is being tackled by @rosbo
I (@rosbo) will probably take over #413 once I am done with #412.

#414 is up for grab. I might take it if I have the bandwidth once I am done with #412 and #413

cc/ @danawillow

PR #738, opened 2017-11-14
Add folder data source

nlopes:
It is common to split infrastructure in GCP by environments, where all are under the same
folder. You can think of it this way: you have a folder named "EnvA" that contains two
different projects, each project managed by a filesystem folder in the terraform
directory.

Currently, one would have to hardcode the folder name in the variables file (as an option)
after it had been created. With this data source it becomes possible to reference the
folder with a given display name (this has to been unique within the same parent).

### DIFF
--- a/google/data_source_google_folder.go
+++ b/google/data_source_google_folder.go
@@ -1,0 +1,59 @@
+package google
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	resourceManagerV2Beta1 "google.golang.org/api/cloudresourcemanager/v2beta1"
+	"google.golang.org/api/googleapi"
+)
+
+func dataSourceGoogleFolder() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGoogleFolderRead,
+
+		Schema: map[string]*schema.Schema{
+			"parent": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"display_name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceGoogleFolderRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	parent := d.Get("parent").(string)
+	displayName := d.Get("display_name").(string)
+
+	queryString := fmt.Sprintf("lifecycleState=ACTIVE AND parent=%s AND displayName=%s", parent, displayName)
+	searchRequest := &resourceManagerV2Beta1.SearchFoldersRequest{
+		Query: queryString,
+	}
+	searchResponse, err := config.clientResourceManagerV2Beta1.Folders.Search(searchRequest).Do()
+	if err != nil {
+		if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 404 {
+			return fmt.Errorf("Folder Not Found : %s", d.Get("name"))
+		}
+
+		return fmt.Errorf("Error reading folders: %s", err)
+	}
+
+	folders := searchResponse.Folders
+	if len(folders) != 1 {
+		return fmt.Errorf("More than one folder found")
+	}
+
+	d.SetId(folders[0].Name)
+	d.Set("name", folders[0].Name)
+	return nil
+}
--- a/google/data_source_google_folder_test.go
+++ b/google/data_source_google_folder_test.go
@@ -1,0 +1,79 @@
+package google
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccDataSourceGoogleFolder(t *testing.T) {
+	skipIfEnvNotSet(t,
+		[]string{
+			"GOOGLE_ORG",
+		}...,
+	)
+
+	parent := fmt.Sprintf("organizations/%s", os.Getenv("GOOGLE_ORG"))
+	displayName := "terraform-test-" + acctest.RandString(10)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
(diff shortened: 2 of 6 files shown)


nlopes:
```bash
» make testacc TEST=./google TESTARGS='-run=TestAccDataSourceGoogleFolder' GOOGLE_USE_DEFAULT_CREDENTIALS=true

TF_ACC=1 go test ./google -v -run=TestAccDataSourceGoogleFolder -timeout 120m
=== RUN   TestAccDataSourceGoogleFolder
--- PASS: TestAccDataSourceGoogleFolder (14.82s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	14.845s
```

nlopes:
Part of better support for folders (#132)

rosbo:
Hi,

I am hesitant about adding this data source as it would lead to unexpected behavior if two folders have the same display name. In a large organization, it is easy for someone to create a new folder without knowing about the other and it would break all your config. I had the same hesitation about adding a google_project data source that would also search for display name. See discussion in #699 for more details.

In the end, I think what you want to achieve would be better and safer using [terraform_remote_state](https://www.terraform.io/docs/providers/terraform/d/remote_state.html) (assuming you are storing your state file remotely).

1. You have a central project creating the folders.
1. Each projects under an environment referenced to the folder via the remote state:

Central project TF config:
```hcl
resource "google_folder" "envA" {
  display_name = "Environment A"
  parent     = "organizations/1234567"
}

output "folder_name" {
  value = "${google_folder.envA.name}"
}
```

Project 1 in envA TF config:
```hcl
data "terraform_remote_state" "central" {
   ... (add the config to connect to the remote state here)
}

resource "google_project" "project1" {
 name = "Project 1
 project-id = "project-1"
 folder_id = "${data.terraform_remote_state.central.folder_name}
}
```

rosbo:
After looking more closely, the folder display name has to be unique for a given parent. So, I am less hesistant to add. However, the question is whether it is useful given you can achieve something similar with the `terraform_remote_state`. Let me know if the approach I described above works for you.

The main benefit of the terraform_remote_state is that you can use it to retrieve arbitrary name, id and so on from any other `terraform project`.

nlopes:
That's a good shout and I understand the reasoning. I'm not too fussed about dropping this in favour of the remote state if you feel that's the right approach. I'll still raise that I feel a bit weird about going across state files for something that could be an explicit data source referencing the component I want. 

I don't think it increases security scope but that data remote state will include way more than I'd like just to reach for a folder id which makes me a bit queasy.

rosbo:
Given that the name is guaranteed to be unique, our users won't accidentally mess up their setup by creating a second folder with the same name. For this reason, I think it is reasonable to add this data source. 

Reviewing...

google/data_source_google_folder.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+}
+
+func dataSourceGoogleFolderRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	parent := d.Get("parent").(string)
+	displayName := d.Get("display_name").(string)
+
+	queryString := fmt.Sprintf("lifecycleState=ACTIVE AND parent=%s AND displayName=%s", parent, displayName)
```
rosbo: could you rename the data source to `google_active_folder` to be explicit about the fact that you can only fetch active folder.

google/data_source_google_folder_test.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func TestAccDataSourceGoogleFolder(t *testing.T) {
+	skipIfEnvNotSet(t,
+		[]string{
+			"GOOGLE_ORG",
+		}...,
+	)
+
+	parent := fmt.Sprintf("organizations/%s", os.Getenv("GOOGLE_ORG"))
```
rosbo: you can use the global test variable `org` for this.

google/data_source_google_folder_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccDataSourceGoogleFolder(t *testing.T) {
+	skipIfEnvNotSet(t,
```
rosbo: this can be simplified to `skipIfEnvNotSet(t, "GOOGLE_ORG)`

website/docs/d/datasource_folder.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+layout: "google"
+page_title: "Google: google_folder"
+sidebar_current: "docs-google-datasource-folder"
+description: |-
+  Get a folder within GCP.
+---
+
+# google\_folder
+
+Get a folder within GCP by display_name and parent.
```
rosbo: add tick around `display_name` and `parent`

google/data_source_google_folder.go
nlopes: Will do, that makes a lot of sense.

google/data_source_google_folder_test.go
nlopes: Can you point me to the right place? Where does that live? I don't seem to be able to find it.

google/data_source_google_folder.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	resourceManagerV2Beta1 "google.golang.org/api/cloudresourcemanager/v2beta1"
+	"google.golang.org/api/googleapi"
+)
+
+func dataSourceGoogleFolder() *schema.Resource {
```
rosbo: Could you rename this method to match the new resource name along with the read method

google/data_source_google_folder_test.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccDataSourceGoogleFolder(t *testing.T) {
```
rosbo: And all these methods here too

rosbo:
```sh
--- PASS: TestAccDataSourceGoogleActiveFolder (13.58s)
```

rosbo:
Merged! Thank you Norberto for your contribution :)

PR #740, opened 2017-11-15
New datasource: google_container_cluster

### LINKED ISSUE #732: GKE cluster data source
### Feature Description
It would be great to have a data source that can output the auth information for a GKE cluster so that it could be plugged into the Kubernetes provider.

This can currently be achieved with the external provider calling the gcloud sdk to retrieve the auth information, but it would be nice to cut out the extra setup steps.

sl1pm4t:
Adds the `google_kubernetes_cluster` datasource.

This is useful for retrieving the cluster credentials and feeding into the Terraform Kubernetes provider, but has other use cases (e.g. feed cluster CIDR range into firewall/VPC route resources).

I've named the datasource `google_kubernetes_cluster` rather than `google_container_cluster` to match the new Google product name, but this means the datasource naming is not consistent with the equivalent resource. Let me know if this is acceptable.

Fixes #732


sl1pm4t:
Hi @danawillow - any chance we could get this reviewed before it gets stale? Thanks!

danawillow:
Hey @sl1pm4t, thanks for the PR! I've been drowning in code reviews lately but I'll try to get to this before the Thanksgiving break (and of course don't let that stop you from continuing to contribute!)

Right now I think keeping the data source name consistent with the resource name is probably the way to go though.

sl1pm4t:
No worries @danawillow, I felt bad asking because I know you're swamped. I really appreciate your massive effort!

sl1pm4t:
Hi @danawillow - I've renamed the datasource and fixed a couple other issues I spotted in the tests, and missing reference to the docs page in `google.erb`.
Thanks!

google/data_source_google_container_cluster.go
```diff
@@ -0,0 +1,187 @@
...
+
+			"additional_zones": {
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"cluster_ipv4_cidr": {
+				Type:     schema.TypeString,
+				Optional: true,
```
danawillow: Except for `name`, `zone`, and `project`, we probably shouldn't allow setting fields since they don't get used to make the request

google/data_source_google_container_cluster.go
```diff
@@ -0,0 +1,187 @@
...
+				Computed: true,
+			},
+
+			"master_auth": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
```
danawillow: Likewise, I don't think `ForceNew` really has meaning for data sources

google/data_source_google_container_cluster.go
```diff
@@ -0,0 +1,187 @@
...
+
+			"node_version": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func datasourceContainerClusterRead(d *schema.ResourceData, meta interface{}) error {
```
danawillow: I'm curious why you decided to have only a subset of the resource schema fields in the data source schema. If you had the same set of fields, then you would be able to reuse the read function (https://github.com/terraform-providers/terraform-provider-google/blob/master/google/data_source_google_compute_instance_group.go is a good example of this). If there is a reason why certain schema elements are excluded, I'd love to see a comment explaining why.

google/data_source_google_container_cluster.go
sl1pm4t: Good point. Will fix this.

google/data_source_google_container_cluster.go
sl1pm4t: Great idea, I didn't considered using the resources Read func, but that makes a lot of sense.
I stripped down the schema to a subset that I thought would be useful in a datasource, based on the notion if it wasn't going to be useful, remove it so there's less to maintain in the read func. Now that you've pointed out we can re-use the read func from the resource it makes sense to keep the whole schema.

sl1pm4t:
Hi @danawillow - sorry for the delay fixing this up.

As per your suggestion, the datasource now uses the existing `google_container_cluster` resource read func rather than being re-implemented for the datasource. 
It also implements the full resource schema rather than just a subset. This is done by generating the datasource schema directly from the resource schema (see `datasourceSchemaFromResourceSchema` in `google/utils.go`) so future maintenance should be minimal if the attributes are added / modified / deleted on the resource.

Thanks!

sl1pm4t:
I rebased on master branch to resolve conflicts.

google/utils.go
```diff
@@ -344,3 +344,65 @@ func extractFirstMapConfig(m []interface{}) map[string]interface{} {
...
 	return m[0].(map[string]interface{})
 }
+
+// datasourceSchemaFromResourceSchema is a recursive func that
+// converts an existing Resource schema to a Datasource schema.
+// All schema elements are copied, but certain attributes are ignored or changed:
+// - all attributes have Computed = true
+// - all attributes have ForceNew, Required = false
+// - Validation funcs and attributes (e.g. MaxItems) are not copied
+func datasourceSchemaFromResourceSchema(rs map[string]*schema.Schema) map[string]*schema.Schema {
```
danawillow: Great idea, thanks so much for adding this! For a bit more visibility within the provider, it might make more sense to move these helper fns into their own file (datasource_helpers.go or something like that)

sl1pm4t:
@danawillow - thanks for the review. The requested changes have been implemented. 👨‍💻

Issue #495, opened 2017-09-29
Cloud KMS secret data source


rochdev:
It would be nice to have a `google_kms_secret` data source similar to [aws_kms_secret](https://www.terraform.io/docs/providers/aws/d/kms_secret.html). This would allow managing secrets with Cloud KMS.

danawillow:
Hey @rochdev, Google's Cloud KMS doesn't store secrets, it encrypts them and allows them to be stored elsewhere. Would a storage bucket / bucket object data source work for your use case? If you could give us a bit more detail about your use case and how the various resources you're planning on using would be interacting, this could help us figure out how to approach this :)

(also in case you haven't seen, #441 is tracking KMS resources)

rochdev:
The use case would be to encrypt secrets with Cloud KMS, which could then be used as a data source in the Terraform files.

Basically it would be: encrypt a secret using Cloud KMS, put that secret as the payload of the data source directly in the .tf file, and when using it the data source would decrypt is automatically using KMS.

See the [aws_kms_secret](https://www.terraform.io/docs/providers/aws/d/kms_secret.html) documentation for an example using the AWS equivalent.

#441 is for the Terraform resources to actually create and manage Cloud KMS resources (which I am also looking forward to :), but this feature request is for a data source.

I understand Cloud KMS was not built for this purpose (secrets are not necessarily keys), but since it's able to encrypt small payloads I think it would still be great to have this feature.

rochdev:
Just as an example of how we have achieved this functionality right now with the `external` provider:

##### example.tf
```hcl
data "external" "secret" {
  program = ["${path.module}/example.sh", "${var.encrypted_secret}"]
}
```

##### example.sh
```sh
secret=$(echo "$1" | base64 --decode | gcloud kms decrypt \
  --ciphertext-file=- \
  --plaintext-file=- \
  --key=key \
  --keyring=keyring \
  --location=global
)

jq -nMcr --arg secret "$secret" '{"secret":$secret}'
```

What we would be looking for instead would be something along these lines:

```hcl
data "google_kms_secret" "secret" {
  key         = "key"
  keyring     = "keyring"
  location    = "global"
  ciphertext  = "${var.encrypted_secret}"
}

// then use ${data.google_kms_secret.secret.plaintext}
```

`${var.encrypted_secret}` is a secret encrypted with Cloud KMS and then base64 encoded.

[...]

PR #741, opened 2017-11-15
Adds google_kms_secret data source

mrparkers:
Discussion in #495 

cc @rochdev and @tragiclifestories who asked for this.

Example:

```hcl
resource "google_kms_key_ring" "key_ring" {
  name     = "kms-secret-keyring"
  location = "us-central1"
}

resource "google_kms_crypto_key" "crypto_key" {
  name     = "kms-secret-key"
  key_ring = "${google_kms_key_ring.key_ring.id}"
}

# After running `gcloud kms encrypt`
data "google_kms_secret" "secret" {
  crypto_key = "${google_kms_crypto_key.crypto_key.id}"
  ciphertext = "CiQAWm5esXYu3WnKrDRi5zY63pw4cZl13InzSGcx4Z7t/6NHneYSOAD3jiTonMNNnUb+iRsD79ir2L91Kfe2VTgD988ixe/VawV/nLHXKD1UOPyXoeJzQWCb96DEluMs"
}

resource "google_sql_database_instance" "master" {
  name = "master-instance"

  settings {
    tier = "D0"
  }
}

resource "google_sql_user" "users" {
  name     = "me"
  instance = "${google_sql_database_instance.master.name}"
  host     = "me.com"
  password = "${data.google_kms_secret.secret.plaintext}"
}
```

Acceptance test:
```
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccGoogleKmsSecret_basic -timeout 120m
=== RUN   TestAccGoogleKmsSecret_basic
--- PASS: TestAccGoogleKmsSecret_basic (55.97s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	55.995s
testing: warning: no tests to run
PASS
ok  	github.com/terraform-providers/terraform-provider-google/scripts	0.026s [no tests to run]
```

### DIFF
--- a/google/data_source_google_kms_secret.go
+++ b/google/data_source_google_kms_secret.go
@@ -1,0 +1,123 @@
+package google
+
+import (
+	"google.golang.org/api/cloudkms/v1"
+
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+	"log"
+	"regexp"
+	"strings"
+	"encoding/base64"
+	"time"
+)
+
+func dataSourceGoogleKmsSecret() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGoogleKmsSecretRead,
+		Schema: map[string]*schema.Schema{
+			"crypto_key": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"ciphertext": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"plaintext": &schema.Schema{
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+			},
+		},
+	}
+}
+
+type kmsCryptoKeyId struct {
+	KeyRingId kmsKeyRingId
+	Name      string
+}
+
+// TODO: Add the info about rotation frequency and start time.
+
+func (s *kmsCryptoKeyId) cryptoKeyId() string {
+	return fmt.Sprintf("%s/cryptoKeys/%s", s.KeyRingId.keyRingId(), s.Name)
+}
+
+func (s *kmsCryptoKeyId) parentId() string {
+	return s.KeyRingId.keyRingId()
+}
+
+func (s *kmsCryptoKeyId) terraformId() string {
+	return fmt.Sprintf("%s/%s", s.KeyRingId.terraformId(), s.Name)
+}
+
+func parseKmsCryptoKeyId(id string, config *Config) (*kmsCryptoKeyId, error) {
+	parts := strings.Split(id, "/")
+
+	cryptoKeyIdRegex := regexp.MustCompile("^([a-z0-9-]+)/([a-z0-9-])+/([a-zA-Z0-9_-]{1,63})+/([a-zA-Z0-9_-]{1,63})$")
+	cryptoKeyIdWithoutProjectRegex := regexp.MustCompile("^([a-z0-9-])+/([a-zA-Z0-9_-]{1,63})+/([a-zA-Z0-9_-]{1,63})$")
+
+	if cryptoKeyIdRegex.MatchString(id) {
+		return &kmsCryptoKeyId{
+			KeyRingId: kmsKeyRingId{
+				Project:  parts[0],
+				Location: parts[1],
+				Name:     parts[2],
+			},
+			Name: parts[3],
+		}, nil
+	}
+
+	if cryptoKeyIdWithoutProjectRegex.MatchString(id) {
+		if config.Project == "" {
+			return nil, fmt.Errorf("The default project for the provider must be set when using the `{location}/{keyRingName}/{cryptoKeyName}` id format.")
+		}
+
+		return &kmsCryptoKeyId{
+			KeyRingId: kmsKeyRingId{
+				Project:  config.Project,
+				Location: parts[0],
+				Name:     parts[1],
+			},
+			Name: parts[2],
+		}, nil
+	}
+
+	return nil, fmt.Errorf("Invalid CryptoKey id format, expecting `{projectId}/{locationId}/{KeyringName}/{cryptoKeyName}` or `{locationId}/{keyRingName}/{cryptoKeyName}.`")
+}
+
+func dataSourceGoogleKmsSecretRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	cryptoKeyId, err := parseKmsCryptoKeyId(d.Get("crypto_key").(string), config)
+
+	if err != nil {
+		return err
+	}
+
+	ciphertext := d.Get("ciphertext").(string)
+
+	kmsDecryptRequest := &cloudkms.DecryptRequest{
+		Ciphertext: ciphertext,
+	}
+
+	decryptResponse, err := config.clientKms.Projects.Locations.KeyRings.CryptoKeys.Decrypt(cryptoKeyId.cryptoKeyId(), kmsDecryptRequest).Do()
+
+	if err != nil {
+		return fmt.Errorf("Error decrypting ciphertext: %s", err)
+	}
+
+	plaintext, err := base64.StdEncoding.DecodeString(decryptResponse.Plaintext)
+
+	if err != nil {
+		return fmt.Errorf("Error decoding base64 response: %s", err)
+	}
+
+	log.Printf("[INFO] Successfully decrypted ciphertext: %s", ciphertext)
+
+	d.Set("plaintext", fmt.Sprintf("%s", plaintext))
+	d.SetId(time.Now().UTC().String())
+
+	return nil
+}
--- a/google/data_source_google_kms_secret.go
+++ b/google/data_source_google_kms_secret.go
@@ -60,7 +60,7 @@ func dataSourceGoogleKmsSecretRead(d *schema.ResourceData, meta interface{}) err
 
 	log.Printf("[INFO] Successfully decrypted ciphertext: %s", ciphertext)
 
-	d.Set("plaintext", fmt.Sprintf("%s", plaintext))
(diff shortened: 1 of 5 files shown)


website/docs/d/google_kms_secret.html.markdown
```diff
@@ -0,0 +1,93 @@
...
+> | base64
+CiQAqD+xX4SXOSziF4a8JYvq4spfAuWhhYSNul33H85HnVtNQW4SOgDu2UZ46dQCRFl5MF6ekabviN8xq+F+2035ZJ85B+xTYXqNf4mZs0RJitnWWuXlYQh6axnnJYu3kDU=
+```
+
+Finally, reference the encrypted ciphertext in your resource definitions:
+
+```hcl
+data "google_kms_secret" "sql_user_password" {
+  crypto_key = "${google_kms_crypto_key.my_crypto_key.id}"
+  ciphertext = "CiQAqD+xX4SXOSziF4a8JYvq4spfAuWhhYSNul33H85HnVtNQW4SOgDu2UZ46dQCRFl5MF6ekabviN8xq+F+2035ZJ85B+xTYXqNf4mZs0RJitnWWuXlYQh6axnnJYu3kDU="
```
tragiclifestories: Nitpick: maybe this base64 text should be truncated? It's going to spill out of the `<code></code>` container on the doc site ... 

`ciphertext = "C1QAqD+x ... "` would probably do.

google/data_source_google_kms_secret_test.go
```diff
@@ -0,0 +1,104 @@
...
+					})
+
+					return nil
+				},
+			},
+		},
+	})
+}
+
+func testAccEncryptSecretDataWithCryptoKey(s *terraform.State, resourceName, plaintext string) (string, *kmsCryptoKeyId, error) {
```
tragiclifestories: Can we rename `resourceName` to refer to the fact that it's a CryptoKey? Given that something else is actually under test I think it would clarify things.

google/data_source_google_kms_secret.go
```diff
@@ -0,0 +1,67 @@
...
+		Ciphertext: ciphertext,
+	}
+
+	decryptResponse, err := config.clientKms.Projects.Locations.KeyRings.CryptoKeys.Decrypt(cryptoKeyId.cryptoKeyId(), kmsDecryptRequest).Do()
+
+	if err != nil {
+		return fmt.Errorf("Error decrypting ciphertext: %s", err)
+	}
+
+	plaintext, err := base64.StdEncoding.DecodeString(decryptResponse.Plaintext)
```
tragiclifestories: Shouldn't these two calls be the other way around - ie, base64 decode the ciphertext first, and then send the ciphertext to KMS?

rochdev:
LGTM functionality wise

google/data_source_google_kms_secret.go
mrparkers: Nope.  `decrypt` expects the encrypted data that was originally returned from `encrypt`, which is already in base64, and it returns some plaintext that is base64 encoded.

[Docs](https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys/decrypt)

website/docs/d/google_kms_secret.html.markdown
mrparkers: I agree that you'd probably have to scroll right a bunch to see the whole thing, but I think that a complete example without truncated strings is worth that cost.

The [docs](https://www.terraform.io/docs/providers/aws/d/kms_secret.html) for `aws_kms_secret` also displays the entire encrypted string.

mrparkers: I could, however, encrypt a shorter string.  That might make it easier to read.

google/data_source_google_kms_secret.go
tragiclifestories: Looks on closer inspection like I got confused by a difference between how the gcloud CLI handles this (ie, it spits raw bytes out to the terminal) and how the REST API does (ie, it base64 encodes everything). Sorry about that.

mrparkers: No worries!

rosbo:
Hi Michael,

Coding wise, your solution is ready to be merged.

However, we are reluctant to accept this change because we think it encourages a bad practice security wise. As you wrote in the documentation for this data source, the plaintext password is saved in the Terraform state (on operators and potentially CI server machines). This defeats the whole purpose of using KMS in the first place. It gives the user a false sense of security...

I can see that this is better than putting the plaintext password in the configuration file itself. At least, the code repository doesn't store the plaintext password.

Have you considered using Vault to dynamically provision your db users/passwords?
https://www.vaultproject.io/docs/secrets/databases/mysql-maria.html

I am happy to discuss this further :)

mrparkers:
Hi @rosbo, thank you for the response.

One of the reasons I decided to add this was because the AWS provider has a [similar data source](https://www.terraform.io/docs/providers/aws/d/kms_secret.html).  I didn't have an immediate need for this data source when I started working on it, but after working with terraform a bit more, I've started to see how a data source like this could be useful.

Perhaps I have a misunderstanding of Terraform state, but I've always treated it as sensitive and encrypted it using Google Cloud KMS before storing it in version control.  This is because all of my projects have GKE clusters, and the Terraform state file includes the master's username and password.  There are probably other examples of this, but as soon as I saw this, I started encrypting my state files.

The problem this data source solves for me is that it removes the need for me to encrypt my terraform variables file.  I have a specific variables file reserved for secrets that could store things like database passwords or pre-shared keys for Cloud VPN, and those are also encrypted with KMS before storing them in version control.

I understand that moving towards a solution like Vault is probably the best way to go.  Unfortunately, I just haven't had the time at my company to set this up, especially when our applications are already using KMS and it's extremely easy to use it for this purpose as well.

I definitely understand your concerns, and I'm not opposed to having this PR closed because of them.  The workflow I'm using now (encrypting terraform variable files) works until I move to something more permanent like Vault.

wyardley:
I personally also find this very useful (though didn't realize the limitation about it being in the state in cleartext). If there's the issue with this, why does the AWS provider have a kms data source?

(unrelated, also looks like there's a merge conflict now).

rochdev:
I agree that the tfstate already contains a lot of sensitive information and should be handled accordingly (until Hashicorp release a better solution for tfstate secrets). I don't see any other security problems with this approach.

wyardley:
FWIW, the equivalent AWS one has this at the top of the docs now:
```
NOTE: Using this data provider will allow you to conceal secret data within your resource definitions but does not take care of protecting that data in the logging output, plan output or state output.
```
maybe a warning like that? It would be nice if resources could be marked sensitive so they're not printed and / or encrypted somehow within the state.

rochdev:
I think they can be marked sensitive (there are quite a few resources that don't display secrets in plans). But there is no way currently to encrypt secrets in the tfstate.

mrparkers:
@wyardley I included a similar warning in the documentation: https://github.com/terraform-providers/terraform-provider-google/pull/741/files#diff-c558c1ef972405ca1d0adcc58f6487b9R17

wyardley:
@rosbo it would seem like using Vault as a backend currently may have some of the same limitations anyway?
https://www.terraform.io/docs/providers/vault/index.html
There are some options like https://github.com/orange-cloudfoundry/terraform-secure-backend that encrypt the backend.

tragiclifestories:
Additionally Google cloud storage encrypts at rest as standard, so using the GCS backend would offer some protection. Then you're left with keeping it out of logs etc.

I no longer have a use case for this, but it seems to me possible to mitigate the security risks on the model of the equivalent aws resource.

rosbo:
@danawillow I am on the fence about this one. What do you think?

danawillow:
I think that given there's precedent for a similar resource in the AWS provider, a way to encrypt state files using remote state, and a prominent warning message on the documentation page, I'm a cautious yes.

mrparkers:
I can fix the merge conflicts here if there's a desire to get this merged.  I've been holding off since I wasn't sure.

google/data_source_google_kms_secret_test.go
```diff
@@ -0,0 +1,104 @@
...
+	t.Parallel()
+
+	skipIfEnvNotSet(t,
+		[]string{
+			"GOOGLE_ORG",
+			"GOOGLE_BILLING_ACCOUNT",
+		}...,
+	)
+
+	projectOrg := os.Getenv("GOOGLE_ORG")
```
rosbo: Please use these instead and delete the skipIfEnvNotSet block above:
```go
projectOrg := getTestOrgFromEnv(t)
projectBillingAccount := getTestBillingAccountFromEnv(t)
```

mrparkers:
@rosbo I rebased to fix the conflicts and included your suggestion.

rosbo:
Failing build:
`google/data_source_google_kms_secret_test.go:13:2: imported and not used: "os"`

mrparkers:
Yup, my bad.  I fixed it.

rosbo:
`--- PASS: TestAccGoogleKmsSecret_basic (60.84s)`

Gabology:
Would you accept a PR on a data source for encrypting the plaintext using Google KMS as well? It's quite a logical flow that you would want to create the KMS keyring and crypto key and then encrypt some secret with that. [Like how they are doing in the GCP Terraform Vault template now.](https://github.com/GoogleCloudPlatform/terraform-google-vault/blob/master/main.tf#L227)

rosbo:
Storing the plaintext secret in your code repo is a bad idea but if the secret is the output of some other resource like in the example for the GCP Terraform Vault template you linked above, I see some value and would accept it.

shree1988:
How to fetch encrypted certificates file from Google cloud storage to deploy vault instead of using self sign certificate with terraform code?

PR #746, opened 2017-11-15
Add documentation for import in google_folder_organization_policy

rosbo:
Import support is not useful since running terraform apply is simpler and the result is the same since no new resource is created.

The documentation did not include anything about import.

Also, the `org_id` field should be `ForceNew`


rosbo:
I agree this might not be intuitive for the users that create and import have the same effect. 

All our other iam/policy resources also do not support import because we can achieve the same by simply running `terraform apply`. However, I do agree with you that this might not be intuitive for our users and that for consistency sake we should support import in the `google_organization_policy` and other iam resources. I filed an issue to add import support for the other resources: #756

rosbo:
I reverted the change to remove the import functionality and added instructions on how to use the import function for this resource.

PR #747, opened 2017-11-15
Add new resource google_folder_organization_policy

### LINKED ISSUE #414: Configure constraints on a GCP folder
Blocked by #412 

Implementation note: Constraints can be set on the org node, folder(s), and the project, thus need a solution that works (or could be extended to work) for all levels.

rosbo:
Fixes #414

### DIFF
--- a/google/resource_google_folder_organization_policy_test.go
+++ b/google/resource_google_folder_organization_policy_test.go
@@ -13,8 +13,7 @@ import (
 func TestAccGoogleFolderOrganizationPolicy_boolean(t *testing.T) {
 	t.Parallel()
 
-	skipIfEnvNotSet(t, "GOOGLE_ORG")
-
+	org := getTestOrgFromEnv(t)
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
@@ -51,8 +50,7 @@ func TestAccGoogleFolderOrganizationPolicy_boolean(t *testing.T) {
 func TestAccGoogleFolderOrganizationPolicy_list_allowAll(t *testing.T) {
 	t.Parallel()
 
-	skipIfEnvNotSet(t, "GOOGLE_ORG")
-
+	org := getTestOrgFromEnv(t)
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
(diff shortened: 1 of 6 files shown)


rosbo:
Rebased with the latest changes.

google/resource_google_folder_organization_policy_test.go
```diff
@@ -0,0 +1,311 @@
...
+		if err != nil {
+			return err
+		}
+
+		if len(policy.ListPolicy.AllowedValues) > 0 || len(policy.ListPolicy.DeniedValues) > 0 {
+			return fmt.Errorf("The `values` field shouldn't be set")
+		}
+
+		if policy.ListPolicy.AllValues != policyType {
+			return fmt.Errorf("Expected the list policy to '%s' all values, got '%s'", policyType, policy.ListPolicy.AllValues)
```
danawillow: I'm not totally sure what this is supposed to say, can you reword it?

google/resource_google_folder_organization_policy_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+	return config.clientResourceManager.Folders.GetOrgPolicy(rs.Primary.Attributes["folder"], &cloudresourcemanager.GetOrgPolicyRequest{
+		Constraint: rs.Primary.Attributes["constraint"],
+	}).Do()
+}
+
+func testAccGoogleFolderOrganizationPolicy_boolean(org string, enforced bool) string {
+	return fmt.Sprintf(`
+resource "google_folder" "orgpolicy" {
+  display_name = "%s"
```
danawillow: nit: indentation isn't consistent between this block and the next

google/resource_google_folder_organization_policy_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+resource "google_folder_organization_policy" "bool" {
+	folder     = "${google_folder.orgpolicy.name}"
+	constraint = "constraints/compute.disableSerialPortAccess"
+
+	boolean_policy {
+		enforced = %t
+	}
+}
+`, acctest.RandomWithPrefix("tf-test"), "organizations/"+org, enforced)
```
danawillow: Even though you _can_ rename a folder, let's use a consistent name per-test (rather than per-test-step) so we guarantee we're testing only the org policy behavior.

google/resource_google_folder_organization_policy_test.go
rosbo: Wooahh. I don't even know what I meant. Reworded. 

FYI, policyType is either ALLOW or DENY.

google/resource_google_folder_organization_policy_test.go
rosbo: Done

google/resource_google_folder_organization_policy_test.go
rosbo: Done

Issue #742, opened 2017-11-15
Cannot create autoscaling clusters without setting (initial_)node_count


drzero42:
### Terraform Version
 - Terraform v0.10.8
 - Terraform-provider-google v1.2.0

### Affected Resource(s)
- google_container_cluster

If I want to create a cluster where the default node pool has autoscaling enabled, I have to specify the `node_pool`, because the `autoscaling` sub is part of the `node_pool` schema. However, when autoscaling is enabled and min/max node count is set, it makes no sense to set `node_count` (with `initial_node_count` being deprecated), as that value changes when the cluster autoscales. Even though `node_count` is marked as optional, the `expandNodePool` function throws this error message: `Node pool default-pool cannot be set with 0 node count` if I don't set `node_count`.
I suggest that `expandNodePool` also checks if autoscaling is enabled, and don't complain about `node_count == 0` if it is. Right now, I basically have to keep using `initial_node_count`, if I want this to work. Maybe have the code automatically set `initial_node_count = 1` or something, when autoscaling is enabled.

Example HCL that will trigger this problem:
```
resource "google_container_cluster" "cluster" {
  name = "cluster"
  zone = "europe-west1-b"

  node_pool {
    name = "default-pool"
    autoscaling {
      min_node_count = 0
      max_node_count = 5
    }
  }
}
```

danawillow:
Fixed in #752.

PR #752, opened 2017-11-16
Allow node pools with size 0

birdayz:
So far the provider forbids empty node pools, however from a google API point of view it appears to be possible to have node pools with size 0 (works for me as of today). This makes it possible to ramp down a node pool without destroying it. I am very happy for feedback if somebody can foresee if this might cause any problems.


danawillow:
Fixes #742. Thanks @birdayz!

PR #760, opened 2017-11-17
the GOOGLE_XPN_HOST_PROJECT variable is no longer mandatory

### LINKED ISSUE #715: Remove `GOOGLE_XPN_HOST_PROJECT` env variable
Even though this is only used in one test, we obnoxiously require it to be set when running any test. Since we have XPN support now, we can actually set up an XPN project in the test that uses this variable so we don't need it anymore. Or if we don't feel like doing that, we can at the very least only check it in the test where it gets used.

eilgin:
Since this variable is only used for 3 tests, it doesn't make sense to declare it as mandatory in the test config. Therefore, concerned tests are now skipped if the variable is absent (the same behavior is used for `GOOGLE_ORG`).

This PR partially closes #715 as we still need to create a project and declare its ID in this variable (_at least, we won't need to include this var for each tests!_).

### DIFF
--- a/google/provider_test.go
+++ b/google/provider_test.go
@@ -70,10 +70,6 @@ func testAccPreCheck(t *testing.T) {
 	if v := multiEnvSearch(regionEnvVars); v != "us-central1" {
 		t.Fatalf("One of %s must be set to us-central1 for acceptance tests", strings.Join(regionEnvVars, ", "))
 	}
-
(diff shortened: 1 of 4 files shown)


eilgin:
Tests are correctly skipped and don't alter the rest of the run but since i'm using an IDE (GoLand), i couldn't capture the console output...

rosbo:
Thanks

PR #766, opened 2017-11-20
Get test org and billing account in a consistent way

rosbo:
Fixes the following issues:
- Consistent behavior: Some tests were skipped if GOOGLE_ORG wasn't set, some would just fail (hard to debug sometimes, the error message would be obscure but then you realize that the org variable was empty.).
- Avoid duplication: The GOOGLE_ORG string was all over the place, if we ever wanted to support two different env variables like for credentials or project, it would be a mess
- Avoid mistake: Typo in GOOGLE_ORG is easy.
- Stop using (and removed) the global `org` variable.

Same applies for billing.

### DIFF
--- a/google/data_source_google_active_folder_test.go
+++ b/google/data_source_google_active_folder_test.go
@@ -10,7 +10,7 @@ import (
 )
 
 func TestAccDataSourceGoogleActiveFolder(t *testing.T) {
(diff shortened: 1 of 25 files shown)


rosbo:
@danawillow The consistent behavior helps me add more coverage to our concourse pipeline.

PR #778, opened 2017-11-22
Make our acceptance tests easier to write and faster to run

rosbo:
Proposition:
* Remove import specific tests (like TestAccComputeGlobalAddress_importBasic)
* Add an import verify step at the end of each regular test cases
* For update tests, we can add import step in between or keep a targeted TestCheckFunc for the updated field. We should be consistent about which approach we choose.
* Remove TestCheckFunc functions.

Advantages:
* Tests run faster by not having to run a test specifically for import (avoid the resource creation delay of an extra resource only to be used by the import test)
* Import is tested for fields (right now, most of the time we only test the basic case which doesn't cover all the fields)
* Less tedious to write acceptance tests. No need to write the boilerplace TestCheckFunc, no need to write a specific test for import.
* All fields in the schema are checked. Right now, we only check the value stored by GCP for few fields via specific TestCheckFunc.

Do we lose coverage? Short answer, no.

By creating a resource and adding an import step at the end we ensure that
* the resource exists since we can import it. Importing reads from the API (testAccCheck*Exists are not needed)
* GCP stores all the value we expect. The import step would fail if GCP didn't store the value we expected since their would be a diff post-import.

For instance, let's say we forgot to set the IpVersion in the create method:
```go
&compute.Address{
  Name:      d.Get("name").(string),
  // Forgot to set this: IpVersion: d.Get("ip_version").(string),
}
```

The test fails with:
```sh
Step 0 error: After applying this step, the plan was not empty:

DIFF:

DESTROY/CREATE: google_compute_global_address.foobar
  address:    "35.186.195.40" => "<computed>"
  ip_version: "" => "IPV6" (forces new resource)
  name:       "address-test-lasybhzfno" => "address-test-lasybhzfno"
  self_link:  "https://www.googleapis.com/compute/v1/projects/[REDACTED]/global/addresses/address-test-lasybhzfno" => "<computed>"

STATE:

google_compute_global_address.foobar:
  ID = address-test-lasybhzfno
  address = 35.186.195.40
  ip_version =
  name = address-test-lasybhzfno
  self_link = https://www.googleapis.com/compute/v1/projects/[REDACTED]/global/addresses/address-test-lasybhzfno
```
The step index in the error message (here 0) tells us that their is a bug in the create method.
 
Another example, assume we forgot to set a field in the read method:
```go
// Forgot to set this: d.Set("ip_version", addr.IpVersion)
```

The test fails with:
```sh
testing.go:435: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

(map[string]string) {
}

(map[string]string) (len=1) {
 (string) (len=10) "ip_version": (string) (len=4) "IPV6"
}
```

The step index in the error message (here 1), tells us that it failed in the import step. Probably means that we should check what is going on in the import/read method.


google/resource_compute_global_address.go
```diff
@@ -96,7 +96,7 @@ func resourceComputeGlobalAddressRead(d *schema.ResourceData, meta interface{})
 	}
 
 	d.Set("name", addr.Name)
-	d.Set("ip_version", addr.IpVersion)
+	//d.Set("ip_version", addr.IpVersion)
```
danawillow: Don't forget this ;)

rosbo: Good catch ;) Fixed

paddycarver:
I've been waffling on this one a bit. Here's my hot take:

* I see no benefit to having the `_basic` tests and the `_import` tests. I think combining them is fine, as `_import` generally already combines them.
* I really just waffle when it comes to replacing TestCheckFuncs with import. Not so much about the loss of coverage, but about the edge cases:
  * Not every resource is or can be importable. Some have information provided at create time that isn't available at read time.
  * Import would be doing more work than strictly necessary, as it would be deleting the state and redownloading all the state between each step, instead of just checking the state.
  * Not every test we want to write worke with "delete and re-bootstrap the state" interlaved throughout it. Errors borne out of certain progressions of state, for example, would be harder to test.

I think the "deprecate import-specific tests" part of this makes perfect sense and is unobjectionable. I think the "deprecate TestCheckFuncs" part of this is a much trickier and more nuanced conversation to have, and I don't know that the gains of it are worth the work.

rosbo:
We got the green light for doing this :) Closing this old PR and we will open new PR progressively to follow this pattern.

PR #779, opened 2017-11-22
WIP: Ability to add/remove node pools from a GKE cluster

danawillow:
This mostly works, with several large caveats:
- You can't add a node pool in the middle of a list, only at the end
- Name prefixes / empty names don't work as expected
- I haven't written any test cases yet, just manual ones

As such, I'm not looking for feedback now, but putting this out there just as a "hey I'm working on this" PR, and also so that nobody ends up duplicating this work.

The real solution for the first point is to convert `node_pool` into a set. The rest I could hypothetically fix now, but I'm going to hold off on new changes until node pools are a Set.

### DIFF
--- a/google/node_config.go
+++ b/google/node_config.go
@@ -10,7 +10,6 @@ var schemaNodeConfig = &schema.Schema{
 	Type:     schema.TypeList,
 	Optional: true,
 	Computed: true,
-	ForceNew: true,
 	MaxItems: 1,
 	Elem: &schema.Resource{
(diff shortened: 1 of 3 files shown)


rileykarson:
Is this outdated @danawillow?

danawillow:
Ah yes very. I'll probably update again later with all the reasons why this is a bad idea, but I'll go ahead and close it for now.

rileykarson:
Sounds good! #780 is probably a better home for updates, I think you pointed to it from #1712 as well.

PR #781, opened 2017-11-22
IAM resources for KMS KeyRIng and CryptoKey

### LINKED ISSUE #762: Support IAM policy bindings for KMS Keyring and CryptoKey resources
As things stand, the only way to manage cloudkms roles is at the project level, which is very coarse grained - especially given the actual function of cryptographic keys in a software system ;-)

Proposal is to have the same sort of interface as in similar resources (eg #481).

```hcl
resource "google_kms_key_ring" "top_secret" {
  name     = "top-secret"
  location = "global"
  project  = "my-project-123"
}

data "google_iam_policy" "top_secret" {
  binding {
    role = "roles/cloudkms.encryptDecrypt"

    members = [
      "ordinary.joe@foo.com",
      "ordinary.jane@foo.com",
    ]
  }
}

resource "google_kms_key_ring_iam_binding" "top_secret" {
  key_ring = "${google_kms_key_ring.top_secret.id}"

  policy_data = "${data.google_iam_policy.top_secret.policy_data}"
}
```

tragiclifestories:
Resolves #762. Many thanks to @rosbo for making it much easier to add this sort of thing.

### DIFF
--- a/google/iam_kms_crypto_key.go
+++ b/google/iam_kms_crypto_key.go
@@ -41,9 +41,7 @@ func (u *KmsCryptoKeyIamUpdater) GetResourceIamPolicy() (*cloudresourcemanager.P
 		return nil, fmt.Errorf("Error retrieving IAM policy for %s: %s", u.DescribeResource(), err)
 	}
 
-	cloudResourcePolicy := &cloudresourcemanager.Policy{}
-
-	err = Convert(p, cloudResourcePolicy)
+	cloudResourcePolicy, err := kmsToResourceManagerPolicy(p)
 
 	if err != nil {
 		return nil, fmt.Errorf("Invalid IAM policy for %s: %s", u.DescribeResource(), err)
@@ -53,8 +51,7 @@ func (u *KmsCryptoKeyIamUpdater) GetResourceIamPolicy() (*cloudresourcemanager.P
 }
 
 func (u *KmsCryptoKeyIamUpdater) SetResourceIamPolicy(policy *cloudresourcemanager.Policy) error {
-	kmsPolicy := &cloudkms.Policy{}
-	err := Convert(policy, kmsPolicy)
+	kmsPolicy, err := resourceManagerToKmsPolicy(policy)
 
 	if err != nil {
 		return fmt.Errorf("Invalid IAM policy for %s: %s", u.DescribeResource(), err)
--- a/google/iam_kms_crypto_key.go
+++ b/google/iam_kms_crypto_key.go
@@ -1,0 +1,84 @@
+package google
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/cloudkms/v1"
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
+var IamKmsCryptoKeySchema = map[string]*schema.Schema{
+	"crypto_key_id": {
+		Type:     schema.TypeString,
+		Required: true,
+		ForceNew: true,
+	},
+}
+
+type KmsCryptoKeyIamUpdater struct {
+	resourceId string
+	Config     *Config
+}
+
+func NewKmsCryptoKeyIamUpdater(d *schema.ResourceData, config *Config) (ResourceIamUpdater, error) {
+	cryptoKey := d.Get("crypto_key_id").(string)
+	cryptoKeyId, err := parseKmsCryptoKeyId(cryptoKey, config)
+
+	if err != nil {
+		return nil, fmt.Errorf("Error parsing resource ID for for %s: %s", cryptoKey, err)
+	}
+
+	return &KmsCryptoKeyIamUpdater{
+		resourceId: cryptoKeyId.cryptoKeyId(),
+		Config:     config,
+	}, nil
+}
+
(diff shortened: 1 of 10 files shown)


google/iam_kms_crypto_key.go
```diff
@@ -0,0 +1,84 @@
...
+func (u *KmsCryptoKeyIamUpdater) GetResourceIamPolicy() (*cloudresourcemanager.Policy, error) {
+	p, err := u.Config.clientKms.Projects.Locations.KeyRings.CryptoKeys.GetIamPolicy(u.resourceId).Do()
+
+	if err != nil {
+		return nil, fmt.Errorf("Error retrieving IAM policy for %s: %s", u.DescribeResource(), err)
+	}
+
+	cloudResourcePolicy := &cloudresourcemanager.Policy{}
+
+	err = Convert(p, cloudResourcePolicy)
```
rosbo: Consider extracting the conversion logic in a function and reuse it for the keys and the rings. See:
https://github.com/terraform-providers/terraform-provider-google/blob/master/google/iam_folder.go#L85

google/resource_google_kms_crypto_key_iam_test.go
```diff
@@ -0,0 +1,218 @@
...
+	"fmt"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"reflect"
+	"sort"
+	"testing"
+)
+
+// Bindings and members are tested serially to avoid concurrent updates of the org's IAM policy.
```
rosbo: You can remove this comment, it only applies to organizations. We cannot create organization on-the-fly the same way we create a different crypto key for each test.

You can split the iam_member test from the iam_binding test. Since a different crypto key is created for each, there won't be concurrent update of the IAM policy between the tests.

I envision two tests: `TestAccGoogleKmsCryptoKeyIamMember` and `TestAccGoogleKmsCryptoKeyIamBinding`. You can also put them in two separate test files.

google/resource_google_kms_key_ring_iam_test.go
```diff
@@ -0,0 +1,203 @@
...
+	"fmt"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"reflect"
+	"sort"
+	"testing"
+)
+
+// Bindings and members are tested serially to avoid concurrent updates of the org's IAM policy.
```
rosbo: Same comment than for crypto keys

website/docs/r/google_kms_crypto_key_iam_binding.html.markdown
```diff
@@ -0,0 +1,47 @@
+---
```
rosbo: You have to update also the `website/google.erb` file to add an entry in the left menu panel

website/docs/r/google_kms_key_ring_iam_binding.html.markdown
```diff
@@ -0,0 +1,46 @@
+---
```
rosbo: No doc entry for the `member` resource?

tragiclifestories:
Thanks. Extra Todo so I remember tomorrow - the tests need to run in their own project so the created resources can be cleaned up after (keyrings etc are immutable).

website/docs/r/google_kms_key_ring_iam_binding.html.markdown
tragiclifestories: They''re in now.

website/docs/r/google_kms_crypto_key_iam_binding.html.markdown
tragiclifestories: Added. There are now enough KMS resources in there that it might be a good idea to pull out another submenu - what do you think? 

Also, how exactly do I build these docs? I'm basically slamming changes in there and hoping it works at this point ...

tragiclifestories:
I haven't actually been able to run the latest version of the acceptance tests as GCP will not allow me to create any more projects. If someone could fire off a test run when they have a moment, that would be just ducky - I do not anticipate that they will fail, but it's possible that I haven't got all the needed services enabled in the configs.

rosbo:
All tests are passing.

```sh
--- PASS: TestAccGoogleKmsCryptoKeyIamBinding (86.97s)
--- PASS: TestAccGoogleKmsCryptoKeyIamMember (82.93s)
--- PASS: TestAccGoogleKmsKeyRingIamMember (79.73s)
--- PASS: TestAccGoogleKmsKeyRingIamBinding (88.74s)
```

website/docs/r/google_kms_crypto_key_iam_binding.html.markdown
rosbo: I am planning on reorganizing the docs around IAM next week. I want to discuss it with @paddycarver and @danawillow  first. I would say for now it is fine the way you did it.

About building the docs, @paddycarver do we have documentation somewhere about how to easily build the docs?

rosbo: @tragiclifestories The documentation on how to build the website is there:
https://github.com/hashicorp/terraform-website

tragiclifestories:
Thanks man.

PR #783, opened 2017-11-23
Update tier setting to use db-f1-micro

nathanleclaire:
The existing setting of "D0" caused an error for me when running
`terraform apply`. This fixes the issue and allows the apply to proceed.

ping @danawillow

### DIFF
--- a/website/docs/r/sql_database_instance.html.markdown
+++ b/website/docs/r/sql_database_instance.html.markdown
@@ -26,7 +26,7 @@ resource "google_sql_database_instance" "master" {
   name = "master-instance"
 
   settings {
-    tier = "D0"
+    tier = "db-f1-micro"
   }
 }
 ```


nathanleclaire:
And thanks for the work on this integration btw!

danawillow:
Ok, so there are a few things going on here.

* The example is just bad generally because it doesn't provide a region, which is a required field.
* v1 vs v2 databases live in different regions: https://cloud.google.com/sql/docs/mysql/instance-locations. The error you got was likely because you tried to create it in a region that only supports second-gen instances (apologies for the bad error message, I just reported it to the team)
* Changing the tier will just cause the same issue for different regions

So instead of this change, let's have two separate examples- one for a v1 database (tier D0), and another for a v2 database (tier db-f1-micro). And each one can set the region to something that's valid for that tier. We should probably also have more fleshed-out examples in general (instead of one that's just required fields), but we can do that another day if you don't feel like doing it now.

How does this sound, @nathanleclaire?

nathanleclaire:
Sure, sounds good. You want me to update this PR to have the two examples you mention?

danawillow:
If you don't mind, that would be great!

nathanleclaire:
Thanks @rosbo , I dropped the 🏀 on this one!

PR #784, opened 2017-11-23
Store project and region in state

rosbo:
As of now, we store the project in region in the state for some resources and not for others. 

This PR addresses this by consistently marking the project and region fields as `Computed: true` and storing it in the state inside the read method.

The main benefits apart from being consistent is that in acceptance tests, we can rely on the project property being set instead of having to do the same fallback logic that we do in the resource again (i.e. check in schema, if not set fallback to config.Project).

### DIFF
--- a/google/resource_runtimeconfig_variable.go
+++ b/google/resource_runtimeconfig_variable.go
@@ -72,15 +72,11 @@ func resourceRuntimeconfigVariableCreate(d *schema.ResourceData, meta interface{
 	}
 	d.SetId(createdVariable.Name)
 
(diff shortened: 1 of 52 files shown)


google/resource_runtimeconfig_variable.go
```diff
@@ -183,11 +184,7 @@ func setRuntimeConfigVariableToResourceData(d *schema.ResourceData, project stri
 	}
 	d.Set("name", name)
 	d.Set("parent", parent)
-
-	if varProject != project {
-		d.Set("project", varProject)
-	}
-
+	d.Set("project", varProject)
```
danawillow: nit: the project parameter is unused now

PR #793, opened 2017-11-27
Validate account_id length in google_service_account

### LINKED ISSUE #791: google_service_account account_id should be between 6 and 30 characters
### Terraform Version
v0.11.0

### Affected Resource(s)
- google_service_account

```hcl
resource "google_service_account" "test" {
  account_id   = "test"
  display_name = "test"
}
```.

### Expected Behavior
Terraform should warn that account_id should be between 6 and 30 characters.

### Actual Behavior
gcloud return bad argument error without saying what the actual error is and what attribute caused the error.

### Steps to Reproduce
1. `terraform apply`

### PR DESCRIPTION
**Pull request by rosbo**
Fixes #791 

Failing at plan time is always better than failing at apply time whenever we can. Additionally, the error returned by the API when the account_id is incorrect is unhelpful: `Error creating service account: googleapi: Error 400: Request contains an invalid argument., badRequest`

cc/ @benbro

### DIFF
--- a/google/resource_google_service_account.go
+++ b/google/resource_google_service_account.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 	"google.golang.org/api/iam/v1"
 )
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: google/resource_google_service_account.go [hunk 1]
```diff
@@ -33,9 +34,10 @@ func resourceGoogleServiceAccount() *schema.Resource {
 				Computed: true,
 			},
 			"account_id": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(6, 30),
```
danawillow: Since we do have some more information on what a valid id is, let's just go ahead and check the regexp: https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts/create
> It is unique within a project, must be 6-30 characters long, and match the regular expression [a-z]([-a-z0-9]*[a-z0-9]) to comply with RFC1035.

rosbo: Done

danawillow:
Looks like this broke TestAccGoogleServiceAccountKey_basic and TestAccGoogleServiceAccountKey_pgp

PR #789, opened 2017-11-24
Detect changes to cloud storage object when using source field.

rosbo:
Fixes #788 

Detect changes to the file when using the `source` field whether the local file content is updated or the file stored on GCP is updated outside of Terraform.

cc/ @rodcloutier

PR #798, opened 2017-11-29
Detect changes to local file or changes made outside of Terraform to the file stored on the server.

rosbo:
Fixes bug introduced by #789.
Add support to detect changes to the file stored on the server outside of Terraform when the `content` field is used. Before, it only worked if you were using the `source` field to specify the object content.

The bug because it is not possible to reference an optional field if the value is not set in the config when the resource is not already in the state. By changing `md5hash` from Computed to Optional, we prevented this valid use case.

The following test was failing because of that:

We were trying to access the "${google_storage_bucket_object.story.md5hash}" field. Since the state is empty at the beginning of the test. We can't reference the md5hash optional field.
```
TestAccStorageSignedUrl_accTest
        * data.google_storage_object_signed_url.story_url_w_md5: Resource 'google_storage_bucket_object.story' does not have attribute 'md5hash' for variable 'google_storage_bucket_object.story.md5hash'
```

### DIFF
--- a/google/resource_storage_bucket_object.go
+++ b/google/resource_storage_bucket_object.go
@@ -80,28 +80,7 @@ func resourceStorageBucketObject() *schema.Resource {
 
 			"md5hash": &schema.Schema{
 				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
-				// Makes the diff message nicer:
-				// md5hash:       "1XcnP/iFw/hNrbhXi7QTmQ==" => "different hash" (forces new resource)
-				// Instead of the more confusing:
-				// md5hash:       "1XcnP/iFw/hNrbhXi7QTmQ==" => "" (forces new resource)
-				Default: "different hash",
-				// If `content` is used, always suppress the diff. Equivalent to Computed behavior.
-				// If `source` is used:
-				// 1. Compute the md5 hash of the local file
-				// 2. Compare the computed md5 hash with the hash stored in Cloud Storage
-				// 3. Don't suppress the diff iff they don't match
-				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					// `old` is the md5 hash we retrieved from the server in the ReadFunc
-					if source, ok := d.GetOkExists("source"); ok {
-						if old != getFileMd5Hash(source.(string)) {
-							return false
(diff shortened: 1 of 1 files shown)


google/resource_storage_bucket_object.go
```diff
@@ -79,6 +79,26 @@ func resourceStorageBucketObject() *schema.Resource {
 			},
 
 			"md5hash": &schema.Schema{
+				Type:     schema.TypeString,
```
rosbo: The diff is not easy to read. I basically reverted "md5hash" field back to its previous state.

And created a new field "detect_md5hash". This field is the one that will detect the changes. I improved the DiffSuppressFunc to also detect changes when the remote file changes and the `content` field is used.

rosbo:
```sh
--- PASS: TestAccGoogleStorageObject_content (3.76s)
--- PASS: TestAccGoogleStorageObject_storageClass (3.85s)
--- PASS: TestAccGoogleStorageObject_basic (3.94s)
--- PASS: TestAccGoogleStorageObject_withContentCharacteristics (3.98s)
--- PASS: TestAccGoogleStorageObject_cacheControl (4.08s)
--- PASS: TestAccGoogleStorageObject_recreate (6.34s)
--- PASS: TestAccStorageSignedUrl_basic (0.09s)
--- PASS: TestAccStorageSignedUrl_accTest (5.82s)
```

google/resource_storage_bucket_object.go
```diff
@@ -79,6 +79,26 @@ func resourceStorageBucketObject() *schema.Resource {
...
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"content"},
+			},
+
+			// Detect changes to local file or changes made outside of Terraform to the file stored on the server.
+			"detect_md5hash": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
```
danawillow: Does this need to be Optional (vs computed-only)? If so can you add a comment why?

rosbo: Computed does not trigger a diff

danawillow: It does if the field has a value

PR #805, opened 2017-11-29
Unremove ipv4_range on networks.

### LINKED ISSUE #609: Legacy VPC network is unsupported, but still supported by Google
Although via the Google Cloud console you can no longer create a legacy VPC network, via `gcloud` it's still possible to create one. Any existing networks of this type are still abundant until there's an easy way to convert them. See https://cloud.google.com/compute/docs/vpc/legacy about the type of network.

It's support seems to be removed from the Terraform Google Cloud provider though.

### Terraform Version
`Terraform v0.10.7`

### Affected Resource(s)
- google_compute_network

```hcl
resource "google_compute_network" "network" {
    name = "tooling"
    ipv4_range = "10.240.0.0/16"
}
```

### Debug Output
https://gist.github.com/JorritSalverda/76552f52bf758b23e206a8702857cdd2

### Expected Behavior
Until Google's legacy VPC network no longer exists out there in the wild I would expect Terraform to support it. I'm fine changing the resource type to cover the fact that it's a legacy network.

### Actual Behavior

When running `terraform plan` I get the following error:

```
1 error(s) occurred:

* google_compute_network.network: "ipv4_range": [REMOVED] Please use google_compute_subnetwork resources instead.
```

### Steps to Reproduce

1. Create a legacy network with `gcloud compute networks create legacy --range 10.240.0.0/16 --mode legacy`
2. Create `legacy-network.tf` with following content:
```
resource "google_compute_network" "legacy_network" {
    name = "legacy"
    ipv4_range = "10.240.0.0/16"
}
```
3. Run `terraform plan`

paddycarver:
We removed ipv4_range, but the API still exists, it's just deprecated.
This breaks configs for users that haven't migrated off yet. I added it
back, added some tests to use it, included it in the docs, and basically
tried to put things back the way they were. The main difference now is
that the auto_create_subnetworks field defaults to true, and we want to
keep that behaviour to avoid a breaking change. So now if users want to
use the lagacy API, they need to set auto_create_subnetworks to false
explicitly.

Fixes #609.

### DIFF
--- a/google/resource_compute_firewall_test.go
+++ b/google/resource_compute_firewall_test.go
@@ -362,6 +362,8 @@ func testAccComputeFirewall_basic(network, firewall string) string {
 	return fmt.Sprintf(`
 	resource "google_compute_network" "foobar" {
 		name = "%s"
+		auto_create_subnetworks = false
+		ipv4_range = "10.0.0.0/16"
 	}
 
 	resource "google_compute_firewall" "foobar" {
@@ -380,6 +382,8 @@ func testAccComputeFirewall_update(network, firewall string) string {
 	return fmt.Sprintf(`
 	resource "google_compute_network" "foobar" {
 		name = "%s"
+		auto_create_subnetworks = false
+		ipv4_range = "10.0.0.0/16"
 	}
 
 	resource "google_compute_firewall" "foobar" {
@@ -399,6 +403,8 @@ func testAccComputeFirewall_priority(network, firewall string, priority int) str
 	return fmt.Sprintf(`
 	resource "google_compute_network" "foobar" {
 		name = "%s"
+		auto_create_subnetworks = false
+		ipv4_range = "10.0.0.0/16"
 	}
 
 	resource "google_compute_firewall" "foobar" {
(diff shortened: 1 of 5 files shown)


google/resource_compute_firewall_test.go
```diff
@@ -362,6 +362,8 @@ func testAccComputeFirewall_basic(network, firewall string) string {
 	return fmt.Sprintf(`
 	resource "google_compute_network" "foobar" {
 		name = "%s"
+		auto_create_subnetworks = false
```
rosbo: I am not sure I follow here... `auto_create_subnetworks` was defaulting to `true`. Why are we changing it to false now?

paddycarver: auto_create_subnetworks was, pre-1.0.0, defaulting to false. As part of 1.0.0, we made it default to true.

To avoid breaking that behaviour outside a major release, we want it to continue to default to true.

In our tests, however, we want to be sure to actually use the legacy VPC API. Because auto_create_subnetworks defaults to true, we need to explicitly set it to false to be able to use ipv4_range, because you can't use ipv4_range while auto_create_subnetworks is true.

PR #807, opened 2017-11-30
Only handle root NS records special.

### LINKED ISSUE #729: Regression/Bug in 1.2.0 - Sub DNS Delegation broken with NS Changes
With Google Provider 1.2.0 I'm unable to create a dns delegation between two zones, details are in the hcl code below, N.B. I've changed the domain names from my code, but have validated that the issue still exists

### Terraform Version
Terraform v0.10.8
Google Cloud Provider 1.2.0 (N.b. issue doesn't exist in 1.1.1)

### Affected Resource(s)
- google_dns_record_set

```hcl
// Configure the Google Cloud provider
provider "google" {
  //version     = "~> 1.1.1"
  version = "1.2.0"
  region      = "europe-west2"
}

// Create Root zone
resource "google_dns_managed_zone" "gcp-root" {
  dns_name    = "gcp.domain.test."
  name        = "gcp-root"
}

// Create dev root zone
resource "google_dns_managed_zone" "gcp-dev" {
  dns_name    = "dev.gcp.domain.test."
  name        = "gcp-dev"
}

// Delegate dev root zone to dev nameservers
resource "google_dns_record_set" "gcp-dev-ns" {
  managed_zone  = "${google_dns_managed_zone.gcp-root.name}"
  name          = "dev.${google_dns_managed_zone.gcp-root.dns_name}"
  rrdatas       = [
    "${google_dns_managed_zone.gcp-dev.name_servers.0}",
    "${google_dns_managed_zone.gcp-dev.name_servers.1}",
    "${google_dns_managed_zone.gcp-dev.name_servers.2}",
    "${google_dns_managed_zone.gcp-dev.name_servers.3}"
  ]
  ttl           = 86400
  type          = "NS"
}
```

### Debug Output
https://gist.github.com/Russell-IO/86ccbeb5188d0eb3cf203117286031be

### Expected Behavior
Terraform should create the NS records on the named record without trying to delete the root records

### Actual Behavior
Terraform tries to delete the root nameservers and create the nonroot nameservers in the same call and google fails the api call

### Steps to Reproduce
1. `terraform init`
2. `terraform apply`

### Important Factoids
I'm using different providers (i.e. google.infra & google.dev) - but the same issues occurs without them present, so I'm assuming... that this isn't related to the provider aliasing at all and an issue with the terraform provider google

### Related Information
https://github.com/terraform-providers/terraform-provider-google/pull/359
https://github.com/terraform-providers/terraform-provider-google/blob/4a342ca8eec32039ad2003d92bc3802834ac910d/google/resource_dns_record_set.go#L80-L105

This code, only checks if the NS is the record type, it doesn't validate that NAME is the apex, which is what is causing this defect
@paddycarver

paddycarver:
We introduced special handling for NS records in 1.2.0 under the
assumption that ALL NS records can't be deleted. This isn't actually
true. Only NS records for the naked domain of the managed zone can't be
removed; all other NS records can be. Because of this, 1.2.0 contains a
bug where all NS records are removed.

This update fixes the situation to only use special handling on NS
records that are for the naked root domain of the managed zone, and
treat all subdomain NS records as normal records. It also adds a test to
ensure this functionality.

Fixes #729.

### DIFF
--- a/google/resource_dns_record_set.go
+++ b/google/resource_dns_record_set.go
@@ -180,15 +180,12 @@ func resourceDnsRecordSetDelete(d *schema.ResourceData, meta interface{}) error 
 	// CAN and MUST be deleted, so we need to retrieve the managed zone,
 	// check if what we're looking at is a subdomain, and only not delete
 	// if it's not actually a subdomain
-	var domain string
 	if d.Get("type").(string) == "NS" {
-		if domain == "" {
-			mz, err := config.clientDns.ManagedZones.Get(project, zone).Do()
-			if err != nil {
-				return fmt.Errorf("Error retrieving managed zone %q from %q: %s", zone, project, err)
-			}
-			domain = mz.DnsName
+		mz, err := config.clientDns.ManagedZones.Get(project, zone).Do()
+		if err != nil {
+			return fmt.Errorf("Error retrieving managed zone %q from %q: %s", zone, project, err)
 		}
+		domain := mz.DnsName
 
 		if domain == d.Get("name").(string) {
 			log.Println("[DEBUG] NS records can't be deleted due to API restrictions, so they're being left in place. See https://www.terraform.io/docs/providers/google/r/dns_record_set.html for more information.")
--- a/google/resource_dns_record_set.go
+++ b/google/resource_dns_record_set.go
@@ -95,7 +95,7 @@ func resourceDnsRecordSetCreate(d *schema.ResourceData, meta interface{}) error 
 		var deletions []*dns.ResourceRecordSet
 
 		for _, record := range res.Rrsets {
-			if record.Type != "NS" {
+			if record.Type != "NS" || record.Name != d.Get("name").(string) {
 				continue
 			}
 			deletions = append(deletions, record)
@@ -164,14 +164,6 @@ func resourceDnsRecordSetRead(d *schema.ResourceData, meta interface{}) error {
 }
 
 func resourceDnsRecordSetDelete(d *schema.ResourceData, meta interface{}) error {
-
-	// NS records must always have a value, so we short-circuit delete
-	// this allows terraform delete to work, but may have unexpected
-	// side-effects when deleting just that record set.
-	if d.Get("type").(string) == "NS" {
-		log.Println("[DEBUG] NS records can't be deleted due to API restrictions, so they're being left in place. See https://www.terraform.io/docs/providers/google/r/dns_record_set.html for more information.")
-		return nil
-	}
 	config := meta.(*Config)
 
 	project, err := getProject(d, config)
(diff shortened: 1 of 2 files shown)


google/resource_dns_record_set.go
```diff
@@ -181,6 +173,29 @@ func resourceDnsRecordSetDelete(d *schema.ResourceData, meta interface{}) error
...
+	// NS records must always have a value, so we short-circuit delete
+	// this allows terraform delete to work, but may have unexpected
+	// side-effects when deleting just that record set.
+	// Unfortunately, you can set NS records on subdomains, and those
+	// CAN and MUST be deleted, so we need to retrieve the managed zone,
+	// check if what we're looking at is a subdomain, and only not delete
+	// if it's not actually a subdomain
+	var domain string
+	if d.Get("type").(string) == "NS" {
+		if domain == "" {
```
rosbo: this condition will always be true... var domain is never updated after being set.

google/resource_dns_record_set.go
```diff
@@ -181,6 +173,29 @@ func resourceDnsRecordSetDelete(d *schema.ResourceData, meta interface{}) error
...
+			mz, err := config.clientDns.ManagedZones.Get(project, zone).Do()
+			if err != nil {
+				return fmt.Errorf("Error retrieving managed zone %q from %q: %s", zone, project, err)
+			}
+			domain = mz.DnsName
+		}
+
+		if domain == d.Get("name").(string) {
```
rosbo: why not comparing with `record.Name` like above?

google/resource_dns_record_set.go
paddycarver: Domain is updated 5 lines below, on line 190.

rosbo: `domain == ""` will always be true. Nothing updates it between 183 and 185.

google/resource_dns_record_set.go
paddycarver: Above, we're comparing if a record exists with the same name we've configured. Here, we're testing if the root of the domain is the same as what we're trying to delete.

Basically, testing for two different things:

1. On create, we want to make sure any NS records that currently exist are removed in the same request we're adding NS records in.
2. On delete, we want to delete the record iff it's not _both_ an NS record _and_ the same name as the zone.

The first is achieved by listing the NS records on a zone, and making note of any that have the same name as the one we're adding, so they can be removed. The second is achieved by retrieving the name of the managed zone, and comparing it to the record we're trying to delete.

google/resource_dns_record_set.go
paddycarver: Ah, yes, for some reason I was trying to reuse that request between zones.

Which makes no sense. o.o

PR #810, opened 2017-11-30
Add an example of use of shared VPC networking.

ndmckinley:
The example brings up four projects:
  - one to host the VPC
  - two to use the VPC
  - one which is outside the VPC

This is based on the diagram in https://cloud.google.com/vpc/docs/shared-vpc and uses the names from there (where possible) for clarity.

It eventually outputs an IP address - on that page you can see a demonstration that one of the VMs on the VPC can ping another VM on the same VPC (but in a different project) using the internal IP.

The page looks like this:
terraform-cloud-page

This is my first contribution to this project, so please let me know if I'm missing something.  I've run `terraform fmt` and I've tested this on my account, but that's about it.  :)

### DIFF
--- a/examples/shared-vpc/README.md
+++ b/examples/shared-vpc/README.md
@@ -8,13 +8,11 @@ four projects:
 
 It is based on the diagram in the overview at [https://cloud.google.com/vpc/docs/shared-vpc](https://cloud.google.com/vpc/docs/shared-vpc).
 
-This example creates projects within an organization - to run it, you will need to have an Organization ID.  To get started using Organizations, read the quickstart [here](https://cloud.google.com/resource-manager/docs/quickstart-organizations).  Since it uses organizations, project-specific credentials won't work, and consequently this example is configured to use [application default credentials](https://developers.google.com/identity/protocols/application-default-credentials).  Ensure that the application default credentials have permission to create and manage projects and Shared VPCs (sometimes called 'XPN').
+Begin by [downloading your credentials from Google Cloud Console](https://www.terraform.io/docs/providers/google/#credentials); the default path for the downloaded file is `~/.gcloud/Terraform.json`.  If you use another path, update the `credentials_file_path` variable.  Ensure that these credentials have Organization-level permissions - this example will create and administer projects.
 
-Since projects require globally unique names, you will need to provide a `project_base_id` variable.  Since project names can't be too long, make sure it's short - if it is too long, Terraform won't be able to create the project.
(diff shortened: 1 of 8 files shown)


ndmckinley:
@danawillow - looks like I can't set assignee.

examples/shared-vpc/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+This is a template showcasing the shared VPC feature in Google Cloud.  It features
+four projects:
+- A host project, which owns a VPC
+- Two service projects, each of which owns a VM connected to the VPC
+- A fourth project, which owns a VM not connected to the VPC.
+
+It is based on the diagram in the overview at [https://cloud.google.com/vpc/docs/shared-vpc](https://cloud.google.com/vpc/docs/shared-vpc).
+
+This example creates projects within an organization - to run it, you will need to have an Organization ID.  To get started using Organizations, read the quickstart [here](https://cloud.google.com/resource-manager/docs/quickstart-organizations).  Since it uses organizations, project-specific credentials won't work, and consequently this example is configured to use [application default credentials](https://developers.google.com/identity/protocols/application-default-credentials).  Ensure that the application default credentials have permission to create and manage projects and Shared VPCs (sometimes called 'XPN').
```
rosbo: It is possible to grant organization-level permissions to project-specific credentials such as a service account.

In the UI, select your org in the top dropdown menu, visit Iam & Admin > Iam. Then click the +Add button at the top to add new members. In the members field, you can put a service account email.

examples/shared-vpc/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Since projects require globally unique names, you will need to provide a `project_base_id` variable.  Since project names can't be too long, make sure it's short - if it is too long, Terraform won't be able to create the project.
+
+The example allows you to specify a billing account, if you want to ensure that the charges (which are minor) are grouped together.
```
rosbo: I think billing account is mandatory to be able to create a VM.

examples/shared-vpc/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+After you run `terraform apply` on this configuration, it will output the IP address of the second service project's VM, which displays a page checking network connectivity to the other two VMs.
+
+Run with a command like:
+```
+terraform apply \
+        -var="region=us-central1" \
+        -var="region_zone=us-central1-f" \
+        -var="org_id=1234567" \
+        -var="project_base_id=unique_string"
```
rosbo: You can use the random provider instead and support full destroy.

See: https://www.terraform.io/docs/providers/random/index.html

danawillow:
I agree with all of @rosbo's comments.

ndmckinley:
Great, I'll make those changes.  Thanks for the pointer to the random provider, that makes this a lot nicer.  :)

ndmckinley:
Made those changes, tested locally.  The service account credentials were kind of a hassle to set up on the Console, but that's probably just my lack of familiarity with the credentials system in Cloud generally.

rosbo:
It is not only you :) They are a hassle. @paddycarver made a screencast to explain how because many people had issues with this.

An easier way to do it is to use Terraform instead of the console for that:
https://www.terraform.io/docs/providers/google/r/google_organization_iam_member.html

ndmckinley:
Thanks for the review!  :)

Issue #794, opened 2017-11-28
Feature request: make secondary_ip_ranges in google_compute_subnetwork changeable


wmuizelaar:
Hi there,

### Terraform Version
Terraform v0.11.0
+ provider.google (unversioned)
for the google provider I build a version based on master-branch, commit c30fdf7, to be able to use some features that aren't released yet

### Affected Resource(s)
- google_compute_subnetwork

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
provider "google" {
  credentials = "${file("/etc/gcp/compute-manager.json")}"
  project     = "bolcom-nonpro-frieps-o44b"
  region      = ""
}

resource "google_compute_network" "bolcom-frieps-network" {
  name = "bolcom-frieps-network"
  project = "bolcom-nonpro-frieps-o44b"

  auto_create_subnetworks = false

}
resource "google_compute_subnetwork" "bolcom-frieps-subnetwork" {
  name    = "bolcom-frieps-subnetwork"
  project = "bolcom-nonpro-frieps-o44b"
  network = "${google_compute_network.bolcom-frieps-network.self_link}"
  region  = "europe-west1"

  private_ip_google_access = true
  ip_cidr_range            = "192.168.0.0/24"

  secondary_ip_range { range_name = "bolcom-frieps-pods" ip_cidr_range = "10.0.0.0/24" }
  secondary_ip_range { range_name = "bolcom-frieps-services" ip_cidr_range = "10.1.0.0/24" }
  secondary_ip_range { range_name = "bolcom-frieps-services2" ip_cidr_range = "10.2.0.0/24" }
  secondary_ip_range { range_name = "bolcom-frieps-services3" ip_cidr_range = "10.3.0.0/24" }
}
```

### Expected Behavior
When I add a 'secondary_ip_range', I would expect the existing subnetwork to be altered to have an extra secondary_ip_range configured.

### Actual Behavior
The existing subnetwork will be deleted, and a new one with the same name will be created.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. Add an extra line to the above configuration with another secondary_ip_range
3. `terraform apply` again.

asterr:
This change breaks if the range_name is changed, which is not supported by Google. 

We broke a kubernetes cluster because:
1. terraform attempted to modify the subnet, but failed because the name change was not possible
[...]

PR #811, opened 2017-11-30
Implement in-place update of secondary IP ranges

koenw:
Google's beta API allows updating a subnetwork's secondary IP ranges without deleting and re-creating the subnetwork, very handy!

I'm very still new to terraform, so please let me know if there are things I should improve. I especially feel the changes to the (...)Read function could be DRY-ed up, but I wasn't sure about the proper way or if it would even be worth the effort.

Thanks!


koenw:
This commit closes #794.

rosbo:
Looks good 👍 

Merging it. Thanks for your contribution :)

PR #815, opened 2017-12-01
Remove all instances of GOOGLE_XPN_HOST_PROJECT environment variable.

### LINKED ISSUE #715: Remove `GOOGLE_XPN_HOST_PROJECT` env variable

ndmckinley:
Instead of GOOGLE_XPN_HOST_PROJECT being required to run some tests, I added the ability to create and tear down the necessary project structure.  This allows us to remove one environment variable, and use two others which are already widely-required: org and billing ID.

After this PR, no instances of GOOGLE_XPN_HOST_PROJECT appear in the repository.  This closes #715.

### DIFF
--- a/google/resource_compute_disk_test.go
+++ b/google/resource_compute_disk_test.go
@@ -5,7 +5,6 @@ import (
 	"regexp"
 	"strconv"
 	"testing"
-	"time"
 
 	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
@@ -28,7 +27,7 @@ func TestAccComputeDisk_basic(t *testing.T) {
 				Config: testAccComputeDisk_basic(diskName),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckComputeDiskExists(
-						"google_compute_disk.foobar", getTestProjectFromEnv(), &disk),
+						"google_compute_disk.foobar", &disk),
 					testAccCheckComputeDiskHasLabel(&disk, "my-label", "my-label-value"),
 					testAccCheckComputeDiskHasLabelFingerprint(&disk, "google_compute_disk.foobar"),
 				),
@@ -66,7 +65,7 @@ func TestAccComputeDisk_update(t *testing.T) {
 				Config: testAccComputeDisk_basic(diskName),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckComputeDiskExists(
-						"google_compute_disk.foobar", getTestProjectFromEnv(), &disk),
+						"google_compute_disk.foobar", &disk),
 					resource.TestCheckResourceAttr("google_compute_disk.foobar", "size", "50"),
 					testAccCheckComputeDiskHasLabel(&disk, "my-label", "my-label-value"),
 					testAccCheckComputeDiskHasLabelFingerprint(&disk, "google_compute_disk.foobar"),
@@ -76,7 +75,7 @@ func TestAccComputeDisk_update(t *testing.T) {
 				Config: testAccComputeDisk_updated(diskName),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckComputeDiskExists(
-						"google_compute_disk.foobar", getTestProjectFromEnv(), &disk),
+						"google_compute_disk.foobar", &disk),
 					resource.TestCheckResourceAttr("google_compute_disk.foobar", "size", "100"),
 					testAccCheckComputeDiskHasLabel(&disk, "my-label", "my-updated-label-value"),
 					testAccCheckComputeDiskHasLabel(&disk, "a-new-label", "a-new-label-value"),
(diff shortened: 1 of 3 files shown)


google/resource_compute_disk_test.go
```diff
@@ -200,6 +200,35 @@ func testAccCheckComputeDiskDestroy(s *terraform.State) error {
 	return nil
 }
 
+func testAccCheckComputeDiskExistsInProject(n, p string, disk *compute.Disk) resource.TestCheckFunc {
```
ndmckinley: I don't like that I had to do this (this is a copy of testAccCheckComputeDiskExists with an extra project parameter, which I hate) but I wasn't sure how I could avoid that.  The other method (here and in all three files) uses the default project.  I'm a little too new to Go to know what the right move here was - I wanted to make the project an argument in testAccCheckComputeDiskExists, but don't know how to get the name of the default project at the call site.  I assume this cannot be right - please let me know what _would_ be right.

ndmckinley:
I think @rosbo is the right person to look at this.

google/resource_compute_disk_test.go
rosbo: `getTestProjectFromEnv()` is the method you are looking for.

ndmckinley: Ahh, thank you!  That seems better.  I'll make that change and ping you again.

ndmckinley:
Okay, I made that change - I'm struggling to run all the tests I've changed, now - running into quota issues on my personal account, and don't have an Organization that I can use in the google.com org.  Do we have CI that will run these tests in a canonical environment?

google/resource_compute_disk_test.go
```diff
@@ -345,29 +345,43 @@ resource "google_compute_disk" "foobar" {
 }`, diskName)
 }
 
-func testAccComputeDisk_fromSnapshot(firstDiskName, snapshotName, diskName, xpn_host string, ref_selector string) string {
+func testAccComputeDisk_fromSnapshot(projectName, firstDiskName, snapshotName, diskName, org, billingId, ref_selector string) string {
```
rosbo: Why was this test using the XPN project in the first place instead of the regular test project?

google/resource_compute_instance_template_test.go
```diff
@@ -350,7 +351,7 @@ func testAccCheckComputeInstanceTemplateDestroy(s *terraform.State) error {
 	return nil
 }
 
-func testAccCheckComputeInstanceTemplateExists(n string, instanceTemplate *compute.InstanceTemplate) resource.TestCheckFunc {
+func testAccCheckComputeInstanceTemplateExists(n, p string, instanceTemplate *compute.InstanceTemplate) resource.TestCheckFunc {
```
rosbo: Consider adding a second method: `testAccCheckComputeInstanceTemplateXPNExists`

```go
func testAccCheckComputeInstanceTemplateExists(n string, instanceTemplate *compute.InstanceTemplate) resource.TestCheckFunc {
  return testAccCheckComputeInstanceTemplateXPNExists(n, getTestProjectFromEnv(), instanceTemplate)
} 
```

google/resource_compute_instance_test.go
```diff
@@ -873,7 +874,7 @@ func testAccCheckComputeInstanceDestroy(s *terraform.State) error {
 	return nil
 }
 
-func testAccCheckComputeInstanceExists(n string, instance *compute.Instance) resource.TestCheckFunc {
+func testAccCheckComputeInstanceExists(n, p string, instance *compute.Instance) resource.TestCheckFunc {
```
rosbo: Same comment than for instance template.

google/resource_compute_instance_template_test.go
ndmckinley: Good idea, done.

google/resource_compute_instance_test.go
ndmckinley: Same.

google/resource_compute_disk_test.go
ndmckinley: Not sure - I can't see any reason that it ought to be.

rosbo: I think we should re-write it using the default test project to avoid having to create extra test project.

ndmckinley: Agreed, and done.

ndmckinley:
Cool, I'm not planning to add to the changelog for this since it's 100% internal.  what should I do to remove the now-unnecessary environment variable from our test environment?  @rosbo or @paddycarver might know?

paddycarver:
I'll take care of it! Thanks for making this happen. 😁

PR #816, opened 2017-12-04
Enable 'zone' to be specified at the provider level instead of per-resource.

### LINKED ISSUE #772: Add provider-level `zone` attribute
Right now our provider config supports Credentials, Project, and Region. Since someone may very well may want to have a default zone for creating resources in, I think it makes sense to add Zone to the list.

This also could potentially have good benefits for our tests- if one zone is low on resources generally, we would be able to change zones for the test suite without having to go into every single resource being tested and fix it there.

ndmckinley:
Fixes #772.  Tested:

```
22:44 $ make test
go test -i $(go list ./... |grep -v 'vendor') || exit 1
echo $(go list ./... |grep -v 'vendor') | \
        xargs -t -n4 go test  -timeout=30s -parallel=4
go test -timeout=30s -parallel=4 github.com/terraform-providers/terraform-provider-google github.com/terraform-providers/terraform-provider-google/google github.com/terraform-providers/terraform-provider-google/scripts 
?       github.com/terraform-providers/terraform-provider-google        [no test files]
ok      github.com/terraform-providers/terraform-provider-google/google 0.065s
ok      github.com/terraform-providers/terraform-provider-google/scripts        0.012s
```

and confirmed I got all the places I needed to replace:

```
22:46 $ grep '.GetOk("zone")' google/*.go
google/utils.go:        res, ok := d.GetOk("zone")
22:46 $ grep '.Get("zone")' google/*.go
google/provider.go:             Zone:        d.Get("zone").(string),
```

(both of which are correct)

### DIFF
--- a/google/resource_compute_disk.go
+++ b/google/resource_compute_disk.go
@@ -45,7 +45,7 @@ func resourceComputeDisk() *schema.Resource {
 
 			"zone": &schema.Schema{
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 				ForceNew: true,
 			},
 
@@ -130,12 +130,16 @@ func resourceComputeDiskCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	// Get the zone
-	log.Printf("[DEBUG] Loading zone: %s", d.Get("zone").(string))
+	z, err := getZone(d, config)
+	if err != nil {
+		return err
+	}
+	log.Printf("[DEBUG] Loading zone: %s", z)
 	zone, err := config.clientCompute.Zones.Get(
-		project, d.Get("zone").(string)).Do()
+		project, z).Do()
 	if err != nil {
 		return fmt.Errorf(
-			"Error loading zone '%s': %s", d.Get("zone").(string), err)
+			"Error loading zone '%s': %s", z, err)
 	}
 
 	// Build the disk parameter
@@ -199,7 +203,7 @@ func resourceComputeDiskCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	op, err := config.clientCompute.Disks.Insert(
-		project, d.Get("zone").(string), disk).Do()
+		project, z, disk).Do()
 	if err != nil {
 		return fmt.Errorf("Error creating disk: %s", err)
 	}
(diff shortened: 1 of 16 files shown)


ndmckinley:
@rosbo This seems to have gone right, but as mentioned I'm not in a good state to run all the acctests.  I'm running the ones I can now and I'll report back with any failures I suspect are related.

ndmckinley:
Aha, there was a problem with the provider config.  Hang on, I'll fix it and report back.

google/utils.go
```diff
@@ -28,14 +28,32 @@ func getRegionFromZone(zone string) string {
 
 // getRegion reads the "region" field from the given resource data and falls
 // back to the provider's value if not given. If the provider's value is not
-// given, an error is returned.
+// given, checks for "zone" in either the given resource data or provider,
+// and extracts region from zone.  If "zone" is not provided, returns an
+// error.
 func getRegion(d *schema.ResourceData, config *Config) (string, error) {
```
rosbo: Probably worth it to add a unit test for getRegion and getZone.

ndmckinley: I'm sure you're right - I'll get on that.  :)

rosbo:
FYI: CI is server is running all the acceptance tests.

google/resource_compute_disk.go
```diff
@@ -279,12 +287,14 @@ func resourceComputeDiskRead(d *schema.ResourceData, meta interface{}) error {
...
 	var disk *compute.Disk
-	if zone, ok := d.GetOk("zone"); ok {
+	if zone, err := getZone(d, config); zone != "" && err == nil {
 		disk, err = config.clientCompute.Disks.Get(
-			project, zone.(string), d.Id()).Do()
+			project, zone, d.Id()).Do()
 		if err != nil {
 			return handleNotFoundError(err, d, fmt.Sprintf("Disk %q", d.Get("name").(string)))
 		}
+	} else if err != nil {
```
rosbo: You should remove this else if block. If no zone are specified, it is fine and the logic in the else block tries to guess the zone.

This comment also applies to group manager and autoscaler.

google/resource_compute_disk.go
```diff
@@ -279,12 +287,14 @@ func resourceComputeDiskRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	var disk *compute.Disk
-	if zone, ok := d.GetOk("zone"); ok {
+	if zone, err := getZone(d, config); zone != "" && err == nil {
```
rosbo: If the zone is empty, the `err` returned by  `getZone` should not be nil and therefor, the `zone != nil` is redundant

google/utils.go
ndmckinley: Done.

google/resource_compute_disk.go
ndmckinley: Ah, shoot, I thought that was idiomatic because of the way that .Get() works in Schema.  :)  I'll fix it.

google/resource_compute_disk.go
ndmckinley: I've done this as well, sorry.

PR #821, opened 2017-12-05
examples: Fix credentials key

philipcristiano:
This is out of date and now uses `credentials` as the key


rosbo:
Thanks Philip!

PR #826, opened 2017-12-06
Make network name unique between each test run

rosbo:
Concurrent runs were failing because this test tries to create a network with a static name.


paddycarver:
Failing a `fmt` check.

paddycarver:
After the fmt check is fixed, 👍

rosbo:
Fixed

PR #835, opened 2017-12-09
Consistent IAM resource imports.

### LINKED ISSUE #756: Add import support for iam resource.
The `google_(folder|project|organization)_iam_(binding|member|policy)` resource should support import for consistency with the other resources even though running `terraform apply` (effectively calling the create method) yields the same result. 

Terraform users may not know the intimacies of the API such that they know importing is the same effect as creating in this particular case. I don't believe it's intuitive to a user that for the iam resources they can simply add the resource block and the apply will be idempotent..

ndmckinley:
- Anything with IAM resources can implement a simple ID parser method and begin working.
- Added ID parser methods for folders, projects, organizations, kms crypto keys, and kms key rings.
- Added docs and tests for the imports.

Fixes #756


ndmckinley:
The import tests pass on my machine with my org / project combo.

ndmckinley:
Oh, and this closes #756.

google/iam.go
```diff
@@ -34,6 +36,83 @@ type ResourceIamUpdater interface {
...
 
+// This method parses identifiers specific to the resource (d.GetId()) into the ResourceData
+// object, so that it can be given to the resource's Read method.  Externally, this is wrapped
+// into schema.StateFunc functions - one each for a _member, a _binding, and a _policy.  Any
+// GCP resource supporting IAM policy might support one, two, or all of these.  Any GCP resource
+// for which an implementation of this interface exists could support any of the three.
+
+type resourceIdParserFunc func(d *schema.ResourceData, config *Config) error
+
+func iamPolicyImport(resourceIdParser resourceIdParserFunc) schema.StateFunc {
```
rosbo: consider moving the `iamPolicyImport`, `iamPolicyBinding` and `iamPolicyMember` functions into their respective `resource_iam_*` file.

ndmckinley: Good call!  Done.

rosbo:
Build is failing because of unused or missing imports after the move.

ndmckinley:
Fixed that - should have tried a test after that.  :)

PR #837, opened 2017-12-10
WIP: Add "internal IP only" support for Dataproc clusters

nickjacques:
This PR creates a `internal_ip_only` boolean argument, nested in the `gce_cluster_config` schema. By default, this value is set to `false`, which is consistent with the current behavior of the API and CLI - Dataproc clusters are created with internal IPs and ephemeral external IPs. When the value is `true`, this passes through to the API and creates a Dataproc cluster with internal IP addresses only (no ephemeral external IP addresses). 

This requires use of subnetworks, as well as Private Google Access enabled in the destination subnetwork. This should be added to the provider docs.

Note: the `internalIpOnly` API argument is [available in the Dataproc v1 API](https://cloud.google.com/dataproc/docs/reference/rest/v1/projects.regions.clusters#GceClusterConfig) (as well as v1beta2)

TODO:
- [x] create feature by adding argument to `gce_cluster_config` schema (737fd7a)
- [x] add test cases
  - [x] default test case: `internal_ip_only` is `false` (a679b18)
  - [x] test case: `internal_ip_only` is `true` (dd4521c)
- [x] validate test cases
- [x] add argument to documentation (fecee25)
- [x] augment resource schema argument to conflict when not using subnets?

### DIFF
--- a/google/resource_dataproc_cluster_test.go
+++ b/google/resource_dataproc_cluster_test.go
@@ -108,7 +108,7 @@ func TestAccDataprocCluster_basic(t *testing.T) {
 					resource.TestCheckResourceAttrSet("google_dataproc_cluster.basic", "cluster_config.0.bucket"),
 
 					// Default behavior is for Dataproc to not use only internal IP addresses
-					resource.TestCheckResourceAttr("google_dataproc_cluster.basic", "cluster_config.0.gce_cluster_config.internal_ip_only", "false"),
+					resource.TestCheckResourceAttr("google_dataproc_cluster.basic", "cluster_config.0.gce_cluster_config.0.internal_ip_only", "false"),
 
 					// Expect 1 master instances with computed values
 					resource.TestCheckResourceAttr("google_dataproc_cluster.basic", "cluster_config.0.master_config.#", "1"),
@@ -152,7 +152,7 @@ func TestAccDataprocCluster_basicWithInternalIpOnlyTrue(t *testing.T) {
 					testAccCheckDataprocClusterExists("google_dataproc_cluster.basic", &cluster),
 
 					// Testing behavior for Dataproc to use only internal IP addresses
-					resource.TestCheckResourceAttr("google_dataproc_cluster.basic", "cluster_config.0.gce_cluster_config.internal_ip_only", "true"),
+					resource.TestCheckResourceAttr("google_dataproc_cluster.basic", "cluster_config.0.gce_cluster_config.0.internal_ip_only", "true"),
 				),
 			},
 		},
--- a/google/resource_dataproc_cluster_test.go
+++ b/google/resource_dataproc_cluster_test.go
@@ -136,6 +136,29 @@ func TestAccDataprocCluster_basic(t *testing.T) {
 	})
 }
 
+func TestAccDataprocCluster_basicWithInternalIpOnlyTrue(t *testing.T) {
+	t.Parallel()
+
+	var cluster dataproc.Cluster
+	rnd := acctest.RandString(10)
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckDataprocClusterDestroy(false),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataprocCluster_basicWithInternalIpOnlyTrue(rnd),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDataprocClusterExists("google_dataproc_cluster.basic", &cluster),
+
+					// Testing behavior for Dataproc to use only internal IP addresses
+					resource.TestCheckResourceAttr("google_dataproc_cluster.basic", "cluster_config.gce_cluster_config.internal_ip_only", "true"),
+				),
+			},
+		},
+	})
+}
+
 func TestAccDataprocCluster_basicWithAutogenDeleteTrue(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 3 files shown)


nickjacques:
@danawillow Hi Dana! Could you give dd4521c a glance? I am not familiar with running the full test suite and don't feel comfortable running it without guidance (I don't want to impact my production projects due to an oversight or missed env var). I believe a679b18 is pretty straightforward and should get the job done, but I'm less confident about dd4521c. Would love your thoughts on this PR as a whole so far. Thanks!!

paddycarver:
Fix for #864.

google/resource_dataproc_cluster_test.go
```diff
@@ -107,6 +107,9 @@ func TestAccDataprocCluster_basic(t *testing.T) {
 					// Default behaviour is for Dataproc to autogen or autodiscover a config bucket
 					resource.TestCheckResourceAttrSet("google_dataproc_cluster.basic", "cluster_config.0.bucket"),
 
+					// Default behavior is for Dataproc to not use only internal IP addresses
+					resource.TestCheckResourceAttr("google_dataproc_cluster.basic", "cluster_config.gce_cluster_config.internal_ip_only", "false"),
```
danawillow: This should be `cluster_config.0.gce_cluster_config.0.internal_ip_only` because of the way nested resources are stored in the Terraform state

google/resource_dataproc_cluster_test.go
```diff
@@ -133,6 +136,29 @@ func TestAccDataprocCluster_basic(t *testing.T) {
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckDataprocClusterDestroy(false),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataprocCluster_basicWithInternalIpOnlyTrue(rnd),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDataprocClusterExists("google_dataproc_cluster.basic", &cluster),
+
+					// Testing behavior for Dataproc to use only internal IP addresses
+					resource.TestCheckResourceAttr("google_dataproc_cluster.basic", "cluster_config.gce_cluster_config.internal_ip_only", "true"),
```
danawillow: likewise

google/resource_dataproc_cluster_test.go
```diff
@@ -648,6 +674,66 @@ resource "google_dataproc_cluster" "basic" {
...
+# certain firewall rules open to allow internal communication. As we
+# are creating a new one here for this test, we need to additionally
+# open up similar rules to allow the nodes to talk to each other
+# internally as part of their configuration or this will just hang.
+#
+resource "google_compute_firewall" "dataproc_network_firewall" {
+	name = "dproc-cluster-test-allow-internal"
+	description = "Firewall rules for dataproc Terraform acceptance testing"
+	network = "${google_compute_network.dataproc_network.name}"
+
```
danawillow: Does it make sense to also include `source_ranges` here to more closely match the default allow-internal firewall rule?

google/resource_dataproc_cluster_test.go
```diff
@@ -648,6 +674,66 @@ resource "google_dataproc_cluster" "basic" {
 `, rnd)
 }
 
+func testAccDataprocCluster_basicWithInternalIpOnlyTrue(rnd string) string {
+	return fmt.Sprintf(`
+resource "google_compute_network" "dataproc_network" {
+	name = "dataproc-internalip-network"
```
danawillow: Can you also add randomness to this and the subnetwork? This helps out with our automated testing so that if the test fails before the resource is deleted, it doesn't collide with the existing one when trying to create it the next time. Feel free to reuse the same rnd var

google/resource_dataproc_cluster_test.go
```diff
@@ -648,6 +674,66 @@ resource "google_dataproc_cluster" "basic" {
...
+
+	allow {
+		protocol = "udp"
+		ports    = ["0-65535"]
+	}
+}
+resource "google_dataproc_cluster" "basic" {
+	name                  = "dproc-cluster-test-%s"
+	region                = "us-central1"
+	depends_on            = ["google_compute_firewall.dataproc_network_firewall","google_compute_subnetwork.dataproc_subnetwork"]
```
danawillow: You're already depending on the subnetwork below, so no need to include it in the `depends_on`

google/resource_dataproc_cluster_test.go
nickjacques: Fixed in 365df40!

google/resource_dataproc_cluster_test.go
nickjacques: Fixed in 365df40!

google/resource_dataproc_cluster_test.go
nickjacques: Fixed in 7d45f5f!

google/resource_dataproc_cluster_test.go
nickjacques: Fixed in 853a9ee!

google/resource_dataproc_cluster_test.go
nickjacques: Fixed in 3e67813!

google/resource_dataproc_cluster_test.go
```diff
@@ -107,6 +107,9 @@ func TestAccDataprocCluster_basic(t *testing.T) {
 					// Default behaviour is for Dataproc to autogen or autodiscover a config bucket
 					resource.TestCheckResourceAttrSet("google_dataproc_cluster.basic", "cluster_config.0.bucket"),
 
+					// Default behavior is for Dataproc to not use only internal IP addresses
+					resource.TestCheckResourceAttr("google_dataproc_cluster.basic", "cluster_config.0.gce_cluster_config.internal_ip_only", "false"),
```
danawillow: You'll still need another .0 after `gce_cluster_config`

google/resource_dataproc_cluster_test.go
```diff
@@ -133,6 +136,29 @@ func TestAccDataprocCluster_basic(t *testing.T) {
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckDataprocClusterDestroy(false),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataprocCluster_basicWithInternalIpOnlyTrue(rnd),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDataprocClusterExists("google_dataproc_cluster.basic", &cluster),
+
+					// Testing behavior for Dataproc to use only internal IP addresses
+					resource.TestCheckResourceAttr("google_dataproc_cluster.basic", "cluster_config.0.gce_cluster_config.internal_ip_only", "true"),
```
danawillow: same

google/resource_dataproc_cluster_test.go
```diff
@@ -648,6 +674,71 @@ resource "google_dataproc_cluster" "basic" {
...
+		protocol = "udp"
+		ports    = ["0-65535"]
+	}
+
+	source_ranges = ["${var.subnetwork_cidr}"]
+}
+resource "google_dataproc_cluster" "basic" {
+	name                  = "dproc-cluster-test-%s"
+	region                = "us-central1"
+
```
danawillow: whoops sorry you'll still want to have the explicit depends on for the firewall rule, just not for the subnetwork

nickjacques:
Right you are @danawillow, sorry about that! I think the last two commits should tidy things up.

danawillow:
Looks good!
```
--- PASS: TestAccDataprocCluster_basic (101.89s)
--- PASS: TestAccDataprocCluster_basicWithInternalIpOnlyTrue (206.42s)
```

Was there anything else you were hoping to add to this? I'm ready to merge if you are.

nickjacques:
Awesome, thanks Dana! The only other thing I’m pondering over is if there should be some sort of conflict if the “subnet” property is not set (since using the internalIp setting requires deploying to a subnet and not a legacy network), or if we should just document that using subnets is required for this option (which is inferred in the added documentation, but maybe should be more directly stated?)

danawillow:
There's no way to say on plan-time "if X field is set, Y field must also be set" so either way, a user running Terraform would get the error when they ran `terraform apply`. In that case, we could catch the error early to save an API call, or we could let the API return an error to save some code. I'd probably lean towards catching the error early, especially if the error the API returns isn't descriptive enough (I haven't checked whether it is or not), but I don't think it's super important to do so.

nickjacques:
@danawillow looks like the API does a pretty good job of handling this itself - here's me trying to use `internal_ip_only` without specifying a subnetwork:

```
Error: Error applying plan:

1 error(s) occurred:

* module.dataproc-cluster.google_dataproc_cluster.dataproc: 1 error(s) occurred:

* google_dataproc_cluster.dataproc: googleapi: Error 400: Network 'default' doesn't contain subnetworks, but 'internal_ip_only' is true. Specify a network which contains subnetworks to enable 'internal_ip_only', or set 'internal_ip_only' to 'false'., badRequest

```

I think this is ready for merge if the above error is acceptable!

danawillow:
Works for me, thanks!

nickjacques:
Thanks Dana!! Appreciate all your help. Happy holidays!

PR #840, opened 2017-12-11
Add service account IAM support

### LINKED ISSUE #819: Add IAM support for service account
Feature request to add google_service_account_iam_binding resource similar to:
[gcloud iam service-accounts add-iam-policy-binding](https://cloud.google.com/sdk/gcloud/reference/iam/service-accounts/add-iam-policy-binding)

### Terraform Version
v0.11.1

### Affected Resource(s)
- google_service_account_iam_binding (missing)
- google_service_account_iam_member (missing)

rosbo:
Fixes #819

When managing IAM roles, you can treat a service account either as a resource or as an identity. This resource is to add iam policy bindings to a service account resource.

Import tests written in the new way proposed in #778.


website/docs/r/google_service_account_iam.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+## Attributes Reference
+
+In addition to the arguments listed above, the following computed attributes are
+exported:
+
+* `etag` - (Computed) The etag of the service account IAM policy.
+
+## Import
+
+IAM member imports use space-delimited identifiers; the resource in question, the role, and the member. This member resource can be imported using the `service_account_id`, role, and member e.g.
```
ndmckinley: If it were me, maybe I'd collapse these three into one "code sample" section and put the explanations above?  I know I didn't do that for the doc I wrote, but now seeing it I think that might be clearer.  Not feeling strongly about it.

rosbo: Done. What do you think? Do you think I should elaborate around the fact that role is only used for member/binding and member only for the member resource? Or does the example speak for itself.

cat5:
I just personally ran into this, seems like it would solve my issue, and supports importing.
Would love to test this out.

PR #850, opened 2017-12-13
Storage bucket doesn't remove lifecycle rules properly

### LINKED ISSUE #845: Last lifecycle rule not removed from google_storage_bucket
Hi there,

When adding lifecycle rules to a bucket and then removing then, they are not removed as no other lifecycle rules remain. So adding two rules and removing one will result in that one rule being removed. However, removing the second one (or removing both at the same time) will result in the rule(s) being kept in place.

### Terraform Version
```$terraform -v
Terraform v0.11.1
+ provider.google v1.4.0
```

### Affected Resource(s)
- google_storage_bucket

```hcl
resource "google_storage_bucket" "mybucket" {
  name          = "mybucket"
  project       = "myproject"
  storage_class = "REGIONAL"

  versioning = {
    enabled = true
  }

  lifecycle_rule = {
    action = {
      type = "SetStorageClass"
      storage_class = "NEARLINE"
    }
    condition = {
      age = "100"
    }
  }

  lifecycle_rule = {
    action = {
      type = "Delete"
    }
    condition = {
      age = "500"
      matches_storage_class = ["NEARLINE", "COLDLINE"]    }
  }

}
```

### Expected Behavior
Lifecycle rules being removed.

### Actual Behavior
Lifecycle rules not always removed

### Steps to Reproduce
Repro path 1:
* When removing the rule with type 'Delete' and running terraform apply, the rule is removed.
* When I then remove the rule with type 'SetStorageClass', the rule is not removed but terraform apply output shows it had updated the bucket resource. 

Repro path 2:
* When removing both rules at the same time, both rules will remain set on the bucket.

rosbo:
Fixes #845

* Added a ForceSendFields to ensure update deletes the rules if we remove all of them from the config.
* Realized that the read method wasn't reading the lifecycle field. Fixed this and added an import test case to catch this.
* Remove the dedicated import test as discussed in #778.
* Updated the storage client to the latest version


rosbo:
```sh
--- PASS: TestAccStorageBucket_cors (2.16s)
--- PASS: TestAccStorageBucket_basic (2.36s)
--- PASS: TestAccStorageBucket_lifecycleRules (2.50s)
--- PASS: TestAccStorageBucket_versioning (2.51s)
--- PASS: TestAccStorageBucket_labels (4.75s)
--- PASS: TestAccStorageBucket_lowercaseLocation (4.90s)
--- PASS: TestAccStorageBucket_customAttributes (4.90s)
--- PASS: TestAccStorageBucket_storageClass (4.86s)
--- PASS: TestAccStorageBucket_forceDestroy (11.08s)
--- PASS: TestAccStorageBucket_update (16.29s)
```

PR #851, opened 2017-12-13
New data source: compute region instance group

### LINKED ISSUE #765: Add data source for: google_compute_region_instance_group
### Terraform Version
Terraform v0.10.8
Google Provider v1.2.0

The `google_compute_instance_group` works with zonal managed instance groups like the `google_compute_instance_group_manager` resource. But we also need a data source that works with the regional instance groups like those created by the `google_compute_region_instance_group_manager` resource.

API is here: https://cloud.google.com/compute/docs/reference/latest/regionInstanceGroups/listInstances

ndmckinley:
This closes #765... I think!  :)  This is my first data source - let me know if I've missed something, please!  The timeout logic I wrote in the manager resource could use some special attention since I'm relatively new to Go.


rosbo:
Vet found suspicious constructs. Please check the reported constructs
and fix them if necessary before submitting the code for review.
make: *** [vet] Error 1
The command "make vet" exited with 2.

google/data_source_google_compute_region_instance_group_manager.go
```diff
@@ -0,0 +1,161 @@
...
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceComputeRegionInstanceGroupRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	var project, region, name string
+	// add self_link parsing here.
```
rosbo: It looks like you've already added self_link parsing below

google/data_source_google_compute_region_instance_group_manager.go
```diff
@@ -0,0 +1,161 @@
...
+		Read: dataSourceComputeRegionInstanceGroupRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+
+			"items": {
```
rosbo: We tend to not pluralize field names (no always followed). Would that make more sense to call it `instance` instead of `item`?

google/resource_compute_region_instance_group_manager.go
```diff
@@ -110,6 +112,12 @@ func resourceComputeRegionInstanceGroupManager() *schema.Resource {
 				Optional: true,
 			},
 
+			"wait_for_instances": &schema.Schema{
```
rosbo: Documentation for this

google/resource_compute_region_instance_group_manager.go
```diff
@@ -182,7 +190,38 @@ func resourceComputeRegionInstanceGroupManagerCreate(d *schema.ResourceData, met
...
+				if _, ok := err.(notFinishedCreatingError); ok {
+					time.Sleep(5 * time.Second)
+					continue
+				} else {
+					c1 <- err
+				}
+			}
+		}
+	}()
+	select {
```
rosbo: why using goroutine if you are blocking the main thread anyway here?

google/data_source_google_compute_region_instance_group_manager.go
ndmckinley: No strong feelings - I thought it would be best to stick to exactly the field names that come back from the /listInstances call.  Do you prefer `instance` to `items` given that `items` is what's used here?
https://cloud.google.com/compute/docs/reference/latest/regionInstanceGroups/listInstances

FWIW I agree that "items" is a lousy name!  It's probably the least specific you can be about what's in the list.

google/data_source_google_compute_region_instance_group_manager.go
ndmckinley: Ha, true, removed.

google/resource_compute_region_instance_group_manager.go
ndmckinley: Done.

google/resource_compute_region_instance_group_manager.go
ndmckinley: In this case, because there are two goroutines - time.After() returns a channel that sends "true" after the duration specified.  This serves as an (apparently) idiomatic timeout mechanism.  I like it, but willing to reformulate it if you prefer.  Definitely need to comment it.

google/data_source_google_compute_region_instance_group_manager.go
rosbo: `instance` is definitely better. I think actually the pluralized version would be better since it is read only.

people will use it for interpolation in their data source this way:
```hcl
instance = "${data.google_compute_region_instance_group_manager.instances.0.self_link}"
```

google/resource_compute_region_instance_group_manager.go
rosbo: You could probably rely on state.WaitForState() and wait until the creating + creatingwhitoutretries is equal to 0.

It would be more consistent with the rest of the waiting logic in Terraform

danisla:
I really like the addition of the `wait_for_instances` argument! Can we get that on the `google_compute_instance_group_manager` resource too?

google/resource_compute_region_instance_group_manager.go
ndmckinley: Aha, yes, state.WaitForState is a much better version of what I've written here.  Will do!

ndmckinley:
I think so!  I've rewritten it to be extensible.  I'll do that in a followup PR.

google/data_source_google_compute_region_instance_group_manager.go
ndmckinley: Agreed, and done.

ndmckinley:
```
TF_ACC=1 go test ./google -v -run=TestAccDataSourceRegionInstanceGroup* -timeout 120m
=== RUN   TestAccDataSourceRegionInstanceGroupManager
--- PASS: TestAccDataSourceRegionInstanceGroupManager (159.80s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 159.811s
```

ndmckinley:
Added docs - how can I preview them to make sure they look right?

Also, 
```
TF_ACC=1 go test ./google -v -run=TestAccRegionInstanceGroup* -timeout 120m    
=== RUN   TestAccRegionInstanceGroupManager_basic                              
=== RUN   TestAccRegionInstanceGroupManager_targetSizeZero                     
=== RUN   TestAccRegionInstanceGroupManager_update                             
=== RUN   TestAccRegionInstanceGroupManager_updateLifecycle                    
=== RUN   TestAccRegionInstanceGroupManager_separateRegions                    
=== RUN   TestAccRegionInstanceGroupManager_autoHealingPolicies                
--- PASS: TestAccRegionInstanceGroupManager_update (180.35s)                   
--- PASS: TestAccRegionInstanceGroupManager_autoHealingPolicies (243.64s)      
--- PASS: TestAccRegionInstanceGroupManager_basic (243.75s)                    
--- PASS: TestAccRegionInstanceGroupManager_targetSizeZero (47.80s)            
--- PASS: TestAccRegionInstanceGroupManager_separateRegions (305.04s)          
--- PASS: TestAccRegionInstanceGroupManager_updateLifecycle (156.64s)          
PASS                                   
ok      github.com/terraform-providers/terraform-provider-google/google 337.010s    
```

rosbo:
You can follow these instructions:
https://github.com/hashicorp/terraform-website#running-the-site-locally

website/docs/d/datasource_compute_region_instance_group.html.markdown
```diff
@@ -0,0 +1,48 @@
+---
+layout: "google"
+page_title: "Google: google_compute_instance_group"
+sidebar_current: "docs-google-datasource-compute-instance-group"
+description: |-
+  Get a Compute Region Instance Group within GCE.
```
rosbo: The big feature is being able to list the instances inside the instance group. We should make that clear here.

ndmckinley: Done!  Took a lot of tiny commits to test the website - sorry about the mess, of course I'll squash before I merge.

PR #853, opened 2017-12-13
Bugfix: add read back in to the KMS crypto key read method.

### LINKED ISSUE #813: impossible condition: nil != nil in resourceKmsCryptoKeyRead
https://github.com/terraform-providers/terraform-provider-google/blob/41a952cdfacce5a08a550f059cb69ed662234fa8/google/resource_kms_crypto_key.go#L113

err is provably nil at this line, thus https://github.com/terraform-providers/terraform-provider-google/blob/41a952cdfacce5a08a550f059cb69ed662234fa8/google/resource_kms_crypto_key.go#L113-L115 has no effect.

I think that https://github.com/terraform-providers/terraform-provider-google/blob/41a952cdfacce5a08a550f059cb69ed662234fa8/google/resource_kms_crypto_key.go#L111-L115 could be removed, but not sure.

ndmckinley:
Impact of this bug: if a user deleted a crypto key or changed the rotation period outside terraform and then reran `apply`, the old data would stay in the plan.  This could cause failures when the key is being used later in the plan, or false `terraform plan` output.

Ran the tests - they still pass, but I confirmed the bug manually and confirmed that the fix is applied.

This closes #813.

### DIFF
--- a/google/resource_kms_crypto_key.go
+++ b/google/resource_kms_crypto_key.go
@@ -110,12 +110,13 @@ func resourceKmsCryptoKeyRead(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] Executing read for KMS CryptoKey %s", cryptoKeyId.cryptoKeyId())
 
+	cryptoKey, err := config.clientKms.Projects.Locations.KeyRings.CryptoKeys.Get(cryptoKeyId.cryptoKeyId()).Do()
 	if err != nil {
 		return fmt.Errorf("Error reading CryptoKey: %s", err)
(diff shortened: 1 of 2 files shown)


ndmckinley:
Good point, it really should have.  Let me figure that out.

ndmckinley:
Ah, because it lacked a rotation period in the config there, so the lack of a read that worked to find out the rotation period didn't matter.  Fixed!

ndmckinley:
Also discovered that the key rotation validation method panics if you give it a format it doesn't agree with - fixed that too.

PR #859, opened 2017-12-14
Removes unsupported identities related to project iam binding

eilgin:
The [docs](https://cloud.google.com/resource-manager/reference/rest/v1/projects/setIamPolicy) is clear about that:

> Project does not support allUsers and allAuthenticatedUsers as members in a Binding of a Policy.

So i've updated the docs to avoid some confusion and head scratching because of the 400 error responses (`Request contains an invalid argument.`)


rosbo:
Thank you Thomas.

rosbo:
I've also filed an issue (#865) to go over the documentation for the other resources and make sure we are not misleading our users.

PR #882, opened 2017-12-20
Rename filename to match resource

rosbo:
The data source is named: "google_compute_region_instance_group" and all the reference in documentation and so on. It is also a data source that takes the name of an instance manager.


rosbo:
Done.

Issue #888, opened 2017-12-20
New data source: Billing Account


ewbankkit:
New data source to get information about a Google Billing Account.
Filter by the resource name (`billingAccounts/{billing_account_id}`) or display name of a billing account.

PR #889, opened 2017-12-20
Add 'google_billing_account' data source

ewbankkit:
Fixes https://github.com/terraform-providers/terraform-provider-google/issues/888.

### DIFF
--- a/google/data_source_google_billing_account.go
+++ b/google/data_source_google_billing_account.go
@@ -15,14 +15,18 @@ func dataSourceGoogleBillingAccount() *schema.Resource {
 		Read: dataSourceBillingAccountRead,
 		Schema: map[string]*schema.Schema{
 			"name": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				Computed:      true,
+				ConflictsWith: []string{"display_name"},
 			},
 			"display_name": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				Computed:      true,
+				ConflictsWith: []string{"name"},
 			},
 			"open": {
 				Type:     schema.TypeBool,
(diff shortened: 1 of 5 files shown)


ewbankkit:
```
make testacc TEST=./google/ TESTARGS='-run=TestAccDataSourceGoogleBillingAccount_'
TF_ACC=1 go test ./google/ -v -run=TestAccDataSourceGoogleBillingAccount_ -timeout 120m
=== RUN   TestAccDataSourceGoogleBillingAccount_basic
--- PASS: TestAccDataSourceGoogleBillingAccount_basic (1.75s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	1.762s
```

website/docs/d/google_billing_account.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+}
+```
+
+## Argument Reference
+
+The arguments of this data source act as filters for querying the available billing accounts.
+The given filters must match exactly one billing account whose data will be exported as attributes.
+The following arguments are supported:
+
+* `name` (Optional) - The resource name of the billing account in the form `billingAccounts/{billing_account_id}`.
```
rosbo: Consider doing like we did for folders where you can specify either "folders/{folder_id}" or "{folder_id}".

Since we would support only "{billing_account_id}" for this field, what do you think or renaming this field `account_id` or `billing_account_id`?

https://github.com/terraform-providers/terraform-provider-google/blob/master/google/iam_folder.go#L81

google/data_source_google_billing_account.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+	}
+}
+
+func dataSourceBillingAccountRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	name, nameOk := d.GetOk("name")
+	displayName, displayNameOk := d.GetOk("display_name")
+	if nameOk == displayNameOk {
```
rosbo: You can remove this check and use ConflictsWith in the schema definition above.

```go
"name": {
  Type: schema.TypeString
  ConflictsWith: []string{"display_name"},
  ...
}
```

google/data_source_google_billing_account.go
```diff
@@ -0,0 +1,111 @@
...
+			if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == http.StatusNotFound {
+				return fmt.Errorf("Billing account not found: %s", name)
+			}
+
+			return fmt.Errorf("Error reading billing account: %s", err)
+		}
+
+		billingAccount = resp
+	} else {
+		resp, err := config.clientBilling.BillingAccounts.List().Do()
```
rosbo: This call has pagination.

Here is an example of how we page through results in other parts of the code:
https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_google_project_services.go#L171

website/docs/d/google_billing_account.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+* `display_name` (Optional) - The display name of the billing account.
+
+~> **NOTE:** One of `name` or `display_name` must be specified.
+
+## Attributes Reference
+
+The following additional attributes are exported:
+
+* `id` - The billing account ID.
+* `open` - `true` if the billing account is open, `false` if the billing account is closed.
```
rosbo: Do you think filtering by "open = true" might be useful?

If you think so, you can define this field both Computed and Optional. This means that if someone specifies it, it will filter, if they don't specify it, you can still set the value from the server.

ewbankkit: Yes, makes sense,

website/docs/d/google_billing_account.html.markdown
ewbankkit: How about keeping `name` as a _Computed_-only attribute and having the argument named `billing_account` specified as either `{billing_account_id}` or `billingAccounts/{billing_account_id}`?
GCP seems exact in its distinction between resource names and resource IDs and this argument can be either.

rosbo: Agree. I like your suggestion better.

google/data_source_google_billing_account.go
ewbankkit: Done.

google/data_source_google_billing_account.go
ewbankkit: Done.

website/docs/d/google_billing_account.html.markdown
ewbankkit: Done.

website/docs/d/google_billing_account.html.markdown
ewbankkit: Done.

ewbankkit:
```
make testacc TEST=./google/ TESTARGS='-run=TestAccDataSourceGoogleBillingAccount_'
TF_ACC=1 go test ./google/ -v -run=TestAccDataSourceGoogleBillingAccount_ -timeout 120m
=== RUN   TestAccDataSourceGoogleBillingAccount_byFullName
--- PASS: TestAccDataSourceGoogleBillingAccount_byFullName (1.90s)
=== RUN   TestAccDataSourceGoogleBillingAccount_byShortName
--- PASS: TestAccDataSourceGoogleBillingAccount_byShortName (1.57s)
=== RUN   TestAccDataSourceGoogleBillingAccount_byFullNameClosed
--- PASS: TestAccDataSourceGoogleBillingAccount_byFullNameClosed (0.25s)
=== RUN   TestAccDataSourceGoogleBillingAccount_byDisplayName
--- PASS: TestAccDataSourceGoogleBillingAccount_byDisplayName (0.29s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	4.021s
```

PR #893, opened 2017-12-22
Don't recreate container_cluster when maintenance_window changes

### LINKED ISSUE #727: Enable non-destructive update of maintenance_window on google_container_cluster
The Google APIs allow for maintenance window to be set, unset or changed on an existing cluster. Enable the same non-destructive behaviour with Terraform.

### Affected Resource(s)
- google_container_cluster

### References
- This is an enhancement to #526

michaelbannister:
Fixes #727

### DIFF
--- a/google/resource_container_cluster_test.go
+++ b/google/resource_container_cluster_test.go
@@ -610,6 +610,7 @@ func TestAccContainerCluster_withNodePoolNodeConfig(t *testing.T) {
 
 func TestAccContainerCluster_withMaintenanceWindow(t *testing.T) {
 	t.Parallel()
+	clusterName := acctest.RandString(10)
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
@@ -617,7 +618,14 @@ func TestAccContainerCluster_withMaintenanceWindow(t *testing.T) {
 		CheckDestroy: testAccCheckContainerClusterDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccContainerCluster_withMaintenanceWindow("03:00"),
+				Config: testAccContainerCluster_withMaintenanceWindow(clusterName, "03:00"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckContainerCluster(
+						"google_container_cluster.with_maintenance_window"),
+				),
+			},
+			{
+				Config: testAccContainerCluster_withMaintenanceWindow(clusterName, ""),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckContainerCluster(
 						"google_container_cluster.with_maintenance_window"),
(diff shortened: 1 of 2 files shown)


google/resource_container_cluster_test.go
```diff
@@ -610,14 +610,22 @@ func TestAccContainerCluster_withNodePoolNodeConfig(t *testing.T) {
...
-				Config: testAccContainerCluster_withMaintenanceWindow("03:00"),
+				Config: testAccContainerCluster_withMaintenanceWindow(clusterName, "03:00"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckContainerCluster(
+						"google_container_cluster.with_maintenance_window"),
+				),
+			},
+			{
+				Config: testAccContainerCluster_withMaintenanceWindow(clusterName, ""),
 				Check: resource.ComposeTestCheckFunc(
```
michaelbannister: I'm not sure if this extra step is worth it. I wanted to add something to prove that the cluster had _not_ been torn down & recreated, but couldn't find a neat way to do this.

michaelbannister: I also tried to add code to `testAccCheckContainerCluster` to prove the absence of a maintenance window:
```go
		if cluster.MaintenancePolicy == nil {
			clusterTests = append(clusterTests, clusterTestField{"maintenance_policy", []string{}})
		} else {
			clusterTests = append(clusterTests, clusterTestField{"maintenance_policy.0.daily_maintenance_window.0.start_time", cluster.MaintenancePolicy.Window.DailyMaintenanceWindow.StartTime})
			clusterTests = append(clusterTests, clusterTestField{"maintenance_policy.0.daily_maintenance_window.0.duration", cluster.MaintenancePolicy.Window.DailyMaintenanceWindow.Duration})
		}
```
but it turned out not really to prove anything: if I failed to correctly implement the "update MaintenanceWindow to nil" call, the state passed into the TestCheckFunc was the _actual_ state (i.e. still with a non-nil MaintenanceWindow) so the test passed even though the apply hadn't done what I wanted.

danawillow: Yeah, our test infrastructure definitely leaves something to be desired. I do think having the test to check that you can remove a maintenance window is good to have. For the "check that it was a real update", https://github.com/hashicorp/terraform/issues/15126. And for proving the absence, you could use `resource.TestCheckNoResourceAttr`- then we have those sorts of checks making sure our state is what we expect it to be, and `testAccCheckContainerCluster` to (mostly) check that the state matches what GCP says.

michaelbannister:
Thanks @danawillow for the comments & suggestions…
Happy Christmas/Holidays/Hanukkah (belated)!

danawillow:
Happy Holidays to you too!

```
=== RUN   TestAccContainerCluster_withMaintenanceWindow
--- PASS: TestAccContainerCluster_withMaintenanceWindow (309.10s)
```

PR #917, opened 2018-01-04
Improve documentation around instance scopes

### LINKED ISSUE #913: scopes param for service_account of google_compute_instance should be optional
_This issue was originally opened by @borgilb as hashicorp/terraform#17004. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
Terraform v0.11.1
+ provider.google v1.4.0

```hcl
resource "google_compute_instance" "instance01" {
  .....
  service_account {
    email = "myserviceaccount@myproject.iam.gserviceaccount.com"
    scopes = []
  }
}
```

### Expected Behavior
"scopes" should not be a required param. I should be able to specify a specific service account and not include any scopes.

### Actual Behavior
"scopes" is a _required_ param- this is wrong. Using scopes like this is deprecated as well. If you specify a service account access is granted based on roles associated with that specific account. If you _don't_ specify a service account your using the default GCE account and then can grant specific access via scopes. Because scopes is required it means if I specify a service account I still have to specify specific scopes which defeats the purpose of a service account! In fact the GCP console doesn't even let you specify scopes if you specify a service account when creating a new instance.

### Suggestion
Perhaps you should have an option to just allow ALL scopes. This seems to be what happens when you create an instance through the console with a specific account. "Cloud API access scopes" is set to "Allow full access to all Cloud APIs". I tried `scopes = ["*"]` but seems that doesn't work.

### References
https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances

rosbo:
Fixes #913

### DIFF
--- a/website/docs/r/compute_instance.html.markdown
+++ b/website/docs/r/compute_instance.html.markdown
@@ -214,7 +214,8 @@ The `service_account` block supports:
     default Google Compute Engine service account is used.
 
 * `scopes` - (Required) A list of service scopes. Both OAuth2 URLs and gcloud
-    short names are supported.
+    short names are supported. To allow full access to all Cloud APIs, use the
+    `cloud-platform` scope. See a complete list of scopes [here](https://cloud.google.com/sdk/gcloud/reference/alpha/compute/instances/set-scopes#--scopes).
 
 The `scheduling` block supports:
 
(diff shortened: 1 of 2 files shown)


rosbo:
This also fixes #88.

PR #928, opened 2018-01-06
Fix typo in error message

ofpiyush:
The error message should have been `source` instead of `string`.


danawillow:
Good catch, thanks @ofpiyush!

PR #929, opened 2018-01-08
Ensure Elem is always either a *Schema or *Resource

jimmy-btn:
This PR prepares this provider for hashicorp/terraform#17037, which will cause Provider.InternalValidate to reject schemas which assign invalid types (anything other than a *Schema or a *Resource) to the Elem field of a Schema.

This mostly addresses issues where the expected type for a Map is provided directly rather than wrapped in schema.Schema{}. Today this is a non-issue, but will cause problems in the future if we enforce type checking on Map types.


danawillow:
Thanks @jimmy-btn!

PR #939, opened 2018-01-09
Standardize resource name extraction from self_link/uri

rosbo:
Doing it manually is error prone and adds extra code. 
Having two methods doing the same thing is bad. 

Better standardize.


rosbo:
`linkDiffSupress` is added as a DiffSuppressFunc for some fields in resource schemas. There is no unit test for it (I will open a new PR to fix this) but this is tested indirectly through the acceptance tests.

PR #945, opened 2018-01-11
 Support subnetwork IP CIDR range expansion

### LINKED ISSUE #57: google_compute_subnetwork resource - support an expansion of CIDR range
_This issue was originally opened by @Dnefedkin as hashicorp/terraform#11689. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
0.8.5

### Affected Resource(s)
- google_compute_subnetwork

Changing ip_cidr_range argument of google_compute_subnetwork resource leads to recreation of the subnetwork object, the update method is not implemented.

GCE API has a method Subnetworks: expandIpCidrRange [1] that allows to modify the existing subnetwork and change a CIDR range in case if a new CIDR range is a superset of the range previously defined before the update.

The proposal is to modify google_compute_subnetwork resource by adding code allowing to expand a CIDR range.  It's not clear to me how resource should behave in a new CIDR range is not a superset of an old one. 

[1] - https://cloud.google.com/compute/docs/reference/latest/subnetworks/expandIpCidrRange

rosbo:
Fixes #57

cc/ @Dnefedkin

### DIFF
--- a/google/resource_compute_subnetwork.go
+++ b/google/resource_compute_subnetwork.go
@@ -6,9 +6,12 @@ import (
 
 	"strings"
 
+	"github.com/apparentlymart/go-cidr/cidr"
+	"github.com/hashicorp/terraform/helper/customdiff"
 	"github.com/hashicorp/terraform/helper/schema"
 	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
+	"net"
 	"time"
 )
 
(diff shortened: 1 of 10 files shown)


rosbo:
```sh
--- PASS: TestAccComputeSubnetwork_update (115.82s)
```

PR #946, opened 2018-01-11
Add support for setting bucket's logging config

skinp:
Mostly self explanatory, this enables setting up which bucket to log access/storage logs to.

In theory, GCP also allows us to set a log_object_prefix. Unfortunately I couldn't easily get this working, so this is a first step. It's also why logging ends up being a TypeList instead of TypeString... to allow for future support of log_object_prefix.

Setting log_object_prefix wasn't obvious for the following reasons... In theory, this should an optional field but GCP sets it to the bucket name by default. It wasn't straightforward to go back and forth from an explicitly set value to the default because the expected TF state is "" but GCP reports it as the bucket name. It seems like the easiest approach would be to set a Default(Func) value for log_object_prefix dynamically to be the bucket's name, but I couldn't figure out how to make this work. Open for suggestions here...

### DIFF
--- a/google/resource_storage_bucket.go
+++ b/google/resource_storage_bucket.go
@@ -223,6 +223,11 @@ func resourceStorageBucket() *schema.Resource {
 							Type:     schema.TypeString,
 							Required: true,
 						},
+						"log_object_prefix": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+						},
 					},
 				},
 			},
@@ -362,8 +367,7 @@ func resourceStorageBucketUpdate(d *schema.ResourceData, meta interface{}) error
 	if d.HasChange("logging") {
 		if v, ok := d.GetOk("logging"); ok {
 			sb.Logging = expandBucketLogging(v.([]interface{}))
-		}
-		if sb.Logging == nil {
+		} else {
 			sb.NullFields = append(sb.NullFields, "Logging")
 		}
 	}
@@ -521,7 +525,8 @@ func expandBucketLogging(configured interface{}) *storage.BucketLogging {
 	logging := loggings[0].(map[string]interface{})
 
 	bucketLogging := &storage.BucketLogging{
-		LogBucket: logging["log_bucket"].(string),
+		LogBucket:       logging["log_bucket"].(string),
+		LogObjectPrefix: logging["log_object_prefix"].(string),
 	}
 
 	return bucketLogging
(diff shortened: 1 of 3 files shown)


amoiseiev:
Added an issue for this pull request, for tracking purposes on our end: https://github.com/terraform-providers/terraform-provider-google/issues/972

skinp:
@rosbo friendly bump?
I'm not sure we need to care about log_object_prefix immediately, it'd be useful for completeness, but just adding log_bucket support by itself is a step in the right direction I think (I just use log_bucket myself, not the prefix).

google/resource_storage_bucket.go
```diff
@@ -342,6 +359,15 @@ func resourceStorageBucketUpdate(d *schema.ResourceData, meta interface{}) error
 		sb.Cors = expandCors(v.([]interface{}))
 	}
 
+	if d.HasChange("logging") {
+		if v, ok := d.GetOk("logging"); ok {
+			sb.Logging = expandBucketLogging(v.([]interface{}))
+		}
+		if sb.Logging == nil {
```
danawillow: this can just be `else`, right?

skinp: Yes, will fix

skinp:
Right, I toyed with Computed for awhile but couldn't get it to revert back to the default if I removed the explicit config like you're describing. I'll go with the "something that works most of the time" for now.

skinp:
Test run fwiw.
```
$ make testacc TEST=./google TESTARGS='-run=TestAccStorageBucket_' GOOGLE_USE_DEFAULT_CREDENTIALS=true
TF_ACC=1 go test ./google -v -run=TestAccStorageBucket_ -timeout 120m
=== RUN   TestAccStorageBucket_basic
=== RUN   TestAccStorageBucket_lowercaseLocation
=== RUN   TestAccStorageBucket_customAttributes
=== RUN   TestAccStorageBucket_lifecycleRules
=== RUN   TestAccStorageBucket_storageClass
=== RUN   TestAccStorageBucket_update
=== RUN   TestAccStorageBucket_forceDestroy
=== RUN   TestAccStorageBucket_versioning
=== RUN   TestAccStorageBucket_logging
=== RUN   TestAccStorageBucket_cors
=== RUN   TestAccStorageBucket_labels
--- PASS: TestAccStorageBucket_lifecycleRules (4.15s)
--- PASS: TestAccStorageBucket_basic (4.17s)
--- PASS: TestAccStorageBucket_cors (2.91s)
--- PASS: TestAccStorageBucket_forceDestroy (15.21s)
--- PASS: TestAccStorageBucket_logging (16.47s)
--- PASS: TestAccStorageBucket_update (25.86s)
--- PASS: TestAccStorageBucket_versioning (12.97s)
--- PASS: TestAccStorageBucket_labels (24.35s)
--- PASS: TestAccStorageBucket_customAttributes (5.43s)
--- PASS: TestAccStorageBucket_lowercaseLocation (5.33s)
--- PASS: TestAccStorageBucket_storageClass (8.70s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	37.028s
```

danawillow:
Thanks @skinp! Merging now.

PR #948, opened 2018-01-11
Read compute instance boot disk initialization params from API

### LINKED ISSUE #87: Terraform refresh does not recognize disk-type change in real-world state
_This issue was originally opened by @aawilson as hashicorp/terraform#14136. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
0.9.3

### Affected Resource(s)
google_compute_instance

```hcl

module "test-prod" {
  source = "../modules/parent"
  disk_size = "10"
  disk_type = "pd-standard"
  zone = "${var.zone}"
  host_name = "test1"
  host_domain = "${var.host_domain}"
}

```

### Expected Behavior
If I update an instance so that it has a new disk-type (changing pd-ssd to pd-standard, for instance), and then update my tfvars file accordingly, terraform plan (with its implicit terraform refresh) should report no changes.

### Actual Behavior
Verify real-world state of the disk is pd-standard:
```
~ gcloud compute instances describe test1 --project=main --zone=us-central1-a
..etc..
disks:
..etc..
  source: https://www.googleapis.com/compute/v1/projects/main/zones/us-central1-a/disks/test1

~ gcloud compute disks describe test1 --project=main --zone=us-central1-a | g
rep type:
type: https://www.googleapis.com/compute/v1/projects/main/zones/us-central1-a/diskTypes/pd-standard
```
Output when plan is run
```
data.terraform_remote_state.prod: Refreshing state...
..etc..
google_compute_instance.parent: Refreshing state... (ID: test1)
..etc..
-/+ module.test-prod.google_compute_instance.parent
..irrelevant, unchanged attributes omitted..
    disk.0.type:                                         "pd-ssd" => "pd-standard" (forces new resource)
    name:                                                "test1" => "test1"
    zone:                                                "us-central1-a" => "us-central1-a"
```
### Steps to Reproduce
1. Have terraform in a state with a google compute instance whose disk type is pd-ssd
2. Using gce commands or a console, put the instance that already exists in a state where its disk has a type pd-standard
3. Change local tfvars to reflect the new state (that is, change pd-ssd to pd-standard)
3. Run `terraform plan` and note the unexpected disk.0.type change

rosbo:
Fixes #87

Pros:
- Fixes import.
- Detect changes happening outside Terraform.

Cons:
- If the image is specified in the config using the image family, then the image names for that family must include the family name in its own name (all official images do) otherwise the diff won't be properly suppress. The same limitation applies to the `google_compute_disk` `image` field.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -79,6 +79,7 @@ func resourceComputeInstance() *schema.Resource {
 						"initialize_params": &schema.Schema{
 							Type:     schema.TypeList,
 							Optional: true,
+							Computed: true,
 							ForceNew: true,
 							MaxItems: 1,
 							Elem: &schema.Resource{
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -86,6 +86,7 @@ func resourceComputeInstance() *schema.Resource {
 									"size": &schema.Schema{
 										Type:         schema.TypeInt,
 										Optional:     true,
+										Computed:     true,
 										ForceNew:     true,
 										ValidateFunc: validation.IntAtLeast(1),
 									},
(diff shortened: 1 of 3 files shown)


rosbo:
Added.

PR #959, opened 2018-01-15
updating Example creating a SQL Database with database_version

discentem:
Including explicit use of database_version to show how to specific postgres. This was unclear to me initially.


discentem:
Thanks @ndmckinley! Hmm... is the official documentation _actually_ served off of this repository? Just curious, as the docs don't yet reflect my change.

rosbo:
The documentation website is updated whenever we release a new version of the google terraform provider plugin. Stay tuned.

PR #965, opened 2018-01-16
Add disable_on_destroy option to google_project_service.

### LINKED ISSUE #5: Add setting to stop google_project_services from being deleted
I'm really enjoying using Google Cloud with Terraform. So far the only blemish has been the wait time for provisioning and removing services, and more generally having to deal with and think about them at all.

I have specified the services my project requires with `google_project_services` so they are created the first time I provision a project. However if I destroy all the resources in my project and then recreate them (with `skip_delete` on the project itself), most of my time is spent waiting for the project services to be destroyed and recreated.

I could just keep the project services out of my Terraform config, but then they wouldn't get created when I'm making a new project (say a temporary staging project).

What are your thoughts on adding a setting for `google_project_services` which provides accumulate only behaviour? It would add any missing services specified in the terraform config, but never delete any which were extra in the console, even when the resource is being destroyed. It also would only show a diff if there were services to create. This is similar to the `google_project` setting `skip_delete`, but I think a different name would be required here, perhaps `accumulate_only`?

If project service APIs took seconds to provision and remove then this probably wouldn't be necessary, but I suspect this would be the easier change to make 😄.

ndmckinley:
Closes #5.

### DIFF
--- a/google/resource_google_project_service.go
+++ b/google/resource_google_project_service.go
@@ -2,6 +2,7 @@ package google
 
 import (
 	"fmt"
+	"log"
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
@@ -12,6 +13,7 @@ func resourceGoogleProjectService() *schema.Resource {
 		Create: resourceGoogleProjectServiceCreate,
 		Read:   resourceGoogleProjectServiceRead,
 		Delete: resourceGoogleProjectServiceDelete,
+		Update: resourceGoogleProjectServiceUpdate,
 
 		Schema: map[string]*schema.Schema{
 			"service": {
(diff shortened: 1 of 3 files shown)


ndmckinley:
```
TF_ACC=1 go test ./google -v -run=TestAccGoogleProjectService_basic -timeout 120m
=== RUN   TestAccGoogleProjectService_basic                                    
                                                                
--- PASS: TestAccGoogleProjectService_basic (64.95s)                           
PASS                                         
ok      github.com/terraform-providers/terraform-provider-google/google 64.958s
```

ndmckinley:
CI's broken - I built locally and all is well.
https://www.traviscistatus.com/incidents/c1s4dlyxd4lf

PR #970, opened 2018-01-17
Add support for Google Cloud IoT Core registry

Ebolon:
This pull request adds support for the Google Cloud IoT Core registry. The config structure reflects the structure of the [registry API](https://cloud.google.com/iot/docs/reference/rest/v1/projects.locations.registries). The downside is a lot of nesting in the config. But hopefully it allows a easy extension if the API evolves.


google/resource_cloudiot_registry_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+  mqtt_config = {
+    mqtt_enabled_state = "MQTT_DISABLED"
+  }
+
+  credentials = [
+    {
+      "public_key_certificate" = {
+        format      = "X509_CERTIFICATE_PEM"
+        certificate = "-----BEGIN CERTIFICATE-----\nMIICnjCCAYYCCQC/5gx7LgJFqTANBgkqhkiG9w0BAQsFADARMQ8wDQYDVQQDDAZ1\nbnVzZWQwHhcNMTgwMTEyMjAxMzQzWhcNMjMwMTExMjAxMzQzWjARMQ8wDQYDVQQD\nDAZ1bnVzZWQwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDStfQvJzmN\nCYLSWpwvTmyCKn8t19cfWZ69wFaB3OSglxXgYe3w9An0QHybDpKITt61PpfsKov3\nEcnzH5IA+Ox+4jUppBL1mSkO/BWig+sd1dG7pQMbGi4nMxW704A0PRUaNIOarOlR\nrNUJZQrsghkMjLayCTJ2HISBBiPnKKB3f3KCc9sDhj2Z7zy7HfeW0apZ1m6xAQCC\neSZNW0IyGIYKTd9F7HEJFzOWg9JHvabbciBEcFWKGVzM8nQr1q8KU8Xi3iN2mpNK\nJkbRLNnqKhvjPyIZ4s4cDSEZN1/OaGQ4XP2mvU03+4UAoMPoJ8IczBKTB0mFxfX8\nlDZZa5IWU9sNAgMBAAEwDQYJKoZIhvcNAQELBQADggEBAHnkTIghRj/cerR9ctji\nkancnjlsdNEuPiVpMj+SOtOH8cvlgl0oWG6segYTVzk4VEHlq3POB67Yjoz829XM\nCEgUxSqGvDrQ7IaPLPryYy8o5azMLnEZDr+Yd6CUKr/pUZzJoZxHj7z3iqeQZnMW\nS6kb6HYvG5PKlJ7+JUIKLou0RQmaM9BQ0Nln/YDRRIerD0MY9k7No2ZEDbywZqQK\nGRIqT+BlN84oHOR44h2RqWhn9O50tkbcmAIKgmeg/mxwmeAm/6hQ8VrOhDHqsFdT\nzh2l6IeCl8EF8MjNrFRcQx21TTqeU6vGIPgM3E0k8PQUc+s+lir8UFsIzKaOFsIh\nuKU=\n-----END CERTIFICATE-----\n"
```
ndmckinley: Consider committing this as a file to google/test-fixtures instead.

google/resource_cloudiot_registry_test.go
```diff
@@ -0,0 +1,206 @@
...
+		_, err := config.clientCloudiot.Projects.Locations.Registries.Get(rs.Primary.ID).Do()
+		if err != nil {
+			return fmt.Errorf("Registry does not exist")
+		}
+
+		return nil
+	}
+}
+
+func testAccCloudiotRegistryUpdate(n string) resource.TestCheckFunc {
```
ndmckinley: This seems elaborate - I think it only tests that the state matches the config, which is implicitly part of a test that has config.  If the state doesn't match the config post-update, the test will fail even without this method, and so removing this method reduces the difficulty of maintenance.

website/docs/r/cloudiot_registry.html.markdown
```diff
@@ -0,0 +1,116 @@
...
+
+  mqtt_config = {
+    mqtt_enabled_state = "MQTT_ENABLED"
+  }
+
+  credentials = [
+    {
+      public_key_certificate = {
+        format      = "X509_CERTIFICATE_PEM"
+        certificate = "<CERTIFICATE>"
```
ndmckinley: Wouldn't hurt to do, maybe, "${file("foo.pem")}" or something like that so the example is more valid?

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,411 @@
...
+								[]string{httpStateUnspecified, httpEnabled, httpDisabled}, false),
+						},
+					},
+				},
+			},
+			"credentials": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: false,
+				MaxItems: 10,
```
ndmckinley: Where does the max of 10 come from?

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,411 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+			"event_notification_configs": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: false,
+				MaxItems: 1,
```
ndmckinley: If it's a max of one, the name should maybe be the singular "config"?

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,411 @@
...
+							ValidateFunc: validation.StringInSlice(
+								[]string{httpStateUnspecified, httpEnabled, httpDisabled}, false),
+						},
+					},
+				},
+			},
+			"credentials": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: false,
```
ndmckinley: ForceNew: false is the default and can be omitted, here and elsewhere.

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,411 @@
...
+			},
+		},
+	}
+}
+
+func expandEventNotificationConfigs(configs []interface{}) []*cloudiot.EventNotificationConfig {
+	eventConfigs := make([]*cloudiot.EventNotificationConfig, len(configs))
+	for i, raw := range configs {
+		data := raw.(map[string]interface{})
+
```
ndmckinley: Remove empty lines where they don't add meaning.

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,411 @@
...
+func expandStateNotificationConfig(config map[string]interface{}) *cloudiot.StateNotificationConfig {
+	if v, ok := config["pubsub_topic_name"]; ok {
+		return &cloudiot.StateNotificationConfig{
+			PubsubTopicName: v.(string),
+		}
+	}
+	return nil
+}
+
+func expandMqttConfig(config map[string]interface{}) *cloudiot.MqttConfig {
```
ndmckinley: I'm not sure these are "expand"s, which we normally use to mean translating config lists -> go API object slices.  Do you think the naming convention should stretch to cover this case or do you have another idea for naming these?

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+
+	region, err := getRegion(d, config)
+	if err != nil {
+		return err
+	}
+
+	parent := fmt.Sprintf("projects/%s/locations/%s", project, region)
+
+	deviceRegistry := expandDeviceRegistry(d)
```
ndmckinley: Here, for instance, createDeviceRegistry would be more accurate.

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,411 @@
...
+	d.Set("name", res.Id)
+
+	if res.EventNotificationConfigs != nil {
+		eventConfigs := make([]map[string]string, len(res.EventNotificationConfigs))
+		for i, config := range res.EventNotificationConfigs {
+			eventConfigs[i] = map[string]string{"pubsub_topic_name": config.PubsubTopicName}
+		}
+		d.Set("event_notification_configs", eventConfigs)
+	}
+	// to keep the data model lean only changes are pushed if not the default is
```
ndmckinley: Could you clean up the wording in this comment a bit?

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,411 @@
...
+	name := d.Id()
+	res, err := config.clientCloudiot.Projects.Locations.Registries.Get(name).Do()
+
+	if err != nil {
+		return handleNotFoundError(err, d, fmt.Sprintf("Registry %q", name))
+	}
+
+	d.Set("name", res.Id)
+
+	if res.EventNotificationConfigs != nil {
```
ndmckinley: If they are nil, is it right to clear them from the ResourceData object?  To stay up to date with, for instance, a deleted config from the console?

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,411 @@
...
+	call := config.clientCloudiot.Projects.Locations.Registries.Delete(name)
+	_, err := call.Do()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func resourceCloudiotRegistryStateImporter(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
ndmckinley: Need to see an import test.

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,411 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	region, err := getRegion(d, config)
+	if err != nil {
+		return nil, err
+	}
+
+	id := fmt.Sprintf("projects/%s/locations/%s/registries/%s", project, region, d.Id())
```
ndmckinley: Problem - in this mode, the user must configure the provider such that the default region and project are the right ones - d.Id() is the only thing you'll have going into the import.

google/resource_cloudiot_registry.go
Ebolon: I found the max 10 in the documentation of [DeviceRegistry](https://cloud.google.com/iot/docs/reference/rest/v1/projects.locations.registries#DeviceRegistry) for the `credentials` field: "No more than 10 credentials can be bound to a single registry at a time.". Hope I got it right.

google/resource_cloudiot_registry.go
Ebolon: You are right. It sounded strange to me when I read: "Only the first configuration is used." in [eventNotificationConfigs[]](https://cloud.google.com/iot/docs/reference/rest/v1/projects.locations.registries#DeviceRegistry.FIELDS.event_notification_configs). Would it make sense to get rid of the list and map with `pubsub_topic_name`?  I would suggest a field like `event_notification_topic` to keep the topic id.

google/provider.go
```diff
@@ -88,6 +88,7 @@ func Provider() terraform.ResourceProvider {
 			"google_bigtable_instance":                     resourceBigtableInstance(),
 			"google_bigtable_table":                        resourceBigtableTable(),
 			"google_cloudfunctions_function":               resourceCloudFunctionsFunction(),
+			"google_cloudiot_registry":                     resourceCloudiotRegistry(),
```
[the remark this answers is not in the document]
Ebolon: Looks much better 👍 Done.

google/resource_cloudiot_registry.go
Ebolon: For now I just removed the list schema.

google/resource_cloudiot_registry.go
Ebolon: 👍

google/resource_cloudiot_registry.go
Ebolon: I just renamed the function to `buildMqttConfig` (and the others). Hoping that this fits better.

google/resource_cloudiot_registry.go
Ebolon: 👍

google/resource_cloudiot_registry_test.go
Ebolon: 👍 Have removed this function. I was not aware of this internal test.

website/docs/r/cloudiot_registry.html.markdown
Ebolon: Yep! Sounds much better!

google/resource_cloudiot_registry_test.go
Ebolon: 👍

google/resource_cloudiot_registry.go
Ebolon: Not thought about that. I now ask for full path `projects/{project}/locations/{region}/registries/{name}` as id. (see also [registry get API](https://cloud.google.com/iot/docs/reference/rest/v1/projects.locations.registries/get)).

google/resource_cloudiot_registry.go
Ebolon: Done

google/resource_cloudiot_registry.go
Ebolon: Done

google/resource_cloudiot_registry.go
Ebolon: Yes I missed that. Now I reset first everything to nil.

google/resource_cloudiot_registry.go
Ebolon: Done

Ebolon:
Thank you very much for your feedback @ndmckinley! Hope I got now everything right. How does it look now?

PS: I added a validation function for the registry id. I hope the function name is ok. This can be reuses later for a device id (see: [Creating or editing a device](https://cloud.google.com/iot/docs/how-tos/devices#creating_or_editing_a_device)).

google/resource_cloudiot_registry_test.go
ndmckinley: Oh, I apologize but I was wrong about this - this is true only when there's an importstate step on the test.  If you can add one of those in, these tests will be easier to maintain, but if that's a hassle for unforeseen reasons you can definitely go back to this method.  My bad!

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,381 @@
...
+
+func resourceCloudIoTRegistryRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	name := d.Id()
+	res, err := config.clientCloudIoT.Projects.Locations.Registries.Get(name).Do()
+	if err != nil {
+		return handleNotFoundError(err, d, fmt.Sprintf("Registry %q", name))
+	}
+
+	d.Set("name", res.Id)
```
ndmckinley: A comment here which explains that you're speculatively setting all attributes to nil and updating the ones you find later might help.

google/validation.go
```diff
@@ -107,3 +109,16 @@ func validateRFC1035Name(min, max int) schema.SchemaValidateFunc {
 
 	return validateRegexp(fmt.Sprintf("^"+RFC1035NameTemplate+"$", min-2, max-2))
 }
+
+func validateCloudIoTId(v interface{}, k string) (warnings []string, errors []error) {
```
ndmckinley: nitpick: "ID" is capitalized when it's short for "identifier"
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

google/validation.go
```diff
@@ -18,6 +19,7 @@ const (
 	SubnetworkLinkRegex = "projects/(" + ProjectRegex + ")/regions/(" + RegionRegex + ")/subnetworks/(" + SubnetworkRegex + ")$"
 
 	RFC1035NameTemplate = "[a-z](?:[-a-z0-9]{%d,%d}[a-z0-9])"
+	CloudIoTIdRegex     = "[a-zA-Z][-a-zA-Z0-9._+~%]{2,254}"
```
ndmckinley: Is this well-tested?  The formatter is trying to complain about the % as if it's a format-string specifier and I'm pretty sure it's a formatter bug and not a bug in your code.  :)

google/resource_cloudiot_registry.go
```diff
@@ -0,0 +1,381 @@
...
+
+	deviceRegistry := createDeviceRegistry(d)
+	deviceRegistry.Id = d.Get("name").(string)
+
+	call := config.clientCloudIoT.Projects.Locations.Registries.Create(parent, deviceRegistry)
+	res, err := call.Do()
+	if err != nil {
+		return err
+	}
+	d.SetId(res.Name)
```
ndmckinley: It can be helpful to do d.SetId(name) before the create call, then do d.SetId(nil) in the event of an error - that way, if the process is killed during the request, the state file will include the resource, and we'll be able to do a Read() on refresh to determine if the call ever actually succeeded.

ndmckinley:
It's looking pretty good to me - I made a mistake in the earlier review, sorry for the back-and-forth, but other than that, and these small changes, I'm happy with it.

google/validation.go
Ebolon: Done.

google/resource_cloudiot_registry.go
Ebolon: After some testing I saw that all the nil checks are not necessary. Hope my changes now makes it more readable.
Just to make sure we have the same thing in mind when we read the code:
This is
```
resource "google_cloudiot_registry" "hello" {
  name = "hello"
}
```
comparable with
```
{
 "id": "hello",
 "name": "projects/<...>/locations/<...>/registries/hello",
 "mqttConfig": {
  "mqttEnabledState": "MQTT_ENABLED"
 },
 "stateNotificationConfig": {
 },
 "httpConfig": {
  "httpEnabledState": "HTTP_ENABLED"
 }
}
```
 because the code ignores default state, mqtt, http config:

google/resource_cloudiot_registry.go
Ebolon: Done.

google/validation.go
Ebolon: Unit test is now included. Found bug and yep this seems to be a formatter bug :-)

Ebolon:
Have implemented the requested changes. Found a bug where the certificates were not properly read into the state, because I forgot the `public_key_certificate` key. Ready for review.

ndmckinley:
Great - I'm running the tests now and if they pass, I'll merge it.

ndmckinley:
Hey there, looks like we're getting errors from the tests, which seem to assume a particular setup:

`* google_cloudiot_registry.foobar: googleapi: Error 403: Topic 'projects/<OUR PROJECT NAME>/topics/psregistry-test-devicestatus-pqzo18advm' was specified, but user 'cloud-iot@system.gserviceaccount.com' does not have 'pubsub.topics.publish' permission to publish events to the topic, or the topic doesn't exist. Make sure that the topic exists, and edit the topic permissions to add the 'Pub/Sub Publisher' role https://cloud.google.com/pubsub/docs/access_control#tbl_roles to the 'cloud-iot@system.gserviceaccount.com' account. You can also run the following command to add publish permission to all topics in your project:\n\ngcloud projects add-iam-policy-binding $PROJECT --member=serviceAccount:cloud-iot@system.gserviceaccount.com --role=roles/pubsub.publisher, forbidden`

Ebolon:
The test creates two pubsub topics. Did I miss somewhere to say what permissions are needed for the test?

I found that the terraform schema validation did not work if `certificate` in `public_key_certificate` is missing, although it is marked as required. Should I catch those errors or wait for terraform to catch those errors?

PS: Sorry for last minute changes. I played a bit more with the API. `UNSPECIFIED_PUBLIC_KEY_CERTIFICATE_FORMAT` only results in an error and `MQTT_STATE_UNSPECIFIED` or `MQTT_STATE_UNSPECIFIED` are never returned. (Only `_ENABLED`is returned) Just thought I remove those before it generates some confusion.

ndmckinley:
The important thing is that the test passes without any special configuration on our side ahead of time - ideally we wouldn't need to manually create a service account and grant it permissions.  The best case would be for you to put that service account creation and permission grants into the test terraform config!  But if that doesn't work out, let me know how to set it up and I'll see what I can do.  :)

ndmckinley:
Thanks!  I'll run the tests again.

Ebolon:
Ouch! This was a RFTM moment. Got it fixed now :-)

ndmckinley:
Hey, wonderful, it passes!  :)  Last thing - the import test is obsolete since your other tests do an import now, and you can remove it without decreasing coverage.  After that, I'll merge it.  :)

Ebolon:
Just removing `TestAccCloudIoTRegistryCreate_basic`?

ndmckinley:
Yes, you can remove the entire import_*_test.go file.

Ebolon:
Done.

ndmckinley:
Thank you!

Issue #64, opened 2017-06-13
provider/google/storage: support for defaultObjectAccessControls


hashibot:
_This issue was originally opened by @renchap as hashicorp/terraform#12481. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

In the same way as `google_storage_bucket_acl` and `google_storage_object_acl` there should be a `google_storage_default_object_acl` to be able to specify the default ACL for new objects in a bucket. This is useful for example in a public bucket where you want all objects to be readable by anybody.

API documentation here : https://cloud.google.com/storage/docs/json_api/v1/defaultObjectAccessControls

Example:
```hcl
resource "google_storage_default_object_acl" "public-read-acl" {
  bucket = "${google_storage_bucket.bucket.name}"
  role_entity = [
    "OWNER:project-owners-123456",
    "READER:allUsers"
  ]
}
```

danawillow:
Fixed in #992

PR #992, opened 2018-01-22
Add google_storage_default_object_acl resource

ishashchuk:
This addresses issue [#64](https://github.com/terraform-providers/terraform-provider-google/issues/64)  and adds support for creating storage default object acls ([API doc](https://cloud.google.com/storage/docs/json_api/v1/defaultObjectAccessControls)

The resource is similar to an existing resource for storage_object_acl, with the exception of setting up permissions on the bucket level


google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+
+func getDefaultObjectAclId(bucket string) string {
+	return bucket + "-default-object-acl"
+}
+
+func resourceStorageDefaultObjectAclCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	bucket := d.Get("bucket").(string)
+	role_entity := make([]interface{}, 0)
```
danawillow: nit: use mixedCaps please for variables (https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps)

google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+		Delete: resourceStorageDefaultObjectAclDelete,
+
+		Schema: map[string]*schema.Schema{
+			"bucket": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"role_entity": &schema.Schema{
```
danawillow: I'd highly recommend just having two separate fields here- one for the role, and the other for the entity. The id for the resource can then be `bucket/entity`, which would allow reading back the resource just from the id. If a Terraform user wants to specify multiple default acls, they can just have multiple resources. I'm going to hold off on reviewing the rest of this PR for now since it'll change a lot if you end up going with this approach.

ishashchuk: I don't necessarily think it is a good idea, because it is very likely that users will have multiple default acls. Default acls get applied to the objects only at object creation time, which means that the explicit dependency will need to be set for "google_storage_bucket_object" resource on "google_storage_default_object_acl" resource to ensure that if both of those resources are created simultaneously, the objects always gets created after the default acl is in place. Having multiple resources will make it more complicated to control for those explicit dependencies. 

Besides, this resource  should be similar in use with google_storage_object_acl, since it essentially does the same thing(sets up object permissions), but on the bucket level. So it seems to me that both should have similar usage.

google/resource_storage_default_object_acl.go
ishashchuk: I'll fix the variable naming in the next review, once I get your thoughts on the comment above
Thanks!

google/resource_storage_default_object_acl.go
danawillow: Ah ok, I can get behind having them be consistent with each other. That's fine then, I'll continue reviewing as-is :)

google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+				Type:     schema.TypeList,
+				Required: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+		},
+	}
+}
+
+func getDefaultObjectAclId(bucket string) string {
+	return bucket + "-default-object-acl"
```
danawillow: That suffix isn't really necessary

google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+			if _, in := re_local_map[entity]; in {
+				role_entity = append(role_entity, fmt.Sprintf("%s:%s", role, entity))
+				log.Printf("[DEBUG]: saving re %s-%s", v.Role, v.Entity)
+			}
+		}
+
+		d.Set("role_entity", role_entity)
+	}
+
+	d.SetId(getDefaultObjectAclId(bucket))
```
danawillow: The id should be set at the end of create, since we want to make sure the state file has some information regardless of whether or not the read succeeded.

google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+	config := meta.(*Config)
+
+	bucket := d.Get("bucket").(string)
+	role_entity := make([]interface{}, 0)
+
+	if v, ok := d.GetOk("role_entity"); ok {
+		role_entity = v.([]interface{})
+	}
+
+	if len(role_entity) > 0 {
```
danawillow: Let's do MinItems: 1 on the schema element for this and then remove this check

google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+	return bucket + "-default-object-acl"
+}
+
+func resourceStorageDefaultObjectAclCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	bucket := d.Get("bucket").(string)
+	role_entity := make([]interface{}, 0)
+
+	if v, ok := d.GetOk("role_entity"); ok {
```
danawillow: You can just do `role_entity := d.Get("role_entity")`, no need for the GetOk

google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	return nil
+}
+
+func resourceStorageDefaultObjectAclRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	bucket := d.Get("bucket").(string)
+
+	if _, ok := d.GetOk("role_entity"); ok {
```
danawillow: same here

google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+		role_entity := make([]interface{}, 0)
+		re_local := d.Get("role_entity").([]interface{})
```
danawillow: and then by this point you'll have already d.gotten, so this var can be removed

google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+	return nil
+}
+
+func resourceStorageDefaultObjectAclRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	bucket := d.Get("bucket").(string)
+
+	if _, ok := d.GetOk("role_entity"); ok {
+		role_entity := make([]interface{}, 0)
```
danawillow: nit: roleEntities since it's a list?

google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+
+		for _, v := range new_re {
+			pair, err := getRoleEntityPair(v.(string))
+
+			ObjectAccessControl := &storage.ObjectAccessControl{
+				Role:   pair.Role,
+				Entity: pair.Entity,
+			}
+
+			// If the old state is missing for this entity, it needs to
```
danawillow: nit: This comment is written as create -> update, but the if statement is update -> create, can it be reworded to match?

google/resource_storage_default_object_acl.go
```diff
@@ -0,0 +1,199 @@
...
+	d.SetId(getDefaultObjectAclId(bucket))
+	return nil
+}
+
+func resourceStorageDefaultObjectAclUpdate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	bucket := d.Get("bucket").(string)
+
+	if d.HasChange("role_entity") {
```
danawillow: Since `role_entity` is currently the only thing that can change, what do you think about:
```
if !d.HasChange("role_entity") {
  return nil
}
```
and then the rest of the logic afterward?

google/resource_storage_default_object_acl_test.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+resource "google_storage_default_object_acl" "acl" {
+	bucket = "${google_storage_bucket.bucket.name}"
+	role_entity = []
+}
+`, bucketName)
+}
+
+func testGoogleStorageDefaultObjectsAclBasic1(bucketName string) string {
```
danawillow: These three tests are exactly the same except for which vars are filled in for the role_entity. How about something like this:
```
func testGoogleStorageDefaultObjectAclBasic(bucketName, roleEntity1, roleEntity2 string) string {
    return fmt.Sprintf(`
// config here
`, bucketName, roleEntity1, roleEntity2)
```

website/docs/r/storage_default_object_acl.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+layout: "google"
+page_title: "Google: google_storage_default_object_acl"
+sidebar_current: "docs-google-storage-default_object-acl"
+description: |-
+  Creates a new default object ACL in Google Cloud Storage.
+---
+
+# google\_storage\_default\_object\_acl
+
+Creates a new default object ACL in Google cloud storage service (GCS). For more information see
```
danawillow: nit: Google Cloud Storage

danawillow:
Looks good! I just pushed a quick commit to fix up the docs so the link shows up in the sidebar. Merging now.

PR #999, opened 2018-01-23
add metadata to gce cluster config in dataproc cluster

### LINKED ISSUE #989: Feature Request: Pass Custom Metadata Key/Values to DataProc Nodes
Hi,

Would be great to have a similar functionality to https://www.terraform.io/docs/providers/google/r/compute_instance.html#metadata (custom key/value metadata) when provisioning Google DataProc nodes (in my use-case I use tags to filter specific nodes by Prometheus).

Thanks,
--Paul

danawillow:
Fixes #989


Bart6114:
Thanks, for this. The `metadata` config param is not mentioned in the docs though.

PR #1005, opened 2018-01-24
Add update support for machine type, min cpu platform, and service accounts

### LINKED ISSUE #728: Non-destructive update for machine-type change in google_compute_instance
Hi, it would be great if changing the machine-type will trigger a workflow where the instance is stopped, resized and started instead of recreating the VM.

This behaviour is what GCP do via console when a resize is requested.

danawillow:
All three of these require the machine to be stopped before updating. I also added a warning in the docs for each of these fields.

Following the example of https://github.com/hashicorp/terraform/pull/11998 of just going ahead and doing the stopping.

Fixes #728.

### DIFF
--- a/google/field_helpers.go
+++ b/google/field_helpers.go
@@ -54,6 +54,10 @@ func ParseAcceleratorFieldValue(accelerator string, d TerraformResourceData, con
 	return parseZonalFieldValue("acceleratorTypes", accelerator, "project", "zone", d, config, false)
 }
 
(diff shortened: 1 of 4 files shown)


rosbo:
Do we want to add an additional protection like I suggested in the issue?

The main problem I see is that we don't have a way to notify our users at plan time that this update will stop their instance (effectively dropping traffic). 

The behavior before was ForceNew. If people used a create_before_destroy, the new instances would be up first and could start serving traffic without downtime.

danawillow:
Done.

PR #1011, opened 2018-01-26
Fix readme to require Go version 1.9

jphalip:
With go 1.8, `make build` returns the following error:

```
# github.com/terraform-providers/terraform-provider-google/vendor/github.com/hashicorp/terraform/config
vendor/github.com/hashicorp/terraform/config/testing.go:9: t.Helper undefined (type *testing.T has no field or method Helper)
```


rosbo:
Thanks again :)

Issue #635, opened 2017-10-26
add ssl certificate outputs resources for cloud sql postgres


andrewchambers:
A feature request for:

1. Outputting the server root certificate as an output of google_sql_database_instance so ssl encryption can be automated.
2. A new set of resources for ssl client certificates so ssl authentication can be automated.

johnsonj:
It'd be great start to export the `serverCaCert` which is exposed as part of the [instance resource](https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/instances#resource).

Lawouach:
Hey folks,

Just wanted to confirm, we still can't automate client certs creation, right?

Thanks,

paddycarver:
I believe we are still missing the necessary resource for that. I do not believe anyone is working on that at the moment. PRs would be accepted, or we'll get to it as soon as we can. Unfortunately, there's a lot going on in the provider right now, and we have only a handful of maintainers, so sometimes we can take longer to get to features than we'd like.

silvpol:
I've been trying to automate handover from google_cloud_sql to postgres provider and came up with the solution below, I hope this workaround helps somebody until proper solution is in place.

1. Before running Teraform shell needs to have env vars set so that Postgres provider can find our certs/keys.
````Shell
# Adjust file names to match Terraform files
PGSSLCERT=terraform-client-cert.pem
PGSSLKEY=terraform-client-key.pe
PGSSLROOTCERT=terraform-ca-cert.pem
````
Shell script is required due to https://github.com/terraform-providers/terraform-provider-postgresql/issues/49

2. Run terraform as usual
````hcl
# Create Terraform user
resource "google_sql_user" "terraform" {
  name = "terraform"
  project = "${module.project.project_id}"
  instance = "${google_sql_database_instance.cluster_db.id}"
  password = "${random_string.tf_tmp.result}"
}

# Specify location of cert/key files
data "null_data_source" "tf_sql_ssl_files" {
  inputs {
    client_key = "${path.root}/${google_sql_user.terraform.name}-client-key.pem"
    client_key_sha = "${path.root}/${google_sql_user.terraform.name}-client-key.sha1"
    client_cert = "${path.root}/${google_sql_user.terraform.name}-client-cert.pem"
    ca_cert = "${path.root}/${google_sql_user.terraform.name}-ca-cert.pem"
  }
}

# Save cert/key files
resource "null_resource" "db_certs" {

  provisioner "local-exec" {
[...]

PR #1020, opened 2018-01-29
Add support for Cloud SQL CA Cert (#635)

johnsonj:
Expose the 'ServerCaCert' object from the Cloud SQL Instance[1]

Related: #635 - Solves first point

[1] https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/instances#resource

### DIFF
--- a/google/resource_sql_database_instance.go
+++ b/google/resource_sql_database_instance.go
@@ -344,7 +344,36 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 					},
 				},
 			},
-
+			"server_ca_cert": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Optional: true,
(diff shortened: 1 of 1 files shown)


google/resource_sql_database_instance.go
```diff
@@ -344,7 +344,36 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 					},
 				},
 			},
-
+			"server_ca_cert": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Optional: true,
```
danawillow: (hello friends I know this is already merged but-)
since this is only set on read (and not something passed in to create or update), it should be only computed and not also optional.

johnsonj: thanks! follow up inbound

PR #1024, opened 2018-01-30
Diff suppress unconventional public image family naming pattern

rosbo:
Help for #988.

There is an acceptance test that will catch new image patterns and fail if we don't support them.

I would highly encourage people to use the `google_compute_image` data source or an image self_link.

I wish we could make a network call in DiffSuppressFunc...

I am starting to wonder if we should just stop reading the image field for the api for disk and instance (import won't work properly and we won't detect image changes outside Terraform). Reading from the image from the API was added in #948.

### DIFF
--- a/google/resource_compute_disk.go
+++ b/google/resource_compute_disk.go
@@ -498,7 +498,7 @@ func diskImageProjectNameEquals(project1, project2 string) bool {
 }
 
 func diskImageEquals(oldImageName, newImageName string) bool {
-	return strings.Contains(oldImageName, newImageName)
+	return oldImageName == newImageName
 }
 
 func diskImageFamilyEquals(imageName, familyName string) bool {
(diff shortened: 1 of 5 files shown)


google/resource_compute_disk.go
```diff
@@ -495,3 +496,91 @@ func diskImageProjectNameEquals(project1, project2 string) bool {
 
 	return project1 == project2
 }
+
+func diskImageEquals(oldImageName, newImageName string) bool {
+	return strings.Contains(oldImageName, newImageName)
```
danawillow: What's the case where it's the image name that's a substring of the other one (as opposed to the family name that's a substring of the image name)?

rosbo: ?

danawillow: Like, could this be `oldImageName == newImageName`?

rosbo: Ah I see :) Yes! Now that I added a specific one for for testing equality for family.

Done. I kept the `diskImageEquals` method even if it is trivial. I find that it reads better in `diskImageDiffSuppress`.

danawillow: Agreed, thanks :)

PR #1032, opened 2018-01-31
Remove unnecessary Exists methods

rosbo:
Only 2 resources uses Exists.

We call `handleNotFound` in the read method which means that if we get a 404 in the read when doing a GET call, we assume the resource do not exists and remove it from the state.

Makes things simpler for MM.


danawillow:
I'm ok with this, but going to call in @paddycarver to confirm it's a better idea to go this route than to add Exists to more resources (I never quite understood the point, since Exists isn't actually supposed to [modify state](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource.go#L81) so we _have_ to check existence in read anyway, but maybe I'm missing something)

Also not sure if you noticed, travis failed (needs make fmt)

rosbo:
Discussed this offline.

Since MM ensures that the 404 is handled properly inside the `Read` method, we can save an extra network call by removing `Exists` and we also save on complexity for MM.

Issue #971, opened 2018-01-17
GCS: Notifications [feature request]


amoiseiev:
Implement Notifications for google_storage_buckets:

https://cloud.google.com/storage/docs/json_api/v1/notifications

danawillow:
Fixed in https://github.com/terraform-providers/terraform-provider-google/pull/1033

PR #1033, opened 2018-01-31
Storage Notification Resource 

ishashchuk:
Support for GSC Notifications [issue 971](https://github.com/terraform-providers/terraform-provider-google/issues/971)

API reference [here](https://cloud.google.com/storage/docs/json_api/v1/notifications)

The resource closely mimics the API functionality. if no event_types are passed, ALL events will be publishing to pubsub topic. If event_types are passed, only specific events will be published

There is no update functionality; updating the notification will always result in new notification config


google/resource_storage_notification.go
```diff
@@ -0,0 +1,157 @@
...
+
+	if err != nil {
+		return fmt.Errorf("Error deleting notification configuration %s for bucket %s: %v", notificationID, bucket, err)
+	}
+
+	return nil
+}
+
+func resourceStorageNotificationImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	parts := strings.Split(d.Id(), "/")
```
emilymye: It might be nicer if we just have this in resourceStorageNotificationParseID, and have resourceStorageNotificationParseID return both bucket and notificationConfig id.

website/docs/r/storage_notification.html.markdown
```diff
@@ -0,0 +1,90 @@
...
+
+* `bucket` - (Required) The name of the bucket.
+
+* `payload_format` - (Required) The desired content of the Payload. One of `"JSON_API_V1"` or `"NONE"`.
+
+* `topic` - (Requred) The Cloud PubSub topic to which this subscription publishes.
+
+- - -
+
+* `custom_attrubutes` - (Optional)  A set of key/value attribute pairs to attach to each Cloud PubSub message published for this notification subscription
```
emilymye: s/custom_attrubutes/custom_attributes

google/resource_storage_notification.go
```diff
@@ -0,0 +1,162 @@
...
+
+	d.SetId(fmt.Sprintf("%s/notificationConfigs/%s", bucket, res.Id))
+
+	return resourceStorageNotificationRead(d, meta)
+}
+
+func resourceStorageNotificationRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	bucket := d.Get("bucket").(string)
```
danawillow: Since the bucket is already part of the id, you could avoid this d.Get here by using the return value from the parse function, which would then allow you to use an ImportStatePassthrough

google/resource_storage_notification.go
```diff
@@ -0,0 +1,162 @@
...
+	_, notificationID := resourceStorageNotificationParseID(d.Id())
+
+	res, err := config.clientStorage.Notifications.Get(bucket, notificationID).Do()
+	if err != nil {
+		return handleNotFoundError(err, d, fmt.Sprintf("Notification configuration %s for bucket %s", notificationID, bucket))
+	}
+
+	topic, _ := getRelativePath(res.Topic)
```
danawillow: Since we have the DiffSuppressFunc for self link or resource name, we should be fine just storing this the way it's returned from the API (either way is really fine, but since that's the case may as well go with the simpler solution)

google/resource_storage_notification_test.go
```diff
@@ -0,0 +1,250 @@
...
+		if val != value {
+			return fmt.Errorf("Custom attribute value did not match for key %s: expected %s but found %s", key, value, val)
+		}
+		return nil
+	}
+}
+
+func testGoogleStorageNotificationBasic(bucketName, topicName, topic string) string {
+	return fmt.Sprintf(`
+		resource "google_storage_bucket" "bucket" {
```
danawillow: nit: we try to have these all the way on the left (mostly a personal preference thing, it just makes it a bit easier to copy/paste into our own configs later, feel free to ignore)

google/resource_storage_notification_test.go
```diff
@@ -0,0 +1,250 @@
...
+		
+		resource "google_pubsub_topic" "topic" {
+			name = "%s"
+		}
+		// We have to provide GCS default storage account with the permission
+		// to publish to a Cloud Pub/Sub topic from this project
+		// Otherwise notification configuration won't work
+		resource "google_pubsub_topic_iam_binding" "binding" {
+			topic = "${google_pubsub_topic.topic.name}"
+			role        = "roles/pubsub.publisher"
```
danawillow: nit: align `=`s (here and elsewhere)

google/resource_storage_notification_test.go
```diff
@@ -0,0 +1,250 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"google.golang.org/api/storage/v1"
+)
+
+var (
+	gscServiceAccount = fmt.Sprintf("serviceAccount:%s@gs-project-accounts.iam.gserviceaccount.com", os.Getenv("GOOGLE_PROJECT"))
```
danawillow: should this be gcs? or does gsc stand for something?

google/resource_storage_notification_test.go
```diff
@@ -0,0 +1,250 @@
...
+	config := testAccProvider.Meta().(*Config)
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "google_storage_notification" {
+			continue
+		}
+
+		bucket := rs.Primary.Attributes["bucket"]
+
+		_, notificationID := resourceStorageNotificationParseID(rs.Primary.ID)
```
danawillow: Since the bucket is in the id also, you could just do `bucket, notificationID` here

google/resource_storage_notification_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+		*notification = *found
+
+		return nil
+	}
+}
+
+func testAccCheckStorageNotificationCheckEventType(notification *storage.Notification, eventTypes []string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+
```
danawillow: nit: extra newline

website/docs/r/storage_notification.html.markdown
```diff
@@ -0,0 +1,90 @@
...
+```hcl
+resource "google_storage_bucket" "bucket" {
+	name = "default_bucket"
+}
+		
+resource "google_pubsub_topic" "topic" {
+	name = "default_topic"
+}
+
+//In order to enable notifications,
```
danawillow: nit: Can you add a space after each `//`?

website/docs/r/storage_notification.html.markdown
```diff
@@ -0,0 +1,90 @@
...
+* `self_link` - The URI of the created resource.
+
+## Import
+
+Storage notifications can be imported using the notification `id` in the format `<bucket_name>/notificationConfigs/<id>` e.g.
+
+```
+$ terraform import google_storage_notification.notification default_bucket/notificationConfigs/102
+```
+
```
danawillow: nit: there are a few extra lines here that can be deleted

website/docs/r/storage_notification.html.markdown
```diff
@@ -0,0 +1,90 @@
...
+## Attributes Reference
+
+In addition to the arguments listed above, the following computed attributes are
+exported:
+
+* `self_link` - The URI of the created resource.
+
+## Import
+
+Storage notifications can be imported using the notification `id` in the format `<bucket_name>/notificationConfigs/<id>` e.g.
```
danawillow: I wonder whether it makes more sense to just have it be `<bucket_name>/<id>` that way people don't have to bother typing out "/notificationConfigs/" (unless there's somewhere where that string tends to be copy/pasted from). What do you think?

Also, do you think most people will generally know the id of the notification? If there can only be one notification per topic per bucket then you could use the topic name as something more easily identifiable. If not though then it probably has to be the id.

ishashchuk: I was debating on the way to  id them too.. The  reason I decided to go with the <bucket_name>/notificationConfigs/<id> is because that's how they are being returned in  CLI util:
```
gsutil  notifications list gs://test_bucket/
projects/_/buckets/test_bucket/notificationConfigs/89
        Cloud Pub/Sub topic: projects/sandbox/topics/test_topic
```
It seems to me that this would be one of the quickest and user friendly ways to get the list of notifications on the bucket, in case they need to be imported and the id can be just copy/pasted. What do you think?

google/resource_storage_notification.go
```diff
@@ -0,0 +1,157 @@
...
+
+	err := config.clientStorage.Notifications.Delete(bucket, notificationID).Do()
+	if err != nil {
+		return fmt.Errorf("Error deleting notification configuration %s for bucket %s: %v", notificationID, bucket, err)
+	}
+
+	return nil
+}
+
+func resourceStorageNotificationImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
danawillow: This function can be removed now (you can just set the importer to schema.ImportStatePassthrough, see other resources for examples), and then you can do the `d.Set` call inside the read fn

website/docs/r/storage_notification.html.markdown
danawillow: Yup, that's a great reason. Thanks for the explanation!

emilymye:
I'm going to approve, but I'd wait for Dana's approval before actually merging.

PR #1035, opened 2018-02-01
Deprecate name_prefix

danawillow:
Since you can generate prefixes with the random provider, there's no need for `name_prefix` anywhere anymore and it only makes things more complicated.

Related: #780.

### DIFF
--- a/google/resource_compute_instance_template.go
+++ b/google/resource_compute_instance_template.go
@@ -51,6 +51,7 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 					}
 					return
 				},
+				Deprecated: "Use the random provider instead",
 			},
 
 			"disk": &schema.Schema{
--- a/google/resource_compute_instance_template.go
+++ b/google/resource_compute_instance_template.go
@@ -40,6 +40,7 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 			"name_prefix": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 				ForceNew: true,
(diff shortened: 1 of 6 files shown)


rosbo:
What's the migration story for our users?

If they change their config to start using the random provider, it will create a diff on the name... For existing state/config, they would need to update the state so that the random provider value matches the one generated by the name prefix field. Is there an easier way to deal with this?

danawillow:
Quick thoughts before I head out- `name` is `Computed` for all of these resources. If we make `name_prefix` computed as well, then a user who's already using `name_prefix` can safely remove it from their config. Then they don't have to start actually using the random provider until they wish to recreate the resource.

sl1pm4t:
Hi @danawillow  + @rosbo.
There is one situation where having a `name_prefix` attribute directly on the resource is very useful.
With `google_container_node_pool` resource, I rely on the `name_prefix` + `create_before_destroy` flag to ensure we can replace a node_pool without an outage. I've done this to update the machine_type or auth scopes of a node pool for example.
If the `name_prefix` attribute was removed, I would have to manually taint the `random_X` resource to get the same behaviour, which will be confusing for our users.

I know there has been a lot of focus on the `google_container_cluster` and `google_container_node_pool` resources lately, which I have not stayed up to date with, so maybe this kind of updates of a node_pool is baked into the resources now?! 

Thanks!

danawillow:
@sl1pm4t I have an answer for you! I'm going to put it on #1054 to try to have a centralized place for discussion (no fault on your part, I created the issue after I had already done the PR)

danawillow:
@paddycarver I'm thinking these changes should probably get in before 2.0.0 so we can remove the feature entirely in 2.0.0, right?

Issue #400, opened 2017-09-08
Support for service account keys


drzero42:
Currently we can create service accounts with Terraform, but we cannot create keys for service accounts. Creating a service account key returns the generated private key and some other information, that should be kept.

My use case for service keys are that I create a service account for the services that I run in container engine to enable usage of various APIs (Cloud Storage, Pub/Sub and Cloud SQL Proxy amongst others). I basically upload the generated JSON file (written by gcloud when using that, or downloaded from Cloud Console when using the browser) as a Kubernetes secret, which can then be consumed by my pods. Being able to create a service account key in Terraform and having the same data as is in such a JSON file available as an attribute would enable me to use the Kubernetes provider to create a secret with this content. This would bring me one step closer to being able to deploy my entire GCP stack with Terraform.

sebglon:
PR#204

danawillow:
Thanks @sebglon, closing this one as a duplicate.

jasonjho:
@danawillow @sebglon I see that you're now able to generate service account keys, but is there a way to generate the furnished JSON key file? As @drzero42 mentioned, these files are typically consumed by other resources, whether referenced on disk or uploaded as a Kubernetes secrets.

Any pointers here would be greatly appreciated. Thanks!

danawillow:
Isn't that the [`private_key`](https://www.terraform.io/docs/providers/google/r/google_service_account_key.html#private_key) field? Or are you looking for something else?

drzero42:
@jasonjho It is easily possible, it is just unintuitively named, and badly documented (IMHO). The attribute named `private_key` on the `google_service_account_key` resource, contains a base64 encoded version of the json-file normally downloaded when you create a service account key through the console or gcloud cli. Here is a simply example of how to use it:

```
resource "google_service_account" "myaccount" {
  account_id   = "myaccount"
  display_name = "My Service Account"
}
resource "google_service_account_key" "mykey" {
  service_account_id = "${google_service_account.myaccount.id}"
}
resource "local_file" "myaccountjson" {
    content     = "${base64decode(google_service_account_key.mykey.private_key)}"
    filename = "${path.module}/myaccount.json"
}
```
Same deal for setting a kubernetes secret with the json-content :)

danawillow:
@drzero42 was that an offer to improve the documentation? ;)

jasonjho:
[...]

PR #1047, opened 2018-02-04
Add example usage and better description of private_key attribute of …

drzero42:
…google_service_account_key.

This should fix my complaints about the documentation not being clear in #400 ;)
I've added a specific example of how to put a Service Account credentials file into a Kubernetes secret, which seems to be what a lot of people want to do, and I've added a sentence or two to the description of the private_key attribute, to make it clear that the content of this attribute is what you normally download when creating keys through the web console.


rosbo:
Thanks for adding the extra example and clarifying the private_key's description.

Issue #912, opened 2018-01-03
SQL Database Instance does not have attribute `ip_address.0.ip_address`


genevieve:
### Terraform Version
Terraform v0.11.1
+ provider.google v0.1.3
+ provider.random v1.0.0
+ provider.tls v1.0.0

### Affected Resource(s)
- google_sql_database_instance

### Terraform Configuration Files
```hcl
resource "google_sql_database_instance" "master" {
  region           = "${var.region}"
  database_version = "MYSQL_5_6"
  name             = "${var.env_name}-${replace(lower(random_id.db-name.b64), "_", "-")}"

  settings {
    tier = "${var.sql_db_tier}"

    ip_configuration = {
      ipv4_enabled = true

      authorized_networks = [
        {
          name  = "all"
          value = "0.0.0.0/0"
        },
      ]
    }
  }

  count = "${var.count}"
}

output "sql_db_ip" {
  value = "${google_sql_database_instance.master.*.ip_address.0.ip_address}"
}

```

### Debug Output
```
* module.external_database.output.sql_db_ip: Resource 'google_sql_database_instance.master' does not have attribute 'ip_address.0.ip_address' for variable 'google_sql_database_instance.master.*.ip_address.0.ip_address'
```

### Expected & Actual Behavior
If I were to change the output value to:

```
output "sql_db_ip" {
  value = "${google_sql_database_instance.master.*.ip_address}"
}
```

I get:

```
sql_db_ip = [
    [
        map[ip_address:35.184.66.76 time_to_retire:]
    ]
]
```

I expect that in the former version of the output syntax (as is specified in the docs), it would print `35.184.66.76` and not an error that the attribute is not available. I additionally expect the output not to cause an error when the instance was not created.

It seems that trying to use the splat syntax and asking for nested attributes (conditionally) is not possible in terraform. If there is a single ip address assigned, why does the attribute have to be gathered by indexing with `ip_address.0.ip_address`? Could the instance have a single, non-nested attribute for returning the address?

### Other Attempts
- Using the `element` syntax is not possible as `master.*.ip_address` will return a nested map and not a flat list. We can flatten this list with `flatten(google_sql_database_instance.master.*.ip_address)` but this is a list of maps and so we cannot select the first map with `element()`. We can't `element(concat(flatten(...), list(SOMETHING)))` because you need to concat the flatted list with another list of maps. `list()` only takes strings.

[...]

PR #1050, opened 2018-02-05
SQL DB Instance has attribute first_ip_address

genevieve:
Related Issues:

https://github.com/terraform-providers/terraform-provider-google/issues/912
https://github.com/hashicorp/terraform/issues/17048

### DIFF
--- a/google/resource_sql_database_instance.go
+++ b/google/resource_sql_database_instance.go
@@ -729,11 +729,17 @@ func resourceSqlDatabaseInstanceRead(d *schema.ResourceData, meta interface{}) e
 	if err := d.Set("settings", flattenSettings(instance.Settings)); err != nil {
 		log.Printf("[WARN] Failed to set SQL Database Instance Settings")
 	}
+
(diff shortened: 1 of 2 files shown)


danawillow:
I see in the other issue that a workaround of using `TF_WARN_OUTPUT_ERRORS=1` until the new syntax comes out was suggested- did that work?

genevieve:
Since the terraform doesn't apply without that environment variable, it means resources that all our teams use in CI/CD pipelines would have to support that, as well as asking it of all ours that go directly to use these terraform templates. It wasn't really something we wanted to recommend since it risks actually swallowing relevant errors. 

When we choose one output syntax or the other either the database-turned-off case works and the database-turned-on fails during `terraform apply`, or choosing a different syntax causes the opposite.

danawillow:
Thanks @genevieve! That sounds reasonable, and the code looks good. Merging now.

danawillow:
Oh wait, I got ahead of myself. I'm going to keep it in, but would you mind sending a follow-up PR for changes to the docs to add this attribute? Please tag me in it and I'll review it right away.

PR #1062, opened 2018-02-08
Return an error if name and name prefix are specified in node pool

rosbo:
Fixes test `TestAccContainerCluster_withNodePoolConflictingNameFields `.

### DIFF
--- a/google/resource_container_node_pool.go
+++ b/google/resource_container_node_pool.go
@@ -307,6 +307,9 @@ func resourceContainerNodePoolStateImporter(d *schema.ResourceData, meta interfa
 func expandNodePool(d *schema.ResourceData, prefix string) (*container.NodePool, error) {
 	var name string
 	if v, ok := d.GetOk(prefix + "name"); ok {
+		if _, ok := d.GetOk(prefix + "name_prefix"); ok {
+			return nil, fmt.Errorf("Cannot specify both name and name_prefix for a node_pool")
+		}
 		name = v.(string)
 	} else if v, ok := d.GetOk(prefix + "name_prefix"); ok {
 		name = resource.PrefixedUniqueId(v.(string))


google/resource_container_node_pool.go
```diff
@@ -307,6 +307,9 @@ func resourceContainerNodePoolStateImporter(d *schema.ResourceData, meta interfa
 func expandNodePool(d *schema.ResourceData, prefix string) (*container.NodePool, error) {
 	var name string
 	if v, ok := d.GetOk(prefix + "name"); ok {
+		if _, ok := d.GetOk(prefix + "name_prefix"); ok {
+			return nil, fmt.Errorf("Cannot specify both name and name_prefix for a node_pool")
+		}
```
paddycarver: Couldn't we add this as `ConflictsWith` on `name` and `name_prefix`?

rosbo: `ConflictsWith` doesn't work with nested fields.

rosbo: Just to clarify, ConflictsWith works on nested field only **IFF** you now the index of the other field to compare with. When you have a list of nested objects, it doesn't work.

For instance, this works:

```
foo = "aa"
bar  {
  baz = "aa"
}

// Definition
foo: {
  ConflictsWith: []string{"bar.0.baz"}
}
```

But if you want to say, two fields under the same entry should be conflicting, then this is not possible:

```
{
 foo = "aaa"
 bar {
  name = "a"
  name_prefix = "b"
 }
 bar {
   name = "c"
   name_prefix = "d"
 }
}
```

For the example above, there is now way to express that name and name_prefix should ConflictsWith each other.

rosbo: That would be nice if ConflictsWith supported a notation (exact notation TBD) like:
```
ConflictsWith: []string{"bar.{i}.baz"}
```

rosbo: Two issues already tracking this:
https://github.com/hashicorp/terraform/issues/13016
https://github.com/hashicorp/terraform/issues/11101

paddycarver: Whoooops. Sorry about that. Ignore this comment then. :)

PR #1064, opened 2018-02-08
Add generated BackendBucket resource + transport.go

danawillow:
```
$ make testacc TEST=./google TESTARGS='-run=TestAccComputeBackendBucket'
TF_ACC=1 go test ./google -v -run=TestAccComputeBackendBucket -timeout 120m
=== RUN   TestAccComputeBackendBucket_basic
=== RUN   TestAccComputeBackendBucket_basicModified
=== RUN   TestAccComputeBackendBucket_withCdnEnabled
--- PASS: TestAccComputeBackendBucket_withCdnEnabled (27.99s)
--- PASS: TestAccComputeBackendBucket_basic (28.40s)
--- PASS: TestAccComputeBackendBucket_basicModified (43.34s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	43.541s
```


google/resource_compute_backend_bucket.go
```diff
@@ -16,40 +29,37 @@ func resourceComputeBackendBucket() *schema.Resource {
...
-			"bucket_name": &schema.Schema{
+			"bucket_name": {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-
-			"description": &schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
```
rosbo: Shouldn't this be ForceNew?

danawillow: Ah yeah, that was broken by the same change that broke import- it'll be fixed once my update changes are in so I don't feel bad fixing it manually for now.

google/resource_compute_backend_bucket.go
```diff
@@ -16,40 +29,38 @@ func resourceComputeBackendBucket() *schema.Resource {
...
+			"description": {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
-
-			"enable_cdn": &schema.Schema{
+			"enable_cdn": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
```
paddycarver: Should we be worried about this Default getting removed?

google/transport.go
```diff
@@ -0,0 +1,176 @@
+package google
```
paddycarver: Is this file auto-generated? If so, should it have a header?

danawillow: Nope, this one is hand-written.

google/resource_compute_backend_bucket.go
danawillow: It's `Default: false` which as far as I know acts exactly the same as if there was no default at all, since `d.GetOk("enable_cdn")` would return (false, false) in both cases.

rosbo: What about the diff, would you get a pertual diff like the one below if you don't specify it?
```
"false" -> ""
  ^               ^
API            config

I am not sure what is the behavior for boolean.

paddycarver:
I missed transport.go was still using the transport from client.go, oops. My bad. :(

I'd love to have the old Print statements, just because doing a print before/after each logical request/response is helpful sometimes when retries happen. Also, in the case of the same request being made for multiple purposes (think create v. update) it's helpful to be explicit about _why_ the request is being made, not just what request is being made.

google/resource_compute_backend_bucket.go
danawillow: Good question! I just checked and it doesn't perpetual diff (and `terraform show` does say "false" for it)

danawillow:
Ok logging is back :)

PR #1065, opened 2018-02-08
Fix TestAccContainerCluster_withIPAllocationPolicy test

rosbo:
Step 2 is failing because an empty `ip_allocation_policy` block doesn't fail.
It fails to cast as a `map[string]interface{}` but this does not return an error. The if throwing the error could not be reached.


rosbo:
I fixed the code instead to throw the error at the right location.

```sh
--- PASS: TestAccContainerCluster_withIPAllocationPolicy (634.96s)
```

PR #1071, opened 2018-02-11
Add VPN gateway data source

jphalip:


google/data_source_google_compute_vpn_gateway.go
```diff
@@ -0,0 +1,84 @@
...
+	return &schema.Resource{
+		Read: dataSourceGoogleComputeVpnGatewayRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"description": &schema.Schema{
```
ndmckinley: We order these fields "Required, Optional, Computed" - A field which is both Optional and Computed counts as Optional for this ordering.

google/data_source_google_compute_vpn_gateway.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+
+	name := d.Get("name").(string)
+
+	vpnGatewaysService := compute.NewTargetVpnGatewaysService(config.clientCompute)
+
+	gateway, err := vpnGatewaysService.Get(project, region, name).Do()
+	if err != nil {
+		if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 404 {
```
ndmckinley: I think you want handleNotFoundError.

jphalip:
Thanks @ndmckinley, I've pushed the requested updates.

google/data_source_google_compute_vpn_gateway.go
ndmckinley: This actually includes the 404 check - it returns the not-found error if the resource doesn't exist, and a regular error otherwise.  Just `if err != nil { return handleNotFoundError(...) }`

jphalip: I've just pushed another change. Does that work?
Also, I had modeled the original error handling code on existing examples such as `data_source_compute_network`, `data_source_compute_subnetwork` or `data_source_compute_global_address`. There might be other such examples. Are those incorrect as well?

google/data_source_google_compute_vpn_gateway.go
```diff
@@ -67,12 +67,8 @@ func dataSourceGoogleComputeVpnGatewayRead(d *schema.ResourceData, meta interfac
 
 	gateway, err := vpnGatewaysService.Get(project, region, name).Do()
 	if err != nil {
-		if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 404 {
-			// The resource doesn't exist anymore
-			return handleNotFoundError(err, d, fmt.Sprintf("VPN Gateway Not Found : %s", name))
-		}
-
-		return fmt.Errorf("Error reading VPN Gateway: %s", err)
+		// The resource doesn't exist anymore
```
ndmckinley: Comment out of date now - sorry for the nitpicks!

ndmckinley:
Thanks for working with me on that!  Running the tests now, and will merge when they're done.

jphalip:
No worries! By the way, did you see my comment regarding the code for the network, subnetwork and global-address data sources? I had modeled my original error handling code based on those. Are they incorrect?

ndmckinley:
Thanks for calling that out!  You're right, those did not move to the new-model 404 handler - we should fix it.  :)

jphalip:
Ok I see, I'm happy to submit a separate PR for that. I'll ping you when it's done.

ndmckinley:
Oh wow, that's great!  Thanks!  :)

PR #1073, opened 2018-02-12
Fix overly restrictive validation for ServiceAccountLinkRegex

KJTsanaktsidis:
A project name of the form `foo.io:foo` is perfectly legal, and a
service account for this project has the form

`projects/foo.io:foo/serviceAccounts/acctname@foo.foo.io.iam.gserviceaccount.com`

The two . characters in foo.foo.io fails the ProjectRegex validation,
making it impossible to create a google_service_account_key resource for
a service account in such a project.

This commit uses a looser validation for the project name the second
time it is used in the ServiceLinkAccountRegex.

Without this patch, when attempting to create a `google_service_account_key` resource with a project of the form `foo.bar:foo`, the following error is obtained:

```
* google_service_account_key.gke_puller_key: "service_account_id" ("projects/foo.io:foo/serviceAccounts/fo-gke-puller@foo.foo.io.iam.gserviceaccount.com") doesn't match regexp "projects/(?:(?:[-a-z0-9]{1,63}\\.)*(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?):)?(?:[0-9]{1,19}|(?:[a-z0-9](?:[-a-z0-9]{0,61}[a-z0-9])?))/serviceAccounts/[a-z](?:[-a-z0-9]{4,28}[a-z0-9])@(?:(?:[-a-z0-9]{1,63}\\.)*(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?):)?(?:[0-9]{1,19}|(?:[a-z0-9](?:[-a-z0-9]{0,61}[a-z0-9])?))\\.iam\\.gserviceaccount\\.com$"

```

With the patch applied here, the service account key is successfully created.


ndmckinley:
Looks good - can you add a quick test with the case you've described there in validation_test.go?

KJTsanaktsidis:
@ndmckinley I've added some test cases for `ServiceAccountLinkRegex`.

Issue #1079, opened 2018-02-13
metadata startup-script for google_compute_instance broken after 1.4.0


mytran:
In my google_compute_instance resource, I have a metadata block with startup script defined instead of the **metadata_startup_script** argument because I do not want the VM to get recreated when it changes.
ie
```
  metadata = {
    startup-script = "MYSCRIPT"
  }
```

This works fine for me on 1.4.0 but both 1.5.0 and 1.60, it keeps on thinking that I have **metadata_startup_script** set when I run terraform plan and tries to destroy it.
```
      metadata.startup-script:  "" => "MYSCRIPT"
      metadata_startup_script:  "MYSCRIPT" => "" (forces new resource)
```

Looking through the source, I believe this line added for import support broke this:
https://github.com/terraform-providers/terraform-provider-google/blob/e1323489baa0e70e135fbecf18675e8bd89ccfc1/google/resource_compute_instance.go#L785

ndmckinley:
I agree that that's broken and that you've found the reason - we'll need a diff-suppress for diffs specifically adding something to metadata.startup-script and removing that exact thing from metadata_startup_script.  I'll look into it.

ndmckinley:
Ah, bummer, I've misunderstood diff suppress.  :-1: I'll figure out another solution.

ndmckinley:
That PR will fix it going forward, but for your existing instances you'll need to either do state munging to get the state to reflect the config, or a full `terraform destroy` / `terraform apply`.  I can't figure a way around that, unfortunately.

gwoo:
`terraform state rm ADDRESS...` worked for me.

PR #1081, opened 2018-02-13
Don't set metadata_startup_script in some cases.

ndmckinley:
If the state contains `metadata.startup-script`, the user likely wants to keep it in there instead of `metadata_startup_script` (which is `ForceNew`, where `metadata.startup-script` is not).  This isn't the standard case - `metadata.startup-script` creates heterogeneous fleets unless managed very carefully, which is why `metadata_startup_script` exists - but it's supported by the API and it's a valid state, so we want it to continue working.

After this PR:
* If you set `metadata_startup_script` and don't set `metadata.startup-script`, you will see no change.
* If you set `metadata.startup-script` and don't set `metadata_startup_script`, you will stop seeing permanent diffs, because `Read` will stop setting `metadata_startup_script`.
* If you set both, your behavior will be strange and you'll probably see permanent diffs - don't set both.
* If you do an import, you'll get `metadata_startup_script`.

If you're affected by something like #1079, this will make sure that you don't have problems going for anything created after this PR goes in, but for your existing instances you'll need to make sure that the `metadata.startup-script` has a non-empty value, probably using the `terraform state` functions.

### DIFF
--- a/google/metadata.go
+++ b/google/metadata.go
@@ -1,6 +1,7 @@
 package google
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


rosbo:
> If you set both, your behavior will be strange and you'll probably see permanent diffs - don't set both.

Should we add a ConflictsWith to ensure people don't set both?

ndmckinley:
Can't, unfortunately - `metadata` is a map.  Probably need to put it in the validator - I'll take a look at that.  If it's easy I'll add it to this PR.

ndmckinley:
Shoot - I've got a problem with updates from `metadata.startup-script` -> `metadata_startup_script`.  I'll try and fix it...

ndmckinley:
Got it.  Running the tests.

ndmckinley:
All the tests passed, merging.

PR #1085, opened 2018-02-14
Add scaffolding to enable beta features in GKE resources

danawillow:
* Replaced ComputeApiVersion with a more generic ApiVersion
* Added allowed versions + ordering for GKE

### DIFF
--- a/google/api_versions.go
+++ b/google/api_versions.go
@@ -6,14 +6,14 @@ import (
 	"strings"
 )
 
-type ComputeApiVersion uint8
+type ApiVersion uint8
 
 const (
-	v1 ComputeApiVersion = iota
+	v1 ApiVersion = iota
 	v0beta
 )
 
-var OrderedComputeApiVersions = []ComputeApiVersion{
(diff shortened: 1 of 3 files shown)


google/api_versions.go
```diff
@@ -42,22 +48,30 @@ type TerraformResourceData interface {
...
 		if feature.InUseByDefault(d) {
 			versions[feature.Version] = struct{}{}
 		}
 	}
 
-	return maxVersion(versions)
+	return maxVersionFunc(versions)
+}
+
+func getComputeApiVersion(d TerraformResourceData, resourceVersion ApiVersion, features []Feature) ApiVersion {
```
rosbo: Why using two separate methods for compute and container? Do you envision any per product family customization happening here?

danawillow: It's more just to hide implementation details. The caller doesn't really need to know that the way we figure out the version to use is via a function that finds the max version used.

rosbo: I missed the `maxComputeVersion` and `maxContainerVersion`. I thought it was only a pass-through. All good then :)

PR #1086, opened 2018-02-15
Fix VPN gateway data source schema for "network" attribute.

jphalip:


ndmckinley:
Thanks, that's correct - merging.

PR #1087, opened 2018-02-15
Changes to handle autogeneration.

rosbo:
* Extend ResourceDataMock and TerraformResourceData to also support `Set(...)`, `Id()` and `SetId()` methods.
* When calling `getRegion` or `getZone`, we should only return the region or zone name when the state value contains a self_link (i.e. API returns a self_link for subnetwork region field)


rosbo:
Friendly ping :)

Issue #549, opened 2017-10-06
Add support for providing zones in google_compute_region_instance_group_manager


zbikmarc:
Google released (currently in beta) capability to select to which zones regional instance group manager should deploy to. It would be awesome to has similar feature in `google_compute_region_instance_group_manager`.

### Terraform Version
- `terraform-0.10.6`
- `terraform-provider-google-1.0.1`

### Affected Resource(s)
- `google_compute_region_instance_group_manager`

### Expected Behavior
It should be possible to provide zones (optional) where regional manager can deploy to. I think it should be list of zones.

### Actual Behavior
It does not support it.

### References
- https://cloud.google.com/compute/docs/instance-groups/distributing-instances-with-regional-instance-groups#selectingzones

danawillow:
Fixed in #1092

PR #1091, opened 2018-02-15
WIP: Support `distributionPolicy` when creating regional instance group managers.

tobz:
This addresses #549.

This one is sort of a doozy because it's both an un-updatable attribute and beta feature.  The format required for the zones to be specified in was also a little tricky.  Definitely looking for thoughts on how my approach could be better.  I chose what I thought was the best user experience -- users specifying the "naked" zone -- and augmenting it behind-the-scenes.

### DIFF
--- a/google/resource_compute_region_instance_group_manager.go
+++ b/google/resource_compute_region_instance_group_manager.go
@@ -1,6 +1,7 @@
 package google
 
 import (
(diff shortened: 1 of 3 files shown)


tobz:
(just realized this breaks the auto-healing policy acceptance test so not finished yet, woops!)

PR #1092, opened 2018-02-15
Support `distributionPolicy` when creating regional instance group managers.

tobz:
This addresses #549.

This one is sort of a doozy because it's both an un-updatable attribute and beta feature. The format required for the zones to be specified in was also a little tricky. Definitely looking for thoughts on how my approach could be better. I chose what I thought was the best user experience -- users specifying the "naked" zone -- and augmenting it behind-the-scenes.

```
TF_ACC=1 go test ./google -v -run=TestAccRegionInstanceGroupManager -timeout 120m
=== RUN   TestAccRegionInstanceGroupManager_basic
=== RUN   TestAccRegionInstanceGroupManager_targetSizeZero
=== RUN   TestAccRegionInstanceGroupManager_update
=== RUN   TestAccRegionInstanceGroupManager_updateLifecycle
=== RUN   TestAccRegionInstanceGroupManager_separateRegions
=== RUN   TestAccRegionInstanceGroupManager_autoHealingPolicies
=== RUN   TestAccRegionInstanceGroupManager_distributionPolicy
--- PASS: TestAccRegionInstanceGroupManager_targetSizeZero (49.02s)
--- PASS: TestAccRegionInstanceGroupManager_updateLifecycle (218.13s)
--- PASS: TestAccRegionInstanceGroupManager_update (241.82s)
--- PASS: TestAccRegionInstanceGroupManager_distributionPolicy (244.57s)
--- PASS: TestAccRegionInstanceGroupManager_autoHealingPolicies (244.72s)
--- PASS: TestAccRegionInstanceGroupManager_separateRegions (255.48s)
--- PASS: TestAccRegionInstanceGroupManager_basic (306.53s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	306.793s
```

### DIFF
--- a/google/resource_compute_region_instance_group_manager.go
+++ b/google/resource_compute_region_instance_group_manager.go
@@ -151,14 +151,18 @@ func resourceComputeRegionInstanceGroupManager() *schema.Resource {
 			},
 
 			"distribution_policy": &schema.Schema{
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Optional: true,
 				ForceNew: true,
 				Computed: true,
-				Set:      hashZoneFromSelfLinkOrResourceName,
-				Elem: &schema.Schema{
-					Type:             schema.TypeString,
-					DiffSuppressFunc: compareSelfLinkOrResourceName,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"zone": &schema.Schema{
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: compareSelfLinkOrResourceName,
+						},
+					},
 				},
 			},
 		},
@@ -183,7 +187,7 @@ func resourceComputeRegionInstanceGroupManagerCreate(d *schema.ResourceData, met
 		NamedPorts:          getNamedPortsBeta(d.Get("named_port").([]interface{})),
 		TargetPools:         convertStringSet(d.Get("target_pools").(*schema.Set)),
 		AutoHealingPolicies: expandAutoHealingPolicies(d.Get("auto_healing_policies").([]interface{})),
-		DistributionPolicy:  expandDistributionPolicy(d.Get("distribution_policy").(*schema.Set)),
+		DistributionPolicy:  expandDistributionPolicy(d.Get("distribution_policy").([]interface{})),
 		// Force send TargetSize to allow size of 0.
 		ForceSendFields: []string{"TargetSize"},
 	}
@@ -530,16 +534,16 @@ func resourceComputeRegionInstanceGroupManagerDelete(d *schema.ResourceData, met
 	return nil
 }
 
-func expandDistributionPolicy(configured *schema.Set) *computeBeta.DistributionPolicy {
-	if configured.Len() == 0 {
+func expandDistributionPolicy(configured []interface{}) *computeBeta.DistributionPolicy {
+	if len(configured) == 0 {
 		return nil
 	}
 
-	distributionPolicyZoneConfigs := make([]*computeBeta.DistributionPolicyZoneConfiguration, 0, configured.Len())
-	for _, raw := range configured.List() {
-		data := raw.(string)
+	distributionPolicyZoneConfigs := make([]*computeBeta.DistributionPolicyZoneConfiguration, 0, len(configured))
+	for _, raw := range configured {
+		data := raw.(map[string]interface{})
 		distributionPolicyZoneConfig := computeBeta.DistributionPolicyZoneConfiguration{
-			Zone: "zones/" + data,
+			Zone: "zones/" + data["zone"].(string),
 		}
 
 		distributionPolicyZoneConfigs = append(distributionPolicyZoneConfigs, &distributionPolicyZoneConfig)
@@ -547,18 +551,19 @@ func expandDistributionPolicy(configured *schema.Set) *computeBeta.DistributionP
 	return &computeBeta.DistributionPolicy{Zones: distributionPolicyZoneConfigs}
 }
 
-func flattenDistributionPolicy(distributionPolicy *computeBeta.DistributionPolicy) *schema.Set {
-	zones := make([]interface{}, 0)
+func flattenDistributionPolicy(distributionPolicy *computeBeta.DistributionPolicy) []map[string]interface{} {
+	policiesSchema := make([]map[string]interface{}, 0)
 
 	if distributionPolicy != nil {
 		for _, zone := range distributionPolicy.Zones {
-			zones = append(zones, zone.Zone)
+			data := map[string]interface{}{
+				"zone": zone.Zone,
+			}
+			policiesSchema = append(policiesSchema, data)
 		}
 	}
 
-	return schema.NewSet(schema.HashSchema(&schema.Schema{
-		Type: schema.TypeString,
-	}), zones)
+	return policiesSchema
 }
 
 func hashZoneFromSelfLinkOrResourceName(value interface{}) int {
--- a/google/resource_compute_region_instance_group_manager.go
+++ b/google/resource_compute_region_instance_group_manager.go
@@ -1,17 +1,18 @@
 package google
 
 import (
+	"fmt"
+	"log"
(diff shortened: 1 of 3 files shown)


tobz:
@rosbo @danawillow @ndmckinley @paddycarver 

How do PRs get assigned to someone for review?  There doesn't seem be any documentation explaining the process, and no activity at all for two weeks is sadface. :(

danawillow:
Sorry @tobz, between the flu that's been going around, some travel, bugs we've needed to fix quickly, and other project work, this one seems to have slipped through the cracks. I'll take a look at it this afternoon :)

google/resource_compute_region_instance_group_manager.go
```diff
@@ -145,6 +149,18 @@ func resourceComputeRegionInstanceGroupManager() *schema.Resource {
 					},
 				},
 			},
+
+			"distribution_policy": &schema.Schema{
```
danawillow: The thing that I'm not totally sold on for this approach is that with a name like "distribution_policy", it should really be a list of policies rather than a list of zones.

As much as the nesting is weird for just one field, it does make the resource more maintainable in the future (especially once we start [generating](https://github.com/terraform-providers/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#autogenerated-resources) resources). If the API maintainers decide to add another field to that block for example, we'd then have to deprecate this field and make it into a nested block, which is a bit of a pain.

Happy to hear a rebuttal though :)

google/resource_compute_region_instance_group_manager.go
```diff
@@ -227,6 +245,7 @@ func getManager(d *schema.ResourceData, meta interface{}) (*computeBeta.Instance
 
 	if err != nil {
 		handleNotFoundError(err, d, fmt.Sprintf("Region Instance Manager %q", d.Get("name").(string)))
+		return nil, err
```
danawillow: Good catch! This should just be `return handleNotFoundError(...)`

google/resource_compute_region_instance_group_manager.go
```diff
@@ -145,6 +149,18 @@ func resourceComputeRegionInstanceGroupManager() *schema.Resource {
 					},
 				},
 			},
+
+			"distribution_policy": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
```
danawillow: I don't think this needs to be computed, does it?

google/resource_compute_region_instance_group_manager.go
```diff
@@ -509,3 +529,41 @@ func resourceComputeRegionInstanceGroupManagerDelete(d *schema.ResourceData, met
...
+func flattenDistributionPolicy(distributionPolicy *computeBeta.DistributionPolicy) *schema.Set {
+	zones := make([]interface{}, 0)
+
+	if distributionPolicy != nil {
+		for _, zone := range distributionPolicy.Zones {
+			zones = append(zones, zone.Zone)
+		}
+	}
+
+	return schema.NewSet(schema.HashSchema(&schema.Schema{
```
danawillow: I think just `return zones` will work (and then update the return type)

google/resource_compute_region_instance_group_manager.go
tobz: I'm not against such a change.

Like I mentioned in the opening comments, I made the choices I did to present a good UX to the end-user: ultimately, this attribute is simply for saying which zones in a region should be used.

I guess I'd be curious about how easily users could splat out a list of zones into such a list of nested blocks.  I'm not sure I've seen that or done it myself.

google/resource_compute_region_instance_group_manager.go
tobz: I'd have to go back and test with/without it, but I believe it needs to be computed because while the API will let you pass in, say, `us-central1-a`, whenever you retrieve the info for the rIGM, you _will_ get back the full zone URL i.e. `https://googlecloud.com/ur_project/zones/us-central1-a`, and then TF would freak out thinking there needed to be an update.

Let me try to retest that one.  I'm also open to ideas on generating the full one URL behind-the-scenes, but it seemed like a bunch of extra work at the time, and I wasn't really familiar with a way to easily nab the current project, etc, to build it.

danawillow: Nah, that's what the diffsuppress is for. Computed means that if it isn't set, the API will provide a value (so don't show a diff when the config doesn't have a value for that field).

google/resource_compute_region_instance_group_manager.go
tobz: So, the signature doesn't actually support that specific code snippet.

What's curious, though, is that we'd be returning `nil, nil` in the case of a not found error if we used the output of `handleNotFoundError`.  This feels totally wrong, but I'm willing to deal if that's simply the way things shook out in the provider.

google/resource_compute_region_instance_group_manager.go
tobz: So, I went back.  This is actually required.

If we use the beta API version, the API actually fills in the distribution policy for you, based on -- what appears to be -- the zones the rIGM seeds itself with during creation.

google/resource_compute_region_instance_group_manager.go
tobz: Alright, so I ended up putting you suggestion into action.  Now it's a list of objects, matching the structure of the API.

tobz:
@danawillow I think this is ready for another pass.

google/resource_compute_region_instance_group_manager.go
danawillow: Sorry to go back and forth on this- your comment about being able to splat is a good one. I asked around a bit and the suggestion I got that I like is to just have a top-level list of strings called `distribution_policy_zones`. That leaves room to deprecate it later in favor of a straight `distribution_policy` object in case the API adds more things, but makes a good UX for the time being.

google/resource_compute_region_instance_group_manager.go
danawillow: So what we want overall is to make sure that if the IGM isn't found when we try to read it, that we remove it from state. This is a totally valid case and shouldn't cause Terraform to error. So we don't want to return the actual error here, because then a not found will cause our terraform command to fail. Here's how we handle it for `resource_compute_instance`:
https://github.com/terraform-providers/terraform-provider-google/blob/ab385ff5740b7a53ca6302d6c6214c273e178aa0/google/resource_compute_instance.go#L773-L776
https://github.com/terraform-providers/terraform-provider-google/blob/ab385ff5740b7a53ca6302d6c6214c273e178aa0/google/resource_compute_instance.go#L588-L603

tobz: Makes sense.  I'll take care of it.

google/resource_compute_region_instance_group_manager.go
tobz: You're killin' me, smalls! :)

Alright, I'll go with that approach.

tobz:
@danawillow Alright, I think we've arrived. :P

google/resource_compute_region_instance_group_manager.go
tobz: FWIW, I tried this, to no avail.

danawillow: I just uploaded a commit where this seems to work. Let me know what you think!

danawillow:
FYI I just pushed a typo fix and the change in flatten that I was talking about. If that seems all right to you, I'm ready to merge this :)

tobz:
Thanks for the fix-ups!  I now see what you meant by changing the flatten method. :)

Tests pass; LGTM.

tobz:
For my own edification, when are releases usually cut?

danawillow:
We aim for every 2 weeks, but sometimes things come up and it stretches a bit. If all goes according to plan, we'll do a release on Monday, but no promises.

tobz:
Late response: yeah, not in a rush, just curious. :)

Thanks again!

PR #1093, opened 2018-02-15
Move method from resource to datasource file

rosbo:
This way, when we autogenerate the subnetwork resource,  we don't break the subnetwork datasource.

Subnetwork is the only datasource using a flatten method defined in a resource file.


ndmckinley:
This has a `go fmt` issue - sending a PR to fix it.

PR #1097, opened 2018-02-16
Fix incorrect result on Read ignoring most authorized networks.

### LINKED ISSUE #1029: Updating master_authorized_networks replaces entired auth list with changes
Hey, 
I've just come across a problem with `master_authorized_networks` on `google_container_clusters`, where when updating the master authorized networks on a cluster, all existing networks are removed and only the changes (additions) are left in place.

Take this example, I have a single network called Darren:
screen shot 2018-01-31 at 11 48 51

I then want to apply 5 new networks, as you can see, terraform correctly detects the `1 => 6`, because I've added five more

screen shot 2018-01-31 at 11 48 46

However when the operation completes, Darren has been removed, and only the 5 new networks remain:

screen shot 2018-01-31 at 11 49 02

Now if i delete the entire `master_authorized_networks` block from my terraform file (to clear the networks out, then paste it back in and apply again, all 6 networks are correctly added.  This is therefore the only way we can currently update the network list, by first clearing it out then re-adding everything.

ndmckinley:
My belief is that this fixes #1029, because the issue is caused by us failing to keep track of all the networks that used to be there.  I have updated the tests to consistently repro that bug (which I believe to be the underlying issue of #1029), and this fix causes those tests to pass.  Put another way, I am certain this addresses *an* issue - I am pretty sure it addresses *that* issue.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -16,6 +16,30 @@ import (
 
 var (
 	instanceGroupManagerURL = regexp.MustCompile("^https://www.googleapis.com/compute/v1/projects/([a-z][a-z0-9-]{5}(?:[-a-z0-9]{0,23}[a-z0-9])?)/zones/([a-z0-9-]*)/instanceGroupManagers/([^/]*)")
+	networkConfig           = &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"cidr_blocks": {
+				Type:     schema.TypeSet,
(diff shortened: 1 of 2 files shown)


ndmckinley:
Ran the new tests locally - running the old tests now.

ndmckinley:
All 27 `ContainerCluster` tests (including the newly modified one) pass on this PR.

PR #1098, opened 2018-02-16
add json omitted fields back when converting

danawillow:
Currently when converting between two structs we lose `ForceSendFields` and `NullFields` because they contain the tag `json:"-"`, which cause `json.Marshal` and `json.Unmarshal` to ignore those fields.

Currently we handle this by adding back those fields immediately after we convert the object. The problems with this are twofold:
1. We have to remember to do so for each resource that gets converted this way
2. If the resource contains nested fields, we have to add the omitted fields back for each nested struct at the time we do the conversion, which means maintaining that list and remembering to update it if we add more fields to the schema

This change recursively checks the input struct for omitted fields and sets the output struct to have those values.

Hypothetically if we got rid of the check for the "-" tag from this change then it would do the full conversion without having to call the json functions. For now I'm very slightly leaning against that since (presumably) the json functions do some sort of error checking that this doesn't do, but I could be convinced.

### DIFF
--- a/google/api_versions.go
+++ b/google/api_versions.go
@@ -4,10 +4,6 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
-
-	// "github.com/davecgh/go-spew/spew"
-	"github.com/fatih/structs"
-	"github.com/mitchellh/mapstructure"
 )
 
 type ApiVersion uint8
@@ -42,15 +38,7 @@ func Convert(item, out interface{}) error {
 	// 	return err
 	// }
 
-	s := structs.Map(item)
-	// fmt.Printf("structs.Map: %#v\n", s)
-	// spew.Dump(s)
-	err := mapstructure.Decode(structs.Map(item), out)
-	if err != nil {
-		return err
-	}
-
-	// setOmittedFields(item, out)
+	setOmittedFields(item, out)
 
 	return nil
 }
--- a/google/api_versions.go
+++ b/google/api_versions.go
@@ -1,6 +1,7 @@
 package google
 
 import (
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
@@ -28,15 +29,15 @@ var OrderedContainerApiVersions = []ApiVersion{
 // between multiple API versions, as they are strict supersets of one another.
 // item and out are pointers to structs
 func Convert(item, out interface{}) error {
-	// bytes, err := json.Marshal(item)
-	// if err != nil {
-	// 	return err
-	// }
+	bytes, err := json.Marshal(item)
+	if err != nil {
+		return err
+	}
 
-	// err = json.Unmarshal(bytes, out)
-	// if err != nil {
-	// 	return err
-	// }
+	err = json.Unmarshal(bytes, out)
+	if err != nil {
+		return err
+	}
 
 	setOmittedFields(item, out)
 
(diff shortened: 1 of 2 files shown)


rileykarson:
Just to weigh in with my 2 cents on this; I agree with @rosbo that we should test to see if we can replace the JSON step entirely. I can't recall any extra error checking the JSON conversion performs (after all, it [regresses to converting the struct to a map[string]interface{} when `ForceSendFields` or `NullFields` are used](https://github.com/google/google-api-go-client/blob/master/gensupport/json.go#L20)) and it feels better to me to reflect struct -> struct in a single pass than to go struct -> json -> struct but then also copy 2 fields struct -> struct if they are used as well.

--- 

As an aside, we could also try leveraging libraries to do the work for us. I haven't seen any single library yet that goes from struct -> struct, but we could consider using [github.com/fatih/structs](https://github.com/fatih/structs) and [github.com/mitchellh/mapstructure](https://github.com/mitchellh/mapstructure). They would use a `map[string]interface` as an intermediary, and should? copy the structs as intended. The code we own would look something like:

```go
func Convert(item, out interface{}) error {
    err := mapstructure.Decode(structs.Map(item), out)
	if err != nil {
		return err
    }
}
```

danawillow:
👋 Hi Riley, good to hear from you!

So I gave both suggestions a try, but I'm still inclined to go with my original solution. I put the commits of trying both in here, so feel free to check them out if you're curious.

For replacing `Convert` entirely, we run into issues when trying to convert structs that don't have exactly the same fields- the exact reason why we needed the Convert function in the first place. It gives us this wonderfully confusing error message:
```
panic: reflect.Set: value of type *compute.InstanceProperties is not assignable to type *compute.InstanceProperties
```
I'm pretty sure in this case one of these is `v1/compute` and the other is `v0beta/compute`. Although this could be worked around with even more custom reflection, I'm just not sure it's worth the effort given that we already have something (the json way) that we're pretty confident works.

I love the idea of using structs/mapstructure, but structs sets unset pointers to nil in the result map and mapstructure doesn't seem to play well with nil fields:
```
--- FAIL: TestSetOmittedFields (0.00s)
panic: reflect: call of reflect.Value.Type on zero Value [recovered]
	panic: reflect: call of reflect.Value.Type on zero Value

goroutine 8 [running]:
testing.tRunner.func1(0xc420404d20)
	/usr/local/opt/go/libexec/src/testing/testing.go:711 +0x2d2
panic(0x23a7f80, 0xc42042e340)
	/usr/local/opt/go/libexec/src/runtime/panic.go:491 +0x283
reflect.Value.Type(0x0, 0x0, 0x0, 0x0, 0x0)
	/usr/local/opt/go/libexec/src/reflect/value.go:1694 +0x1a3
github.com/terraform-providers/terraform-provider-google/vendor/github.com/mitchellh/mapstructure.(*Decoder).decodeStruct(0xc42000ec48, 0x2173d6b, 0x5, 0x222ace0, 0x0, 0x2434920, 0xc42040f500, 0x199, 0x2434920, 0xbe1e0)
	/Users/danahoffman/go/src/github.com/terraform-providers/terraform-provider-google/vendor/github.com/mitchellh/mapstructure/mapstructure.go:650 +0xa9
github.com/terraform-providers/terraform-provider-google/vendor/github.com/mitchellh/mapstructure.(*Decoder).decode(0xc42000ec48, 0x2173d6b, 0x5, 0x222ace0, 0x0, 0x2434920, 0xc42040f500, 0x199, 0x20, 0x4002042e320)
```
Again, this could be fixed by either modifying the upstream libraries or adding some extra processing to get rid of these nil fields, but I think leaving this solution as-is gives us the best ROI.

rileykarson:
Hey Dana! 👋 I agree, using your changes as-is seems like the best option here. Thanks for trying out both approaches!

I'd forgotten how fun the nested structs like `compute.InstanceProperties` made converting between types. Like you said, the effort of adding even more custom reflection isn't worth it when we already have a working solution with the JSON.

PR #1102, opened 2018-02-19
Move method out of resource file to prepare for autogen

rosbo:


rosbo:
I was unsure about this. My thought process was to put them in the `_migrate.go` because once the `resource_*.go` files are autogenerated, these methods will only be used inside each `_migrate.go` file. If we put them in `compute_address.go`, it seems like these are used across the codebase. 

I am okay with moving them. Or leaving them there.

Thoughts?

danawillow:
Ah ok, if they're going to be used inside the migrate files only then that's fine. I'm curious to see how the generation turns out such that these will still be necessary but only in the migrate files :)

Issue #1105, opened 2018-02-19
Add GCS project account data source


amoiseiev:
Expose GCS project service account as data source

amoiseiev:
Implemented in #1110. Closing

PR #1110, opened 2018-02-20
Datasource for retrieving GCS service account

ishashchuk:
Closing issue [1105](https://github.com/terraform-providers/terraform-provider-google/issues/1105)


google/data_source_google_storage_project_service_account.go
```diff
@@ -0,0 +1,37 @@
...
+func dataSourceGoogleStorageProjectServiceAccountRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	serviceAccount, err := config.clientStorage.Projects.ServiceAccount.Get(project).Do()
+	if err != nil {
```
danawillow: This block here can be replaced with:
```
if err != nil {
  return handleNotFoundError(err, d, "GCS service account not found")
}
```

ishashchuk: Should be all set now

danawillow:
Thanks @ishashchuk!

Issue #910, opened 2018-01-03
Allow defining a bigquery_table schema directly


dasch:
Currently, one must write BigQuery table schemas as JSON, typically in a separate file that's loaded with `file()`. This is a rather odd user experience, as the rest of the table and most other resources in Terraform are defined using HCL.

I'd like to suggest that some DSL be added to the resource that allows defining the schema of the table. I'm not terribly opinionated about how that DSL should look, as long as I don't have to write JSON.

### Affected Resource(s)
- google_bigquery_table

danawillow:
Hey @dasch, the reason behind this is that BigQuery schemas can have arbitrary depth (i.e. a schema can contain another schema: https://godoc.org/google.golang.org/api/bigquery/v2#TableFieldSchema), which isn't something that HCL supports. See https://github.com/hashicorp/terraform/pull/13743 for the original discussion about this.

Defining another language to support just this use case isn't really a sustainable option when JSON works perfectly well, so I'm going to go ahead and close this but thanks @dasch for the suggestion and feel free to comment if there's something I've missed :)

dasch:
Do we need to support _all_ schemas if there's still a JSON fallback option? It would be a much better UX if there was at least support for schemas that can be represented in HCL...

danawillow:
Are you saying that BigQuery schemas _can_ be represented in HCL? HCL doesn't parse arbitrary text, it has to be part of a specific schema that we specify in the resource definition. Since this field is self-referential (a TableFieldSchema contains a list of other TableFieldSchemas), there's no way to represent it in the resource definition. If you have a workaround I'd be happy to hear it though!

dasch:
If the HCL version simply disallows nested schemas it wouldn't be a problem, right? As long as there's a JSON fallback you'd be able to cover most use cases with a nice DSL instead of forcing everyone to use JSON.

danawillow:
If you have a proposal for a "nice DSL" I'd be happy to hear it- it's not really in our best interest to have to come up with a new language along with a parser/compiler for it in order to support one resource for people that don't wish to use JSON.

One thing we _could_ do is have a schema field that allows you to define a flat list of `TableFieldSchema`s with no nesting, meaning people only need to fallback to json if they need nested fields. I'll reopen this issue in case anybody wants to take that on, but at least right now it won't be a high priority for us since there's a perfectly usable workaround.

dasch:
My preference would be for something like this:

```hcl
[...]

PR #1113, opened 2018-02-21
Allow defining BigQuery table fields inline

dasch:
An implementation of #910.

I'm pretty new to Go, so I wanted to push this PR even if it's a bit rough.

- [x] Add a repeated `field` to `google_bigquery_table` resources, allowing for inline definition of table schemas.
- [x] Support setting the name of the field.
- [x] Support setting the type of the field.
- [x] Support setting the description of the field.
- [x] Support making a field required.
- [x] Support making a field repeated.

### DIFF
--- a/google/resource_bigquery_table.go
+++ b/google/resource_bigquery_table.go
@@ -83,13 +83,45 @@ func resourceBigQueryTable() *schema.Resource {
 
 			// Schema: [Optional] Describes the schema of this table.
 			"schema": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Computed:     true,
-				ValidateFunc: validation.ValidateJsonString,
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ValidateFunc:  validation.ValidateJsonString,
+				ConflictsWith: []string{"field"},
 				StateFunc: func(v interface{}) string {
 					json, _ := structure.NormalizeJsonString(v)
 					return json
+				},
+			},
+
+			// Field: A field in the table's schema.
+			"field": &schema.Schema{
+				Type:          schema.TypeSet,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"schema"},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"type": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"description": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+						},
+						"mode": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Default:      "NULLABLE",
+							ValidateFunc: validation.StringInSlice([]string{"NULLABLE", "REQUIRED", "REPEATED"}, false),
+						},
+					},
 				},
 			},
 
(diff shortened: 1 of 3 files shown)


ndmckinley:
This looks like a great idea.

google/resource_bigquery_table.go
```diff
@@ -268,6 +285,17 @@ func resourceTable(d *schema.ResourceData, meta interface{}) (*bigquery.Table, e
 		table.TimePartitioning = expandTimePartitioning(v)
 	}
 
+	fields := d.Get("binding").(*schema.Set)
+	table.Schema = &bigquery.TableSchema{}
+
+	for i, v := range fields.List() {
+		field := v.(map[string]interface{})
+		table.Schema.Fields[i] = &bigquery.TableFieldSchema{
```
danawillow: This will segfault because table.Schema is nil at this point- take a look around the codebase for some examples of how to fill this in.

google/resource_bigquery_table.go
```diff
@@ -268,6 +285,17 @@ func resourceTable(d *schema.ResourceData, meta interface{}) (*bigquery.Table, e
 		table.TimePartitioning = expandTimePartitioning(v)
 	}
 
+	fields := d.Get("binding").(*schema.Set)
```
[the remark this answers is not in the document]
dasch: Ah, copy-paste error. I _think_ we need to use `Get` because `field` can be used multiple times – I copied this code from another place that had that behavior.

google/resource_bigquery_table.go
dasch: I set `table.Schema` on line 289 – can it still be nil?

google/resource_bigquery_table.go
```diff
@@ -93,6 +93,33 @@ func resourceBigQueryTable() *schema.Resource {
...
+						},
+						"description": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"mode": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Default:      "NULLABLE",
+							ValidateFunc: validation.StringInSlice([]string{"NULLABLE", "REQUIRED", "REPEATED"}, false),
```
dasch: @danawillow this API doesn't really abstract away the underlying BQ API – should it? Basically, a field can _either_ be optional, required, or repeated. But maybe using a string is bad?

google/resource_bigquery_table.go
danawillow: Oh sorry, it's a different reason- the Fields array needs to be initialized: https://play.golang.org/p/2AsZv6v55tP is a stripped down version of what's happening here.

google/resource_bigquery_table.go
danawillow: Nah, this is how we usually do enum-like values.

google/resource_bigquery_table.go
dasch: Ah, thanks!

dasch:
@danawillow sorry for all the questions, but I don't seem to be able to make the tests actually run – I try to break a test be e.g. screwing with the inline config, but `make test` reports success:

```
$ make test
go test -i $(go list ./... |grep -v 'vendor') || exit 1
echo $(go list ./... |grep -v 'vendor') | \
		xargs -t -n4 go test  -timeout=30s -parallel=4
go test -timeout=30s -parallel=4 github.com/terraform-providers/terraform-provider-google github.com/terraform-providers/terraform-provider-google/google github.com/terraform-providers/terraform-provider-google/scripts
ok  	github.com/terraform-providers/terraform-provider-google/google	0.090s
ok  	github.com/terraform-providers/terraform-provider-google/scripts	0.025s
```

dasch:
@danawillow one more: BQ only allows making a `REQUIRED` column `NULLABLE`, [all other column changes are forbidden](https://cloud.google.com/bigquery/docs/managing-table-schemas). With `TypeSet`, it's difficult to work with columns as logical units, since the `name` has no special role. Would it make more sense to use `TypeMap` here, and handle field diffs specially, e.g. not allowing changes that are illegal?

danawillow:
For running tests, you want `testacc`: check out the instructions at https://github.com/terraform-providers/terraform-provider-google/blob/master/.github/CONTRIBUTING.md and let me know if they can be improved!

As for your other comments, are you trying to add apply-time logic to catch these? If so, I would probably not bother- if the API throws an error, that'll happen at the exact same spot that it would if we threw the error from inside Terraform. If the API takes a long time to return or has a confusing error message, then I'd reconsider. For plan-time though, just make everything except for description and mode as `ForceNew`.

dasch:
Thanks! Will `ForceNew` work for `Elem` items in a `TypeSet`? If e.g. the `description` field is changed, Terraform doesn't recognize the item anymore – it's effectively a _new_ item, since the hash has changed. As such, does it even make sense to say a field has "changed"? That's why I asked about using `TypeMap` instead, since each item would be stable even if some fields on it changes.

danawillow:
Hmm, without spending some time playing around with it I'm not sure. You may also be interested in looking at setting a custom hash function on the set.

(also unrelated fyi- if you could avoid squashing your commits, that would be great just so I can see what changed in between reviews 🙂)

dasch:
> Hmm, without spending some time playing around with it I'm not sure. You may also be interested in looking at setting a custom hash function on the set.

I'll try that.

> (also unrelated fyi- if you could avoid squashing your commits, that would be great just so I can see what changed in between reviews 🙂)

Sure thing!

paddycarver:
Hey @dasch! I just looked through this, and it's looking really good. Is this something you're still looking to get merged?

As for the `REQUIRED` to `NULLABLE` change, you could probably accomplish that with [`CustomizeDiff`](https://godoc.org/github.com/hashicorp/terraform/helper/schema#Resource.CustomizeDiff): loop through all fields in a set, and track changes. You could also, as Dana mentioned, use a custom [`SchemaSetFunc`](https://godoc.org/github.com/hashicorp/terraform/helper/schema#Schema.Set) that only used the `name` of the column, making the keys stable as long as the name doesn't change.

dasch:
@danawillow @paddycarver would you be willing to merge this as-is and complete the changes yourselves? I'm afraid I've forgotten whatever Go I learned doing this PR...

rileykarson:
Closing PR as stale

Issue #1067, opened 2018-02-09
Allow specifying accelerators in cluster node_config


kmontag:
Beta support for GPUs on Kubernetes clusters has been released: https://cloud.google.com/kubernetes-engine/docs/concepts/gpus

It looks like they can be added via REST API using the `accelerators` field in a node pool configuration, e.g. (when creating a cluster):

```
POST https://container.googleapis.com/v1/projects/foo/zones/us-west1-a/clusters
{
  "cluster": {
    "name": "gpu-test",
    "zone": "us-west1-a",
    "nodePools": [
      {
        "name": "default-pool",
        "initialNodeCount": 1,
        "config": {
          // ...
          "accelerators": [
            {
              "acceleratorCount": 1,
              "acceleratorType": "nvidia-tesla-k80"
            }
          ]
        },
        // ...
      }
    ],
    // ...
  }
}
```

Would be great if this could be supported in the `node_config` block for `google_container_cluster` / `google_container_node_pool`.

Thanks!

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

```
Terraform v0.11.3
+ provider.google v1.5.0
```

### Affected Resource(s)
- google_container_cluster
- google_container_node_pool

dhague:
I have the same requirement - currently hacking on the code for it, unless someone else has already started work on this?

ndmckinley:
That's great!  Tag me in the PR and I'll expedite the review.

dhague:
Almost done with this - in case anyone is desperate for it, the code [here](https://github.com/dhague/terraform-provider-google/commit/9af59a1d0f95dcb7c4034bc147d25d7b18f50b43) works, I just need to tweak the tests to pass (so caveat emptor!). I'm on vacation for a long weekend, so I expect to finish this on Tuesday.

thomasjungblut:
I can also pick this up if there's anything urgent, synced with Darren yesterday on what's missing.

PR #1115, opened 2018-02-21
Allow specifying accelerators in cluster node_config (#1067)

dhague:


dhague:
Tagging @ndmckinley for expedited review :-)

google/node_config.go
```diff
@@ -128,6 +150,22 @@ func expandNodeConfig(v interface{}) *container.NodeConfig {
 		nc.MachineType = v.(string)
 	}
 
+	if v, ok := nodeConfig["guest_accelerator"]; ok {
+		accels := v.([]interface{})
+		guestAccelerators := make([]*container.AcceleratorConfig, 0, len(accels))
+		for _, raw := range accels {
+			data := raw.(map[string]interface{})
```
ndmckinley: It seems like you could make accels a []map[string]interface{}, which would mean you'd only have to do one of these casts.

google/node_config.go
```diff
@@ -128,6 +150,22 @@ func expandNodeConfig(v interface{}) *container.NodeConfig {
...
+			if data["count"].(int) == 0 {
+				continue
+			}
+			guestAccelerators = append(guestAccelerators, &container.AcceleratorConfig{
+				AcceleratorCount: int64(data["count"].(int)),
```
ndmckinley: Should be able to do data["count"].(int64) - why do it this way?

google/resource_container_node_pool_test.go
```diff
@@ -92,6 +92,24 @@ func TestAccContainerNodePool_withNodeConfig(t *testing.T) {
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckContainerNodePoolDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccContainerNodePool_withGPU(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckContainerNodePoolMatches("google_container_node_pool.np_with_gpu"),
+				),
+			},
```
ndmckinley: Looks great - can you stick an ImportState step on there so we know that accelerated node pools can be imported?

website/docs/r/container_node_pool.html.markdown
```diff
@@ -45,6 +45,10 @@ resource "google_container_cluster" "primary" {
       "https://www.googleapis.com/auth/logging.write",
       "https://www.googleapis.com/auth/monitoring",
     ]
+    guest_accelerator = [{
+      type="nvidia-tesla-k80"
+      count=1
+    }]
   }
```
ndmckinley: You'll need to add the details of this structure to the documentation below here.

dhague: Done.

google/node_config.go
dhague: I lazily lifted this section of code from at https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_instance.go#L1384, so it may be better to maintain consistency

google/node_config.go
dhague: See above

google/resource_container_node_pool_test.go
dhague: It would be the only ImportState in this file - why introduce it for just this case?

google/node_config.go
dhague: Making accels a `[]map[string]interface{}` leads to an error:
```
panic: interface conversion: interface {} is []interface {}, not []map[string]interface {}

goroutine 297 [running]:
github.com/terraform-providers/terraform-provider-google/google.expandNodeConfig(0x16797c0, 0xc4205361c0, 0xb)
	/home/i056593/Code/go/src/github.com/terraform-providers/terraform-provider-google/google/node_config.go:154 +0x12e0
```

google/node_config.go
dhague: Using `data["count"].(int64)` leads to an error:
```
panic: interface conversion: interface {} is int, not int64

goroutine 296 [running]:
github.com/terraform-providers/terraform-provider-google/google.expandNodeConfig(0x16797c0, 0xc42035a040, 0xb)
	/home/i056593/Code/go/src/github.com/terraform-providers/terraform-provider-google/google/node_config.go:162 +0x12b4
```

google/node_config.go
dhague: (Side note: Line 154 is basically copied from https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_instance_template.go#L504)

google/node_config.go
dhague: (Side note: line 162 is basically copied from https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_instance_template.go#L512)

google/resource_container_node_pool_test.go
ndmckinley: We've talked to Hashicorp about best practices recently, and so it's a new thing we're doing with new tests.

google/node_config.go
ndmckinley: Ah, all right, that makes sense.

ndmckinley:
Go ahead and add that test and it'll be good to merge.

dhague:
ImportState test added & passing - looking OK?

ndmckinley:
Running the tests now, and then I'll merge it.

ndmckinley:
Some difficulties with the test environment but it should only be another few hours.  Monday at the latest.

ndmckinley:
Aha, great, tests pass now.  Merging.

dhague:
Awesome - thanks!

PR #1119, opened 2018-02-22
Datasource for GCE service account

### LINKED ISSUE #1104: Add default GCE account data source
For IAM related purposes, we need to expose default GCE account as a data source.

ishashchuk:
Fixes #1104

### DIFF
--- a/google/data_source_google_compute_default_service_account.go
+++ b/google/data_source_google_compute_default_service_account.go
@@ -7,6 +7,16 @@ import (
 func dataSourceGoogleComputeDefaultServiceAccount() *schema.Resource {
 	return &schema.Resource{
 		Read: dataSourceGoogleComputeDefaultServiceAccountRead,
+		Schema: map[string]*schema.Schema{
+			"email": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"project_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+		},
 	}
 }
 
@@ -18,12 +28,17 @@ func dataSourceGoogleComputeDefaultServiceAccountRead(d *schema.ResourceData, me
 		return err
 	}
 
+	if v, ok := d.GetOk("project_id"); ok {
+		project = v.(string)
+	}
+
 	projectCompResource, err := config.clientCompute.Projects.Get(project).Do()
 	if err != nil {
 		return handleNotFoundError(err, d, "GCE service account not found")
 	}
 
 	d.SetId(projectCompResource.DefaultServiceAccount)
-
+	d.Set("email", projectCompResource.DefaultServiceAccount)
+	d.Set("project_id", project)
 	return nil
 }
--- a/google/data_source_google_compute_default_service_account.go
+++ b/google/data_source_google_compute_default_service_account.go
@@ -12,7 +12,7 @@ func dataSourceGoogleComputeDefaultServiceAccount() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
-			"project_id": {
+			"project": {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
@@ -28,10 +28,6 @@ func dataSourceGoogleComputeDefaultServiceAccountRead(d *schema.ResourceData, me
 		return err
 	}
 
-	if v, ok := d.GetOk("project_id"); ok {
-		project = v.(string)
-	}
-
 	projectCompResource, err := config.clientCompute.Projects.Get(project).Do()
 	if err != nil {
 		return handleNotFoundError(err, d, "GCE service account not found")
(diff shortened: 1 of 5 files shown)


google/data_source_google_compute_default_service_account.go
```diff
@@ -0,0 +1,29 @@
+package google
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceGoogleComputeDefaultServiceAccount() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGoogleComputeDefaultServiceAccountRead,
```
rosbo: Could you add an optional project field to the schema? This way, users can find the default service account for projects that are not necessarily set as their default Terraform provider project.

website/docs/d/google_compute_default_service_account.html.markdown
```diff
@@ -0,0 +1,32 @@
...
+## Argument Reference
+
+There are no arguments available for this data source.
+
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - Email address of the default service account used by VMs running in this project
```
rosbo: Add a field named `service_account` in the schema instead.

ishashchuk: I can do that, but the data source itself is named data "google_compute_default_service_account". Does it make sense to add a field "service_account"? "google_compute_default_service_account.service_account" seems a little repetitive to me

rosbo: @danawillow what are your thoughts on this?

To give you some context. The DefaultServiceAccount is retrieved by calling the compute Project GET API. Our `google_project` data source is getting the project details from the resource manager GET API. The response **does not** include the DefaultServiceAccount.

I see different options:
1. Create a `google_compute_default_service_account` resource (this PR). In that case, do we only set the id field or do we call the field a different name (e.g. service_account)?
2. Create a new resource to get all the compute GET API project details. In addition to the DefaultServiceAccount, we can also get the quotas, enabledFeatures and commoninstancemetadata. But how do we call this data source since we already have one for the resource manager GET API called `google_project`?
3. Make a call to the compute GET API in the project data source and include the data there.

danawillow: 1. Interpolating based on ids isn't a good idea, since we've had many examples of changing ids for a resource over time. If we go with option one, then we should have it as a separate field even though it seems weird. In the `google_service_account` resource we have fields called `email` and `account_id`, so this one could match that.
2. If we did this, we could call it `google_compute_project` or `google_compute_project_info`. I don't love either of those names, but metadata is a totally separate thing we can't call it that.
3. The thing that makes me a bit nervous about this is that the info the compute projects resource returns is compute specific. For example, the quotas it returns are only the GCE quotas. I don't know if that will ever change.

Right now I'm leaning towards option 1 (and then just have separate data sources for each item in the project resources as needed), but could definitely be persuaded that option 2 is the way to go.

rosbo: Thanks for your input Dana :)

I am also leaning towards option 1. To be consistent with the `google_service_account` resource, we should add a new field called `email` and set the default service account email to this field.

In the future, if users ask for the default service account id, we can add a new field called `account_id` to this `google_compute_default_service_account` data source  and extract this value from the email.

ishashchuk: @rosbo @danawillow thanks for you input. I agree that having those to be consistent is a better approach.  commit with the requested changes is on its way.

google/data_source_google_compute_default_service_account.go
ishashchuk: I think it's ready for another pass

google/data_source_google_compute_default_service_account.go
```diff
@@ -0,0 +1,44 @@
...
+		Schema: map[string]*schema.Schema{
+			"email": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"project_id": {
```
rosbo: For consistency with all the other resource, could you rename this field "project"

google/data_source_google_compute_default_service_account.go
```diff
@@ -0,0 +1,44 @@
...
+
+func dataSourceGoogleComputeDefaultServiceAccountRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	if v, ok := d.GetOk("project_id"); ok {
```
rosbo: the getProject method checks first if the "project" field is set and then fallback on the default provider project. This means you won't need this check here.

website/docs/d/google_compute_default_service_account.html.markdown
```diff
@@ -0,0 +1,34 @@
...
+output "default_account" {
+  value = "${google_compute_default_service_account.default.email}"
+} 
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project_id` - (Optional) The project ID. If it is not provided, the provider project is used.
```
rosbo: Rename here too.

rosbo:
Merged! Thank you @ishashchuk for your contribution :)

PR #1121, opened 2018-02-22
s/TestAccGoogle/TestAcc/

danawillow:
Some of our tests have the extra "Google" and others don't. Since we're in the google provider already, all our tests are implicitly Google-based, and so we should standardize on not putting the extra "Google" in.

### DIFF
--- a/google/data_source_google_kms_secret_test.go
+++ b/google/data_source_google_kms_secret_test.go
@@ -5,14 +5,15 @@ import (
 	"fmt"
 	"testing"
 
+	"log"
+
 	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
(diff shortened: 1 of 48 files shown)


google/data_source_google_kms_secret_test.go
```diff
@@ -5,14 +5,15 @@ import (
 	"fmt"
 	"testing"
 
+	"log"
+
```
ndmckinley: Spacing is a little off.

google/resource_kms_key_ring_test.go
```diff
@@ -4,10 +4,11 @@ import (
 	"fmt"
 	"testing"
 
+	"log"
```
ndmckinley: Spacing again ... but maybe there's something I don't know about the `"log"` package?

danawillow: I think it's just gofmt being confused. It doesn't try to change it if I reorder them manually though, so done.

PR #1122, opened 2018-02-22
Let the compute zones data source accept a `project` parameter

jphalip:


rosbo:
Thanks for adding this.

PR #1133, opened 2018-02-27
Update `cluster_ipv4_cidr` For `google_container_cluster`

### LINKED ISSUE #1132: Resource `google_container_engine` Enforcing Un-necessary Check
The `google_container_cluster` resource is enforcing an unnecessary check to make sure the CIDR is valid for `cluster_ipv4_cidr` https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_container_cluster.go#L141 however the API actually accepts values other than CIDRs (https://github.com/terraform-providers/terraform-provider-google/blob/master/vendor/google.golang.org/api/container/v1/container-gen.go#L338).

kylegoch:
This resolves #1132.

- remove CIDR validation check as the field can accept non-cidr values

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -138,7 +138,6 @@ func resourceContainerCluster() *schema.Resource {
 				Optional:     true,
 				Computed:     true,
 				ForceNew:     true,
-				ValidateFunc: validateRFC1918Network(8, 32),
 			},
 
 			"description": {
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -134,10 +134,10 @@ func resourceContainerCluster() *schema.Resource {
 			},
 
 			"cluster_ipv4_cidr": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Computed:     true,
-				ForceNew:     true,
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
 			},
 
 			"description": {


ndmckinley:
Hi, I mentioned this in the issue, but I see the documentation string:

```
        // ClusterIpv4Cidr: The IP address range of the container pods in this
	// cluster,
	// in
	// [CIDR](http://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing)
	//
	// notation (e.g. `10.96.0.0/14`). Leave blank to have
	// one automatically chosen or specify a `/14` block in `10.0.0.0/8`.
```

Meaning that the only valid values are CIDRs or blank, right?

kylegoch:
Correct. We tried to make a module for this, but terraform didnt like it when we tried to pass `""` in the event that our module didnt want to set this field.

So maybe rather than striking it, it should check for cidr or blank?

ndmckinley:
Absolutely!  I'd approve and merge that.

ndmckinley:
Do you still want to add that?  I'm happy to leave this open if you do!  If you're not interested anymore or you don't need it anymore, I'll close it.

rosbo:
We could create a method called `OrEmpty` to easily accept empty values for a given ValidateFunc.

```go
func OrEmpty(f schema.SchemaValidateFunc) schema.SchemaValidateFunc {
  return func(i interface{}, k string) ([]string, []error) {
    v, ok := i.(string)
    if ok && v == "" {
        return nil, nil
    }
    return f(i,k)
  }
}

// Usage
*schema.schema {
  ...
  ValidateFunc: OrEmpty(validateRFC1918Network(8, 32))
}
```

The benefit of creating a `OrEmpty` generic method is to be able to reuse the same validation logic whether the field is Optional or Required. By composing `OrEmpty` with another validate function, it makes that validate function simpler because it does not have to handle the "" case.

ndmckinley:
You're right, I prefer that, and you've already written it.  You want credit for the PR, or want me to do it?  :)

rosbo:
I can open a quick PR.

rosbo:
Closed in favor of #1446

Issue #51, opened 2017-06-13
Feature Request: support rolling updates for google instance group manager.


hashibot:
_This issue was originally opened by @qvallance-ctc as hashicorp/terraform#10647. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Google has added functionality for performing rolling updates for managed instance groups.
https://cloud.google.com/compute/docs/instance-groups/updating-managed-instance-groups

Would like Terraform's google_compute_instance_group_manager to support rolling updates as an update policy.

(Functionality is currently in Alpha - opening request to track progress for when it goes GA)

InstanceGroupManagerUpdatePolicy (v0.alpha)
https://code.googlesource.com/google-api-go-client/+/master/compute/v0.alpha/compute-gen.go#7175
https://code.googlesource.com/google-api-go-client/+/master/compute/v0.alpha/compute-api.json#4339

rileykarson:
Linking to the general alpha/beta issue: https://github.com/terraform-providers/terraform-provider-google/issues/93

danielcompton:
One thing that could be useful to have (maybe in a separate PR, but thought I'd mention it here), is the ability to keep the last n instance group templates around. This would mean you could do things like canary deploys, or rollback to previous templates with the instance group updater.

domudall:
Giving this a bump as it's now in beta (as instructed [here](https://github.com/terraform-providers/terraform-provider-google/issues/93#issuecomment-321395960)) =)

zbikmarc:
Any chances to have it done in near future?

shaharmor:
I would be happy to see this as well

amarlot:
Confirmed. It would be great to support rolling update.

seanmalloy:
@danawillow this would be a very nice feature to have. I sent you brief email with some additional details.

danawillow:
@paddycarver looks like this is assigned to you. Have you started thinking about this? Also @ishashchuk it looks like you might be doing some work around this. If you are could you claim the issue here so we don't end up duplicating work? Thanks!

paddycarver:
I've investigated this and it's on my short list. I'll either have an update on 3/1 or will unassign myself from the issue at that time to free it up for someone else to nab. Sorry for the slowness on my end. :(

ishashchuk:
[...]

PR #1137, opened 2018-02-28
Rolling update support for instance group manager

ishashchuk:
Addresses [51](https://github.com/terraform-providers/terraform-provider-google/issues/51)

Supports beta functionality for rolling updates.

When rolling_update is selected as update_strategy, it requires "rolling_update_policy" block.
I'm storing rolling_update_policy block locally, and not applying it to computeBeta.InstanceGroupManager. There is no way to make an update call for this block on the InstanceGroupManager. It is only updated when the patch call for the instances is made. That's the reason I'm storing it locally and only applying it on the update calls that are changing the instance template. That allows us to change rolling_update_policy arguments post creating InstanceGroupManager and making sure that they will be applied at the next  patch call.

Once I get feedback on this, I will add similar support to region_instance_group_manager

### DIFF
--- a/google/resource_compute_instance_group_manager.go
+++ b/google/resource_compute_instance_group_manager.go
@@ -602,6 +602,11 @@ func resourceComputeInstanceGroupManagerUpdate(d *schema.ResourceData, meta inte
 		}
 
 		if d.Get("update_strategy").(string) == "ROLLING_UPDATE" {
+			// UpdatePolicy is set for InstanceGroupManager on update only, because it is only relevant for `Patch` calls.
+			// Other tools(gcloud and UI) capable of executing the same `ROLLING UPDATE` call
+			// expect those values to be provided by user as part of the call
+			// or provide their own defaults without respecting what was previously set on UpdateManager.
+			// To follow the same logic, we provide policy values on relevant update change only.
 			manager := &computeBeta.InstanceGroupManager{
 				UpdatePolicy: expandUpdatePolicy(d.Get("rolling_update_policy").([]interface{})),
 			}
--- a/google/resource_compute_instance_group_manager.go
+++ b/google/resource_compute_instance_group_manager.go
@@ -14,7 +14,10 @@ import (
 )
 
 var InstanceGroupManagerBaseApiVersion = v1
-var InstanceGroupManagerVersionedFeatures = []Feature{Feature{Version: v0beta, Item: "auto_healing_policies"}}
+var InstanceGroupManagerVersionedFeatures = []Feature{
+	Feature{Version: v0beta, Item: "auto_healing_policies"},
+	Feature{Version: v0beta, Item: "rolling_update_policy"},
+}
 
 func resourceComputeInstanceGroupManager() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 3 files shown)


apenney:
@ishashchuk I don't have any code feedback, I just wanted to thank you for tackling this (as we'll be early adopters of it, soon as it's merged).

amoiseiev:
@paddycarver friendly reminder :)

google/resource_compute_instance_group_manager.go
```diff
@@ -184,6 +241,12 @@ func resourceComputeInstanceGroupManagerCreate(d *schema.ResourceData, meta inte
 		return err
 	}
 
+	_, ok := d.GetOk("rolling_update_policy")
```
ndmckinley: Nit: Seems like it'd be neater as part of the if block below.

google/resource_compute_instance_group_manager.go
```diff
@@ -380,6 +443,11 @@ func resourceComputeInstanceGroupManagerUpdate(d *schema.ResourceData, meta inte
 
 	d.Partial(true)
 
+	_, ok := d.GetOk("rolling_update_policy")
```
ndmckinley: same comment - include in the if statement if possible.

ishashchuk:
> I like this... but I wonder, what happens if someone:
> 
> - creates the resource with RESTART
> - changes it to ROLLING_UPDATE
> - does an update
> - changes it back to RESTART
> - does another update
> 
> In that case, would this code act correctly? It seems like it wouldn't send a PATCH request to remove the strategy, but I might be off.

@ndmckinley 
There is no need to remove the strategy, since it only matters  on PATCH requests. PATCH request is using a different API endpoint(`clientComputeBeta.InstanceGroupManagers.Patch`) and is only used when the resource is set to ROLLING_UPDATE. And we will always get the most up-to-date strategy from the config, because we are passing it explicitly on template change

When resource is set to RESTART, `config.clientCompute.InstanceGroupManagers.RecreateInstances` is  used and that endpoint doesn't care about update_strategy attributes. It reboots all the machines at the same time.

ndmckinley:
Ah!  That's quite nice, then, neat trick.  I _think_ (but am not 100% sure) that I'd like it to be set during `Create` as well as during `Update`, then, because in the event that someone creates a resource with Terraform then runs a `ROLLING_UPDATE` manually or using some other tool, I think the most sensible behavior is to default to what they specified in Terraform.  Does that seem right to you?

ishashchuk:
@ndmckinley 
I don't think there is much value in setting update on `CREATE`, because most of the tools are either expect the user to provide those values, or use their own defaults.

In UI, the values for `ROLLING RESTART/REPLACE` are pre-populated with defaults, which means that the values instantiated on terraform create won't be respected

Running `gcloud beta compute instance-groups managed rolling-action` without specifying strategy, use their own defaults too without respecting anything that was previously set on instance manager

ndmckinley:
Fair enough - can you add a comment to that effect?  Just that and I'll be ready to merge.

ishashchuk:
Should be good to go

ndmckinley:
Perfect.  I'm running the tests now, and I expect to merge in an hour or so assuming no problems.

Issue #862, opened 2017-12-14
google_dns_record_set always wants to update DNS records if you use uppercase letters


abevoelker:
### Terraform Version
```
Terraform v0.11.1
+ provider.google v1.4.0
```

### Affected Resource(s)
- google_dns_record_set

### Terraform Configuration Files
```hcl
resource "google_dns_record_set" "mx" {
  name = "example.com."
  type = "MX"
  ttl  = 300
  managed_zone = "foo-zone"

  rrdatas = ["1 ASPMX.L.GOOGLE.COM.",
             "5 ALT1.ASPMX.L.GOOGLE.COM.",
             "5 ALT2.ASPMX.L.GOOGLE.COM.",
             "10 ASPMX2.GOOGLEMAIL.COM.",
             "10 ASPMX3.GOOGLEMAIL.COM."]
}
```

### Debug Output
https://gist.github.com/abevoelker/d59e926a52aea181009ea724ec563154

### Panic Output
N/A

### Expected Behavior
After the initial `terraform apply`, subsequent `terraform plan` operations should say "No changes. Infrastructure is up-to-date."

### Actual Behavior
`terraform plan` continually wants to update the DNS records because the casing doesn't match

### Steps to Reproduce
1. `terraform apply`
2. `terraform plan`

### Important Factoids
The uppercase examples came straight from G Suite's setup instructions. Maybe the `google_dns_record_set` should downcase rrdatas before comparing to remote state.

### References
N/A

PR #1141, opened 2018-03-01
Fix perpetual diff in #863.

### LINKED ISSUE #863: google_dns_record_set always wants to update TXT records that are not escape-quoted
### Terraform Version
```
Terraform v0.11.1
+ provider.google v1.4.0
```

### Affected Resource(s)
- google_dns_record_set

```hcl
resource "google_dns_record_set" "foo" {
  name = "example.com."
  type = "TXT"
  ttl  = 300
  managed_zone = "foo-zone"

  rrdatas = ["foo"]
}
```

### Debug Output
https://gist.github.com/abevoelker/e35529ee72f8b0619a9cae0130c8ce03

### Panic Output
N/A

### Expected Behavior
After the initial `terraform apply`, subsequent `terraform plan` operations should say "No changes. Infrastructure is up-to-date."

### Actual Behavior
`terraform plan` continually wants to update the DNS record.

### Steps to Reproduce
1. `terraform apply`
2. `terraform plan`

### Important Factoids
I can work around the issue by escape-quoting the value (`rrdatas = ["\"foo\""]`), however the documentation makes it sound like that should only be required for values with spaces in them.

### References
- #232
- hashicorp/terraform#13463

paddycarver:
This treats record set RRdata values as equivalent if they're quoted and
unquoted, as the GCP backend likes to return TXT records quoted.

It also updates the DNS name we're using to be a subdomain of
hashicorptest.com, as the tests started failing when run in parallel if
I tried adding the same zone domain multiple times.

Fixes #863

### DIFF
--- a/google/resource_dns_record_set.go
+++ b/google/resource_dns_record_set.go
@@ -40,7 +40,7 @@ func resourceDnsRecordSet() *schema.Resource {
 					Type: schema.TypeString,
 				},
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					return strings.Trim(old, `"`) == strings.Trim(new, `"`)
+					return strings.ToLower(strings.Trim(old, `"`)) == strings.ToLower(strings.Trim(new, `"`))
 				},
 			},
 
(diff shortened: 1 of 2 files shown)


paddycarver:
(I'm going to actually lump the fix for #862 in with this, as they're modifying the same exact thing.)

PR #1143, opened 2018-03-02
Add docs for first_ip_address on sql db instance.

genevieve:
@danawillow I saw you already updated the changelog, so I only updated the doc for the sql db instance. Let me know if I missed anything or if you prefer a different wording!

Thank you!!! 

Gen


danawillow:
Would you mind adding a one-sentence description of what the use case for this is? (just in case someone wonders in the future why this particular list has its first element broken out) You can link to the issues if you want.

genevieve:
Does that sound better?

genevieve:
I can also add the issue from the terraform core repo

danawillow:
That's perfect, thanks!

PR #1145, opened 2018-03-02
computeDiskUserRegexString in resource_compute_disk.go should also accept projectIds that contain ":" or "."

derfloh:
I ran into problems, when I had a project that was using a project ID in the format of "hashicorptest.net:my-project". 
The problem came up, when terraform tried to delete attached disks and threw an error like "Unknown user XXX of disk YYY"

This PR adds ":" and "." to the projectID part of computeDiskUserRegexString

### DIFF
--- a/google/resource_compute_disk.go
+++ b/google/resource_compute_disk.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	computeDiskUserRegexString = "^(?:https://www.googleapis.com/compute/v1/projects/)?([-_a-zA-Z0-9:.]*)/zones/([-_a-zA-Z0-9]*)/instances/([-_a-zA-Z0-9]*)$"
+	computeDiskUserRegexString = "^(?:https://www.googleapis.com/compute/v1/projects/)?(" + ProjectRegex + ")/zones/([-_a-zA-Z0-9]*)/instances/([-_a-zA-Z0-9]*)$"
 )
 
 var (
--- a/google/resource_compute_disk.go
+++ b/google/resource_compute_disk.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	computeDiskUserRegexString = "^(?:https://www.googleapis.com/compute/v1/projects/)?([-_a-zA-Z0-9]*)/zones/([-_a-zA-Z0-9]*)/instances/([-_a-zA-Z0-9]*)$"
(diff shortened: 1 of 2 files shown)


google/resource_compute_disk_test.go
```diff
@@ -372,6 +372,36 @@ func TestAccComputeDisk_deleteDetach(t *testing.T) {
...
+		"https://www.googleapis.com/compute/v1/projects/project-id/zones/us-central1/instances/123",
+		"https://www.googleapis.com/compute/v1/projects/123123/zones/us-central1/instances/123",
+		"https://www.googleapis.com/compute/v1/projects/hashicorptest.net:project-123/zones/us-central1/instances/123",
+		"https://www.googleapis.com/compute/v1/projects/123/zones/456/instances/789",
+	}
+
+	shouldFail := []string{
+		"https://www.googleapis.com/compute/v1/projects/project#/zones/us-central1/instances/123",
+		"https://www.googleapis.com/compute/v1/projects/project/zones/us-central#/instances/123",
+		"https://www.googleapis.com/compute/v1/projects/project/zones/us-central1/instances/?",
```
ndmckinley: Looks right - can you add a shouldFail test for "foo.com:bar:baz", and "foo.com:"?

google/resource_compute_disk.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	computeDiskUserRegexString = "^(?:https://www.googleapis.com/compute/v1/projects/)?([-_a-zA-Z0-9]*)/zones/([-_a-zA-Z0-9]*)/instances/([-_a-zA-Z0-9]*)$"
+	computeDiskUserRegexString = "^(?:https://www.googleapis.com/compute/v1/projects/)?([-_a-zA-Z0-9:.]*)/zones/([-_a-zA-Z0-9]*)/instances/([-_a-zA-Z0-9]*)$"
```
ndmckinley: This is just a little bit too loose: you probably want this regex right here:
https://github.com/terraform-providers/terraform-provider-google/blob/411017c35f96ff73b951e49607093b442cad360a/google/validation.go#L15:2

google/resource_compute_disk_test.go
ndmckinley: Sorry, though of course those should fail, I meant as project names:
"https://www.googleapis.com/compute/v1/projects/foo.com:bar:baz/zones/us-central1/instances/123".

Issue #149, opened 2017-06-21
Feature: Backend Services as Data


johnsonj:
### Affected Resource(s)
- google_compute_backend_service
- google_compute_region_backend_service

### Terraform Configuration Files
```hcl
data "google_compute_backend_service" "global" {
	name        = "global-backend"
}

data "google_compute_region_backend_service" "regional" {
	name        = "regional-backend"
        region      = "us-east1"
}

output "backend_service_global" {
	value = "${data.google_compute_backend_service.global.self_link}"
}

output "backend_service_regional" {
	value = "${data.google_compute_region_backend_service.regional.self_link}"
}

// Now I can create LBs!
resource "google_compute_target_http_proxy" "http-lb-proxy" {
  name = "tf-http-lb-proxy"
  url_map = "${google_compute_url_map.web-map.self_link}"
}

resource "google_compute_global_forwarding_rule" "default" {
  name = "tf-http-content-gfr"
  target = "${google_compute_backend_service.global-backend.self_link}"
  ip_address = "${google_compute_global_address.external-address.address}"
  port_range = "80"
}

resource "google_compute_url_map" "web-map" {
  name = "tf-web-map"
  default_service = "${google_compute_backend_service.global-backend.self_link}"

  host_rule {
    hosts = ["*"]
    path_matcher = "tf-allpaths"
  }

  path_matcher {
    name = "tf-allpaths"
    default_service = "${google_compute_backend_service.global-backend.self_link}"

    path_rule {
      paths = ["/*", "/"]
      service = "${google_compute_backend_service.global-backend.self_link}"
    }
  }
}
// OMITTED: 
// - Firewall rule for LB -> Instance Group
// - Firewall rule for Healthcheck
// - Global Address
```

### Desired Behavior
I'd like to create load balancers (HTTP/HTTPS/TCP/Internal TCP) out of band of instance group/backend service creation by referencing the backend services as data.

ndmckinley:
That merged PR took care of this a few months ago.

PR #1150, opened 2018-03-05
Datasource for Backend Services

bpineau:
Exposes existing `google_compute_backend_service` as data sources.
This addresses #149 (except for `google_compute_region_backend_service`,
which is not covered here).

This allows, for instance, to collect a backend service's self_link and
use it from an other workspace/tfstate, sharing most of the
loadbalancers definition.

```
$ env GOOGLE_REGION=us-central1 make testacc TEST=./google TESTARGS='-run=TestAccGoogleComputeBackendService_'
TF_ACC=1 go test ./google -v -run=TestAccGoogleComputeBackendService_ -timeout 120m
=== RUN   TestAccGoogleComputeBackendService_basic
--- PASS: TestAccGoogleComputeBackendService_basic (45.09s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	45.105s
```

### DIFF
--- a/google/data_source_google_compute_backend_service_test.go
+++ b/google/data_source_google_compute_backend_service_test.go
@@ -1,0 +1,85 @@
+package google
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccDataSourceComputeBackendService_basic(t *testing.T) {
+	t.Parallel()
+
+	serviceName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	checkName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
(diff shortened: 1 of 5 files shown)


google/data_source_google_compute_backend_service_test.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccGoogleComputeBackendService_basic(t *testing.T) {
```
ndmckinley: nit: please remove "Google" from test names?  We're trying to do that as of recently.

ndmckinley:
Tests pass - happy to merge once you make that quick change.

google/data_source_google_compute_backend_service_test.go
bpineau: Thanks for the review @ndmckinley !
I replaced the "Google" with "DataSource" in the tests names,
since just removing this string would name-clash with the resource's own tests, I hope that's ok?

PR #1155, opened 2018-03-05
[docs] - Improve docs for google_container_{node_pool,cluster}

nyodas:
Greetings
Following a feeling that we were missing a bit of docs.
I tried to add a few things that i thought were missing
Feel free to point where i might have been overzealous.

### DIFF
--- a/website/docs/r/container_cluster.html.markdown
+++ b/website/docs/r/container_cluster.html.markdown
@@ -94,11 +94,14 @@ output "cluster_ca_certificate" {
 * `enable_legacy_abac` - (Optional) Whether the ABAC authorizer is enabled for this cluster.
     When enabled, identities in the system, including service accounts, nodes, and controllers,
     will have statically granted permissions beyond those provided by the RBAC configuration or IAM.
-
+    Defaults to `true`
+    
 * `initial_node_count` - (Optional) The number of nodes to create in this
     cluster (not including the Kubernetes master). Must be set if `node_pool` is not set.
 
 * `ip_allocation_policy` - (Optional) Configuration for cluster IP allocation. As of now, only pre-allocated subnetworks (custom type with secondary ranges) are supported.
+    This will activate IP aliases. See the [official documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/ip-aliases)
+    Structure is documented below.
 
 * `logging_service` - (Optional) The logging service that the cluster should
     write logs to. Available options include `logging.googleapis.com` and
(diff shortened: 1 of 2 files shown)


website/docs/r/container_cluster.html.markdown
```diff
@@ -202,6 +207,8 @@ The `ip_allocation_policy` block supports:
     ClusterIPs. This must be an existing secondary range associated with the cluster
     subnetwork.
 
+> Please note that this activate Ip aliases. See the [official documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/ip-aliases)
```
danawillow: I think a better place for this would be in the section on `ip_allocation_policy` at line 102. Then you wouldn't need for it to be a separate note, but just part of the description.

(also nit: activates, IP)

website/docs/r/container_node_pool.html.markdown
```diff
@@ -14,19 +14,19 @@ and
...
 ## Example usage
-
+### Standard usage
 ```hcl
 resource "google_container_node_pool" "np" {
   name               = "my-node-pool"
   zone               = "us-central1-a"
   cluster            = "${google_container_cluster.primary.name}"
-  initial_node_count = 3
+  node_count = 3
```
danawillow: The spacing on this is a bit off now- the lines above should have the `=` aligned with this one (same comment for the other examples too)

website/docs/r/container_node_pool.html.markdown
```diff
@@ -52,7 +52,42 @@ resource "google_container_cluster" "primary" {
...
+  name               = "marcellus-wallace"
+  zone               = "us-central1-a"
 
+  lifecycle {
+    ignore_changes = [
+      "node_pool"
+    ]
+  }
+  
+  node_pool = [{
```
danawillow: This can just be:
```
node_pool {
  name = "default-pool"
}
```

nyodas: Fixed.

website/docs/r/container_node_pool.html.markdown
nyodas: Done

website/docs/r/container_cluster.html.markdown
nyodas: I moved it to line 102 as you suggested
I changed the wording slightly to fit w/h what was used line 96 (I'm not satisfied with it so if you have any suggestion)

website/docs/r/container_node_pool.html.markdown
danawillow: The `= [` is also unnecessary

nyodas: Indeed my bad,i kinda misread that.
reFixed :)

danawillow:
Thanks! I made a few small changes that were easier to just make than to try to write out. Merging now!

PR #1171, opened 2018-03-08
Dataproc autogen bucket changes

danawillow:
This PR does two things:
1. Deprecates `delete_autogen_bucket`. This shouldn't really be an option- clusters in the same region all share the same autogenerated bucket, so if we delete it then it's unclear what would happen to other dataproc clusters in that region. There's not really much harm in keeping it around, since there's only one per region.
2. In the meantime, adds a bit of waiting for objects in the bucket to be deleted. Object deletion is asynchronous, but doesn't provide any sort of wait operation. Instead, we'll list the objects until it returns that it's empty. Since we're deprecating the feature anyway, we don't necessarily need this part, but I left it in just in case.


google/resource_dataproc_cluster.go
```diff
@@ -92,9 +93,10 @@ func resourceDataprocCluster() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 
 						"delete_autogen_bucket": {
-							Type:     schema.TypeBool,
-							Optional: true,
-							Default:  false,
+							Type:       schema.TypeBool,
+							Optional:   true,
+							Default:    false,
+							Deprecated: "see documentation",
```
paddycarver: A link to the documentation here would be great.

danawillow:
Added more context to the deprecation message.

PR #1177, opened 2018-03-09
Add generated resources for http(s)_health_check, updates to backend_bucket, and import helpers

danawillow:
(I separated the changes out into logical commits, but let me know if you want me to separate out the PRs too)


google/resource_compute_http_health_check.go
```diff
@@ -94,159 +109,260 @@ func resourceComputeHttpHealthCheckCreate(d *schema.ResourceData, meta interface
...
 	if err != nil {
 		return err
 	}
 
-	hchk, err := config.clientCompute.HttpHealthChecks.Get(
-		project, d.Id()).Do()
+	url, err := replaceVars(d, config, "https://www.googleapis.com/compute/v1/projects/{{project}}/global/httpHealthChecks/{{name}}")
 	if err != nil {
-		return handleNotFoundError(err, d, fmt.Sprintf("HTTP Health Check %q", d.Get("name").(string)))
+		return err
```
rosbo: We should update our magic module to handle 404 in delete properly

danawillow: We don't usually do 404 checking in deletes, since hypothetically the refresh would have caught it and removed the resource from state. Do you have an example resource that does?

rosbo: this resource, look at the diff

danawillow: It's diffing it against read() because the order of the functions changed

rosbo: :)

danawillow:
Pushed a quick commit with some fixes that came up in the test run. All `TestAccCompute` tests are passing now. Going to go ahead and merge because the fixes are just putting back things that were accidentally left out.

Issue #1179, opened 2018-03-11
Support for new Cloud Functions Storage trigger events


ssaqua:
The current `providers/cloud.storage/eventTypes/object.change` trigger event type is marked as legacy with new finer-grained storage triggers available:

- `google.storage.object.finalize`
- `google.storage.object.delete`
- `google.storage.object.archive`
- `google.storage.object.metadataUpdate`

https://cloud.google.com/functions/docs/calling/storage#legacy_cloud_storage_triggers

After February 28, the legacy event type will only receive the object finalize event. It would be great to get support for these new event types although I suspect it would need a new option on the cloudfunctions resource.

Snippets from GCP email I received highlighting these changes:

> New more powerful storage triggers
We've heard your feedback and have introduced finer grained storage triggers that allow developers to invoke functions on only a subset of Cloud Storage change events as opposed to responding to all changes on a bucket. For example, you can now write a Cloud Function that only executes when a file is created or one that only executes when a file is deleted. See our documentation to learn more about how to use these new triggers. 

> Google Cloud Storage triggers
Google Cloud storage triggers will use the new finer grained event types mentioned above. As a result, the format of the events will change, and the behavior of the 'gcloud beta functions deploy --trigger-bucket' command will be changing shortly to use the new triggers. After February 28, 2018, deploying a new function with this command will no longer cause it to receive Cloud Storage object change notifications. Instead, it will cause your function to only trigger on OBJECT_FINALIZE events--when a new file is created.

> What happens to my running functions?
Your running functions will not be affected by either of these changes at this time. They will continue to receive events as usual and with the current schema. These older, legacy event types, however, are no longer being actively supported and will remain in beta indefinitely.

> What if I want to update one of my functions that is using a legacy event type?
After February 28, 2018, attempting to update a function that is using a legacy event type via the --trigger-bucket or --trigger-topic flag in gcloud will not succeed. If you want to update one of your existing functions, we have preserved the legacy Cloud Storage and legacy Pub/Sub events. These will allow you to continue to maintain your code even after the gcloud SDK behavior is changed to take advantage of the new Cloud Storage and Cloud Pub/Sub events.

### Terraform Version

`v0.11.3`

### Affected Resource(s)

- google_cloudfunctions_function

sudomilk:
This would be a breaking change in syntax compatibility, but perhaps something like this would be the proper way of doing this?

```HCL
resource "google_storage_bucket" "example_trigger_bucket" {
[...]

PR #1180, opened 2018-03-11
Update to Cloud Functions Pub/Sub trigger event

ssaqua:
The current `providers/cloud.pubsub/eventTypes/topic.publish` trigger event is marked as a legacy event type. There is no issue using it at the moment but is stated that it could be removed at a future date.

https://cloud.google.com/functions/docs/calling/pubsub#legacy_cloud_pubsub_triggers

There are some changes to the event object format that comes through on Pub/Sub messages. Although this doesn't matter for terraform, one thing to note is that functions that have been deployed using the legacy event type (before Feb 28) need to be deleted and redeployed in order to support the new event type. Any new deployments that occur after Feb 28 will use the new event type.

I've created an issue separately for updates to Cloud Storage event triggers: #1179

Here are some snippets from a GCP email I received highlighting these changes:

> Google Cloud Pub/Sub event format
Google Cloud Pub/Sub events will move to a new format to better and more logically organize related attributes. Specifically, all metadata such as [eventId] and [timestamp] will be moved exclusively into the context property of the event parameter. See our documentation on how to use the new events. After February 28, 2018, deploying a new function with the 'gcloud beta functions deploy --trigger-topic' command will cause it to receive this new format.

> What happens to my running functions?
Your running functions will not be affected by either of these changes at this time. They will continue to receive events as usual and with the current schema. These older, legacy event types, however, are no longer being actively supported and will remain in beta indefinitely.

> What if I want to update one of my functions that is using a legacy event type?
After February 28, 2018, attempting to update a function that is using a legacy event type via the --trigger-bucket or --trigger-topic flag in gcloud will not succeed. If you want to update one of your existing functions, we have preserved the legacy Cloud Storage and legacy Pub/Sub events. These will allow you to continue to maintain your code even after the gcloud SDK behavior is changed to take advantage of the new Cloud Storage and Cloud Pub/Sub events.

### DIFF
--- a/google/resource_cloudfunctions_function.go
+++ b/google/resource_cloudfunctions_function.go
@@ -267,7 +267,7 @@ func resourceCloudFunctionsCreate(d *schema.ResourceData, meta interface{}) erro
 		// Make PubSub event publish as in https://cloud.google.com/functions/docs/calling/pubsub
 		function.EventTrigger = &cloudfunctions.EventTrigger{
 			// Other events are not supported
-			EventType: "providers/cloud.pubsub/eventTypes/topic.publish",
+			EventType: "google.pubsub.topic.publish",
 			// Must be like projects/PROJECT_ID/topics/NAME
 			// Topic must be in same project as function
 			Resource: fmt.Sprintf("projects/%s/topics/%s", project, v.(string)),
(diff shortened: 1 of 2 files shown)


ndmckinley:
Ah, bummer, looks like the pubsub test doesn't pass:

```
------- Stdout: -------
=== RUN   TestAccCloudFunctionsFunction_pubsub
--- FAIL: TestAccCloudFunctionsFunction_pubsub (27.56s)
    testing.go:513: Step 0 error: Check failed: Check 4/7 error: Expected cloud.pubsub EventType, found google.pubsub.topic.publish
FAIL
```

Can you fix that up?

ndmckinley:
Re-running tests.

ndmckinley:
Tests look good!  Waiting on Travis now.

PR #1181, opened 2018-03-12
Add support for regional GKE clusters in google_container_cluster

ashish-amarnath:
This is a work in progress PR.
An initial set of changes to share and get feedback on the approach.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -511,19 +511,16 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 	}
 
 	if v, ok := d.GetOk("additional_zones"); ok {
-		locationsList := v.(*schema.Set).List()
-		locations := []string{}
-		for _, v := range locationsList {
-			loc := v.(string)
-			locations = append(locations, loc)
-			if loc == location {
-				return fmt.Errorf("additional_zones should not contain the original 'zone'")
-			}
+		locationsSet := v.(*schema.Set)
+		if locationsSet.Contains(location) {
+			return fmt.Errorf("additional_zones should not contain the original 'zone'")
 		}
 		if isZone(location) {
-			locations = append(locations, location)
+			// GKE requires a full list of locations (including the original zone),
+			// but our schema only asks for additional zones, so append the original.
+			locationsSet.Add(location)
 		}
-		cluster.Locations = locations
+		cluster.Locations = convertStringSet(locationsSet)
 	}
 
 	if v, ok := d.GetOk("cluster_ipv4_cidr"); ok {
@@ -730,14 +727,8 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 
 	d.Set("zone", cluster.Zone)
 
-	locations := []string{}
-	if len(cluster.Locations) > 1 {
-		for _, location := range cluster.Locations {
-			if location != cluster.Zone {
-				locations = append(locations, location)
-			}
-		}
-	}
+	locations := schema.NewSet(schema.HashString, convertStringArrToInterface(cluster.Locations))
+	locations.Remove(cluster.Zone) // Remove the original zone since we only store additional zones
 	d.Set("additional_zones", locations)
 
 	d.Set("endpoint", cluster.Endpoint)
@@ -1010,11 +1001,10 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 		if azSet.Contains(location) {
 			return fmt.Errorf("additional_zones should not contain the original 'zone'")
 		}
-		azs := convertStringArr(azSet.List())
-		locations := append(azs, location)
+		azSet.Add(location)
 		req := &container.UpdateClusterRequest{
 			Update: &container.ClusterUpdate{
-				DesiredLocations: locations,
+				DesiredLocations: convertStringSet(azSet),
 			},
 		}
 
@@ -1024,8 +1014,7 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 			return err
 		}
 
-		log.Printf("[INFO] GKE cluster %s locations have been updated to %v", d.Id(),
-			locations)
+		log.Printf("[INFO] GKE cluster %s locations have been updated to %v", d.Id(), azSet.List())
 
 		d.SetPartial("additional_zones")
 	}
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -1126,7 +1126,7 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 
 			// Wait until it's updated
 			err = containerSharedOperationWait(config, op, project, location, "updating GKE cluster network policy", timeoutInMinutes, 2)
-			log.Println("[DEBUG] emilyye done updating network_policy")
+			log.Println("[DEBUG] done updating network_policy")
 			return err
 		}
 
@@ -1545,11 +1545,19 @@ func resourceContainerClusterStateImporter(d *schema.ResourceData, meta interfac
 
 	switch len(parts) {
 	case 2:
-		d.Set("zone", parts[0])
+		if loc := parts[0]; isZone(loc) {
+			d.Set("zone", loc)
+		} else {
+			d.Set("region", loc)
+		}
 		d.Set("name", parts[1])
 	case 3:
 		d.Set("project", parts[0])
-		d.Set("zone", parts[1])
+		if loc := parts[1]; isZone(loc) {
+			d.Set("zone", loc)
+		} else {
+			d.Set("region", loc)
+		}
 		d.Set("name", parts[2])
 	default:
 		return nil, fmt.Errorf("Invalid container cluster specifier. Expecting {zone}/{name} or {project}/{zone}/{name}")
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -464,6 +464,11 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		return err
 	}
 
+	location, err := getLocation(d, config)
+	if err != nil {
+		return err
+	}
+
 	clusterName := d.Get("name").(string)
 
 	cluster := &containerBeta.Cluster{
@@ -505,26 +510,18 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		}
 	}
 
-	var location string
-	locations := []string{}
-	if regionName, isRegionalCluster := d.GetOk("region"); !isRegionalCluster {
-		location, err = getZone(d, config)
-		if err != nil {
-			return err
-		}
-		locations = append(locations, location)
-	} else {
-		location = regionName.(string)
-	}
-
 	if v, ok := d.GetOk("additional_zones"); ok {
 		locationsList := v.(*schema.Set).List()
+		locations := []string{}
 		for _, v := range locationsList {
 			loc := v.(string)
 			locations = append(locations, loc)
 			if loc == location {
 				return fmt.Errorf("additional_zones should not contain the original 'zone'")
 			}
+		}
+		if isZone(location) {
+			locations = append(locations, location)
 		}
 		cluster.Locations = locations
 	}
@@ -671,7 +668,7 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 			op, err = config.clientContainer.Projects.Zones.Clusters.NodePools.Delete(
 				project, location, clusterName, "default-pool").Do()
 		case v1beta1:
-			parent := fmt.Sprintf("projects/%s/locations/%s/clusters/%s/nodePools/%s", project, location, clusterName, "default-pool")
+			parent := fmt.Sprintf("%s/nodePools/%s", containerClusterFullName(project, location, clusterName), "default-pool")
 			op, err = config.clientContainerBeta.Projects.Locations.Clusters.NodePools.Delete(parent).Do()
 		}
 		if err != nil {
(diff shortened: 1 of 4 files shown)


google/resource_container_cluster.go
```diff
@@ -393,7 +412,13 @@ func resourceContainerCluster() *schema.Resource {
 }
 
 func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) error {
-	containerApiVersion := getContainerApiVersion(d, ContainerClusterBaseApiVersion, ContainerClusterVersionedFeatures)
+	_, isRegionalCluster := d.GetOk("region")
+	if isRegionalCluster {
```
danawillow: This block isn't necessary- you can just add the feature to the var as defined at the top of the file. The name of it should be the name of the field in the schema, so "region"

google/resource_container_cluster.go
```diff
@@ -447,19 +472,35 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 	}
 
 	if v, ok := d.GetOk("additional_zones"); ok {
+		if isRegionalCluster {
+			return fmt.Errorf("'additional_zones' is incompatible with regional clusters")
```
danawillow: This should be doable with `ConflictsWith` in the schema

google/resource_container_cluster.go
```diff
@@ -94,6 +94,25 @@ func resourceContainerCluster() *schema.Resource {
...
+			},
+
+			"node_locations": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"num_nodes": {
```
danawillow: This doesn't seem to get used anywhere

google/resource_container_cluster.go
```diff
@@ -447,19 +472,35 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
...
+				return fmt.Errorf("additional_zones should not contain the original 'zone'")
 			}
 		}
 		locations = append(locations, zoneName)
 		cluster.Locations = locations
 	}
 
+	if isRegionalCluster {
+		locations := []string{}
+		locations = append(locations, d.Get("region").(string))
```
danawillow: I think the locations field is still supposed to be just a list of zones- this will likely still work without this entire block (and the `node_locations` field)

google/resource_dataproc_cluster.go
```diff
@@ -797,7 +797,7 @@ func flattenPreemptibleInstanceGroupConfig(d *schema.ResourceData, icg *dataproc
 func flattenInstanceGroupConfig(d *schema.ResourceData, icg *dataproc.InstanceGroupConfig) []map[string]interface{} {
 	disk := map[string]interface{}{}
 	data := map[string]interface{}{
-	//"instance_names": []string{},
+		//"instance_names": []string{},
```
danawillow: this can be reverted, it's not really relevant here

google/resource_container_cluster.go
ashish-amarnath: I see that this is not a new field added to support regional clusters. So this is not relevant to this PR. Removing this. But we'll have to add this later.

google/resource_container_cluster.go
ashish-amarnath: Got it! Thanks.

google/resource_container_cluster.go
ashish-amarnath: Thank you. That should simplify this.

google/resource_container_cluster.go
ashish-amarnath: I see that this is not a new field added to support regional clusters. So this is not relevant to this PR. Removing this. But we'll have to add this later. As `additional_zones` is getting deprecated.

google/resource_dataproc_cluster.go
ashish-amarnath: Reverted.

ashish-amarnath:
@danawillow  I took another stab at this and I need some help to update the `vendor/google.golang.org/api/container/v1beta1/container-gen.go` file with a field for `region` and also use the right API to create regional clusters so that the unit test can pass.

danawillow:
The API as it is in the vendor directory should work. Instead of `.Projects.Zones` you'll want`.Projects.Locations` and then call the API as documented at https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters/create
(actual code at https://github.com/terraform-providers/terraform-provider-google/blob/master/vendor/google.golang.org/api/container/v1beta1/container-gen.go#L3623)

ashish-amarnath:
Thanks for those pointers. I should be able to make more progress with this.

ashish-amarnath:
The test is now able to create regional GKE clusters.
`POST /v1beta1/projects/nordstrom-k8s-dev/locations/us-central1/clusters?alt=json HTTP/1.1
Host: container.googleapis.com
User-Agent: google-api-go-client/0.5 (darwin amd64) Terraform/0.11.3-dev
Content-Length: 464
Content-Type: application/json
Accept-Encoding: gzip`
this request is successful. The test however, still fails when trying to 
`---[ REQUEST ]---------------------------------------
GET /v1/projects/nordstrom-k8s-dev/zones//operations/operation-1521589157203-edadd8b8?alt=json HTTP/1.1
Host: container.googleapis.com
User-Agent: google-api-go-client/0.5 (darwin amd64) Terraform/0.11.3-dev
Accept-Encoding: gzip`

This is most likely because the support is only implemented to create regional clusters but not to update, read or destroy. Next step: Make this work!

Stono:
@ashish-amarnath I just wanted to say thanks for your efforts with this, I don't know enough golang to help out with the implementation but if i (we - people on my team) can help with testing in any way just ask

ashish-amarnath:
@Stono Thanks for offering to help! I will surely ask for help validating this change. Right now I am trying to a checkpoint where this is can spin-up and teardown regional clusters. This is not the top priority for us right now so I am trying to backfill my time working on this. For that reason, and my unfamiliarity with this codebase are the reasons for the slow progress.  I am not blocked. @danawillow has been super helpful in unblocking me :)

ashish-amarnath:
Made more progress on this. I now have some code and unit test that creates and destroys regional GKE clusters. Next, I need to work on the Update operation and then there are merge conflicts that I have to resolve, of course.

ashish-amarnath:
@danawillow this PR is ready for more feedback. I need to add some more tests. Suggestions are welcome :)
@Stono  Feel free to take a look and make comments.

google/resource_container_cluster.go
```diff
@@ -594,26 +609,31 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
...
 		err = Convert(req, reqV1Beta)
 		if err != nil {
 			return err
 		}
-		op, err = config.clientContainerBeta.Projects.Zones.Clusters.Create(
-			project, zoneName, reqV1Beta).Do()
-		if err != nil {
-			return err
+
+		if isRegionalCluster {
```
danawillow: Since the location could also be the zone, I don't think we actually need this conditional- it should be fine to just always do the `Projects.Locations` call

google/resource_container_cluster.go
```diff
@@ -637,43 +657,49 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
...
+	if !isRegionalCluster {
+		zoneName, err = getZone(d, config)
+		if err != nil {
+			return err
+		}
+		locations = append(locations, zoneName)
 	}
 
 	cluster := &containerBeta.Cluster{}
+	var clust interface{}
```
danawillow: Did this work for you when you tested it? I've tried this approach (just doing the `Convert` from an interface at the end) in the past and it failed for me because Convert needed the parameter passed to be a struct of a certain type (so it can recurse through the fields).

google/resource_container_cluster.go
```diff
@@ -785,28 +815,49 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
...
 
+	var req *container.UpdateClusterRequest
+	var updateDescription string
+	updateF := func() error {
+		var op interface{}
+		location := zoneName
+		switch containerAPIVersion {
+		case v1:
+			op, err = config.clientContainer.Projects.Zones.Clusters.Update(
+				project, zoneName, clusterName, req).Do()
```
danawillow: This change makes me nervous- it feels a bit brittle to me to have this depending on what the value of req is at the time this function gets called. If you'd like to refactor it out, I think wrapping it in another function that takes in req as a parameter and returns this func would be a safer choice.

google/resource_container_cluster.go
```diff
@@ -1001,15 +1031,33 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 		}
 
 		updateF := func() error {
-			log.Println("[DEBUG] emilyye updating enable_legacy_abac")
-			op, err := config.clientContainer.Projects.Zones.Clusters.LegacyAbac(project, zoneName, clusterName, req).Do()
+			log.Println("[DEBUG] updating enable_legacy_abac")
```
danawillow: thanks :)

google/resource_container_cluster_test.go
```diff
@@ -243,6 +243,23 @@ func TestAccContainerCluster_withMasterAuthorizedNetworksConfig(t *testing.T) {
...
+	clusterName := fmt.Sprintf("cluster-test-regional-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckContainerClusterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccContainerCluster_regional(clusterName),
+			},
```
danawillow: Can you add another test step that does an import? There are a bunch of examples elsewhere in this file.

google/resource_container_cluster.go
```diff
@@ -485,17 +490,26 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		}
 	}
 
+	var zoneName string
+	locations := []string{}
+	regionName, isRegionalCluster := d.GetOk("region")
+	if !isRegionalCluster {
```
danawillow: Here's an idea- what if instead of separate "region" and "zone" fields, we just had a "location" field? Then once regional clusters have gone GA we can deprecate zone. I'm not totally set on this, but it does seem like it would simplify the logic. What do you think?

google/resource_container_cluster.go
ashish-amarnath: That made sense. But didn't what that change to have unknown side-effects. I will give that a try.

google/resource_container_cluster.go
ashish-amarnath: Well.. It compiled right and the tests passed. From that, I figured it works. Any other tests I can run to be doubly sure?

google/resource_container_cluster.go
ashish-amarnath: I agree. For that exact reason, I tried going the function route but quickly realised that that number of args to that function became too many and that made me a little uncomfortable. That function would need `config, project, zone, region, isRegionalCluster, request, clusterName, timeout, apiVersion`.
Total of 9 args. But, collapsing zone and region into `location` that number could go down to 8.

google/resource_container_cluster_test.go
ashish-amarnath: Will do.

google/resource_container_cluster.go
ashish-amarnath: Thought about that. But because that would increase the surface area of this change I decided against it. I'll do this if that is not too bad. If that makes sense :)

google/resource_container_cluster.go
ashish-amarnath: Fixed.

google/resource_container_cluster.go
ashish-amarnath: @danawillow What did you to break this? I'd like to make sure this will not break later.

danawillow: Looks like it works! I must have done something slightly differently when I tried doing it this way in the past.

google/resource_container_cluster.go
danawillow: Ah yeah, I was more just thinking that only `req` would be the parameter since it's the only thing that would change between requests.

google/resource_container_cluster.go
danawillow: Yup, that's fair. I do think that having it as two separate fields probably makes it a bit clearer to users what to put, so I'm fine with that.

ashish-amarnath:
I've had other things taking up my time and haven't had a chance to work on this for a while. Just so you don't think I've vanished off the face of the Earth!

Stono:
Hey, 
If there is anything we can do to help progress this please let us know!

danawillow:
@ashish-amarnath, I think there were just a few small changes that were remaining, right? I don't mind making them and pushing them to your branch if you're all right with that. Seems like people are pretty eager for this :)

@drubin, we usually aim to release every 2 weeks or so, with one aimed for the next few days. So if this gets in today it'll be in the next release, if not probably not. Even though I think the schema for this is set, I'm also pretty hesitant about solution 1. If this doesn't make it into the next release, then I think solution 3 will be totally fine.

ashish-amarnath:
I haven't had the time to pick-up where I left off on this.
@danawillow If you want to pick this up I would totally be OK with it. I don't want to hold this up for everyone. The last bit that was left on this was adding a test step that does an import. I remember taking an initial stab at it but somewhere while executing the test it would fail trying to use a v1 API to access the test cluster. I had to figure out where and why and then fix it.

danawillow:
hey @rosbo, since the changes I made are non-trivial I wouldn't mind having you look at this as well if you don't mind

google/resource_container_cluster.go
```diff
@@ -504,13 +514,15 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		locationsList := v.(*schema.Set).List()
 		locations := []string{}
 		for _, v := range locationsList {
-			location := v.(string)
-			locations = append(locations, location)
-			if location == zoneName {
-				return fmt.Errorf("additional_zones should not contain the original 'zone'.")
+			loc := v.(string)
```
rosbo: Consider simplifying this by using convertStringSet and schema.Set#Contains.

Also I am not sure why we have this block below:

```go
if isZone(location) {
   locations = append(locations, location)
}
```
Why are we adding the "main" zone to the list of additional zones... Consider adding a comment to clarify this.

PR #1188, opened 2018-03-13
Force validation of region for cloudfunctions.

### LINKED ISSUE #1173: 400 Error when creating Google Cloud Function
_This issue was originally opened by @kaifresh as hashicorp/terraform#17522. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

```
v0.11.3
```

```hcl
resource "google_storage_bucket" "bucket" {
  name = "testing-cf-bucket"
}

resource "google_storage_bucket_object" "archive" {
  name   = "index.zip"
  bucket = "${google_storage_bucket.bucket.name}"
  source = "./index.zip"
}

resource "google_cloudfunctions_function" "function" {
  name                  = "function-test"
  description           = "Testing the 400 error when uploading a function"
  available_memory_mb   = 128
  source_archive_bucket = "${google_storage_bucket.bucket.name}"
  source_archive_object = "${google_storage_bucket_object.archive.name}"
  trigger_http          = true
#   timeout               = 60
  entry_point           = "helloHttp"
#   labels {
#     my-label = "my-label-value"
#   }
}

...
```

### Debug Output

Error: Error applying plan:

1 error(s) occurred:

* google_cloudfunctions_function.function: 1 error(s) occurred:

* google_cloudfunctions_function.function: googleapi: Error 400: The request has errors, badRequest

Or from the debug output 
`
 {
   "error": {
     "code": 400,
     "message": "The request has errors",
     "errors": [
       {
         "message": "The request has errors",
         "domain": "global",
        "reason": "badRequest"
       }
     ],
     "status": "INVALID_ARGUMENT"
   }
 }
`

### Expected Behavior

The cloud function should have uploaded correctly to GCP. I have used the same zipped code from the google_storage_bucket_object to create a cloud function via the GCP Console, and can do so successfully. So we know that the source code works.

### Actual Behavior

Upon running `terraform apply`, a (400) error message is returned from the Google API saying that the cloud function could not be created. `( googleapi: Error 400: The request has errors, badRequest )`

### Steps to Reproduce

1. Paste the config above into a terraform file, then run:
2. `terraform init`
3. `terraform apply`

ndmckinley:
This fixes #1173.  The issue is that neither the user nor I knew that `australia-southeast1` is not a valid region for cloud functions, which meant that the 400 error they got back was confusing to both of us (and my lack of repro, since my test cases run in `us-central1`, was doubly confusing).  Since `ValidateFunc`s don't run on `d.Set()`, calling the function explicitly is the way to go to make this error obvious right away.

I tested this locally with the user's repro case, and the error is clear and helpful.

### DIFF
--- a/google/resource_cloudfunctions_function.go
+++ b/google/resource_cloudfunctions_function.go
@@ -26,6 +26,10 @@ var functionAllowedMemory = map[int]bool{
 	2048: true,
 }
 
+// For now CloudFunctions are allowed only in us-central1
+// Please see https://cloud.google.com/about/locations/
+var validCloudFunctionRegion = validation.StringInSlice([]string{"us-central1"}, true)
+
 const functionDefaultAllowedMemoryMb = 256
 
 type cloudFunctionId struct {
@@ -201,13 +205,11 @@ func resourceCloudFunctionsFunction() *schema.Resource {
 			},
 
 			"region": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
-				ForceNew: true,
-				// For now CloudFunctions are allowed only in us-central1
-				// Please see https://cloud.google.com/about/locations/
-				ValidateFunc: validation.StringInSlice([]string{"us-central1"}, true),
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
+				ValidateFunc: validCloudFunctionRegion,
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


google/resource_cloudfunctions_function.go
```diff
@@ -225,6 +227,12 @@ func resourceCloudFunctionsCreate(d *schema.ResourceData, meta interface{}) erro
 	if err != nil {
 		return err
 	}
+	// We do this extra validation here since most regions are not valid, and the
```
rosbo: I see, the provider-level region skips the field validation. May be add a mention about this here?

ndmckinley:
Thanks, rerunning tests to be double-sure.

Issue #843, opened 2017-12-12
Add Bucket IAM policy resource


rolandkool:
Hi,

Version 1.4.0 comes with support for Bucket IAM member and binding, which is great! It would be nice to also have the policy variant where you apply one policy to a bucket overriding all other applied policies. Similar to what you can do on a project with the google_project_iam_policy.

Our use case requires that we don't want allow to any manual changes to policies so if one is to manually change the buckets IAM it would be reverted with a terraform run.

Thanks

rosbo:
Hi @rolandkool 

The reason I did not add a `google_storage_bucket_iam_policy` resource is because there is a bunch of legacy roles that are automatically added when creating a policy.

There is different ways we can handle these:
1. Always ignore these roles. You cannot set or remove them when using the authoritative iam_policy resource.
2. A user has to specify these legacy roles in their config otherwise they will get deleted.

The main advantage of 2 is that users can manage legacy roles if they want to. The main disadvantage is that it is tiny bit more verbose to keep them.

I personally tend to prefer 2. I am curious about your opinion on the subject.

Thank you

rolandkool:
Hi @rosbo 

For me, option 2 is fine and expected. For default rules we want to keep, we will just make sure to add them in the policy as well. For project policy bindings, we are doing the same.

paddycarver:
I believe this was addressed by #1190. If not, feel free to comment, and we can reopen this!

PR #1190, opened 2018-03-14
#843: Add policy support to storage buckets

ReneKroon:
#843 is fixed by this line, after the consensus that legacy roles should not be implicitly supported.

Since this is my first commit to the project, i'd like to have some pointers on testing. For example this change only wraps existing policy functionality on a bucket. Does this require a separate testcase anyway? 

Since using this feature could remove legacy rows, how would a documentation update for this change work?

Thanks in advance,

Rene Kroon


ndmckinley:
You'll need to add documentation for this under `website/docs/r/storage_bucket_iam.html.markdown`: https://github.com/terraform-providers/terraform-provider-google/blob/master/website/docs/r/storage_bucket_iam.html.markdown

Please be sure to explicitly call out the legacy roles that might be auto-removed!

As for testing, let me look into that and get back to you.

ndmckinley:
You'll just need to add a simple test to https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_storage_bucket_iam_test.go - similar to what's already there.

ReneKroon:
@ndmckinley I adressed the documentation and test. It also explains what happens if the legacy roles are removed and how that can be restored. The test is alike to IAM Binding. As far as i can tell this policy is now on par with the other two methods of assigning permissions.

website/docs/r/storage_bucket_iam.html.markdown
```diff
@@ -8,10 +8,11 @@ description: |-
 
 # IAM policy for Google storage bucket
 
-Two different resources help you manage your IAM policy for storage bucket. Each of these resources serves a different use case:
+Three different resources help you manage your IAM policy for storage bucket. Each of these resources serves a different use case:
 
 * `google_storage_bucket_iam_binding`: Authoritative for a given role. Updates the IAM policy to grant a role to a list of members. Other roles within the IAM policy for the storage bucket are preserved.
 * `google_storage_bucket_iam_member`: Non-authoritative. Updates the IAM policy to grant a role to a new member. Other members for the role for the storage bucket are preserved.
+* `google_storage_bucket_iam_policy`: Authoritative for a given policy. Resets any other permission on the bucket, including default Google permissions that you inherit upon bucket creation. See the usage example on how to handle this.
```
ndmckinley: Could you make this a little bit more cautious?  "Setting a policy removes all other permissions on the bucket, and if done incorrectly, there's a real chance you will lock yourself out of the bucket.  If possible for your use case, using multiple `google_storage_bucket_iam_binding` resources will be much safer.

ndmckinley:
Bummer: 
```

------- Stdout: -------
=== RUN   TestAccStorageBucketIamPolicy
--- FAIL: TestAccStorageBucketIamPolicy (0.01s)
    testing.go:513: Step 0 error: Error loading configuration: Error loading /opt/teamcity-agent/temp/buildTmp/tf-test827642825/main.tf: Error reading config for google_storage_bucket_iam_policy[foo]: google_storage_bucket.tf-test-3276351694613004242: resource variables must be three parts: TYPE.NAME.ATTR in:
        
        ${google_storage_bucket.tf-test-3276351694613004242}
FAIL
```

ReneKroon:
I updated the readme, and i think i found the issue with the test. When running the PR against real google cloud it works, but i have not yet replicated the acceptance test suite environment.

ndmckinley:
I'll give it another run.  :)

google/resource_storage_bucket_iam_test.go
```diff
@@ -86,6 +115,71 @@ func testAccCheckGoogleStorageBucketIam(bucket, role string, members []string) r
...
+    role = "roles/storage.objectViewer"
+
+    members = [
+      "serviceAccount:${google_service_account.test-account-1.email}",
+    ]
+  }
+}
+
+resource "google_storage_bucket_iam_policy" "fooPolicy" {
+  bucket      = "${google_storage_bucket.%s}"
```
ndmckinley: This is the problem - you'll need to append `.name`.  You can't use the entire resource here, you need to access the right attribute on the resource, which is `.name`.

ndmckinley:
Sorry, no dice:
```
------- Stdout: -------
=== RUN   TestAccStorageBucketIamPolicy
--- FAIL: TestAccStorageBucketIamPolicy (0.01s)
    testing.go:513: Step 0 error: Error loading configuration: Error loading /opt/teamcity-agent/temp/buildTmp/tf-test827642825/main.tf: Error reading config for google_storage_bucket_iam_policy[foo]: google_storage_bucket.tf-test-3276351694613004242: resource variables must be three parts: TYPE.NAME.ATTR in:
        
        ${google_storage_bucket.tf-test-3276351694613004242}
FAIL
```

I commented the problem in a review.

google/resource_storage_bucket_iam_test.go
```diff
@@ -144,7 +144,7 @@ data "google_iam_policy" "fooPolicy" {
 }
 
 resource "google_storage_bucket_iam_policy" "foo" {
-  bucket      = "${google_storage_bucket.%s}"
+  bucket      = "${google_storage_bucket.%s.name}"
```
ndmckinley: This still isn't quite right - you just want "${google_storage_bucket.bucket.name}", right?  You just want the name of that bucket?

ndmckinley: `%s` here is the name of the bucket (the value of `google_storage_bucket.bucket.name`), right?

ReneKroon:
```
resource "google_storage_bucket" "bolcom-sbx-kroonapp-010-kroon" {
  name          = "bolcom-sbx-kroonapp-010-kroon"
  project       = "bolcom-sbx-kroonapp-010"
  storage_class = "REGIONAL"
  force_destroy = false

  labels = {
    opex            = "teamprovisioning"
    stores_pii_data = "false"
  }

  versioning = {
    enabled = false
  }
}

data "google_iam_policy" "kroontest" {
  binding {
    role = "roles/storage.objectAdmin"

    members = [
      "user:rkroon@bol.com",
    ]
  }
}

resource "google_storage_bucket_iam_policy" "kroontest" {
  bucket      = "${google_storage_bucket.bolcom-sbx-kroonapp-010-kroon.name}"
  policy_data = "${data.google_iam_policy.kroontest.policy_data}"
}
```

This is the generated terraform file that works for me.

ndmckinley: Right, because the bucket resource is called "bolcom-sbx-kroonapp-010-kroon", but that's also its name attribute.  In the test, the bucket resource is called "bucket", but its name attribute is a random string.  You want to access that random string here.  I'm rerunning the tests now.

ndmckinley:
```
------- Stdout: -------
=== RUN   TestAccStorageBucketIamPolicy
--- FAIL: TestAccStorageBucketIamPolicy (0.00s)
    testing.go:513: Step 0 error: config is invalid: resource 'google_storage_bucket_iam_policy.fooPolicy' config: unknown resource 'google_storage_bucket.tf-test-8494430595842574891' referenced in variable google_storage_bucket.tf-test-8494430595842574891.name
FAIL
```

The bucket's resource name isn't the same as the bucket name, which is the problem.  Sorry!

ReneKroon:
@ndmckinley I have run the integration tests myself. It turns out that removing all policies (i.e. setting an empty iam policy) is non trivial. I found that gsutil is able to do this by just setting the Etag value and i copied this behaviour.

ndmckinley:
I'll run the tests now.

ndmckinley:
Sorry:

```

------- Stdout: -------
=== RUN   TestAccStorageBucketIamPolicy
--- FAIL: TestAccStorageBucketIamPolicy (2.67s)
    testing.go:513: Step 0 error: Error applying: 1 error(s) occurred:
        
        * google_storage_bucket_iam_policy.bucketBinding: 1 error(s) occurred:
        
        * google_storage_bucket_iam_policy.bucketBinding: Error retrieving IAM policy for Storage Bucket "tf-test-5427762820494880885": googleapi: Error 403: terraform-acceptance-tests@hc-terraform-testing.iam.gserviceaccount.com does not have storage.buckets.getIamPolicy access to tf-test-5427762820494880885., forbidden
    testing.go:573: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.
        
        Error: Error refreshing: 1 error(s) occurred:
        
        * google_storage_bucket.bucket: 1 error(s) occurred:
        
        * google_storage_bucket.bucket: google_storage_bucket.bucket: Error reading Storage Bucket "tf-test-5427762820494880885": googleapi: Error 403: terraform-acceptance-tests@hc-terraform-testing.iam.gserviceaccount.com does not have storage.buckets.get access to tf-test-5427762820494880885., forbidden
        
        State: <nil>
FAIL
```

If I'm reading that right, I think you can configure it when you create the bucket.

ReneKroon:
@ndmckinley  I ran the test in a separate cloud, so i ran under 'Owner' permissions. What user can i expect for these integration tests? It probably means i have to set up roles and permissions for the service account user first which makes the test config a dependant on other parts of the application as well.

paddycarver:
I'm not following here, sorry--why should a contributor have to add our service account to anything? Our service account should have all the permissions it needs when running tests against our CI environment. If it doesn't, I would say that's a problem with the CI setup, not with the PR, and if we think it's a permissions problem, I can investigate the CI setup and make sure things are right.

ReneKroon:
i run with the acceptance with the following environment variables set:

`GOOGLE_PROJECT=lustrous-fabric-165106`
`GOOGLE_REGION=us-central1`
`GOOGLE_USE_DEFAULT_CREDENTIALS=1`
`TF_ACC=1`
`TF_LOG=TRACE`
`TF_LOG_PATH=/tmp/tf.log`
`TF_LOG_PATH_MASK=/tmp/%s`

You are right that there it's non standard to delete the bucket which has it's policy removed. However it's not impossible, but i can imagine that my use of GOOGLE_USE_DEFAULT_CREDENTIALS which makes me owner on the project has some effect.

ndmckinley:
That's a good idea, thanks.  @ReneKroon - sorry to keep asking for more here.  I'm happy to add that if you'd prefer not to, you've definitely spent enough time on this already, just let me know.

ReneKroon:
@ndmckinley Well i've learned at least to manage the full development cycle with this change :-) I would like to ask you to make this CI related adjustment. It will be a lot of roundtrips again if something would not work right away, since i have no insight into your CI infrastructure.

ndmckinley:
Okay, there we go.  @ReneKroon please be aware!  To continue with this PR I had to push to your master branch because that's the branch you used to open the PR.  Because of some merge conflicts, I had to merge master into the PR ... so your master branch just jumped by a couple versions.  If you need your old master branch back, you should be able to do this:

```
git checkout d96138b
git branch -f master
git push origin master -f
```

Of course that'll screw up this PR, but I still have all this stuff on my fork, so you don't need to worry about that - I'll just open a new one if you need that.

ndmckinley:
Test passed, but since my code's in here, I probably ought to get another review on that stuff.  Hang on, I'll get this merged yet.  :)

website/docs/r/storage_bucket_iam.html.markdown
```diff
@@ -38,6 +40,22 @@ resource "google_storage_bucket_iam_member" "member" {
...
+When applying a policy that does not include the roles listed below, you lose the default permissions which google adds to your bucket:
+* `roles/storage.legacyBucketOwner`
+* `roles/storage.legacyBucketReader`
+
+If this happens only an entity with `roles/storage.admin` privileges can repair this bucket's policies. It is recommended to include the above roles in policies to get the same behaviour as with the other two options.
+
+```hcl
+resource "google_storage_bucket_iam_policy" "member" {
+  bucket = "your-bucket-name"
+  policy_data = "policy_data"
```
danawillow: Looks like a string snuck in here- I think you meant to have this refer to a policy data source.

google/resource_storage_bucket_iam_test.go
```diff
@@ -86,6 +116,87 @@ func testAccCheckGoogleStorageBucketIam(bucket, role string, members []string) r
...
+  binding {
+    role = "roles/storage.objectViewer"
+
+    members = [
+      "serviceAccount:${google_service_account.test-account-1.email}",
+      "serviceAccount:${google_service_account.test-account-2.email}",
+    ]
+  }
+
+	binding {
```
danawillow: nit: tabs/spaces consistency

google/resource_storage_bucket_iam_test.go
```diff
@@ -86,6 +116,87 @@ func testAccCheckGoogleStorageBucketIam(bucket, role string, members []string) r
...
+  display_name = "Iam Testing Account"
+}
+
+resource "google_service_account" "test-account-2" {
+  account_id   = "%s-2"
+  display_name = "Iam Testing Account"
+}
+
+
+data "google_iam_policy" "fooPolicy" {
```
danawillow: nit: let's aim for consistency between the naming case for the resources, so this should be foo-policy and then below would be bucket-binding

ndmckinley: Done.

google/resource_storage_bucket_iam_test.go
ndmckinley: Done.

website/docs/r/storage_bucket_iam.html.markdown
ndmckinley: Done.

ndmckinley:
Storage bucket tests all pass.

PR #1192, opened 2018-03-14
Add support for pod security policy

### LINKED ISSUE #1168: Add support for podsecuritypolicies in google_container_cluster
### Terraform Version
```
Terraform v0.11.3
+ provider.google v1.6.0
```

### Affected Resource(s)
- google_container_cluster

### Expected Behavior
This is a feature request to add support for PodSecurityPolicies on creating/updating a container cluster.

### References
This document describes what needs to be done
https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.PodSecurityPolicyConfig

danawillow:
Fixes #1168

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -341,7 +341,7 @@ func resourceContainerCluster() *schema.Resource {
 				Type:     schema.TypeList,
 				Optional: true,
 				MaxItems: 1,
-				// Remove update support for now: https://issuetracker.google.com/74063492
+				// Disable update support for now: https://issuetracker.google.com/74063492
 				ForceNew: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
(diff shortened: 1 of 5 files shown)


google/resource_container_cluster.go
```diff
@@ -1063,6 +1090,32 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 		d.SetPartial("logging_service")
 	}
 
+	// Disable update support for now: https://issuetracker.google.com/74063492
```
ndmckinley: Is this the best way?  I feel like maybe we might want to just use regular git / github for this, pull it out into its own branch, and leave the PR open with a note that it should be merged when the bug is closed.  Commented-out-code makes me nervous.

danawillow: Yeah, that's a better idea. I'll do that now.

PR #1194, opened 2018-03-14
Properly delete dataflow jobs in the event of terraform destroy.

### LINKED ISSUE #1191: Terraform destroy does not stop dataflow jobs
_This issue was originally opened by @kaifresh as hashicorp/terraform#17578. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
Terraform v0.11.3
+ provider.google v1.7.0

```
...
```

```hcl

resource "google_dataflow_job" "data_cleaning" {
  name              = "${var.project-name}-${var.environment}-dataflow-${var.pipeline-name}"
  template_gcs_path = "${local.dataflow_template_gcs_path}"
  temp_gcs_location = "gs://${google_storage_bucket.dataflow-storage-bucket.name}/temp"
  on_delete         = "cancel"
```

### Expected Behavior

When I call `terraform destroy` the dataflow job should be cancelled

### Actual Behavior

Nothing happens. The DataFlow job remains in a Running state.

### Steps to Reproduce

1. Upload your terraform job to a bucket
2. Set variables appropriately
3. `terraform init`
4. `terraform apply`

ndmckinley:
This fixes #1191 - in the event that the job wasn't properly canceled, we'll get an error message, and now we'll actually tear down the jobs.

### DIFF
--- a/google/resource_dataflow_job.go
+++ b/google/resource_dataflow_job.go
@@ -177,16 +177,20 @@ func resourceDataflowJobDelete(d *schema.ResourceData, meta interface{}) error {
 				// If we have an error and it's not a google-specific error, we should go ahead and return.
 				return err
 			} else if err_ok && strings.Contains(gerr.Message, "not yet ready for canceling") {
-				time.Sleep(5 * time.Second)
+				// We'll sleep below to wait for the job to be ready to cancel.
 			} else {
 				return err
 			}
 		}
+
 		err = resourceDataflowJobRead(d, meta)
-		if _, ok := dataflowTerminalStatesMap[d.Get("state").(string)]; !ok {
+		postReadState := d.Get("state").(string)
+		log.Printf("[DEBUG] Job state: '%s'.", postReadState)
+		if _, ok := dataflowTerminalStatesMap[postReadState]; !ok {
+			// If we're not yet in a terminal state, we need to sleep a few seconds so we don't
+			// exhaust our update quota with repeated attempts.
 			time.Sleep(5 * time.Second)
 		}
-		log.Printf("[DEBUG] Job state: '%s'.", d.Get("state").(string))
 		if err != nil {
 			return err
 		}
--- a/google/resource_dataflow_job.go
+++ b/google/resource_dataflow_job.go
@@ -184,9 +184,15 @@ func resourceDataflowJobDelete(d *schema.ResourceData, meta interface{}) error {
 			return err
 		}
 	}
-	d.SetId("")
 
-	return nil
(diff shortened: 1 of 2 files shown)


ndmckinley:
Hm, this makes the test fail.  Interesting!  I wonder if that reveals the problem in there...  Looking into it.

ndmckinley:
Good news!  It *did* reveal that the test was failing to catch this sort of error, and I'm updating the description.

ndmckinley:
The modified test does pass.  :)

google/resource_dataflow_job.go
```diff
@@ -165,28 +166,40 @@ func resourceDataflowJobDelete(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return err
 	}
-	for _, ok := dataflowTerminalStatesMap[d.Get("state").(string)]; ok; _, ok = dataflowTerminalStatesMap[d.Get("state").(string)] {
+	for _, ok := dataflowTerminalStatesMap[d.Get("state").(string)]; !ok; _, ok = dataflowTerminalStatesMap[d.Get("state").(string)] {
```
rosbo: consider extracting `d.Get("state").(string)` to a local variable

google/resource_dataflow_job.go
```diff
@@ -165,28 +166,40 @@ func resourceDataflowJobDelete(d *schema.ResourceData, meta interface{}) error {
...
+				return err
+			} else if err_ok && strings.Contains(gerr.Message, "not yet ready for canceling") {
+				time.Sleep(5 * time.Second)
+			} else {
+				return err
+			}
 		}
 		err = resourceDataflowJobRead(d, meta)
+		if _, ok := dataflowTerminalStatesMap[d.Get("state").(string)]; !ok {
+			time.Sleep(5 * time.Second)
```
rosbo: Can you add a comment explaining why we are waiting here.

ndmckinley: Done.

google/resource_dataflow_job.go
ndmckinley: Done.

danawillow: Looks like you did so below, but not up here

danawillow: Ok I see it'll be different on the first read and on each successive pass, carry on

ndmckinley:
Thanks!  Rerunning tests before merging since a few things have gone in recently.

PR #1195, opened 2018-03-14
Add update support for pod security policy

danawillow:
Continuation of #1192. Test/submit once https://issuetracker.google.com/74063492 is fixed.


google/resource_container_cluster.go
```diff
@@ -1125,6 +1123,31 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
...
+		}
+
+		updateF := func() error {
+			op, err := config.clientContainerBeta.Projects.Zones.Clusters.Update(
+				project, zoneName, clusterName, req).Do()
+			if err != nil {
+				return err
+			}
+			// Wait until it's updated
+			return containerSharedOperationWait(config, op, project, zoneName, "updating GKE cluster pod security policy config", timeoutInMinutes, 2)
```
ndmckinley: Question - I see that this makes a call to change the security config - was this particular part of the update method chosen for a reason?  Are there dependencies up or down - things that this needs to go before or after?  If so, could you add a comment so we don't break it later?

danawillow: Done

PR #1197, opened 2018-03-14
Reduce flakiness by ensuring three successful fetches before returning.

### LINKED ISSUE #1131: add ability to wait for operation completion
### Affected Resource(s)
- all

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

```hcl
esource "google_project_iam_member" "datastore_user" {
  role    = "roles/datastore.user"
  member  = "serviceAccount:${google_service_account.todo_account.email}"
  // Await for 1min before provisioning; would be nice to be able to wait for the actual completion.
  provisioner "local-exec" {
      command = "echo sleep 60s to propagate all permissions; sleep 60"
 }
}

// Create Pod
resource "kubernetes_pod" "test" {
 depends_on = ["google_project_iam_member.datastore_user"]
 count = "${var.mode=="pod" ? 1 : 0}"
 metadata {
   name = "todo-backend"
   namespace = "${kubernetes_namespace.todo_backend.metadata.0.name}"
 }
```

### Expected Behavior
What should have happened? Rather than wait for unknown number of seconds (which is flaky), provide option to wait for operation completion.

### Actual Behavior
What actually happened? Service account provisioning operation is kicked off, but the terraform resource returns immediately as if it's created since it's an async operation. Provide option to wait for operation completion.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Are there anything atypical about your accounts that we should know? N/A

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? N/A

ndmckinley:
Fixes #1131.

### DIFF
--- a/google/iam.go
+++ b/google/iam.go
@@ -2,10 +2,12 @@ package google
 
 import (
 	"fmt"
-	"github.com/hashicorp/terraform/helper/schema"
-	"google.golang.org/api/cloudresourcemanager/v1"
 	"log"
 	"time"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/cloudresourcemanager/v1"
+	"google.golang.org/api/googleapi"
 )
 
 // The ResourceIamUpdater interface is implemented for each GCP resource supporting IAM policy.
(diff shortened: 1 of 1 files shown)


ndmckinley:
There's a serious risk of this making us run out of iam policy fetch quota, making the tests flakier, so most of this new code is to try to keep things on track even when we run out of quota.

google/iam.go
```diff
@@ -64,6 +69,36 @@ func iamPolicyReadModifyWrite(updater ResourceIamUpdater, modify iamPolicyModify
 		log.Printf("[DEBUG]: Setting policy for %s to %+v\n", updater.DescribeResource(), p)
 		err = updater.SetResourceIamPolicy(p)
 		if err == nil {
+			fetchBackoff := 5 * time.Second
```
danawillow: I assume you came to this value from trial and error, right? It feels a bit high since that means that at minimum we've just added 15s to this resource, but I'll defer to your findings

ndmckinley: I can try starting at 1 - increasing the backoff as we run out of quota might work out okay.

ndmckinley:
Yep, that seems to work fine.  :)

PR #1200, opened 2018-03-15
Add network policy config.

### LINKED ISSUE #860: Error when trying to update network_policy on google_container_cluster
When trying to enable network policy support on existing clusters, the following error is returned:

```
* google_container_cluster.cluster: googleapi: Error 400: the network policy addon must be enabled before updating the nodes., badRequest
```

The change we have made to the cluster resource was to add the following block:

```yaml
network_policy {
  enabled  = true
  provider = "CALICO"
}
```

Adding the policy manually with the CLI and re-running `terraform apply` works. It seems the `network_policy` block attempts to enable network policy on the cluster without first adding the corresponding addon.

See https://cloud.google.com/sdk/gcloud/reference/beta/container/clusters/update#--enable-network-policy

terraform 0.11.1
terraform-provider-google 1.3.0

ndmckinley:
Fixes #860

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -157,6 +157,20 @@ func resourceContainerCluster() *schema.Resource {
 								},
 							},
 						},
+						"network_policy_config": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Computed: true,
+							MaxItems: 1,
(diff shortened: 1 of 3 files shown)


ndmckinley:
Oh, whoops, need to add docs.

PR #1202, opened 2018-03-15
Remove commented out code.

paddycarver:
We don't really need dead code, and this is breaking our `make fmt` test
because it gets indented differently in go 1.10 from in go 1.9.


danawillow:
At some point it may be worth discussing whether to keep those tests. I like that it keeps the code at a certain standard, but it also goes against recommendations: https://golang.org/doc/go1.10#gofmt
> In general, we recommend against building systems that check that source code matches the output of a specific version of gofmt. For example, a continuous integration test that fails if any code already checked into a repository is not “properly formatted” is inherently fragile and not recommended.

paddycarver:
Agreed! I'd much rather see something like a GitHub indicator for it that it failed that check, or something.

PR #1204, opened 2018-03-15
importify gke tests

danawillow:
Also fix a bug with the way we were setting `guest_accelerators` that got caught by this, and the bug with `workload_metadata_config` that was causing a whole bunch of tests to fail already.


ndmckinley:
Congrats on commit number 2,000!

danawillow:
🙌

PR #1210, opened 2018-03-16
cdn_policy should also be computed.

ndmckinley:


ndmckinley:
I haven't gotten a perfectly clean build on this because of quota issues, but I have had at least a few of the tests that failed in the last `master` run pass, and the only failures here are quota.

ndmckinley:
Got a clean build!  Thanks, Paddy, for taking care of the quota issues.

PR #1205, opened 2018-03-16
Add deletion protection to resource_compute_instance

nickjacques:
- [x] Added `deletion_protection` boolean parameter for `google_compute_instance` resources (in `resource_compute_instance.go`).
  - Parameter is optional, default value is false
- [x] Updated documentation to describe functionality
- [x] Add unit tests to confirm functionality

This resolves #956.

PR #1211, opened 2018-03-16
Add deletion_protection wrapper for deleting instance

nickjacques:
- Throws a friendly error to the user immediately instead of waiting for the API to return a 400

Previously:
```
* google_compute_instance.instance_delprotect: Error deleting instance: googleapi: Error 400: Invalid resource usage: 'Resource cannot be deleted if it's protected against deletion.'., invalidResourceUsage
```

Now:
```
* google_compute_instance.instance_delprotect: Cannot delete instance instance-delprotect: instance Deletion Protection is enabled. Set deletion_protection to false for this resource and run "terraform apply" before attempting to delete it.
```

@ndmckinley this is a follow-on to #1205 to add friendly handling when you try to delete an instance with `deletion_protection = true`.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -1495,19 +1495,24 @@ func resourceComputeInstanceDelete(d *schema.ResourceData, meta interface{}) err
 		return err
 	}
 	log.Printf("[INFO] Requesting instance deletion: %s", d.Id())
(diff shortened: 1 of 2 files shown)


nickjacques:
Also updated the docs to call this out to users.

ndmckinley:
Tests pass!  Merging.

PR #1222, opened 2018-03-19
Add wait_for_instances field to IGM and self_link option to the IG data source

### LINKED ISSUE #1213: Add operation to wait for MIG resize before completion
### Terraform Version
`Terraform v0.11.2`

### Affected Resource(s)
- google_compute_instance_group_manager
- google_compute_region_instance_group_manager

```hcl

resource "google_compute_instance_template" "fe1" {
  # Network tags
  tags = ["allow-http"]

  labels = {
    example = "dependent_mig"
  }

  disk {
    source_image = "cos-cloud/cos-stable"
    auto_delete  = true
    boot         = true
  }

  network_interface {
    network = "default"

    # use a random IP
    access_config = {}
  }

  service_account {
    scopes = ["storage-ro"]
  }

  metadata {
    gce-container-declaration = <<CONTAINER
spec:
  containers:
  - name: nginx-template-1
    image: 'gcr.io/cloud-marketplace/google/nginx1:1.12'
    stdin: false
    tty: false
  restartPolicy: Always
CONTAINER
  }

  lifecycle {
    create_before_destroy = true
  }
}

resource "google_compute_instance_group_manager" "fe1" {
  name               = "fe1-manager"
  instance_template  = "${google_compute_instance_template.fe1.self_link}"
  base_instance_name = "fe1-manager-instance"
  zone               = "us-central1-f"
  target_size        = "10"
}
```

### Expected Behavior
When I change the target size of a MIG, update the instance template, or otherwise change the instance group property, I expect that my Terraform shouldn't return until the underlying change is complete. Critically, any resources which depend on the MIG (including via a depends_on argument) shouldn't be triggered/created until the MIG is completely updated/resized.

### Actual Behavior
The MIG resource is considered done as soon as the initial API request is resolved, which is usually long before the MIG is actually resized/changed.

### Steps to Reproduce
1. `terraform apply`
2. See it returns very quickly.

### Important Factoids
I think this behavior should probably be configurable. Some users want the MIG to be considered updated as soon as the resize request is completed.

### References
- GH-1131 for Google

danawillow:
Fixes #1213.

(even though it's two changes, doing it in a single PR enables the test I wanted to write)

```
$ make testacc TEST=./google TESTARGS='-run=TestAccDataSourceGoogleComputeInstanceGroup_fromIGM'
TF_ACC=1 go test ./google -v -run=TestAccDataSourceGoogleComputeInstanceGroup_fromIGM -timeout 120m
=== RUN   TestAccDataSourceGoogleComputeInstanceGroup_fromIGM
=== PAUSE TestAccDataSourceGoogleComputeInstanceGroup_fromIGM
=== CONT  TestAccDataSourceGoogleComputeInstanceGroup_fromIGM
--- PASS: TestAccDataSourceGoogleComputeInstanceGroup_fromIGM (121.64s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	121.848s
```

### DIFF
--- a/google/resource_compute_instance_group_manager.go
+++ b/google/resource_compute_instance_group_manager.go
@@ -450,8 +450,9 @@ func resourceComputeInstanceGroupManagerRead(d *schema.ResourceData, meta interf
 			Timeout: d.Timeout(schema.TimeoutCreate),
 		}
 		_, err := conf.WaitForState()
-		// If err is nil, success.
-		return err
+		if err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/google/resource_compute_instance_group_manager.go
+++ b/google/resource_compute_instance_group_manager.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
 
@@ -197,6 +198,11 @@ func resourceComputeInstanceGroupManager() *schema.Resource {
 					},
 				},
 			},
+			"wait_for_instances": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
 		},
 	}
 }
@@ -312,18 +318,18 @@ func flattenNamedPortsBeta(namedPorts []*computeBeta.NamedPort) []map[string]int
 
 }
 
-func resourceComputeInstanceGroupManagerRead(d *schema.ResourceData, meta interface{}) error {
+func getManager(d *schema.ResourceData, meta interface{}) (*computeBeta.InstanceGroupManager, error) {
 	computeApiVersion := getComputeApiVersion(d, InstanceGroupManagerBaseApiVersion, InstanceGroupManagerVersionedFeatures)
 	config := meta.(*Config)
 
 	project, err := getProject(d, config)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	region, err := getRegion(d, config)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	manager := &computeBeta.InstanceGroupManager{}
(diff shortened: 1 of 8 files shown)


google/resource_compute_instance_group_manager.go
```diff
@@ -422,6 +442,18 @@ func resourceComputeInstanceGroupManagerRead(d *schema.ResourceData, meta interf
...
 
+	if d.Get("wait_for_instances").(bool) {
+		conf := resource.StateChangeConf{
+			Pending: []string{"creating", "error"},
+			Target:  []string{"created"},
+			Refresh: waitForInstancesRefreshFunc(getManager, d, meta),
+			Timeout: d.Timeout(schema.TimeoutCreate),
+		}
+		_, err := conf.WaitForState()
+		// If err is nil, success.
```
ndmckinley: If someone later adds some code below this, it'll get missed - might be good to add a comment saying not to do that?  Not sure.

google/resource_compute_region_instance_group_manager.go
```diff
@@ -257,17 +257,17 @@ func waitForInstancesRefreshFunc(f getInstanceManagerFunc, d *schema.ResourceDat
 			log.Printf("[WARNING] Error in fetching manager while waiting for instances to come up: %s\n", err)
 			return nil, "error", err
 		}
-		if creatingCount := m.CurrentActions.Creating + m.CurrentActions.CreatingWithoutRetries; creatingCount > 0 {
-			return creatingCount, "creating", nil
+		if done := m.CurrentActions.None; done < m.TargetSize {
```
ndmckinley: _slaps forehead_

I think I remember writing this code, so thanks for catching my mistake there - obviously the instances shouldn't report ready if they're `FAILED`.

google/resource_compute_instance_group_manager.go
danawillow: Just updated it to return the error only if there is one, that seemed like a much more standard way to go about this.

Issue #1193, opened 2018-03-14
add support for a google_project_organization_policy resource


lawrenae:
I'd like to be able to do the equivelant of the following command in terraform:

```bash
cat allow-external-ip.json
{ "constraint": "constraints/compute.vmExternalIpAccess", "listPolicy": { "allValues": ALLOW }}

gcloud beta resource-manager org-policies set-policy allow-external-ip.json --project my-project-id
```

The bigger use case is to define an org policy at a folder level like so:

```hcl
resource "google_folder_organization_policy" "no-external-ips" {
  folder     = "${google_folder.cool_folder.name}"
  constraint = "compute.vmExternalIpAccess"

  list_policy {
    deny {
      all = true
    }
  }
}
```

and then be able to "override" it on a per project basis:
```hcl
resource "google_project_organization_policy" "cool-project-allow-external-IPs" {
  project    = "${google_project.cool_project.project_id}"
  constraint = "compute.vmExternalIpAccess"

  list_policy {
    allow {
      all = true
    }
  }
}
```

Thoughts?  I'd be interested in alternatives to solving this problem, or implementing the solution in the provider, so please let me know.

Thanks!

-- Andy

lawrenae:
Closing this issue with the PR being merged

PR #1226, opened 2018-03-20
google_project_organization_policy

lawrenae:
this to implement #1193 

Feedback most welcome -- this is my 1st PR for this project.

### DIFF
--- a/google/iam_project.go
+++ b/google/iam_project.go
@@ -5,7 +5,6 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/schema"
 	"google.golang.org/api/cloudresourcemanager/v1"
-	"strings"
 )
 
 var IamProjectSchema = map[string]*schema.Schema{
@@ -71,11 +70,4 @@ func (u *ProjectIamUpdater) GetMutexKey() string {
 
 func (u *ProjectIamUpdater) DescribeResource() string {
 	return fmt.Sprintf("project %q", u.resourceId)
-}
-
-func canonicalProjectId(project string) string {
-	if strings.HasPrefix(project, "projects/") {
-		return project
-	}
-	return fmt.Sprintf("projects/%s", project)
 }
--- a/google/iam_project.go
+++ b/google/iam_project.go
@@ -5,6 +5,7 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/schema"
 	"google.golang.org/api/cloudresourcemanager/v1"
+	"strings"
 )
 
 var IamProjectSchema = map[string]*schema.Schema{
@@ -70,4 +71,11 @@ func (u *ProjectIamUpdater) GetMutexKey() string {
 
 func (u *ProjectIamUpdater) DescribeResource() string {
 	return fmt.Sprintf("project %q", u.resourceId)
+}
+
+func canonicalProjectId(project string) string {
+	if strings.HasPrefix(project, "projects/") {
+		return project
+	}
+	return fmt.Sprintf("projects/%s", project)
 }
--- a/google/resource_google_project_organization_policy_test.go
+++ b/google/resource_google_project_organization_policy_test.go
@@ -1,0 +1,321 @@
+package google
+
+import (
+	"fmt"
+	"reflect"
+	"sort"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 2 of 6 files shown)


google/resource_google_project_organization_policy_test.go
```diff
@@ -0,0 +1,321 @@
...
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
+func TestAccProjectOrganizationPolicy_boolean(t *testing.T) {
+	t.Parallel()
+
+	folder := acctest.RandomWithPrefix("tf-test")
```
danawillow: Looks like a lot of this test was copied directly from the "folder" test and a fair number of "folder" references stuck behind. Mind cleaning those up?

website/docs/r/google_project_organization_policy.html.markdown
```diff
@@ -0,0 +1,106 @@
...
+documentation](https://cloud.google.com/resource-manager/docs/organization-policy/overview) and
+[API](https://cloud.google.com/resource-manager/reference/rest/v1/projects/setOrgPolicy).
+
+## Example Usage
+
+To set policy with a [boolean constraint](https://cloud.google.com/resource-manager/docs/organization-policy/quickstart-boolean-constraints):
+
+```hcl
+resource "google_project_organization_policy" "serial_port_policy" {
+  project    = "123456789"
```
danawillow: I think the resource wants the project id right, not the number? If so, mind doing something with text in it so people don't think they have to figure out what their project number is?

website/docs/r/google_project_organization_policy.html.markdown
```diff
@@ -0,0 +1,106 @@
...
+  }
+}
+```
+
+
+Or to deny some services, use the following instead:
+
+```hcl
+resource "google_project_organization_policy" "services_policy" {
+  project    = "folders/123456789"
```
danawillow: found a folder :)

website/docs/r/google_project_organization_policy.html.markdown
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project` - (Required) The project id of the project to set the policy for. Its format is simply {project_id}.
```
danawillow: I'm not sure the second sentence is really necessary here since we use `project` all over the google terraform provider to mean the exact same thing. Up to you though, I won't block on it.

lawrenae:
Hi @danawillow -- great feedback. I've updated accordingly!

google/resource_google_project_organization_policy_test.go
```diff
@@ -0,0 +1,321 @@
...
+	if !ok {
+		return nil, fmt.Errorf("Not found: %s", rn)
+	}
+
+	if rs.Primary.ID == "" {
+		return nil, fmt.Errorf("No ID is set")
+	}
+
+	config := testAccProvider.Meta().(*Config)
+	//TODO: fix this to return an ACTUAL projectID from state
```
danawillow: Can you fix this? (There's also a similar issue in the destroy() function)

Also, I'd highly recommend running the tests on your own if you can- this will make the review go a bit faster so we don't need a full back-and-forth if the tests are failing. There are some instructions at https://github.com/terraform-providers/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#tests, let me know if you have any questions!

google/iam_project.go
```diff
@@ -71,3 +72,10 @@ func (u *ProjectIamUpdater) GetMutexKey() string {
 func (u *ProjectIamUpdater) DescribeResource() string {
 	return fmt.Sprintf("project %q", u.resourceId)
 }
+
+func canonicalProjectId(project string) string {
```
danawillow: Since this is only used in the test, should it live in that file?

google/resource_google_project_organization_policy_test.go
```diff
@@ -0,0 +1,289 @@
...
+	"fmt"
+	"reflect"
+	"sort"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
```
danawillow: Mind adding a comment here explaining why the tests don't run in parallel? (that way nobody tries to re-parallelize them)

PR #1234, opened 2018-03-20
Update container_cluster documentation for the zone field

rosbo:
The `zone` field is optional. It takes the provider-level zone if not specified.


danawillow:
Also congrats on PR #1234 :)

rosbo:
🤣

PR #1235, opened 2018-03-21
Set wait_for_instances when importing IGMs, make defaults match

danawillow:
This fixes `TestAccInstanceGroupManager_basic`, which is currently failing in CI.

While doing so, I realized the default for `wait_for_instances` was true for IGMs but false for regional IGMs so I adjusted it to false to match.


google/resource_compute_instance_group_manager.go
```diff
@@ -201,7 +201,7 @@ func resourceComputeInstanceGroupManager() *schema.Resource {
 			"wait_for_instances": &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  true,
+				Default:  false,
```
paddycarver: Calling out for future-Paddy--this doesn't require a migration or anything because this hasn't been released yet.

PR #1240, opened 2018-03-21
Add time partitioning field to google_bigquery_table resource

amatellanes:
Add time partitioning field to [google_bigquery_table](https://www.terraform.io/docs/providers/google/r/bigquery_table.html) resource. This field is available on type [`TimePartitioning`](https://godoc.org/cloud.google.com/go/bigquery#TimePartitioning) from package [`bigquery`](https://godoc.org/cloud.google.com/go/bigquery):

```go
type TimePartitioning struct {
    ...

    // If empty, the table is partitioned by pseudo column '_PARTITIONTIME'; if set, the
    // table is partitioned by this field. The field must be a top-level TIMESTAMP or
    // DATE field. Its mode must be NULLABLE or REQUIRED.
    Field string
}
```

### DIFF
--- a/google/resource_bigquery_table.go
+++ b/google/resource_bigquery_table.go
@@ -440,10 +440,9 @@ func expandTimePartitioning(configured interface{}) *bigquery.TimePartitioning {
 
 func flattenTimePartitioning(tp *bigquery.TimePartitioning) []map[string]interface{} {
 	result := map[string]interface{}{"type": tp.Type}
-	result["field"] = tp.Type
 
-	if tp.Type != "" {
-		result["field"] = tp.Type
+	if tp.Field != "" {
+		result["field"] = tp.Field
 	}
 
 	if tp.ExpirationMs != 0 {
--- a/google/resource_bigquery_table.go
+++ b/google/resource_bigquery_table.go
@@ -148,6 +148,7 @@ func resourceBigQueryTable() *schema.Resource {
 						"field": {
 							Type:     schema.TypeString,
 							Optional: true,
+							ForceNew: true,
 						},
 					},
 				},
--- a/google/resource_bigquery_table.go
+++ b/google/resource_bigquery_table.go
@@ -141,6 +141,14 @@ func resourceBigQueryTable() *schema.Resource {
 							Required:     true,
 							ValidateFunc: validation.StringInSlice([]string{"DAY"}, false),
 						},
+
+						// Type: [Optional] The field used to determine how to create a time-based
+						// partition. If time-based partitioning is enabled without this value, the
+						// table is partitioned based on the load time.
+						"field": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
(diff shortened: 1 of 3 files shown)


google/resource_bigquery_table.go
```diff
@@ -428,6 +440,11 @@ func expandTimePartitioning(configured interface{}) *bigquery.TimePartitioning {
 
 func flattenTimePartitioning(tp *bigquery.TimePartitioning) []map[string]interface{} {
 	result := map[string]interface{}{"type": tp.Type}
+	result["field"] = tp.Type
```
danawillow: shouldn't this be tp.Field?

amatellanes:
@danawillow I have fix `flattenTimePartitioning` function to retrieve `field`. Also I have created `TestAccBigQueryTable_TimePartitioningField` test to check time partitioning field but I'm not really sure if I should have added this test to an existing one... perhaps `TestAccBigQueryTable_Basic`...

danawillow:
Thanks! Yeah I think having it in `basic` would work fine. Do you mind also adding an import step to the test? We're trying to standardize on as many tests as possible having that step since it lets us check that all the fields match. It would look like this:
```
			{
				ResourceName:        "google_bigquery_table.test",
				ImportState:         true,
				ImportStateVerify:   true,
			},
```

amatellanes:
Hi @danawillow 

Changes done.

danawillow:
Hey @amatellanes, I'm not sure if you were intending to have the test update the existing table, but by having the two steps in the same test that's what it tried to do. It sounds like from this error message that you can't update that field. If that's true, can you make it `ForceNew`?

```
$ make testacc TEST=./google TESTARGS='-run=TestAccBigQueryTable_Basic'
TF_ACC=1 go test ./google -v -run=TestAccBigQueryTable_Basic -timeout 120m
=== RUN   TestAccBigQueryTable_Basic
=== PAUSE TestAccBigQueryTable_Basic
=== CONT  TestAccBigQueryTable_Basic
--- FAIL: TestAccBigQueryTable_Basic (6.64s)
	testing.go:513: Step 2 error: Error applying: 1 error(s) occurred:

		* google_bigquery_table.test: 1 error(s) occurred:

		* google_bigquery_table.test: googleapi: Error 400: Cannot change partitioning spec for a partitioned table., invalid
FAIL
FAIL	github.com/terraform-providers/terraform-provider-google/google	6.847s
make: *** [testacc] Error 1
```

amatellanes:
I've just marked the field as `ForceNew`. Since we cannot update partitioning field for a partitioned table, should I create a new test or modify `testAccBigQueryTable` to add the field `Field` to `time_partitioning` block?

danawillow:
I would just modify `testAccBigQueryTable` but I don't care particularly strongly

amatellanes:
Finally I've chosen to modify `testAccBigQueryTable`.

Issue #1230, opened 2018-03-20
[enhancement] Add SSL Policy support to Target SSL Proxies


nickjacques:
Now that [SSL Policies](https://cloud.google.com/compute/docs/load-balancing/ssl-policies) have [been released](https://cloudplatform.googleblog.com/2018/02/announcing-SSL-policies-for-HTTPS-and-SSL-proxy-load-balancers.html), let's add SSL Policy support to Target SSL Proxies.

Parent issue for creating SSL Policy resources is #1228.

rosbo:
Note to implementer, the change should be similar to the one for the https proxy:
https://github.com/GoogleCloudPlatform/magic-modules/pull/151/files

Issue #1229, opened 2018-03-20
[enhancement] Add SSL Policy support to Target HTTPS Proxies


nickjacques:
Now that [SSL Policies](https://cloud.google.com/compute/docs/load-balancing/ssl-policies) have [been released](https://cloudplatform.googleblog.com/2018/02/announcing-SSL-policies-for-HTTPS-and-SSL-proxy-load-balancers.html), let's add SSL Policy support to Target HTTPS Proxies.

Parent issue for creating SSL Policy resources is #1228.

rosbo:
This feature was added in #1466 and release in 1.13.0

PR #1247, opened 2018-03-23
New resource/datasource: SSL Policy

### LINKED ISSUE #1228: [new resource] Support SSL Policies
Now that [SSL Policies](https://cloud.google.com/compute/docs/load-balancing/ssl-policies) have [been released](https://cloudplatform.googleblog.com/2018/02/announcing-SSL-policies-for-HTTPS-and-SSL-proxy-load-balancers.html), let's support them in the TF provider so that teams can perform CRUD ops on SSL Policies.

Not in scope (separate issues/enhancements): 
- add SSL Policy support to Target HTTPS Proxies #1229 
- add SSL Policy support to Target SSL Proxies #1230 

I'll see if I can take a crack at creating the resource definition and related tests.

nickjacques:
- Add new datasource for SSL Policy
- Add new resource for SSL Policy
- Add docs for datasource and resource
- Add unit tests for datasource and resource

Open questions: 

- In scenarios where the user has not provided an alphabetized list of `custom_features`, TF will show a "perpetual diff" of the same contents in a different order. Since I'm a bit new to both developing TF resources and developing in Go, I'm seeking suggestions in how to best handle this.
  - example terraform:
    ```
    resource "google_compute_ssl_policy" "test" {
      name = "test"
      profile         = "CUSTOM"
      custom_features = ["TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384", "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384", "TLS_RSA_WITH_3DES_EDE_CBC_SHA"]
    }
    ```
  - example `terraform plan` output, post-creation of resource:
    ```
    
      ~ google_compute_ssl_policy.test
          custom_features.0: "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384" => "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"
          custom_features.1: "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384" => "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384"
    
    
    ```
- In scenarios where the user attempts to update two resource attributes at once, `terraform apply` will fail with:
    ```
    ~ google_compute_ssl_policy.nlj-test2
        min_tls_version:   "TLS_1_2" => "TLS_1_0"
        profile:           "RESTRICTED" => "COMPATIBLE"

    ...

      Error: Error applying plan:

    1 error(s) occurred:

    * google_compute_ssl_policy.test: 1 error(s) occurred:

    * google_compute_ssl_policy.test: Error updating SSL Policy: googleapi: Error 412: Invalid fingerprint., conditionNotMet
    ```
  - Presumably this is related to partial updates and the fingerprint changing between the two. Is there a good reference for updates that are not partial? I think a "consolidated" or full update may prevent this issue, as it will build the entire object (including all changed parameters) and thus the fingerprint won't change until that update has been processed by the API.

This PR will resolve #1228 and paves the way for #1229 and #1230


google/resource_compute_ssl_policy.go
```diff
@@ -0,0 +1,281 @@
...
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	name := d.Get("name").(string)
+
+	d.Partial(true)
+
+	if d.HasChange("min_tls_version") {
```
danawillow: To answer your other question, yeah I think it makes sense to just do one update call for this instead of breaking it down. It's a bit weird since this resource seems to only support PATCH and not UPDATE, but there should also be no harm in calling PATCH with unchanged fields. `resource_compute_backend_service.go` is an example of a resource with a full update call

google/resource_compute_ssl_policy_test.go
```diff
@@ -0,0 +1,308 @@
...
+		CheckDestroy: testAccCheckComputeSslPolicyDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccComputeSslUpdate1(sslPolicyName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeSslPolicyExists(
+						"google_compute_ssl_policy.update", &sslPolicy),
+					resource.TestCheckResourceAttr(
+						"google_compute_ssl_policy.update", "profile", "MODERN"),
+					//resource.TestCheckResourceAttr(
```
danawillow: should this be uncommented?

website/docs/d/datasource_compute_ssl_policy.html.markdown
```diff
@@ -0,0 +1,50 @@
+
```
danawillow: nit: extra newline

website/docs/d/datasource_compute_ssl_policy.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+# google\_compute\_ssl\_policies
+
+Gets an SSL Policy within GCE from its name, for use with Target HTTPS and Target SSL Proxies.
+    For more information see [the official documentation](https://cloud.google.com/compute/docs/load-balancing/ssl-policies).
+
+## Example Usage
+
+```tf
+data "google_compute_ssl_policy" "my-ssl-policy" {
+  name   = "production-ssl-policy"
```
danawillow: nit: extra spaces between `name` and `=`

website/docs/d/datasource_compute_ssl_policy.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+* `description` - Description of this SSL Policy.
+
+* `min_tls_version` - The minimum supported TLS version of this policy.
+
+* `profile` - The Google-curated or custom profile used by this policy.
+
+* `custom_features` - If the `profile` is `CUSTOM`, these are the custom encryption
+    ciphers supported by the profile. If the `profile` is *not* `CUSTOM`, this
+    attribute will be nil.
```
danawillow: nit: since nil isn't really a terraform concept, let's say empty instead

google/resource_compute_ssl_policy.go
```diff
@@ -0,0 +1,281 @@
...
+			"custom_features": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
danawillow: this is surprising to me, can you confirm that the description can't be changed?

google/resource_compute_ssl_policy_test.go
nickjacques: Ah, this was a result of the Fingerprint issue (the test was originally written to update two attributes). I'll make the update a full instead of partial and these tests should work properly with two attributes.

google/resource_compute_ssl_policy.go
nickjacques: I went off of the details provided in `gcloud beta compute ssl-policies update`, which doesn't state that description is an updatable field, but I'll try against the API directly and see if it can be patched.

nickjacques: I just tried updating two fields via API Explorer: `minTlsVersion` and `description`. The change was applied to `minTlsVersion` but not `description`, so I believe the `ForceNew: true` is valid. If someone wants to change the resource description, a new resource must be created.

nickjacques:
Okay, I think these latest commits resolve most of the comments - thanks for the tip on using Sets @danawillow, that worked perfectly! The only issue I'm encountering now is when updating a resource from:

```tf
profile         = "CUSTOM"
custom_features = ["TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384", "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384", "TLS_RSA_WITH_3DES_EDE_CBC_SHA"]
```

to something like:
```tf
profile = "MODERN"
```

In this scenario, I receive an error from the underlying API:
```
google_compute_ssl_policy.test: Error updating SSL Policy: googleapi: Error 400: Invalid value for field 'resource.customFeatures': ''. Custom features cannot be specified when the SslPolicy profile is not 'CUSTOM', invalid
```

So it seems that I need to delete the `sslPolicy.CustomFeatures` node before I send the request to the patch API, but I'm unsure of how to do that. I tried setting `sslPolicy.CustomFeatures = nil` if the length is 0, but I'm still hitting the same error. Once this defect is fixed, I can write a test to ensure we don't regress in the future.

google/data_source_google_compute_ssl_policy.go
```diff
@@ -0,0 +1,29 @@
...
+
+func dataSourceGoogleComputeSslPolicy() *schema.Resource {
+	// Generate datasource schema from resource
+	dsSchema := datasourceSchemaFromResourceSchema(resourceComputeSslPolicy().Schema)
+
+	// Set 'Required' schema elements
+	addRequiredFieldsToSchema(dsSchema, "name")
+
+	// Set 'Optional' schema elements
+	addOptionalFieldsToSchema(dsSchema, "custom_features", "description", "min_tls_version", "profile", "project")
```
danawillow: You probably just want "project" here- an optional field in a data source is one that you can choose to set in the data source itself, rather than all the fields that it might return. Setting something like custom_features wouldn't have any effect, because the data source is read-only.

google/resource_compute_ssl_policy.go
```diff
@@ -0,0 +1,226 @@
...
+				Optional: true,
+				ForceNew: true, // currently, the beta patch API call does not allow updating the description
+			},
+
+			"min_tls_version": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "TLS_1_0",
+				// Although compute-gen.go says that TLS_1_3 is a valid value, the API currently (26 Mar 2018)
+				// responds with an HTTP 200 but doesn't actually create/update the policy.
```
danawillow: Oh that's distressing. Is there an issue in the public issue tracker that we could link to here?

google/resource_compute_ssl_policy.go
```diff
@@ -0,0 +1,226 @@
...
+		MinTlsVersion:  d.Get("min_tls_version").(string),
+		CustomFeatures: convertStringSet(d.Get("custom_features").(*schema.Set)),
+	}
+
+	op, err := config.clientComputeBeta.SslPolicies.Patch(project, name, sslPolicy).Do()
+	if err != nil {
+		return fmt.Errorf("Error updating SSL Policy: %s", err)
+	}
+
+	err = computeSharedOperationWaitTime(config.clientCompute, op, project, int(d.Timeout(schema.TimeoutCreate).Minutes()), "Updating SSL Policy")
```
danawillow: This should probably be schema.TimeoutUpdate

google/resource_compute_ssl_policy_test.go
```diff
@@ -0,0 +1,308 @@
...
+			config.Project, rs.Primary.ID).Do()
+		if err == nil {
+			return fmt.Errorf("SSL Policy still exists")
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckComputeSslPolicyCustomFeatures(sslPolicy *computeBeta.SslPolicy, n string) resource.TestCheckFunc {
```
danawillow: You can actually get rid of this test function- the import check will verify this on its own (it'll call read and then try to populate state from it and compare it to the config, so if custom_features wasn't set upstream then the import would fail)

website/docs/d/datasource_compute_ssl_policy.html.markdown
```diff
@@ -0,0 +1,49 @@
+---
+layout: "google"
+page_title: "Google: google_compute_ssl_policy"
+sidebar_current: "docs-google-datasource-compute-ssl-policy"
+description: |-
+  Gets an SSL Policy within GCE, for use with Target HTTPS and Target SSL Proxies.
+---
+
+# google\_compute\_ssl\_policies
```
danawillow: s/policies/policy

website/docs/r/compute_ssl_policy.html.markdown
```diff
@@ -0,0 +1,82 @@
+---
+layout: "google"
+page_title: "Google: google_compute_ssl_policy"
+sidebar_current: "docs-google-compute-ssl-policy"
+description: |-
+  Manages an SSL Policy within GCE, for use with Target HTTPS and Target SSL Proxies.
+---
+
+# google\_compute\_ssl\_policies
```
danawillow: and here

google/resource_compute_ssl_policy.go
nickjacques: Not that I'm aware of, but I can file one if you'd like! Relevant lines are 21109-21118 in compute/v0.beta/compute-gen.go.

danawillow: If you don't mind, that would be great! That way if someone stumbles across this part of the code and sees a comment with a link to an issue that's been fixed, we'd be able to add TLS_1_3 real quick

nickjacques:
Thanks @danawillow `NullFields` was indeed exactly what I needed! Updates that enable and disable `custom_features` now work as expected, and I also have included tests for these scenarios. I've also added `enabled_features` as a computed field that folks can use to output what cipher suites resulted from their policy configuration. I think this is ready to go if things look good to you!

google/resource_compute_ssl_policy.go
nickjacques: Ok, can do! Which product should I file it under? Cloud SDK?

danawillow: I'm not super familiar with how the products are laid out in components, but if there's one for compute admin api, or networking api, or something along those lines. I'm not going to block this PR on getting that link in there though. I'll finish up the review now :)

google/resource_compute_ssl_policy.go
```diff
@@ -0,0 +1,244 @@
...
+		return err
+	}
+
+	name := d.Get("name").(string)
+
+	sslPolicy := &computeBeta.SslPolicy{
+		Fingerprint: d.Get("fingerprint").(string),
+	}
+
+	if v, ok := d.GetOk("profile"); ok {
```
danawillow: nit: There isn't actually any difference between doing it this way and just setting it in the struct, since if the value is empty it'll be the empty string in the struct regardless.

google/resource_compute_ssl_policy.go
nickjacques: No worries, I just added a link to the issue that Aaron created. (Thanks Aaron!)

google/resource_compute_ssl_policy_test.go
```diff
@@ -0,0 +1,387 @@
...
+				Config: testAccComputeSslUpdate1(sslPolicyName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeSslPolicyExists(
+						"google_compute_ssl_policy.update", &sslPolicy),
+					resource.TestCheckResourceAttr(
+						"google_compute_ssl_policy.update", "profile", "MODERN"),
+					resource.TestCheckResourceAttr(
+						"google_compute_ssl_policy.update", "min_tls_version", "TLS_1_0"),
+				),
+			},
```
danawillow: let's also add the import test step here (and similarly after all test steps) juuuuuuuust in case

nickjacques:
No worries at all! I've added `Profile` and `minTlsVersion` back into the struct, and dropped in imports between steps in the multi-step tests.

danawillow:
👍

PR #1254, opened 2018-03-26
Add tracking labels to GKE clusters and instance templates

mkjelland:
Instance templates have a configurable label field through Terraform already. I added a label to the list of labels set now, and then had to remove the label in the read method to prevent a change from being detected on a second run.

For GKE clusters, there is currently no option to add a user label to the cluster from Terraform, so the additional label is not detected as a change on a second run.


danawillow:
adding @paddycarver just to double-check on the wording

PR #1259, opened 2018-03-27
Set isn't a valid key for a TypeMap

ndmckinley:
This line wasn't doing anything so we might as well get rid of it.  :)


ndmckinley:
Running the tests now.

ndmckinley:
I can't explain it, but:

```
fatal error: all goroutines are asleep - deadlock!
```

I'll see if I can figure out what's happened there.

ndmckinley:
The deadlock was unrelated and it's passing now.  Can't explain that!  Mystery for the ages.  Since it passes, I'm merging.

PR #1263, opened 2018-03-27
Add DiffSuppress for Target HTTP(S) Proxies

nickjacques:
Prevents no-op diff/change during plan/apply. Example of such a no-op diff:

```
  ~ update in-place

  ~ google_compute_target_https_proxy.foobar
      url_map: "https://www.googleapis.com/compute/beta/projects/***/global/urlMaps/httpsproxy-test-url-map" => "https://www.googleapis.com/compute/v1/projects/***/global/urlMaps/httpsproxy-test-url-map"


```

Example with DiffSuppress:
```
No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, no
actions need to be performed.
```

### DIFF
--- a/google/resource_compute_target_http_proxy.go
+++ b/google/resource_compute_target_http_proxy.go
@@ -28,8 +28,9 @@ func resourceComputeTargetHttpProxy() *schema.Resource {
 			},
 
 			"url_map": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
+				Type:             schema.TypeString,
+				Required:         true,
+				DiffSuppressFunc: compareSelfLinkOrResourceName,
 			},
 
 			"description": &schema.Schema{
--- a/google/resource_compute_target_http_proxy.go
+++ b/google/resource_compute_target_http_proxy.go
@@ -30,7 +30,7 @@ func resourceComputeTargetHttpProxy() *schema.Resource {
 			"url_map": &schema.Schema{
 				Type:             schema.TypeString,
 				Required:         true,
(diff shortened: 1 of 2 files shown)


google/resource_compute_target_http_proxy.go
```diff
@@ -28,8 +28,9 @@ func resourceComputeTargetHttpProxy() *schema.Resource {
 			},
 
 			"url_map": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
+				Type:             schema.TypeString,
+				Required:         true,
+				DiffSuppressFunc: compareSelfLinkOrResourceName,
```
danawillow: I don't think the API would accept the resource name, would it? We have a lot of logic around the codebase (see the fns in field_helpers.go and where they get called) to convert between names and self links so users can supply either one, which would enable the use of this diffsuppress. So if you wanted to add that logic that would totally work (and maybe make some people's experience better), but a much easier solution is to use compareSelfLinkRelativePaths instead.

nickjacques: Ah, I just took another glance at this, and it looks like there actually _is_ a difference: `.../compute/beta/` vs `.../compute/v1/...`. I was indeed using the self-link of the URL map and not the resource name alone. So it looks like `compareSelfLinkRelativePaths` looks like the better option here. I'll update to use that.

danawillow:
Thanks @nickjacques!

PR #1266, opened 2018-03-28
Add support for node pool versions

### LINKED ISSUE #633: Add support for version upgrades in `google_container_node_pool`

danawillow:
Fixes #633.

```
--- PASS: TestAccContainerNodePool_version (918.80s)
```

### DIFF
--- a/google/resource_container_node_pool.go
+++ b/google/resource_container_node_pool.go
@@ -143,6 +143,12 @@ var schemaNodePool = map[string]*schema.Schema{
 		Computed:     true,
 		ValidateFunc: validation.IntAtLeast(0),
 	},
+
+	"version": {
+		Type:     schema.TypeString,
+		Optional: true,
+		Computed: true,
+	},
 }
 
 func resourceContainerNodePoolCreate(d *schema.ResourceData, meta interface{}) error {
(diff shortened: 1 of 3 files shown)


danawillow:
This felt way too easy though so please do try to find something I might have missed :P

PR #1272, opened 2018-03-29
Fix organization policy tests.

paddycarver:
This updates the organization policy tests to be run sequentially,
instead of in parallel, as they share a resource that they're modifying.
It also updates them to use a separate organization than the one all our
other tests are running in, which prevents other tests from failing
because they're run in parallel to the organization policy changing
under them.

### DIFF
--- a/google/resource_google_organization_policy_test.go
+++ b/google/resource_google_organization_policy_test.go
@@ -18,9 +18,25 @@ var DENIED_ORG_POLICIES = []string{
 
 // Since each test here is acting on the same organization, run the tests serially to
 // avoid race conditions and aborted operations.
+func TestAccOrganizationPolicy(t *testing.T) {
+	testCases := map[string]func(t *testing.T){
+		"boolean":        testAccOrganizationPolicy_boolean,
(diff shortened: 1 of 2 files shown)


paddycarver:
CI has been updated with the correct value for the new environment variable, so running this PR in CI _should_ work.

google/resource_google_organization_policy_test.go
```diff
@@ -18,22 +18,38 @@ var DENIED_ORG_POLICIES = []string{
...
+	testCases := map[string]func(t *testing.T){
+		"boolean":        testAccOrganizationPolicy_boolean,
+		"list_allowAll":  testAccOrganizationPolicy_list_allowAll,
+		"list_allowSome": testAccOrganizationPolicy_list_allowSome,
+		"list_denySome":  testAccOrganizationPolicy_list_denySome,
+		"list_update":    testAccOrganizationPolicy_list_update,
+	}
+
+	for name, tc := range testCases {
+		tc := tc
```
danawillow: is this on purpose? if so can you add a comment for what's going on?

danawillow: 👍

Issue #1269, opened 2018-03-28
Add folder information retrieval by folder id


zamir0109:
### Terraform Version
Terraform v0.10.8

### Affected Resource(s)
- (data source) google_active_folder

## Feature request
We'd like to be able to retrieve the folder information by its id, actually that data source requires the display name and the parent (another folder or the organization), but it'll be a lot easier if we could retrieve the folder info by its id.

### Terraform configuration files
Maybe like the following:

```hcl
data "google_active_folder" "folder" {
  id = "123456789"
}
```

And then retrieve information like:
- Parent id
- Name, display name

Thanks in advance

PR #1280, opened 2018-04-01
google folder data source

lnesci:
Created to solve #1269 .
The code is based on current organization and active_folder data source but to unify it looks like organization data source.
It support get by id, search by fields (for example ACTIVE + parent + display name) and lookup organization (it searches up the tree in case there are sub folders, not just the parent).
This is my first contribution, I hope is welcome and please let me know if I need to modify something.
Thanks!

### DIFF
--- a/google/data_source_google_folder_test.go
+++ b/google/data_source_google_folder_test.go
@@ -1,0 +1,173 @@
+package google
+
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccDataSourceGoogleFolder_byFullName(t *testing.T) {
+	folderId := getTestFolderFromEnv(t)
+	name := "folders/" + folderId
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckGoogleFolder_byName(name),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.google_folder.folder", "id", folderId),
+					resource.TestCheckResourceAttr("data.google_folder.folder", "name", name),
+				),
+			},
+		},
+	})
+}
+
+func TestAccDataSourceGoogleFolder_byShortName(t *testing.T) {
+	folderId := getTestFolderFromEnv(t)
+	name := "folders/" + folderId
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckGoogleFolder_byName(folderId),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.google_folder.folder", "id", folderId),
+					resource.TestCheckResourceAttr("data.google_folder.folder", "name", name),
+				),
+			},
+		},
+	})
+}
+
+func TestAccDataSourceGoogleFolder_lookupOrganization(t *testing.T) {
+	orgId := getTestOrgFromEnv(t)
+	orgName := "organizations/" + orgId
+	folderId := getTestFolderFromEnv(t)
+	name := "folders/" + folderId
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckGoogleFolder_lookupOrganization(folderId),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.google_folder.folder", "id", folderId),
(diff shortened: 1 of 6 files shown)


rosbo:
Hi @Inesci,

The reason why we called the data source to search by display name `google_active_folder` is because you can have multiple folders with the same display name under the same parent but only one ACTIVE folder with the same display name under that parent. To make sure we get a consistent result from the API and a result consistent with the user expectation, this is why we are filtering on ACTIVE state and make it explicit in the data source name.

For that reason, I would keep two datasources. One to retrieve the folder by id and another one to retrieve it by name.

I am curious about your use case to retrieve the folder by id. If you have the id, what do you use the display_name for in your config?

Thank you

lnesci:
Hi @rosbo,

It is for a module that has only the folder id, from it we need to retrieve the organization and from the organization infer the domain to then integration with gsuite.

Thanks

morgante:
I think it's sufficient for our use case to simply add a new data source to retrieve folder by ID.

lnesci:
@rosbo just to double check, the only pending task is to remove the query part of this new data source?
Thanks

rosbo:
@lnesci your use case makes sense. Thanks for the clarification. Yes, removing the query part of the new data source is the only thing left before I give my 👍

lnesci:
$ make testacc TEST=./google TESTARGS='-run=TestAccDataSourceGoogleFolder_*' GOOGLE_USE_DEFAULT_CREDENTIALS=true GOOGLE_FOLDER=886025180534 GOOGLE_PROJECT=base-project-196723 GOOGLE_REGION=us-central1 GOOGLE_ORG=65779779009
TF_ACC=1 go test ./google -v -run=TestAccDataSourceGoogleFolder_* -timeout 120m
=== RUN   TestAccDataSourceGoogleFolder_byFullName
--- PASS: TestAccDataSourceGoogleFolder_byFullName (2.72s)
=== RUN   TestAccDataSourceGoogleFolder_byShortName
--- PASS: TestAccDataSourceGoogleFolder_byShortName (1.58s)
=== RUN   TestAccDataSourceGoogleFolder_lookupOrganization
--- PASS: TestAccDataSourceGoogleFolder_lookupOrganization (1.20s)
=== RUN   TestAccDataSourceGoogleFolder_byFullNameNotFound
--- PASS: TestAccDataSourceGoogleFolder_byFullNameNotFound (0.25s)
=== RUN   TestAccDataSourceGoogleFolder_attributesCheck
--- PASS: TestAccDataSourceGoogleFolder_attributesCheck (15.26s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	(cached)

lnesci:
@rosbo code updated, please let me know if something needs to be modified.
Thanks!

website/docs/d/google_folder.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+}
+
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `folder` (Required) - The name of the Folder in the form `{folder_id}` or `folders/{folder_id}`.
+* `lookup_organization` (Optional) - `true` to find the organization that the folder belongs, `false` to avoid the lookup. It searches up the tree.
```
rosbo: Can you document whether it is set to true or false by default?

google/data_source_google_folder_test.go
```diff
@@ -0,0 +1,163 @@
...
+	return fmt.Sprintf(`
+data "google_folder" "folder" {
+  folder = "%s"
+  lookup_organization = true
+}`, name)
+}
+
+func testAccCheckGoogleFolder_attributesCheckConfig(parent string, displayName string) string {
+	return fmt.Sprintf(`
+resource "google_folder" "foobar" {
```
rosbo: You should create a new folder in the other tests above too instead of using GOOGLE_FOLDER. 

To make test run faster, and reduce the number of new folders our tests generate, using GOOGLE_FOLDER is a good idea. However, I suggest we do this in a separate PR since we will need to update our CI server config and GCP account to create and reuse that folder. We should also update the active_folder_data source test.

I filed an issue #1295 to track this.

lnesci: done in new commit

website/docs/d/google_folder.html.markdown
lnesci: done in new commit

lnesci:
requested changes done in commit 432a9de8767ab25c608af2046a92d92cdd97992b.

$ make testacc TEST=./google TESTARGS='-run=TestAccDataSourceGoogleFolder_' GOOGLE_USE_DEFAULT_CREDENTIALS=true GOOGLE_PROJECT=base-project-196723 GOOGLE_REGION=us-central1 GOOGLE_ORG=65779779009
TF_ACC=1 go test ./google -v -run=TestAccDataSourceGoogleFolder_ -timeout 120m
=== RUN   TestAccDataSourceGoogleFolder_byFullName
--- PASS: TestAccDataSourceGoogleFolder_byFullName (17.28s)
=== RUN   TestAccDataSourceGoogleFolder_byShortName
--- PASS: TestAccDataSourceGoogleFolder_byShortName (14.73s)
=== RUN   TestAccDataSourceGoogleFolder_lookupOrganization
--- PASS: TestAccDataSourceGoogleFolder_lookupOrganization (15.46s)
=== RUN   TestAccDataSourceGoogleFolder_byFullNameNotFound
--- PASS: TestAccDataSourceGoogleFolder_byFullNameNotFound (0.34s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	48.000s

rosbo:
Thanks for your contribution! Merging.

PR #1285, opened 2018-04-03
fix: include object versions when deleting all objects in GCS bucket

matthewrj:
This fixes a bug where force_destory fails to delete a GCS bucket when there are versioned objects in the bucket. This is caused by the force_destroy logic only deleting the current versions of objects and not all previous versions. This PR ensures all previous versions of objects are also deleted.

### DIFF
--- a/google/resource_storage_bucket.go
+++ b/google/resource_storage_bucket.go
@@ -446,7 +446,7 @@ func resourceStorageBucketDelete(d *schema.ResourceData, meta interface{}) error
 	bucket := d.Get("name").(string)
 
 	for {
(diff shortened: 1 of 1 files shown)


rosbo:
Can you add an acceptance test with the following 2 steps to prove the fix works and prevent regression?

Step 1: Creates a versioned bucket and add object A with content Z
Step 2: Update object A to content Y

The test tear down will delete the bucket. The test will pass if the bucket is successfully deleted and fail otherwise.

Thank you

matthewrj:
Done

PR #1297, opened 2018-04-05
Fix failing project_services import test and merge it

rosbo:


rosbo:
```sh
--- PASS: TestAccProjectServices_authoritative2 (67.10s)
--- PASS: TestAccProjectServices_authoritative (67.52s)
--- PASS: TestAccProjectServices_basic (102.41s)
--- PASS: TestAccProjectServices_manyServices (142.37s)
--- PASS: TestAccProjectServices_ignoreUnenablableServices (291.68s)
```

google/resource_google_project_services_test.go
```diff
@@ -176,34 +183,9 @@ func TestAccProjectServices_manyServices(t *testing.T) {
 	billingId := getTestBillingAccountFromEnv(t)
 	pid := "terraform-" + acctest.RandString(10)
 	services := []string{
```
danawillow: I think the reason we had all of these in the first place was to test that pagination would work correctly (which probably could have used a better test name or comment but oh well)

rosbo: Do you want to keep it? I removed them because the test was taking a whooping 723s to run but if we think it is important to test the pagination part, I am fine with putting it back.

rosbo: Done

PR #1301, opened 2018-04-06
fix error introduced when attached disks are deleted out of band

### LINKED ISSUE #763: State gets corrupt when using attached disks
### Terraform Version
Terraform v0.10.8

### Affected Resource(s)
- google_compute_instance

```hcl
resource "google_compute_disk" "default" {
  name    = "boot-${var.project_name}"
  type    = "${var.pd_type}"
  zone    = "${var.zone}"
  image   = "${var.boot_image}"
  project = "${var.google_project}"
  size    = "${var.pd_size}"
}

resource "google_compute_disk" "files_pd" {
  name = "files-${var.project_name}"
  type = "${var.files_pd_type}"
  size = "${var.files_pd_size}"
  zone = "${var.zone}"

  project = "${var.google_project}"
}

resource "google_compute_instance" "default" {
  name         = "instance-${var.project_name}"
  zone         = "${var.zone}"
  project      = "${var.google_project}"

  network_interface {
    subnetwork = "${var.google_subnetwork_name}"

    access_config {
      // Ephemeral IP
    }
  }

  boot_disk {
    source      = "${google_compute_disk.default.id}"
    auto_delete = false
  }

  attached_disk {
    source      = "${google_compute_disk.files_pd.id}"
    device_name = "files-disk"
  }
}
```

### Expected Behavior
The disk is always available and the state looks like this:

```
                "google_compute_instance.default": {
                    "type": "google_compute_instance",
                    "depends_on": [
                        "google_compute_disk.default",
                        "google_compute_disk.files_pd"
                    ],
                    "primary": {
                        "id": "project-name",
                        "attributes": {
                            "attached_disk.#": "1",
                            "attached_disk.0.device_name": "files-disk",
                            "attached_disk.0.disk_encryption_key_raw": "",
                            "attached_disk.0.disk_encryption_key_sha256": "",
                            "attached_disk.0.source": "https://www.googleapis.com/compute/v1/projects/project-name/zones/europe-west1-d/disks/files-project-name",
...
```

### Actual Behavior
The plan commands throws an error:
```
Error: Error refreshing state: 1 error(s) occurred:

* module.xxx.google_compute_instance.default: 1 error(s) occurred:

* module.xxx.google_compute_instance.default: google_compute_instance.default: The zonal field for resource disks cannot be empty.
```

Inspecting the state shows that it seems to be corrupted, the attached disk metadata are incomplete see below
```
                "google_compute_instance.default": {
                    "type": "google_compute_instance",
                    "depends_on": [
                        "google_compute_disk.default",
                        "google_compute_disk.files_pd",
                    ],
                    "primary": {
                        "id": "project-name",
                        "attributes": {
                            "attached_disk.#": "1",
....
```
### Steps to Reproduce
It's really happening randomly 😞  and so far couldn't figure much of what was triggering it ...

EDIT:
I was able to fix the state manually by completing the attached_disk metadata, plan then looks like:

```
  ~ update in-place

  ~ module.xxx.google_compute_instance.default
      attached_disk.0.device_name: "" => "files-disk"
      attached_disk.0.source:      "" => "files-project-name"


```

Applying this will corrupt the state again.

danawillow:
Previously, if a disk was deleted out of band, we would still try to store the same number of disks in state as we previously had, and `ParseDiskFieldValue(d.Get(fmt.Sprintf("attached_disk.%d.source", i)).(string), d, config)` would throw an error because there was no disk at index i. This makes sure we ignore any supposed disk fields that aren't really there (that occurred as a result of this bug), and fixes the bug by removing nils before we set the attached disks.

Fixes #763.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -852,26 +852,26 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("label_fingerprint", instance.LabelFingerprint)
 	}
 
-	attachedDisksCount := d.Get("attached_disk.#").(int)
-	attachedDiskSources := make(map[string]int, attachedDisksCount)
-	for i := 0; i < attachedDisksCount; i++ {
-		source, err := ParseDiskFieldValue(d.Get(fmt.Sprintf("attached_disk.%d.source", i)).(string), d, config)
+	attachedDiskSources := make(map[string]int)
+	for i, v := range d.Get("attached_disk").([]interface{}) {
+		if v == nil {
+			continue
+		}
+		disk := v.(map[string]interface{})
+		source, err := ParseDiskFieldValue(disk["source"].(string), d, config)
 		if err != nil {
 			return err
 		}
 		attachedDiskSources[source.RelativeLink()] = i
 	}
 
-	sIndex := 0
-	attachedDisks := make([]map[string]interface{}, attachedDisksCount)
+	attachedDisks := make([]map[string]interface{}, d.Get("attached_disk.#").(int))
 	scratchDisks := []map[string]interface{}{}
-	extraAttachedDisks := []map[string]interface{}{}
 	for _, disk := range instance.Disks {
 		if disk.Boot {
 			d.Set("boot_disk", flattenBootDisk(d, disk, config))
 		} else if disk.Type == "SCRATCH" {
 			scratchDisks = append(scratchDisks, flattenScratchDisk(disk))
-			sIndex++
 		} else {
 			source, err := ParseDiskFieldValue(disk.Source, d, config)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


google/resource_compute_instance.go
```diff
@@ -852,26 +852,26 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("label_fingerprint", instance.LabelFingerprint)
 	}
 
-	attachedDisksCount := d.Get("attached_disk.#").(int)
-	attachedDiskSources := make(map[string]int, attachedDisksCount)
-	for i := 0; i < attachedDisksCount; i++ {
-		source, err := ParseDiskFieldValue(d.Get(fmt.Sprintf("attached_disk.%d.source", i)).(string), d, config)
+	attachedDiskSources := make(map[string]int)
+	for i, v := range d.Get("attached_disk").([]interface{}) {
+		if v == nil {
```
rosbo: Can you add a quick comment about why we need to do this null check here.

PR #1316, opened 2018-04-10
Delete the default network created by the project.

### LINKED ISSUE #1312: Add option to delete default VPC on the new project creation
Hi,

This "issue" is more like a new feature, rather than a bug.

What I am missing so far in this provider is a capability to get rid off the default VPC, when creating a new GCP project and enabling "compute.googleapis.com" service API.

Google behalf of the requestor creates the default VPC and enables "Auto Network creation" option.
For our usecase we do not require to have a subnets in every single region, where GCP is present and we want to be in the control of the subnet creation process (in addition Google documentation claims, that for production workloads "Auto Network creation" option should be disabled)

As default VPC creation is baked in into the backend APIs - I see, only one way to implement it is delete it blindly within the "google_compute_network" resource etc.

```
resource "google_compute_network" "vpc_per_project" {
  name                    = "control-freaks-default-vpc-in-us"
  auto_create_subnetworks = false
  routing_mode            = "REGIONAL"
  description             = "Auto pre-created VPC for all US regions"
  project                 = "${google_project.new_project.id}"

# by having implemented smth like
# drop_default_vpc = true
# where the resource could call API and try to delete the default VPC (if it does exists)
}
```
as I cannot do the "terraform import" of the network before running a plan and apply new project setup - as project does not exists at that moment

It is really gets annoying, as for now only using UI or gcloud CLI I can delete this default VPC.

Hope that make sense.

### Terraform Version
Terraform v0.11.5
+ provider.google v1.8.0
+ provider.random v1.2.0

nat-henderson:
This will fix #1312.

### DIFF
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -338,7 +338,8 @@ func resourceGoogleProjectDelete(d *schema.ResourceData, meta interface{}) error
 }
 
 func resourceProjectImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-	// Explicitly set to default as a workaround for `ImportStateVerify` tests.
+	// Explicitly set to default as a workaround for `ImportStateVerify` tests, and so that users
+	// don't see a diff immediately after import.
 	d.Set("auto_create_network", true)
 	return []*schema.ResourceData{d}, nil
 }


ndmckinley:
Tests pass, except for `TestAccProjectOrganizationPolicy_list_denySome`, which is flaky.

google/resource_google_project.go
```diff
@@ -310,3 +336,9 @@ func resourceGoogleProjectDelete(d *schema.ResourceData, meta interface{}) error
 	d.SetId("")
 	return nil
 }
+
+func resourceProjectImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	// Explicitly set to default as a workaround for `ImportStateVerify` tests.
```
rosbo: It is not only for ImportStateVerify but also to ensure people don't see a diff after importing this resource.

ndmckinley: Great point, I'll modify that.

PR #1336, opened 2018-04-13
Fix auto-delete default network in google_project.

### LINKED ISSUE #1331: Error on auto_create_network = false
### Terraform Version
v0.11.7

### Affected Resource(s)
- google_project

Hi there,
I'm trying to do the following:
```hcl
resource "google_project" "folder" {
  name                = "pf-test-smoke"
  project_id          = "pf-test-smoke-april13-2"
  org_id              = "${var.org_id}"
  billing_account     = "${var.billing}"
  auto_create_network = "false"
}
```

But when I perform the terraform apply command, it creates the project but throws the following error:

```hcl
google_project.folder: Still creating... (10s elapsed)

Error: Error applying plan:

1 error(s) occurred:

* google_project.folder: 1 error(s) occurred:

* google_project.folder: Error deleting network: googleapi: Error 404: Failed to find project pf-test-smoke, notFound

```

I think it's searching project by name and not by id. project_id is "pf-test-smoke-april13-2" but the error says the project's name (pf-test-smoke).

I'm using the last terraform version and the compiled code from this provider.
I've also tested it with folder (not org_id) and fails too.

```hcl
Thanks in advance!
```

rosbo:
- Use ProjectId when calling the APIs instead of the project name
- The compute API has to be enabled before we can delete the network
- The firewall rules inside the default network must be deleted before we delete the network.
- Update documentation to clarify this side effect.

Fixes #1331

### DIFF
--- a/google/resource_compute_network.go
+++ b/google/resource_compute_network.go
@@ -195,9 +195,12 @@ func resourceComputeNetworkDelete(d *schema.ResourceData, meta interface{}) erro
 		return err
 	}
 
-	// Delete the network
+	return deleteComputeNetwork(project, d.Id(), config)
+}
+
+func deleteComputeNetwork(project, network string, config *Config) error {
 	op, err := config.clientCompute.Networks.Delete(
-		project, d.Id()).Do()
+		project, network).Do()
 	if err != nil {
 		return fmt.Errorf("Error deleting network: %s", err)
 	}
@@ -206,7 +209,5 @@ func resourceComputeNetworkDelete(d *schema.ResourceData, meta interface{}) erro
 	if err != nil {
 		return err
 	}
-
-	d.SetId("")
 	return nil
 }
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -151,13 +151,17 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
 	// people if we don't have to.  The GCP Console is doing the same thing - creating
 	// a network and deleting it in the background.
 	if !d.Get("auto_create_network").(bool) {
-		op, err := config.clientCompute.Networks.Delete(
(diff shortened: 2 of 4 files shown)


rosbo:
Test is failling because the default network is created with 4 default firewall rules and these have to be deleted before we can delete the network:

```sh
--- FAIL: TestAccProject_deleteDefaultNetwork (118.51s)
    testing.go:513: Step 0 error: Error applying: 1 error(s) occurred:
        
        * google_project.acceptance: 1 error(s) occurred:
        
        * google_project.acceptance: The network resource 'projects/terraform-619kodgqiy/global/networks/default' is already being used by 'projects/terraform-619kodgqiy/global/firewalls/default-allow-rdp'  
```

ndmckinley:
Ah, dang, that's totally my fault for not testing my code.  I see the test failure but I'll take it from here if you like, no reason for you to keep working fixing my mistake.

google/resource_compute_network.go
```diff
@@ -195,9 +195,12 @@ func resourceComputeNetworkDelete(d *schema.ResourceData, meta interface{}) erro
 		return err
 	}
 
-	// Delete the network
+	return deleteComputeNetwork(project, d.Id(), config)
```
ndmckinley: You'll still want a d.SetId("") if err == nil, right?

rosbo:
```sh
--- PASS: TestAccProject_deleteDefaultNetwork (206.61s)
```

@ndmckinley tests are passing. PTAL

google/resource_compute_network.go
rosbo: The `d.SetId("")` is actually redundant: https://github.com/hashicorp/terraform/blob/master/helper/schema/resource.go#L205

ndmckinley: I did not know that!

PR #1337, opened 2018-04-14
WIP: Fix IAM tests

danawillow:
Fixing tests failures introduced in #1322 and #1327

### DIFF
--- a/google/iam_kms_crypto_key.go
+++ b/google/iam_kms_crypto_key.go
@@ -2,6 +2,7 @@ package google
 
 import (
 	"fmt"
+
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/schema"
 	"google.golang.org/api/cloudkms/v1"
@@ -35,8 +36,13 @@ func NewKmsCryptoKeyIamUpdater(d *schema.ResourceData, config *Config) (Resource
 	}, nil
 }
 
-func CryptoIdParseFunc(d *schema.ResourceData, _ *Config) error {
-	d.Set("crypto_key_id", d.Id())
+func CryptoIdParseFunc(d *schema.ResourceData, config *Config) error {
+	cryptoKeyId, err := parseKmsCryptoKeyId(d.Id(), config)
+	if err != nil {
+		return err
+	}
+	d.Set("crypto_key_id", cryptoKeyId.cryptoKeyId())
+	d.SetId(cryptoKeyId.cryptoKeyId())
 	return nil
 }
 


ndmckinley:
Tests pass!  :)

google/resource_kms_crypto_key.go
```diff
@@ -42,6 +43,16 @@ func resourceKmsCryptoKey() *schema.Resource {
 	}
 }
 
+func kmsCryptoKeyRingsEquivalent(k, old, new string, d *schema.ResourceData) bool {
```
rosbo: Hum, I thought the whole point of #1322 was to have a consistent id.

ndmckinley: And we still do, but the user might provide the long form in their config.  See how this only tries to parse the long-form out of `new` (e.g. the value in the config) instead of `old` (e.g. the value in state).

google/resource_kms_key_ring.go
```diff
@@ -79,7 +80,7 @@ func resourceKmsKeyRingCreate(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] Created KeyRing %s", keyRing.Name)
 
-	d.SetId(keyRingId.terraformId())
+	d.SetId(keyRingId.keyRingId())
```
rosbo: Is #terraformId() now unused? If so, we should remove it

google/resource_kms_crypto_key.go
```diff
@@ -42,6 +43,16 @@ func resourceKmsCryptoKey() *schema.Resource {
...
+	keyRingIdWithSpecifiersRegex := regexp.MustCompile("^projects/([a-z0-9-]+)/locations/([a-z0-9-])+/keyRings/([a-zA-Z0-9_-]{1,63})$")
```
rosbo: Rely on `parseKmsKeyRingId` method instead. Parse the new with that method and compare the value with the old  value which should be canonical.

ndmckinley: Can't - that requires a `Config` object not available at `DiffSuppressFunc` time.

google/resource_kms_key_ring.go
ndmckinley: Done.

ndmckinley: Ah, got it - just convenience in the key tests.  Since we have to build up the key ring before we build the key, and since the key's ID isn't known back in the test method, it's just easier to call terraformId() on the key ring then add a slash.

PR #1340, opened 2018-04-16
Fix Instance Group Manager and Region Instance Group Managers plan crash

dwatson:
When either an Instance Group Manager or Region Instance Group Manager exists in tfstate but the resource itself no longer exists, running a terraform plan causes a crash.  This fix ensures that terraform plan correctly indicates that the resource should be created.

### DIFF
--- a/google/resource_compute_instance_group_manager.go
+++ b/google/resource_compute_instance_group_manager.go
@@ -419,7 +419,7 @@ func resourceComputeInstanceGroupManagerRead(d *schema.ResourceData, meta interf
 	}
 
 	manager, err := getManager(d, meta)
-	if err != nil {
+	if err != nil || manager == nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


google/resource_compute_instance_group_manager.go
```diff
@@ -419,7 +419,7 @@ func resourceComputeInstanceGroupManagerRead(d *schema.ResourceData, meta interf
 	}
 
 	manager, err := getManager(d, meta)
-	if err != nil {
+	if err != nil || manager == nil {
```
ndmckinley: This additional condition will cause some unexpected results, I think - you'll return `nil` as the error, where I think you probably want something like `fmt.Errorf("Instance Group Manager %s does not exist", d.Id())`.  What do you think?

dwatson: If we return an error from the `getManager` function, this will be returned here in `resourceComputeInstanceGroupManagerRead` and terraform will display the error to the user. 
 In order for `terraform plan` to work `resourceComputeInstanceGroupManagerRead` needs to return `nil`.

ndmckinley: Aha, you do the `d.SetId("")` in the `getManager` function, I see it now.  Yes, you are right.

ndmckinley:
I'm running the tests now.

ndmckinley:
The test runner is dying with a deadlock - I doubt that's your fault, but it's delaying me getting these tests run.

PR #1349, opened 2018-04-17
Add DNS PTR record to Compute Instance

### LINKED ISSUE #1333: [feature request] - Set instance Public DNS PTR Record
### Terraform Version
0.11.7

### Affected Resource(s)
- google_compute_instance

While this isn't a process that is 100% automate-able, as you have to go to the google webmaster console to claim the name with a TXT record first, it would be helpful if I could set the Public DNS PTR Record for an instance via Terraform.

https://cloud.google.com/compute/docs/instances/create-ptr-record

ndmckinley:
Fixes #1333.


ndmckinley:
Required a revendor - I'm going to run all the compute tests on this one.

ndmckinley:
```
[TestAccComputeInstance_PTRRecord] [Test Output]
=== RUN   TestAccComputeInstance_PTRRecord
--- PASS: TestAccComputeInstance_PTRRecord (80.11s)
PASS
```

google/resource_compute_instance.go
```diff
@@ -1120,6 +1125,10 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
 					Type:  "ONE_TO_ONE_NAT",
 					NatIP: d.Get(acPrefix + ".nat_ip").(string),
 				}
+				if ptr, ok := d.GetOk(acPrefix + ".public_ptr_domain_name"); ok {
+					ac.SetPublicPtr = true
```
rosbo: Do you need to set `SetPublicPtr` explicitely to true to turn it off if some was using `public_ptr_domain_name` and now removes it from their config?

ndmckinley: I don't really know why the API requires that `SetPublicPtr` be true *and* `PublicPtrDomainName` have the value, rather than just doing a null check on the latter, but the docs are clear about that.  Creating the new `AccessConfig` means that it won't be set unless that's set in the current config - we do a full replace.

rosbo: We got bit a few times by that when in order to disable a feature in update, we had to use ForceSend false and nil values otherwise, the server would keep the value it currently had.

ndmckinley: Makes sense!  I don't think it can happen here at all since we do a full replace.  Let me add that to the test to be 100% sure.

ndmckinley:
Test passes, even with imports, so it's definitely not a problem.

PR #1360, opened 2018-04-19
Bump GKE version.

paddycarver:
Tests were failing with an invalid version error, so bumping the version
like this should fix the tests.


google/resource_container_node_pool_test.go
```diff
@@ -635,8 +635,8 @@ resource "google_container_cluster" "cluster" {
 	name = "tf-cluster-nodepool-test-%s"
 	zone = "us-central1-c"
 	initial_node_count = 1
-  node_version = "1.9.2-gke.1"
-  min_master_version = "1.9.2-gke.1"
+  node_version = "1.9.6-gke.1"
```
rosbo: indentation is funky

PR #1380, opened 2018-04-25
Use DNS Name to retrieve DNS Managed Zone

### LINKED ISSUE #1255: Feature Request: DNS Managed Zone data source by DNS Name, not just Zone Name
### Affected Resource
- data google_dns_managed_zone

### Expected Behavior
A dns managed zone can be found by it's dns name, not just the zone name.

### Actual Behavior
The only way to retrieve a dns managed zone with a terraform google data source is via the zone name. With the aws terraform provider, you retrieve the zone with the dns name.

Would you be willing to accept a PR for this functionality?

genevieve:
Fixes #1255

### DIFF
--- a/google/data_source_dns_managed_zone.go
+++ b/google/data_source_dns_managed_zone.go
@@ -15,6 +15,7 @@ func dataSourceDnsManagedZone() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"dns_name": &schema.Schema{
 				Type:     schema.TypeString,
+				Computed: true,
 				Optional: true,
 			},
 
--- a/google/data_source_dns_managed_zone.go
+++ b/google/data_source_dns_managed_zone.go
@@ -1,6 +1,12 @@
 package google
 
-import "github.com/hashicorp/terraform/helper/schema"
+import (
+	"errors"
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	dns "google.golang.org/api/dns/v1"
+)
 
 func dataSourceDnsManagedZone() *schema.Resource {
 	return &schema.Resource{
@@ -9,12 +15,12 @@ func dataSourceDnsManagedZone() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"dns_name": &schema.Schema{
 				Type:     schema.TypeString,
-				Computed: true,
+				Optional: true,
 			},
 
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 			},
 
 			"description": &schema.Schema{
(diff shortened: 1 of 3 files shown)


google/data_source_dns_managed_zone.go
```diff
@@ -9,12 +15,12 @@ func dataSourceDnsManagedZone() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"dns_name": &schema.Schema{
 				Type:     schema.TypeString,
-				Computed: true,
```
emilymye: I think this should still be computed as Computed values are still Optional, but DNS name is probably a required value for a managed zone. Even though we set the value to the GCP resource's value right now, Optional can potentially cause this value to not be set.

emilymye:
Nope, sorry that was a confusing comment - this (with "Optional: true") is good.

google/data_source_dns_managed_zone_test.go
```diff
@@ -60,13 +64,23 @@ func testAccDataSourceDnsManagedZoneCheck(dsName, rsName string) resource.TestCh
...
 resource "google_dns_managed_zone" "foo" {
-	name		= "qa-zone-%s"
-	dns_name	= "qa.test.com."
-	description	= "QA DNS zone"
+	name		= "foo-zone-%s"
+	dns_name	= "foo.test.com."
+	description	= "Foo DNS zone"
 }
 
-data "google_dns_managed_zone" "qa" {
```
emilymye: I think we wanted to keep this a different name, to differentiate between resource "google_dns_managed_zone.foo" and data source "google_dns_managed_zone.qa" - same with below. You can name to data-foo or something like that if you're against the "qa" name.

genevieve: Ok! I can change that back

google/data_source_dns_managed_zone.go
```diff
@@ -10,11 +16,12 @@ func dataSourceDnsManagedZone() *schema.Resource {
 			"dns_name": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
+				Optional: true,
 			},
 
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
```
paddycarver: You may want to take advantage of [`ConflictsWith`](https://godoc.org/github.com/hashicorp/terraform/helper/schema#Schema.ConflictsWith) to ensure both these values can't be set.

google/data_source_dns_managed_zone_test.go
```diff
@@ -68,5 +72,15 @@ resource "google_dns_managed_zone" "foo" {
 data "google_dns_managed_zone" "qa" {
 	name	= "${google_dns_managed_zone.foo.name}"
 }
-`, acctest.RandString(10))
+
+resource "google_dns_managed_zone" "bar" {
+	name		= "bar-zone-%s"
+	dns_name	= "bar.test.com."
```
paddycarver: Could we use `tf-test.club` instead of `test.com`? It helps to make sure we don't run into domain validation errors when running the tests.

website/docs/d/dns_managed_zone.html.markdown
```diff
@@ -32,7 +32,9 @@ resource "google_dns_record_set" "dns" {
 
 ## Argument Reference
 
-* `name` - (Required) A unique name for the resource.
+* `name` - (Optional) A unique name for the resource. Required if `dns_name` is not set.
+
+* `dns_name` - (Optional) The fully qualified DNS name of this zone, e.g. `terraform.io.`.  Required if `name` is not set.
```
paddycarver: Should this mention that if multiple zones have the same DNS name, it will use the first?

PR #1396, opened 2018-04-27
Add issue_client_certificate to cluster

### LINKED ISSUE #1378: [google_container_cluster] Support not issuing Client Key on K8S cluster creation
`gcloud beta container cluster create` has option `--no-issue-client-certificate`, which avoids creating a client key and fully leverages gcloud IAM.

### Terraform Version
```
Terraform v0.11.7
+ provider.google v1.10.0
```

### Affected Resource(s)
- google_container_cluster

### Expected Behavior
Have a way to disable cert creation in `masterAuth` attribute

### Actual Behavior
Client Key / Cert is created along with the cluster and the state file contains sensitive, long term , not easy to rotate credentials !!!

emilymye:
Fixes #1378. 
* Also changed test name so go test filter doesn't conflict with  MasterAuthorization...Tests

make testacc TEST=./google TESTARGS='--run="TestAccContainerCluster_withMasterAuthConfig"'
TF_ACC=1 go test ./google -v --run="TestAccContainerCluster_withMasterAuthConfig" -timeout 120m
=== RUN   TestAccContainerCluster_withMasterAuthConfig
=== PAUSE TestAccContainerCluster_withMasterAuthConfig
=== RUN   TestAccContainerCluster_withMasterAuthConfig_NoCert
=== PAUSE TestAccContainerCluster_withMasterAuthConfig_NoCert
=== CONT  TestAccContainerCluster_withMasterAuthConfig
=== CONT  TestAccContainerCluster_withMasterAuthConfig_NoCert
--- PASS: TestAccContainerCluster_withMasterAuthConfig_NoCert (394.82s)
--- PASS: TestAccContainerCluster_withMasterAuthConfig (396.11s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	396.281s

TF_ACC=1 go test ./google -v --run="TestAccContainerClusterDatasource_basic" -timeout 120m
=== RUN   TestAccContainerClusterDatasource_basic
=== PAUSE TestAccContainerClusterDatasource_basic
=== CONT  TestAccContainerClusterDatasource_basic
--- PASS: TestAccContainerClusterDatasource_basic (311.69s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	311.870s

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -282,6 +282,24 @@ func resourceContainerCluster() *schema.Resource {
 							ForceNew: true,
 						},
 
+						"client_certificate_config": {
+							Type:             schema.TypeList,
+							MaxItems:         1,
+							Optional:         true,
+							DiffSuppressFunc: masterAuthClientCertCfgSuppress,
+							ForceNew:         true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"issue_client_certificate": {
+										Type:             schema.TypeBool,
+										Required:         true,
+										ForceNew:         true,
+										DiffSuppressFunc: masterAuthClientCertCfgSuppress,
+									},
+								},
+							},
+						},
+
 						"client_certificate": {
 							Type:     schema.TypeString,
 							Computed: true,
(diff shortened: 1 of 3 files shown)


google/resource_container_cluster.go
```diff
@@ -747,6 +774,11 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 			"cluster_ca_certificate": cluster.MasterAuth.ClusterCaCertificate,
 		},
 	}
+	if len(cluster.MasterAuth.ClientCertificate) == 0 {
+		masterAuth[0]["client_certificate_config"] = []map[string]interface{}{
+			{"issue_client_certificate": false},
+		}
```
rosbo: If you set `issue_client_certificate` to `true` in a else statement to this if,  I think it would simplify `masterAuthClientCertCfgSuppress` since it will only need to handle one case:
`[{ "issue_client_certificate": true}] --> []`

emilymye: I wanted to do it like that first :/ The problem with that approach is that we don't actually have access to the old data (i.e. issue_client_certificate=true) in the diff suppress function, and because the new value is nil, it doesn't attempt to do the diff of issue_client_certificate.

rosbo: I see 👍

ellipticasec:
Thanks @emilymye you rock !

PR #1397, opened 2018-04-28
Generate resource entries in ResourcesMap in provider.go for generated resources.

rosbo:
Once https://github.com/GoogleCloudPlatform/magic-modules/pull/132 is merged, the `provider_compute_gen.go` file will be automatically generated.


google/provider.go
```diff
@@ -93,130 +93,127 @@ func Provider() terraform.ResourceProvider {
 			"google_compute_backend_service":         dataSourceGoogleComputeBackendService(),
 		},
 
-		ResourcesMap: map[string]*schema.Resource{
-			"google_bigquery_dataset":                      resourceBigQueryDataset(),
```
rosbo: The only different is that I removed the 6 resources in GeneratedComputeResourcesMap from this list

PR #1422, opened 2018-05-02
WIP: remove switch statements between api versions for compute

danawillow:
(TODO: put actual description/explanation here when ready)


rileykarson:
Very exciting! I'm a big supporter of this change, I'm glad that we can get rid of the scaffolding and clean up the logic inside beta resources. As you said, this change simplifies development on resources with beta features again; in my opinion, this is especially beneficial for Terraform users who need a beta GCP feature and are interested in becoming community contributors to add it. They won't need to decipher what all the switch cases and `Convert` calls are for anymore!

In that vein, I'm curious if there are any beta compute features in the backlog that could have support added soon after merging this change. Previously, when introducing the versioned resources with beta scaffolding, @rosbo and I made a handful of changes introducing beta features to non-beta resources whose commit contents were (hopefully!) helpful to the couple of community contributors who worked with versioned resources since then.

By making a simple commit adding a beta feature to a non-beta resource, we'll be able to easily show anyone who contributes to the provider what changes are necessary for a beta feature now, such as changing the API client version, guarding the self links, and using `computeSharedOperationWait` over `computeOperationWait`. 

---

We can also probably just eliminate `computeSharedOperationWait` and replace occurrences with `computeBetaOperationWait` for a small code complexity win too! Since afaik we should always know the version of `Operation` we are working with now.

danawillow:
Merged, but @paddycarver if you have any comments I'm happy to make follow-up PRs to address them.

paddycarver:
From chat:

> Looks like my beta-ifying prs broke a bunch of tests because it uncovered places where we weren't setting doing the self link conversion. feel free to roll them back

I think we're going to roll this back quickly just so we can get the tests in a good spot, but I'm in favor of the idea, in general.

PR #1431, opened 2018-05-04
Store region name only for vpn gateway

rosbo:
Once @ndmckinley's work to support custom in magic-modules is in, we can go back to autogenerating this resource


rosbo:
```sh
--- PASS: TestAccComputeVpnGateway_basic (61.43s)
```

PR #1452, opened 2018-05-08
add ability to retry cloud functions on failure

### LINKED ISSUE #1164: Failure to create resilient background functions (i.e. retry on failure)
Hello,

We are triggering CloudFunctions upon new messages being inserted in to a PubSub topic. The team spent time investigating why the CloudFunctions were not being re-invoked upon failure. Turns out we can't configure the `retry on failure` property of Background CloudFunctions. Can a retry upon failure parameter be added to `google_cloudfunctions_function`. Or are we missing something in the current setup?

danawillow:
Fixes #1164.

I opted to top-level this because we already have precedent in this resource for top-leveling instead of nesting, but I could be persuaded otherwise. Also, the API for this is just to send an empty Retry object if you want to enable retries, which would lead to tf configs that look like this if we kept to the API:
```tf
resource "google_cloudfunctions_function" "function" {
  name = "whatever"
  // other properties
  failure_policy {
    retry {
    }
  }
}
```
since right now, the only thing in a failure policy that can be set is retry, which has no attributes of its own.
(see https://cloud.google.com/functions/docs/reference/rest/v1/projects.locations.functions)

### DIFF
--- a/google/resource_cloudfunctions_function.go
+++ b/google/resource_cloudfunctions_function.go
@@ -197,6 +197,12 @@ func resourceCloudFunctionsFunction() *schema.Resource {
 				Computed: true,
 			},
 
+			"retry_on_failure": {
+				Type:          schema.TypeBool,
+				Optional:      true,
+				ConflictsWith: []string{"trigger_http"},
+			},
+
 			"project": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


Farhie:
Awesome, thanks @danawillow!

PR #1468, opened 2018-05-10
fix: replica_configuration.password  to be sensitive

orisano:
currently, google_sql_database_instance.replica_configuration.password is not sensitve.
i think that sensitive is better.

thank you


ndmckinley:
I think you're right.  Running the tests now.

ndmckinley:
Just a timeout!  Merging.

PR #1474, opened 2018-05-10
Check current status during apply-time diff.

### LINKED ISSUE #54: Recreating a google_compute_instance removes it from instance groups
_This issue was originally opened by @danawillow as hashicorp/terraform#10896. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

**To reproduce:**

* Run example in https://github.com/hashicorp/terraform/pull/10844 with `terraform apply`
* Make a change to `startup.sh`
* Run `terraform apply` again

**Expected result:**

* Instances are deleted and recreated, and re-added to their instance groups

**Actual result:**

* Instances are deleted and recreated but not re-added to their instance groups.

@evandbrown let's go over this tomorrow

ndmckinley:
This fixes #54!

This adds three reads to the `terraform apply` cycle - one at plan-time in the `CustomizeDiff`, one at apply-time in the `CustomizeDiff`, and one full `Read` execution at Update time.

### DIFF
--- a/google/resource_compute_instance_group.go
+++ b/google/resource_compute_instance_group.go
@@ -177,20 +177,19 @@ func customDiffInstanceGroupInstancesField(diff *schema.ResourceDiff, meta inter
 			// Any other errors return them
 			return fmt.Errorf("Error reading InstanceGroup Members: %s", err)
 		}
-	} else {
-		for _, member := range members.Items {
-			memberUrls = append(memberUrls, member.Instance)
-		}
-		sort.Strings(memberUrls)
-		sort.Strings(oldInstances)
-		sort.Strings(newInstances)
-		log.Printf("[DEBUG] InstanceGroup members: %#v.  OldInstances: %#v", memberUrls, oldInstances)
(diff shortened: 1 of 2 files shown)


ndmckinley:
Tests are passing!  :)

google/resource_compute_instance_group.go
```diff
@@ -94,7 +96,104 @@ func resourceComputeInstanceGroup() *schema.Resource {
...
+		}).Do()
+	if err != nil {
+		if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 404 {
+			// This is where we'll end up at either plan time or apply time on first creation.
+			return nil
+		} else {
+			// Any other errors return them
+			return fmt.Errorf("Error reading InstanceGroup Members: %s", err)
+		}
+	} else {
```
danawillow: nit: this else is unnecessary since the if statement it's attached to will always return

PR #1475, opened 2018-05-10
Add some more test cases to subnetwork flow logs.

paddycarver:
Test what happens when adding enable_flow logs to compute_subnetwork and
what happens when it's removed.

### DIFF
--- a/google/resource_compute_subnetwork_test.go
+++ b/google/resource_compute_subnetwork_test.go
@@ -171,7 +171,69 @@ func TestAccComputeSubnetwork_secondaryIpRanges(t *testing.T) {
 	})
 }
 
-func TestAccComputeSubnetwork_flowLogs(t *testing.T) {
+func TestAccComputeSubnetwork_createFlowLogsEnabled(t *testing.T) {
+	t.Parallel()
+
+	var subnetwork compute.Subnetwork
+
+	cnName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	subnetworkName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeSubnetworkDestroy,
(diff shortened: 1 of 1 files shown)


google/resource_compute_subnetwork_test.go
```diff
@@ -184,6 +184,15 @@ func TestAccComputeSubnetwork_flowLogs(t *testing.T) {
 		Providers:    testAccProviders,
 		CheckDestroy: testAccCheckComputeSubnetworkDestroy,
 		Steps: []resource.TestStep{
+			{
+				Config: testAccComputeSubnetwork_flowLogsUnset(cnName, subnetworkName),
```
rosbo: Since flow is not set in this first step, the following steps are updating the resource. We are not testing the case where we want to enable flow at `Create` time. 

I think the best would be to have:
Step 1: flow enabled (test the create path)
Step 2: flow disabled
Step 3: flow enabled (test the update path)

paddycarver: You're right, thanks! I'll add another test to follow the Create path.

paddycarver:
Got around to updating this, sorry for the delay. I'll be honest, I forgot about it. @danawillow or @ndmckinley, are either of you available to take a look at this, as @rosbo isn't an active maintainer (I believe?) anymore?

google/resource_compute_subnetwork_test.go
```diff
@@ -198,6 +198,77 @@ func TestAccComputeSubnetwork_flowLogs(t *testing.T) {
...
+	cnName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	subnetworkName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeSubnetworkDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccComputeSubnetwork_flowLogs(cnName, subnetworkName, true),
```
danawillow: shouldn't this be false? (and also in the check below)

PR #1487, opened 2018-05-13
Compare region_backend_service.backend[].group as a relative path

tmshn:
In a `region_backend_service` resource, `backend[].group` argument is compared using full url. Therefore, if you pass a `self_link` of `instance_group` with beta feature enabled, you'll suffer from persisting perpetual diff.

E.g.:

```hcl
resource "google_compute_instance_group_manager" "foo" {
  // ...configs (omitted)...
  // ↓ This is beta feature
  update_strategy       = "ROLLING_UPDATE" 
  rolling_update_policy = {
    type           = "PROACTIVE"
    minimal_action = "REPLACE"
  }
}

resource "google_compute_region_backend_service" "foo" {
  // ...configs (omitted)...
  backend = [
    {
        group = "${google_compute_instance_group_manager.foo.self_link}"
    },
  ]
}
```

With this configuration, following diff always occurs.
```
------------------------------------------------------------------------

  ~ update in-place

  ~ google_compute_region_backend_service.foo
      backend.1111111111.description: "" => ""
      backend.1111111111.group:       "https://www.googleapis.com/compute/v1/projects/my-project/zones/us-central1-a/instanceGroups/igm-foo" => ""
      backend.2222222222.description: "" => ""
      backend.2222222222.group:       "" => "https://www.googleapis.com/compute/beta/projects/my-project/zones/us-central1-a/instanceGroups/igm-foo"

------------------------------------------------------------------------
```

To suppress this, I set `compareSelfLinkRelativePaths` as `DiffSuppressFunc` for `backend[].group`, like non-regional `backend_service` does.
https://github.com/terraform-providers/terraform-provider-google/blob/v1.12.0/google/resource_compute_backend_service.go#L80

### DIFF
--- a/google/resource_compute_region_backend_service.go
+++ b/google/resource_compute_region_backend_service.go
@@ -39,8 +39,9 @@ func resourceComputeRegionBackendService() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"group": &schema.Schema{
-							Type:     schema.TypeString,
-							Optional: true,
+							Type:             schema.TypeString,
+							Optional:         true,
+							DiffSuppressFunc: compareSelfLinkRelativePaths,
 						},
 						"description": &schema.Schema{
 							Type:     schema.TypeString,


tmshn:
Should I add a test for this?

emilymye:
I actually think this is the appropriate behavior - if a resource is changed to a beta resource, it should be recreated. If after the initial v1 --> beta change the diff persists, that's more an issue around how we compute the instance group for the initial diff.

danawillow:
@emilymye that's not actually true- v1 and beta resources are the same underneath. the self_link only describes which API we used to access it. I'm going to go ahead and merge this, hope you don't mind :)

danawillow:
@tmshn whoops, merged before realizing you'll also need to change the hash function. take a look at how we do it for regular backend services- it's already been fixed there.

tmshn:
@danawillow Oh I missed that! Thanks to point out. I fixed it in #1491.

PR #1489, opened 2018-05-14
compute address: use the specified project and region fields

vishen:
If the project or region fields are specified in the compute address resource use them as the fallback for parseComputeAddressId(). 

The current implementation ignored the project and region fields and used whatever was in the config as defaults.

### DIFF
--- a/google/data_source_google_compute_address_test.go
+++ b/google/data_source_google_compute_address_test.go
@@ -80,7 +80,7 @@ func testAccCheckDataSourceComputeAddressDestroy(resource_name string) resource.
 			return fmt.Errorf("can't find %s in state", resource_name)
 		}
 
-		addressId, err := parseComputeAddressId(rs.Primary.ID, nil)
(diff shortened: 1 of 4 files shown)


vishen:
When I run an import on a compute address that is in a different project I was getting errors because it was trying to get it from the current project as opposed to the project I specified in the resource.

```hcl
# In project A
resource "google_compute_address" "my-ip" {
    name = "myip"
    project = "project-b"
}
```
`terraform import google_compute_address.my-ip myip` will try and get the compute address from `project-a` as opposed to `project-b`.

I think the change only needs to happen in the `resourceComputeAddressImportState` but I added the others for consistency sake.

danawillow:
Just taking a look at this, I'm not sure it'll work. Import functions don't usually have any reference to the fields that are set. Have you tested it?

By the way, you can already import address resources in a different project: https://www.terraform.io/docs/providers/google/r/compute_address.html#import.

PR #1490, opened 2018-05-14
Fix order of getting project for storage notification topic

### LINKED ISSUE #1486: google_storage_notification has no 'project' field
`google_storage_notification` has no project field. 
And the only way to make the terraform succeed is to explicitly configure the google provider project in your terraform project. 
it will fail with:
```
Error: Error applying plan:

1 error(s) occurred:

* google_storage_notification.notification: 1 error(s) occurred:

* google_storage_notification.notification: project: required field is not set

above and apply again to incrementally change your infrastructure
```

Which comes from copy & pasting from the official doc.
Also the [Documentation](https://www.terraform.io/docs/providers/google/r/storage_notification.html)
is wrong (    event_types       = ["%s","%s"])

emilymye:
Fixes #1486.

* See if the topic includes the project name before checking config.  
* Also, make documentation use actual constants rather than '%s', since it might be confusing 

Should I add a project field for the topic as well? Worried that if I add it, it will get confused with a project associated with the bucket/notification.

### DIFF
--- a/google/resource_storage_notification.go
+++ b/google/resource_storage_notification.go
@@ -79,12 +79,15 @@ func resourceStorageNotificationCreate(d *schema.ResourceData, meta interface{})
 
 	bucket := d.Get("bucket").(string)
 
-	project, err := getProject(d, config)
-	if err != nil {
-		return err
+	topicName := d.Get("topic").(string)
+	computedTopicName := getComputedTopicName("", topicName)
+	if computedTopicName != topicName {
+		project, err := getProject(d, config)
+		if err != nil {
+			return err
+		}
+		computedTopicName = getComputedTopicName(project, topicName)
 	}
-
-	computedTopicName := getComputedTopicName(project, d.Get("topic").(string))
 
 	storageNotification := &storage.Notification{
 		CustomAttributes: expandStringMap(d, "custom_attributes"),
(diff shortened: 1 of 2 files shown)


rosbo:
Can you connect a topic and a storage bucket from different projects?

google/resource_storage_notification.go
```diff
@@ -79,13 +79,16 @@ func resourceStorageNotificationCreate(d *schema.ResourceData, meta interface{})
 
 	bucket := d.Get("bucket").(string)
 
-	project, err := getProject(d, config)
-	if err != nil {
-		return err
+	topicName := d.Get("topic").(string)
+	computedTopicName := getComputedTopicName("", topicName)
```
rosbo: Would that make sense to move this logic inside `getComputedTopicName`?

emilymye:
make testacc TEST=./google TESTARGS='--run="TestAccStorageNotification*"'
TF_ACC=1 go test ./google -v --run="TestAccStorageNotification*" -timeout 120m
=== RUN   TestAccStorageNotification_basic
=== PAUSE TestAccStorageNotification_basic
=== RUN   TestAccStorageNotification_withEventsAndAttributes
=== PAUSE TestAccStorageNotification_withEventsAndAttributes
=== CONT  TestAccStorageNotification_basic
=== CONT  TestAccStorageNotification_withEventsAndAttributes
--- PASS: TestAccStorageNotification_withEventsAndAttributes (12.50s)
--- PASS: TestAccStorageNotification_basic (15.04s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	15.218s

PR #1491, opened 2018-05-15
Fixed region_backend_service to calc hash using relative path

tmshn:
Along the change on region_backend_service.backend[].group to use DiffSuppressFunc (#1487), we also need to change hash function, as [pointed out](https://github.com/terraform-providers/terraform-provider-google/pull/1487#issuecomment-389000239).

Fixed this in a same way regular backend services do.
https://github.com/terraform-providers/terraform-provider-google/blob/v1.12.0/google/resource_compute_backend_service_migrate.go#L102-L103

### DIFF
--- a/google/resource_compute_region_backend_service.go
+++ b/google/resource_compute_region_backend_service.go
@@ -331,7 +331,8 @@ func resourceGoogleComputeRegionBackendServiceBackendHash(v interface{}) int {
 	var buf bytes.Buffer
 	m := v.(map[string]interface{})
 
-	buf.WriteString(fmt.Sprintf("%s-", m["group"].(string)))
+	group, _ := getRelativePath(m["group"].(string))
+	buf.WriteString(fmt.Sprintf("%s-", group))
 
 	if v, ok := m["description"]; ok {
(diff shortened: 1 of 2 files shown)


google/resource_compute_region_backend_service.go
```diff
@@ -331,7 +331,8 @@ func resourceGoogleComputeRegionBackendServiceBackendHash(v interface{}) int {
 	var buf bytes.Buffer
 	m := v.(map[string]interface{})
 
-	buf.WriteString(fmt.Sprintf("%s-", m["group"].(string)))
+	group, _ := getRelativePath(m["group"].(string))
```
emilymye: This should probably still use m["group"] if getRelativePath fails.

danawillow:
I agree with @emilymye! @tmshn, mind making that change here and also in backend_service? Let me know if you won't have time to do it; since this is a pretty important bugfix I'd like to have it in before our next release (hoping for Monday)

tmshn:
Fixed the point: dc71aaf (if I understand you correctly)

danawillow:
👍 from me! I'll let @emilymye have final say.

PR #1496, opened 2018-05-15
Allow bigquery regional location

### LINKED ISSUE #1492: Support BigQuery dataset regional location
BigQuery supports [regional location](https://cloud.google.com/bigquery/docs/dataset-locations?hl=en) which is not supported according to [the documentation](https://www.terraform.io/docs/providers/google/r/bigquery_dataset.html).

### Terraform Version

```
$ terraform -v
Terraform v0.11.7
+ provider.google v1.12.0
```

### Affected Resource(s)

- google_bigquery_dataset

### Expected Behavior

Able to specify regional locations.

emilymye:
Fixes #1492

### DIFF
--- a/google/resource_bigquery_dataset.go
+++ b/google/resource_bigquery_dataset.go
@@ -72,7 +72,7 @@ func resourceBigQueryDataset() *schema.Resource {
 				Optional:     true,
 				ForceNew:     true,
 				Default:      "US",
-				ValidateFunc: validation.StringInSlice([]string{"US", "EU"}, false),
+				ValidateFunc: validation.StringInSlice([]string{"US", "EU", "asia-northeast1"}, false),
 			},
 
 			// DefaultTableExpirationMs: [Optional] The default lifetime of all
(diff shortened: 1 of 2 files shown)


google/resource_bigquery_dataset.go
```diff
@@ -72,7 +72,7 @@ func resourceBigQueryDataset() *schema.Resource {
 				Optional:     true,
 				ForceNew:     true,
 				Default:      "US",
-				ValidateFunc: validation.StringInSlice([]string{"US", "EU"}, false),
+				ValidateFunc: validation.StringInSlice([]string{"US", "EU", "asia-northeast1"}, false),
```
rosbo: Can you remove the possible values from the comment at line 68. The comment is stale and the ValidateFunc makes it obvious what are the valid values. Same applies for the bit in the comment talking about the default value.

emilymye: Done!

PR #1512, opened 2018-05-20
website/docs/r/google_service_account_key.html.markdown: fixed docs

vishen:
Fixed some _possibly_ copy-and-pasted documentation in the `google_service_account_key` resource. I've kept it similar to the other resource definition values in `google_service_account_key` docs.


danawillow:
Whoops, thanks @vishen!

PR #1514, opened 2018-05-21
Fix cloudiot example

magiconair:
The topic names for state and event notifications are the other way around.


danawillow:
Thanks @magiconair!

magiconair:
@danawillow That was quick. Thx!

PR #1516, opened 2018-05-21
Update PostGRES cloudsql docs

PepperTeasdale:
This is now GA, so removing that caveat.


danawillow:
Thanks @PepperTeasdale!

PR #1522, opened 2018-05-22
Use new serviceusage API for `google_project_service[s]`

danawillow:
This migrates us to the new API, which has a `batchEnable` method 🙌 

While I was here, I fixed a test by adding "bigquery-json" to the list of APIs that get enabled automatically, and added import support to `google_project_service` since it was pretty straightforward (cc @mrparkers)

### DIFF
--- a/google/config.go
+++ b/google/config.go
@@ -37,6 +37,7 @@ import (
 	"google.golang.org/api/redis/v1beta1"
 	"google.golang.org/api/runtimeconfig/v1beta1"
 	"google.golang.org/api/servicemanagement/v1"
+	"google.golang.org/api/serviceusage/v1beta1"
 	"google.golang.org/api/sourcerepo/v1"
 	"google.golang.org/api/spanner/v1"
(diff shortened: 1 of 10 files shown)


danawillow:
Ok, so turns out it won't _actually_ be a silent upgrade because users are going to need to enable the new API.  However, @paddycarver gave me permission to do it in a non-major release, so I'll put a warning in the changelog and call it a day

paddycarver:
Chiming in late to say that because Terraform historically has not enabled the APIs it needs for you (except in exceptional circumstances), and because any user that can use this resource must also be able to enable the new service, I think this change preserves the fundamental spirit of the contract Terraform provides: it will make the API calls, but you must turn the APIs on. It's unfortunate in this case that you can't do that from within Terraform, but the error message has a direct link to a button that fixes the issue. So on balance, I'd say the benefits of switching to this are enough and the remediation pain is limited enough to warrant shipping it before 2.0.

ctavan:
I just found #1538. Sorry fo commenting here.

PR #1524, opened 2018-05-23
Guard against eventually consistent services

### LINKED ISSUE #1393: Wait and verify google_project_service is enabled before continuing
### Terraform Version

```
Terraform v0.11.7
+ provider.google v1.10.0
+ provider.random v1.2.0
+ provider.tls v1.1.0
```

### Affected Resource(s)
- `google_project_service`
- `google_project_services`

```hcl
resource "google_project" "demo" {
  name            = "my-project"
  project_id      = "my-project-abcd1234"
  org_id          = "..."
  billing_account = "..."
}

resource "google_project_service" "demo" {
  project = "${google_project.demo.project_id}"
  service = "cloudkms.googleapis.com"
}

# This will fail because Terraform will immediately try to create it after 
# "google_project_service.demo" returns success, but the upstream API
# is still provisioning.
resource "google_kms_key_ring" "demo" {
  name     = "demo"
  project  = "${google_project.demo.project_id}"
}
```

### Expected Behavior
Terraform should not consider `google_project_service(s)` finished until the API is enabled **AND** ready.

### Actual Behavior
Terraform assumes a successful response from the API means the API is finished.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
One could argue this is an API bug, we do long-poll to wait for SSH to be ready on an instance. I think we should do the same here.

/cc @danawillow @rosbo

paddycarver:
When enabling services, after the waiter returns, list the enabled
services and ensure the ones we enabled are in there. If not, retry. May
not always resolve #1393, but should help. Unfortunately, the real
answer is probably either:

1. For us to try and get the API updated to only return the waiter when
the service will consistently be available. I don't know how feasible
this is, but I'm willing to open a ticket.
2. For us to build retries into ~all our resources to retry for a set
amount of time when a service not enabled error is returned. This would
greatly slow down the provider in the case of the service legitimately
not being enabled, but is how other providers handle this class of
problem.

Unfortunately, due to the eventual consistency at play, this is a hard
issue to reproduce and prove, though it matches with my
experience--while testing this patch, one of the tests failed with the
error that the serviceusage API hadn't been enabled, but only on step 4
of the test, when calls had already succeeded. Which suggests eventual
consistency, to me. Regardless, this patch shouldn't _hurt_ and should
mostly be an imperceptible change to users, and should make instances
like #1393 less likely.

### DIFF
--- a/google/resource_google_project_services.go
+++ b/google/resource_google_project_services.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/api/serviceusage/v1beta1"
 )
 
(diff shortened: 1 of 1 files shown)


morgante:
Dana, do you have an API where it says its enabled but it's not actually? I'm happy to help follow up with the relevant teams - I know some of my customers have complained about this in the past.

danawillow:
I've been told compute, but honestly in all my testing I've never run into the issue. I know @sethvargo has though :)

paddycarver:
I've seen it in GKE and the serviceusage API. It seems to happen mostly late at night (10pm-3am) PST, in my experience, but this is all speculation, because it's so hard to reproduce :/

PR #1528, opened 2018-05-23
Add GKE Shared VPC support

danawillow:
All this needed was to allow setting the network/subnetwork to a self link.

~WIP for now because I want to run the tests in CI.~ Tests are passing.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -328,11 +328,11 @@ func resourceContainerCluster() *schema.Resource {
 			},
 
 			"network": {
-				Type:      schema.TypeString,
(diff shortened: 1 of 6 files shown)


morgante:
Does the test actually work? I believe that Kubernetes Engine also needs to be granted permission on the host network to work properly.

danawillow:
Right, that's why it's still a WIP (Work In Progress)- I'm running the tests to figure out the minimum I can do for them, but our CI infrastructure requires running against a PR (there might be another way to do it but this is the easiest). Once I'm confident this is ready, I'll remove the WIP.

Issue #936, opened 2018-01-09
Support AuditLogConfigs in Policy


uschen:
Currently the `google_iam_policy` data source / object doesn't expose `AuditConfigs` which prevents 
enabling`DATA_READ` Audit Log mentioned in https://cloud.google.com/kms/docs/logging

uschen:
wonder it is just expose AuditLog (https://godoc.org/google.golang.org/api/cloudresourcemanager/v1#AuditConfig) in  https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_iam_policy.go#L147

saleh-sedighi-ck:
+1

gkohen:
+1

Farhie:
I would like to see support for this. I see this is labelled as an enhancement. @rosbo is this an enhancement? Enabling data access logs for all services is important for my current team and would argue it is likely a high priority for others.

https://cloud.google.com/logging/docs/audit/configure-data-access

rosbo:
The `enhancement` tag simply means it will update existing resources. We use one of the three tags: `enhancement`, `new resource` or `new datasource`.

Farhie:
Ah, my misunderstanding. Thanks for clearing that up. I should RTFM more 😎

rileykarson:
This is present in `2.0.0`. https://www.terraform.io/docs/providers/google/d/google_iam_policy.html#audit_config

PR #1531, opened 2018-05-24
#936 adding support for AuditLogConfigs in IAM Policies

adamenger:
Hi,

We have a use case for accessing [AuditConfigLogs](https://cloud.google.com/kms/docs/logging#enabling_data_access_logs) at Etsy. This PR addresses #936 and adds support for AuditLogConfigs through a `google_iam_policy` data block. 

See below for an example:

```
resource "google_project_iam_policy" "project" {
  project     = "${google_project.project.id}"
  policy_data = "${data.google_iam_policy.audit_config.policy_data}"
}

data "google_iam_policy" "audit_config" {
  binding {
    role = "roles/owner"

    members = [
      "user:you@domain.com",
    ]
  }

  audit_config {
    service = "cloudkms.googleapis.com"
    audit_log_configs = [
      {
        log_type = "DATA_READ",
        exempted_members = [
          "user:you@domain.com,
        ]
      },
    ]
  }

}
```

Please let me know how I can improve this PR to make it easier to merge.

Best,

Adam

### DIFF
--- a/google/data_source_google_iam_policy.go
+++ b/google/data_source_google_iam_policy.go
@@ -28,6 +28,36 @@ var iamBinding *schema.Schema = &schema.Schema{
 	},
 }
 
+var auditConfig *schema.Schema = &schema.Schema{
+	Type:     schema.TypeSet,
+	Optional: true,
+	Elem: &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"service": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"audit_log_configs": {
+				Type:     schema.TypeSet,
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"log_type": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"exempted_members": {
+							Type:     schema.TypeList,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							Optional: true,
+						},
+					},
+				},
+			},
+		},
+	},
+}
+
 // dataSourceGoogleIamPolicy returns a *schema.Resource that allows a customer
 // to express a Google Cloud IAM policy in a data resource. This is an example
 // of how the schema would be used in a config:
@@ -49,6 +79,7 @@ func dataSourceGoogleIamPolicy() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"audit_config": auditConfig,
 		},
 	}
 }
@@ -58,13 +89,17 @@ func dataSourceGoogleIamPolicy() *schema.Resource {
 func dataSourceGoogleIamPolicyRead(d *schema.ResourceData, meta interface{}) error {
 	var policy cloudresourcemanager.Policy
 	var bindings []*cloudresourcemanager.Binding
+	var audit_configs []*cloudresourcemanager.AuditConfig
 
 	// The schema supports multiple binding{} blocks
 	bset := d.Get("binding").(*schema.Set)
+	aset := d.Get("audit_config").(*schema.Set)
 
 	// All binding{} blocks will be converted and stored in an array
 	bindings = make([]*cloudresourcemanager.Binding, bset.Len())
+	audit_configs = make([]*cloudresourcemanager.AuditConfig, aset.Len())
 	policy.Bindings = bindings
+	policy.AuditConfigs = audit_configs
 
 	// Convert each config binding into a cloudresourcemanager.Binding
 	for i, v := range bset.List() {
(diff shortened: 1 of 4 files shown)


endemics:
Hi,

@danawillow  @paddycarver, any chance a maintainer could review this PR?

I would love to see this feature shipping in the next version of the provider.

Cheers,
Gildas

danawillow:
@paddycarver, I remember you had some ideas about audit configs being a separate resource instead (search our slack convo for "audit" to find it). do you think you'll be able to get to this review this week?

adamenger:
Oh that's an interesting idea! I'd be happy to take a crack at that if that's the path the maintainers decide is best. 

Let me know.

paddycarver:
Ah, ok, yes, I apologise for the delay here, but I believe I remember what @danawillow is referencing.

So how we manage IAM resources in the GCP provider is a bit weird. That's because we offer three separate levels of control over IAM, as explained on https://www.terraform.io/docs/providers/google/r/google_project_iam.html

Essentially, you can control a policy entirely, in which case the entire policy for the resource under management is owned by Terraform. Anything added outside of Terraform is considered drift, and Terraform removes it on the next run. This can get you locked out of your own projects, for example, and can remove access the GCP automatically adds for other resources in your account, but gives you the most locked-down option.

You can also control a specific _binding_ on a policy, ensuring that certain users have a role, and nobody else does. If anyone else does, Terraform considers it drift, and corrects it on the next run. This is less foot-gun-y than the policy control option, because you're only managing one role, so you can tiptoe around the changes that GCP automatically makes to your policy, but it's also less locked down.

Finally, you can control a specific _member_ on a policy, ensuring that they have a specific role. This version is the least locked down, and is only concerned with making sure that user has the role; it won't detect any drift except the user not having that role. It doesn't ensure the user _only_ has that role, etc.

The way this PR is implemented, from what I can tell, only gives the first way of working with IAM support for audit log configs. Which is fine, and a good place to start, but is likely not going to be sufficient, as people using the binding and member options are likely going to want this feature.

My proposal was to:

1. Add the feature to the IAM level, as you have.
2. Add a new resource, similar to member and binding, that only controls the audit log configs. This would allow Terraform to _just_ own the audit log config, and not interfere with any of the permissions managed outside of Terraform. It could, therefore, be used in conjunction with the binding and member versions of the IAM resources.

What do you think about this? Do you have any thoughts on this approach? Is that work you'd like to take on, or would you prefer we treat this PR as only addressing the first point above, and a separate PR be opened to address the second point?

google/data_source_google_iam_policy.go
```diff
@@ -28,6 +28,36 @@ var iamBinding *schema.Schema = &schema.Schema{
...
+				Type:     schema.TypeSet,
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"log_type": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"exempted_members": {
+							Type:     schema.TypeList,
```
paddycarver: Does the order of the exempted members matter? Would you ever want to reference them using interpolation? If the answer to both is "no", this would probably be better off as a Set.

google/iam.go
```diff
@@ -154,3 +154,67 @@ func rolesToMembersMap(bindings []*cloudresourcemanager.Binding) map[string]map[
 	}
 	return bm
 }
+
+// Merge multiple Audit Configs such that configs with the same service result in
+// a single exemption list with combined members
+func mergeAuditConfigs(audit_configs []*cloudresourcemanager.AuditConfig) []*cloudresourcemanager.AuditConfig {
```
paddycarver: As a general rule of thumb, camelCase, not snake_case, is preferred. :)

adamenger: Thank you! I'll get this fixed up.

morgante:
@danawillow @paddycarver @adamenger Could we get someone to take a look at this? It'd be very useful to configure this in Terraform.

paddycarver:
I'll take over this PR and pick it up on Monday. Can't promise an ETA, but I'll have an update on it next week.

paddycarver:
I'm late on my update to this, _and_ it's very dissatisfying! Sorry. The `paddy_audit_log_configs` branch on this repo has my updated code from this PR: fixing some style issues I mentioned, adding support for AuditConfigs in our diff suppress function, and adding support to our test checks for AuditConfigs. The only step remaining is to add tests that exercise AuditConfigs and then get PR review. My day today is going to be primarily bug fixes, test fixes, and issue triage, but I'll try to add tests tonight and update this PR tomorrow with a reference to the new PR. Sorry for the delay here, there's turning out to be a bit more code than I hoped.

morgante:
@paddycarver Gentle ping. Would love to see this added soon.

stepanstipl:
@ndmckinley can you please elaborate on what's this waiting on the GCP side?

Looks to me this PR is missing tests - is this something @paddycarver is working on? Would there be interest in some help with the tests?

ndmckinley:
Yes, Paddy is on it, and you can follow along at https://github.com/GoogleCloudPlatform/magic-modules/pull/1066.  :)

paddycarver:
My PR has been merged, and audit configs should be configurable when setting an entire project's IAM policy (e.g., using `google_project_iam_policy`) in the 2.0.0 release.

I'm currently working to add a `google_project_iam_audit_config` resource, which will allow configuring just the audit config for a project, without needing to overwrite all your configured bindings. I'm hoping to open a PR for that shortly.

paddycarver:
GoogleCloudPlatform/magic-modules#1094 brings that fine-grained support, which I think should be enough support for the moment, until we can get a better understanding of how people are using this and want to use it.

paddycarver:
Both PRs have been merged, so as of 2.0.0, you should be able to manage audit configs as part of managing an entire project's policy, or you should be able to manage _just_ the audit configs for a project.

I'm going to close this now, as the functionality has been implemented.

PR #1534, opened 2018-05-24
Adding attributes in google_container_cluster documentation

mclavel:
We can use this attributes in a google_container_cluster resource.

https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_container_cluster.go#L254


ndmckinley:
Thanks for the docs update!  Merging.  :)

danawillow:
(just popping in here to say that those attributes were already in the docs: https://www.terraform.io/docs/providers/google/r/container_cluster.html#password)

mclavel:
If you don't set a custom username or password in cluster, you don't have documentation about who get this info like a attribute (this was the main reason for the PR).

PR #1535, opened 2018-05-24
datasource: added service account and service account key

vishen:
Added datasources for;` service account` and `service account key`. Both datasources should be somewhat similar to their resource counterparts, but they accept a much more flexible `account_id` and `service_account_id`.

This is my first time with acceptance tests (or a datasource), so please let me know if I need to add more, or have done something weird :)

### DIFF
--- a/google/data_source_google_service_account.go
+++ b/google/data_source_google_service_account.go
@@ -10,19 +10,16 @@ import (
 func dataSourceGoogleServiceAccount() *schema.Resource {
 	return &schema.Resource{
 		Read: dataSourceGoogleServiceAccountRead,
-
 		Schema: map[string]*schema.Schema{
-			// Required
 			"account_id": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validateRFC1035Name(6, 30),
 			},
-			// Optional
 			"project": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
 			},
-			// Computed
 			"email": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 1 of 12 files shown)


google/data_source_google_service_account.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceGoogleServiceAccount() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGoogleServiceAccountRead,
+
+		Schema: map[string]*schema.Schema{
+			// Required
```
ndmckinley: The comments here aren't necessary since they duplicate the code below, and it's generally good practice not to include comments that duplicate code.  Thanks for putting the schema elements in our preferred order, though!

google/data_source_google_service_account.go
```diff
@@ -0,0 +1,84 @@
...
+	sa, err := config.clientIAM.Projects.ServiceAccounts.Get(accountID).Do()
+	if err != nil {
+		return handleNotFoundError(err, d, fmt.Sprintf("Service Account %q", accountID))
+	}
+
+	d.SetId(sa.Name)
+	d.Set("email", sa.Email)
+	d.Set("unique_id", sa.UniqueId)
+	d.Set("project", sa.ProjectId)
+	d.Set("account_id", strings.Split(sa.Email, "@")[0])
```
ndmckinley: _Probably_ not ideal to set `account_id` - users will be a little surprised if `account_id` changes from the value they set it to, on the off chance they choose to interpolate it.  In a resource this would be a permanent diff (unless we had a `DiffSuppressFunc`), which we try to avoid.  I'm 50/50 on this - what do you think?

google/data_source_google_service_account_key.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func dataSourceGoogleServiceAccountKey() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGoogleServiceAccountKeyRead,
+
+		Schema: map[string]*schema.Schema{
+			// Required
```
ndmckinley: Same comment - probably best to remove duplicate comments.

google/data_source_google_service_account_key.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func dataSourceGoogleServiceAccountKeyRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	// Get the service account id as the fully qualified name
+	serviceAccountID := d.Get("service_account_id").(string)
+
+	// If the service account id isn't already the fully qualified name
+	if !strings.HasPrefix(serviceAccountID, "projects/") {
```
ndmckinley: How much of this can be shared?  Possibly even with the resources?

google/data_source_google_service_account_key_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccDatasourceGoogleServiceAccountKey(account),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGoogleServiceAccountKeyExists(resourceName),
+					resource.TestCheckResourceAttrSet(resourceName, "name"),
```
ndmckinley: It'd be great to test that they're set to the right values, also.

google/data_source_google_service_account.go
vishen: No worries! I was trying to keep it similar to other datasource code, but I will remove it :)

google/data_source_google_service_account_key.go
vishen: Yup, I can share it with the resource `google_sevice_account_key`, but I don't know if I can do the same with `google_service_account` as it has `vadliateRFC1035Name(6, 30)` in the schema validation for `account_id` which doesn't allow '@' or `/`

google/data_source_google_service_account_key_test.go
vishen: The service account key endpoint returns the name as `projects/PROJECT-ID/serviceAccounts/SA-NAME@PROJECT-ID.iam.gserviceaccount.com/keys/<key_id>` - should I check that it starts with the expected service account?

ndmckinley: Well, right now I think this just checks that the resource has any attribute as `name` - if you could check that it's the value you expect somehow, that would help.

google/data_source_google_service_account.go
ndmckinley: Hm.  That must have a `DiffSuppressFunc`, then?  I suppose it's better to maintain consistency.

vishen: OK, I was misusing the service account id for this, the service account resource only accepts a service account id, so I have changed the datasource to mimic this; as such the account_id should always be the same. The resource for service account still has this line, so I left it in, but if it is confusing I can remove it?

ndmckinley:
Okay, that looks all right to me.  I'll run the tests!

ndmckinley:
Hm.  I re-ran the tests (forgot about the resource tests) and now they're crashing with a tooling error.  Let me see what's going on there!

ndmckinley:
Eh, you know what, I ran them locally and I'm sure it works.  Merging!

PR #1536, opened 2018-05-25
add disabled property to firewall

danawillow:


danawillow:
Me, 6pm: Man, what a drag it is to write all these tests that only fail because the tests are wrong.
Me, 7pm: This API does not behave the way I thought it did, good thing I had a test.

🤦‍♀️ 

Anyway, the answer to your question is yes. Turns out the update method only supports updating certain fields, but patch lets you update any of them so I just used patch instead.

Issue #1541, opened 2018-05-26
Support Stackdriver-Kubernetes


kim0:
### Terraform Version
Terraform v0.11.7

### Affected Resource(s)
- google_container_cluster

### Expected Behavior
I am able to create a GKE cluster with the new stackdriver-kubernetes mode monitoring as detailed [here](https://cloud.google.com/monitoring/kubernetes-engine/installing#creating_the_cluster)

### Actual Behavior
I tried to create a new GKE cluster and set `logging_service = "logging.googleapis.com/kubernetes"` but that was not accepted as a valid value

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create new GKE cluster with Terraform
1. Set `logging_service = "logging.googleapis.com/kubernetes"`

danawillow:
Fixed in #1589.

PR #1544, opened 2018-05-28
Add support for Stackdriver Monitoring on GKE clusters

tanordheim:
Using Strackdriver to monitor GKE clusters is currently available in beta ([see docs](https://cloud.google.com/monitoring/kubernetes-engine/)) and I've done some attempts at letting us provision GKE clusters with this enabled via Terraform.

This PR solves that by allowing you to set the following properties:

```
logging_service = "logging.googleapis.com/kubernetes"
monitoring_service = "monitoring.googleapis.com/kubernetes"
```

It does, however, come with some caveats:

- `google_container_cluster.min_master_version` must be set to `1.10.2-gke.1` or higher, or the API call fails with a validation error (`googleapi: Error 400: Stackdriver kubernetes resource model is not supported for cluster version 1.8.10-gke.0; version 1.10.2 or newer is required., badRequest`).
- The `logging_service` and `monitoring_service` changes can only take place when creating a new cluster. Attempting to update these on an already running cluster (even if it satisfies the version requirement) yields this error: `googleapi: Error 400: desired_monitoring_service was "monitoring.googleapis.com/kubernetes" but must be one of "", "none", or "monitoring.googleapis.com"., badRequest`. According to [the migration docs](https://cloud.google.com/monitoring/kubernetes-engine/migration), it's for the time being only possible to enable this when creating a new cluster.

I'm not sure how this should be handled on a Terraform level. I'm not terribly familiar with Terraform internals but I'm happy to contribute if someone has some idea of to solve this.

This PR is also related to #1541.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -219,7 +219,7 @@ func resourceContainerCluster() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Computed:     true,
-				ValidateFunc: validation.StringInSlice([]string{"logging.googleapis.com", "none"}, false),
+				ValidateFunc: validation.StringInSlice([]string{"logging.googleapis.com", "logging.googleapis.com/kubernetes", "none"}, false),
 			},
 
 			"maintenance_policy": {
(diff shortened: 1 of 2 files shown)


nikitashalnov:
We are really need this feature. +1

xenozauros:
Would be great to implement it

danawillow:
Oh no, I saw the open issue about this (#1577) before realizing there was a PR out already to fix it. Sorry to steal your thunder! Closing in favor of #1589 since that's been merged already.

PR #1545, opened 2018-05-28
Fix link to app engine locations page

nownabe:
Correct URL is https://cloud.google.com/appengine/docs/locations


rosbo:
Thanks for fixing the docs :)

PR #1550, opened 2018-05-28
Remove label from bucket

### LINKED ISSUE #1350: Removing a label on a bucket doesn't work
Hi there,

When a bucket contains a label that is not part of the terraform file or state, terraform tries to remove it. However this doesn't work so with every run it tries to remove the label.

### Terraform Version
Terraform 0.11.5
Google Provider 1.9.0

### Affected Resource(s)
- google_storage_bucket

```hcl
resource "google_storage_bucket" "testbucket" {
  name          = "myproject-testbucket"
  project       = "myproject"
  storage_class = "REGIONAL"
  force_destroy = true

  labels = {
    opex = "teamxyz"
    application = "myapp"
  }
}
```

### Debug Output
When adding a label to this bucket called 'remove' with value 'me', terraform shows this:

> google_storage_bucket.testbucket:
 Modifying... (ID: myproject-testbucket)
  labels.%:      "3" => "2"
  labels.remove: "me" => ""

### Expected Behavior
Label should be removed

### Actual Behavior
Label remains on the bucket

rosbo:
Fixes #1350 

In order to remove a label from a bucket using the `patch` method, you have to set the label value explicitly to `null` for the label key you want to remove.

### DIFF
--- a/google/resource_storage_bucket.go
+++ b/google/resource_storage_bucket.go
@@ -378,6 +378,15 @@ func resourceStorageBucketUpdate(d *schema.ResourceData, meta interface{}) error
 		if len(sb.Labels) == 0 {
 			sb.NullFields = append(sb.NullFields, "Labels")
 		}
+
+		// When using PATCH for this resource, in order to delete a label, we
+		// have to explicitly set its value to null
+		old, _ := d.GetChange("labels")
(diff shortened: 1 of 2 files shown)


rosbo:
```sh
--- PASS: TestAccStorageBucket_labels (6.25s)
```

PR #1557, opened 2018-05-29
Allow using in repo configuration for cloudbuild trigger

jamielennox:
Cloudbuild triggers have a complex configuration that can be defined
from the API. When using the console, the more typical way of doing this
is to defined the configuration within the repository and point the
configuration to the file that defines the config.

This can be supported by sending the filename parameter instead of the
build parameter, however only one can be sent.

### DIFF
--- a/website/docs/r/cloudbuild_trigger.html.markdown
+++ b/website/docs/r/cloudbuild_trigger.html.markdown
@@ -33,6 +33,21 @@ resource "google_cloudbuild_trigger" "build_trigger" {
 }
 ```
 
+OR
+
+```hcl
+resource "google_cloudbuild_trigger" "build_trigger" {
+  project  = "my-project"
+  trigger_template {
+    branch_name = "master"
+    project     = "my-project"
+    repo_name   = "some-repo"
+  }
+  filename = "cloudbuild.yaml"
+}
+```
+
+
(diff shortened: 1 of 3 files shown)


website/docs/r/cloudbuild_trigger.html.markdown
```diff
@@ -59,6 +74,10 @@ will be expanded when the build is created:
     or resolved from the specified branch or tag.
   * `$SHORT_SHA`: first 7 characters of `$REVISION_ID` or `$COMMIT_SHA`.
 
+* `filename` - (Optional) Specify the path to a Cloud Build configuration file
+in the Git repo. This is mutually exclusive with `build`. By default this is
```
ndmckinley: It doesn't seem like there's a default in here, so if you change 'by default' to 'Typically', that will be a bit clearer.

jamielennox: Ah, agreed. There's no default on the API, you have to specify a value. It's the web portal that suggests you use cloudbuild.yaml.

mhamrah:
Can this be merged?

rosbo:
Merged. Thanks Jamie for your contribution :)

mhamrah:
+1

Issue #1562, opened 2018-05-29
Add retry logic when reading enabled services


danawillow:
Got a report from a customer that's getting 429 errors. We already retry on enabling/disabling services, but we don't retry when reading them back (which could still throw these errors)

sethvargo:
@danawillow it looks like this still got auto-closed for some reason

PR #1565, opened 2018-05-30
Lions, tigers, and services being enabled with "precondition failed", oh my!

sethvargo:
This PR does a few things:

1. It updates the "list enabled services" API call to be more efficient

1. It batches services to be enabled 20 at a time. The API fails if you try to enable more than 20 services, and this is documented in the SDK and API. I learned this the hard way. I think Terraform should "do the right thing" here and batch them in series' of twenty, which is what this does. Each batch is tried in serial, but I think making it parallelized is not worth the complexity tradeoffs.

1. Handle the precondition failed error that occurs randomly. This just started happened, but it affects at least two APIs consistently, and a rudimentary test showed that it failed 78% of the time (78/100 times in an hour). We should fix this upstream, but that failure rate also necessitates (in my opinion) some mitigation on the Terraform side until a fix is in place at the API level.

1. Use errwrap on errors for better tracing. It was really difficult to trace exactly which error was being throw. That's fixed.

/cc @danawillow @rosbo @paddycarver

### DIFF
--- a/google/resource_google_project_services.go
+++ b/google/resource_google_project_services.go
@@ -185,25 +185,34 @@ func getConfigServices(d *schema.ResourceData) (services []string) {
 
 // Retrieve a project's services from the API
 func getApiServices(pid string, config *Config, ignore map[string]struct{}) ([]string, error) {
-	apiServices := make([]string, 0)
-	// Get services from the API
-	token := ""
-	for paginate := true; paginate; {
-		svcResp, err := config.clientServiceUsage.Services.List("projects/" + pid).PageToken(token).Filter("state:ENABLED").Do()
-		if err != nil {
-			return apiServices, err
-		}
-		for _, v := range svcResp.Services {
-			// names are returned as projects/{project-number}/services/{service-name}
-			nameParts := strings.Split(v.Name, "/")
-			name := nameParts[len(nameParts)-1]
-			if _, ok := ignore[name]; !ok {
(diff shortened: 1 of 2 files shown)


google/resource_google_project_services.go
```diff
@@ -212,58 +223,110 @@ func enableService(s, pid string, config *Config) error {
...
+			var sop *serviceusage.Operation
+			var err error
+
+			if len(services) < 1 {
+				// No more services to enable
+				return nil
+			} else if len(services) == 1 {
+				// Use the singular enable - can't use batch for a single item
+				name := fmt.Sprintf("projects/%s/services/%s", pid, services[0])
+				op := &serviceusage.EnableServiceRequest{}
```
danawillow: nit: can this (and below) be named req instead of op? Op tends to refer to Operation objects

rosbo:
Fixes #1562

danawillow:
I don't think this works for #1562- it does a bunch of other fixes, but not retry logic on read (I have one that I'll send out that does it once this is in though)

rosbo:
My bad, I added it because I thought they were related. I removed the linked issue.

sethvargo:
Okay @danawillow @rosbo updated to fix comments. Let me know if there's anything else (I don't have merge permissions).

rosbo:
LGTM. Thanks Seth!

PR #1573, opened 2018-05-31
Retry while listing enabled services

sethvargo:
This fixes GH-1562 for realz


sethvargo:
The [whitespace suppress diff](https://github.com/terraform-providers/terraform-provider-google/pull/1573/files?w=1) makes this a bit easier to read

PR #1575, opened 2018-05-31
Add link to sidebar for liens.

paddycarver:
When we added lien support, we forgot to add a link to the docs in the sidebar navigation. This adds that link.


rosbo:
Thanks Paddy!

PR #1587, opened 2018-06-04
update auth docs

danawillow:
* Added a link to the console page where you can download a file
* Removed instructions on how to get to that page, since now you can just click on the link
* Added caveat for application default credentials

@sethvargo @theacodes @kimcam let me know if this seems reasonable / you have any suggestions!

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -35,9 +35,9 @@ resource "google_compute_instance" "default" {
 The following keys can be used to configure the provider.
 
 * `credentials` - (Optional) Contents of a file that contains your service
-  account private key in JSON format. You can download this file from the
-  [Google Cloud Console](https://console.cloud.google.com/apis/credentials/serviceaccountkey)
-  by picking an existing service account or creating a new one for Terraform.
+  account private key in JSON format. You can download your existing
+  [Google Cloud service account file]
+  from the Google Cloud Console, or you can create a new one from the same page.
 
   Credentials can also be specified using any of the following environment
   variables (listed in order of precedence):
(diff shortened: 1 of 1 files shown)


website/docs/index.html.markdown
```diff
@@ -58,7 +59,8 @@ The following keys can be used to configure the provider.
   details. On your computer, if you have made your identity available as the
   Application Default Credentials by running [`gcloud auth application-default
   login`](https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login),
-  the provider will use your identity.
+  the provider will use your identity. However, this method is not guaranteed to
```
rosbo: Consider putting it in bold or as a `Note`. I think this is a very important caveat.

sethvargo: Not sure if you know this, but if you put `~>` and then text like:

```
~> This is a note
```

It'll get called out with the right CSS to actually look like a note. There's also:

- `!>` for big red warnings
- `->` for info
- `=>` for green success

I would like to link to the `service accounts` section or some kind of documentation. We say "this is the easiest way, but it might break in edge cases", name the replacement, but then don't provide a way to get more information on the replacement.

website/docs/index.html.markdown
```diff
@@ -36,7 +36,8 @@ The following keys can be used to configure the provider.
 
 * `credentials` - (Optional) Contents of a file that contains your service
   account private key in JSON format. You can download this file from the
```
sethvargo: For SEO purposes, we should try to make the link descriptive, like:

```
You can download your existing [Google Cloud service account file](LINK) from the Google Cloud Console, or you can create a new one from the same page.
```

website/docs/index.html.markdown
rosbo: Yeah, by `Note`, I meant `~>` like we do [here](https://www.terraform.io/docs/providers/google/r/compute_disk.html).

But I didn't know about `=>` and `->`. Thanks for sharing this tip :)

danawillow:
Thanks @sethvargo for the ideas, should be better now.

PR #1591, opened 2018-06-05
Support filename field for cloudbuild_trigger resource

mhamrah:
Cloud Build Triggers[ supports a one of build_template that can be a build object (currently supported by Terraform) or a filename string](https://cloud.google.com/container-builder/docs/api/reference/rest/v1/projects.triggers) that points to a file in the source repository. This PR supports the filename attribute of a build trigger definition.


mhamrah:
Didn't see #1557 which is a more robust implementation.

PR #1598, opened 2018-06-06
Fix network_tier tests.

paddycarver:
Make network_tier computed, so when it's not specified we don't get a
permadiff.

Always store v1 self_links, so we don't get a permadiff.


rosbo:
Thanks Paddy

PR #1620, opened 2018-06-07
Update schema to 0.11.7.

paddycarver:
Update our helper/schema to v0.11.7, and update all its dependencies to
their latest, as well.

PR #1621, opened 2018-06-07
Make App Engine applications updatable

paddycarver:
No longer ForceNew on every little change to an App Engine application. Now adding an App Engine application to a project doesn't force a new project, only removing an application from a project forces a new project. Also, updating the location_id forces a new project, but all other fields can be updated in place.

This depends on #1620 and should only be merged after that is.

### DIFF
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -105,15 +105,6 @@ func appEngineResource() *schema.Resource {
 	return &schema.Resource{
 		CustomizeDiff: resourceGoogleProjectAppEngineCustomizeDiff,
 		Schema: map[string]*schema.Schema{
-			"name": &schema.Schema{
-				Type:     schema.TypeString,
-				Computed: true,
-			},
-			"url_dispatch_rule": &schema.Schema{
-				Type:     schema.TypeList,
-				Computed: true,
-				Elem:     appEngineURLDispatchRuleResource(),
-			},
 			"auth_domain": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
@@ -137,10 +128,6 @@ func appEngineResource() *schema.Resource {
 					"australia-southeast1",
 				}, false),
 			},
-			"code_bucket": &schema.Schema{
-				Type:     schema.TypeString,
-				Computed: true,
-			},
 			"serving_status": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
@@ -150,6 +137,26 @@ func appEngineResource() *schema.Resource {
 					"USER_DISABLED",
 					"SYSTEM_DISABLED",
 				}, false),
+				Computed: true,
+			},
+			"feature_settings": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem:     appEngineFeatureSettingsResource(),
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"url_dispatch_rule": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     appEngineURLDispatchRuleResource(),
+			},
+			"code_bucket": &schema.Schema{
+				Type:     schema.TypeString,
 				Computed: true,
 			},
 			"default_hostname": &schema.Schema{
@@ -163,13 +170,6 @@ func appEngineResource() *schema.Resource {
 			"gcr_domain": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
-			},
-			"feature_settings": &schema.Schema{
-				Type:     schema.TypeList,
-				Optional: true,
-				Computed: true,
-				MaxItems: 1,
-				Elem:     appEngineFeatureSettingsResource(),
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


google/resource_google_project.go
```diff
@@ -103,6 +103,7 @@ func resourceGoogleProject() *schema.Resource {
 
 func appEngineResource() *schema.Resource {
 	return &schema.Resource{
+		CustomizeDiff: resourceGoogleProjectAppEngineCustomizeDiff,
 		Schema: map[string]*schema.Schema{
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
```
danawillow: Not actually related to this change, but can you order the properties in Required -> Optional -> Computed order? I'm a big fan of that to make it easy to immediately see which fields are settable.

google/resource_google_project.go
```diff
@@ -215,8 +206,18 @@ func appEngineFeatureSettingsResource() *schema.Resource {
...
+		return diff.ForceNew("app_engine")
+	} else if old, _ := diff.GetChange("app_engine.0.location_id"); diff.HasChange("app_engine.0.location_id") && old != nil && old.(string) != "" {
+		// if location_id was already set, and has a new value, that forces a new app
+		// if location_id wasn't set, don't force a new value, as we're just enabling app engine
+		return diff.ForceNew("app_engine.0.location_id")
+	}
+	return nil
+}
+
+func resourceGoogleProjectAppEngineCustomizeDiff(diff *schema.ResourceDiff, meta interface{}) error {
```
danawillow: This can probably be removed

google/resource_google_project.go
```diff
@@ -492,10 +501,39 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
...
+
+				// Wait for the operation to complete
+				waitErr := appEngineOperationWait(config.clientAppEngine, op, pid, "App Engine app to update")
+				if waitErr != nil {
+					return waitErr
+				}
+				log.Printf("[DEBUG] Updated App Engine App")
+			}
+		}
+	}
```
danawillow: May as well d.SetPartial here in case someone comes and adds a new HasChange after this

google/resource_google_project_test.go
```diff
@@ -216,6 +216,52 @@ func TestAccProject_appEngineBasic(t *testing.T) {
...
+
+	org := getTestOrgFromEnv(t)
+	pid := acctest.RandomWithPrefix("tf-test")
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccProject_appEngineNoApp(pid, org),
+				Check: resource.ComposeTestCheckFunc(
```
danawillow: If you want to keep the checks I won't tell you no, but they aren't necessary if you check by importing

google/resource_google_project_test.go
```diff
@@ -417,6 +483,21 @@ resource "google_project" "acceptance" {
 }`, pid, pid, org)
 }
 
+func testAccProject_appEngineUpdate(pid, org string) string {
+	return fmt.Sprintf(`
+resource "google_project" "acceptance" {
+  project_id = "%s"
```
danawillow: nit: can you align the `=`s here? (as if you were to run terraform fmt)

google/resource_google_project_test.go
```diff
@@ -402,6 +459,15 @@ resource "google_folder" "folder1" {
 `, pid, projectName, folderName, org)
 }
 
+func testAccProject_appEngineNoApp(pid, org string) string {
```
danawillow: this is pretty much identical to `testAccProject_create` (which lives in `resource_google_project_iam_policy_test.go`) if you wanted to reuse that (a few other resources are doing so)

google/resource_google_project.go
paddycarver: 👍

google/resource_google_project.go
paddycarver: Oops, sorry, this is a remnant of my futzing around with customize diff -_-

google/resource_google_project.go
paddycarver: 👍

google/resource_google_project_test.go
paddycarver: I think in this case import wouldn't actually check them, because they're computed fields not set in the config, so I believe without the checks, if the fields just aren't being set, ImportStateVerify won't complain. Because (from what I understand) ImportStateVerify just creates a new, empty state, imports the resource again, and checks it against the state from creating the resource. If there was a bug where these weren't being set, in both cases it would have them not set, so ImportStateVerify would think that's fine. And because they're computed and not config fields, there'd be no diff.

google/resource_google_project_test.go
paddycarver: 👍

google/resource_google_project_test.go
paddycarver: I have a weakly-held opinion that it's better to have the configs for semantically different things separated, even if the implementation ends up looking the same, to minimize unintended consequences of changes. But if you feel strongly about it, I'm happy to change this.

google/resource_google_project_test.go
danawillow: Cool, makes sense! Regardless, you can still take out `testAccCheckGoogleProjectExists`

google/resource_google_project_test.go
danawillow: Cool that's fine!

paddycarver:
Think I addressed all the feedback here. :)

PR #406, opened 2017-09-11
Export connection_name on sql instance resource

selmanj:
Closes #387

PR #1650, opened 2018-06-13
Fix connection_name parameter sql_database_instance docs

tomelliff:
connection_name is a read only parameter that was added in https://github.com/terraform-providers/terraform-provider-google/pull/406.

Looks like it was confusing someone at https://stackoverflow.com/q/50838772/2291321


danawillow:
Good catch, thank you @tomelliff!

PR #1652, opened 2018-06-14
add new compute_instance_from_template resource

### LINKED ISSUE #1582: Feature Request: support instance templates in the google_compute_instance provider
GCP Supports creating instances from instance templates

https://cloud.google.com/compute/docs/instances/create-vm-from-instance-template

It would be nice if Terraform could offer the same support by providing a template name to the google_compute_instance provider.

danawillow:
This was done as its own resource as suggested in slack, since we don't have the option of making all fields Computed in google_compute_instance. There's precedent in the aws provider for this sort of thing (see ami_copy, ami_from_instance).

When I started working on this I assumed I could do it in the compute_instance resource and so I went ahead and reordered the schema to make it easier to work with in the future. Now it's not quite relevant, but I left it in as its own commit that can be looked at separately from the other changes.

Fixes #1582.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -620,50 +620,36 @@ func getDisk(diskUri string, d *schema.ResourceData, config *Config) (*compute.D
 	return disk, err
 }
 
-func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) error {
-	config := meta.(*Config)
-
-	project, err := getProject(d, config)
-	if err != nil {
-		return err
-	}
-
-	// Get the zone
-	z, err := getZone(d, config)
-	if err != nil {
-		return err
-	}
-	log.Printf("[DEBUG] Loading zone: %s", z)
-	zone, err := config.clientCompute.Zones.Get(
-		project, z).Do()
-	if err != nil {
-		return fmt.Errorf(
-			"Error loading zone '%s': %s", z, err)
-	}
-
+func computeInstance(project string, zone *compute.Zone, d *schema.ResourceData, config *Config) (*computeBeta.Instance, error) {
 	// Get the machine type
-	log.Printf("[DEBUG] Loading machine type: %s", d.Get("machine_type").(string))
-	machineType, err := config.clientCompute.MachineTypes.Get(
-		project, zone.Name, d.Get("machine_type").(string)).Do()
-	if err != nil {
-		return fmt.Errorf(
-			"Error loading machine type: %s",
-			err)
+	var machineTypeUrl string
+	if mt, ok := d.GetOk("machine_type"); ok {
+		log.Printf("[DEBUG] Loading machine type: %s", mt.(string))
+		machineType, err := config.clientCompute.MachineTypes.Get(
+			project, zone.Name, mt.(string)).Do()
+		if err != nil {
+			return nil, fmt.Errorf(
+				"Error loading machine type: %s",
+				err)
+		}
+		machineTypeUrl = machineType.SelfLink
 	}
 
 	// Build up the list of disks
 
 	disks := []*computeBeta.AttachedDisk{}
-	bootDisk, err := expandBootDisk(d, config, zone, project)
-	if err != nil {
-		return err
+	if _, hasBootDisk := d.GetOk("boot_disk"); hasBootDisk {
+		bootDisk, err := expandBootDisk(d, config, zone, project)
+		if err != nil {
+			return nil, err
+		}
+		disks = append(disks, bootDisk)
 	}
-	disks = append(disks, bootDisk)
 
 	if _, hasScratchDisk := d.GetOk("scratch_disk"); hasScratchDisk {
 		scratchDisks, err := expandScratchDisks(d, config, zone, project)
 		if err != nil {
-			return err
+			return nil, err
 		}
 		disks = append(disks, scratchDisks...)
 	}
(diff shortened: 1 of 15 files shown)


paddycarver:
I'm reviewing the code right now, but in case I get distracted while doing that, just wanted to jump in here and say I agree with making this a separate resource.

google/resource_compute_instance_from_template.go
```diff
@@ -0,0 +1,136 @@
...
+			instance.ForceSendFields = append(instance.ForceSendFields, strcase.UpperCamelCase(f))
+		}
+	}
+
+	// Read create timeout
+	// Until "create_timeout" is removed, use that timeout if set.
+	createTimeout := int(d.Timeout(schema.TimeoutCreate).Minutes())
+	if v, ok := d.GetOk("create_timeout"); ok && v != 4 {
+		createTimeout = v.(int)
+	}
```
paddycarver: Is there a reason to port over this behavior we want to get rid of to a new resource? I think we could probably just not use `create_timeout` here.

google/resource_compute_instance_from_template.go
```diff
@@ -0,0 +1,136 @@
...
+	for f, s := range computeInstanceFromTemplateSchema() {
+		// It seems that GetOkExists always returns true for sets.
+		// TODO: confirm this and file issue against Terraform core.
+		// In the meantime, don't force send sets.
+		if s.Type == schema.TypeSet {
+			continue
+		}
+
+		if _, exists := d.GetOkExists(f); exists {
+			instance.ForceSendFields = append(instance.ForceSendFields, strcase.UpperCamelCase(f))
```
paddycarver: :/ I can see this breaking when someone adds a field to `compute_instance` that isn't a direct 1:1 snake_case version of the API field.

google/resource_compute_instance.go
```diff
@@ -620,50 +620,36 @@ func getDisk(diskUri string, d *schema.ResourceData, config *Config) (*compute.D
...
-	}
-	log.Printf("[DEBUG] Loading zone: %s", z)
-	zone, err := config.clientCompute.Zones.Get(
-		project, z).Do()
-	if err != nil {
-		return fmt.Errorf(
-			"Error loading zone '%s': %s", z, err)
-	}
-
+func computeInstance(project string, zone *compute.Zone, d *schema.ResourceData, config *Config) (*computeBeta.Instance, error) {
```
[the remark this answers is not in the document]
danawillow: Done

google/resource_compute_instance.go
```diff
@@ -831,9 +844,7 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("tags", convertStringArrToInterface(instance.Tags.Items))
 	}
 
-	if len(instance.Labels) > 0 {
-		d.Set("labels", instance.Labels)
-	}
+	d.Set("labels", instance.Labels)
```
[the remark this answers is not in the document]
danawillow: Done

google/resource_compute_instance_from_template.go
```diff
@@ -0,0 +1,136 @@
...
+		// as a google_compute_instance.
+
+		Timeouts: resourceComputeInstance().Timeouts,
+
+		Schema:        computeInstanceFromTemplateSchema(),
+		CustomizeDiff: resourceComputeInstance().CustomizeDiff,
+	}
+}
+
+func computeInstanceFromTemplateSchema() map[string]*schema.Schema {
```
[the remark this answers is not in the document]
danawillow: Yeah, it was either this or copy/paste and then have to keep them both up-to-date manually, and this seemed to me like the better option.

google/resource_compute_instance_from_template.go
danawillow: Done

google/resource_compute_instance_from_template.go
danawillow: Fair. I took a look at our current fields, and as far as I can tell, things that would potentially be overridden to their zero value are indeed snake_case versions of the API fields. Given that we're generally trying to do that sort of thing anyway for new fields, I'm not too concerned, but I added a comment.

google/resource_compute_instance_from_template.go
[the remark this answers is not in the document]
paddycarver: Agreed. The nice thing about this is we know we want those fields, so if we decide in the future to go a different direction with this, we can do it without a breaking change. But yeah, this is the best way I can think of to do it right now.

google/resource_compute_instance_from_template.go
paddycarver: I assume the API returns a good error if a field that doesn't match a field is set in ForceSendFields? I'm gonna test that real quick. As long as a field that doesn't match up isn't silently ignored, I think this is a manageable solution. Nested values scare me a little, and I think over time we may want to add a test that, e.g., checks all the properties work out to the same strings as ForceSendFields, but that's probably not a project for today.

danawillow: Nope, it's silently ignored. I could instead directly set the fields that we expect to be able to be overridden to zero values, but I honestly think that's just about as error prone as this is.

paddycarver: The API does, actually, just silently ignore invalid fields. :/ I've opened google/google-api-go-client#280 to see if I can't get the client to throw an error on invalid fields. Otherwise, this may be something worth bringing in internally, but I suppose I'm happy to punt on that, as we have test coverage, and don't have perfect `ForceSendFields` support _elsewhere_, anyways, so the class of bugs this would enable is a class of bugs we already accept.

Sorry, this is basically a long comment explaining for future-Paddy/future maintainers why this was an acceptable risk, in case someone's doing archaeology for a bug. :)

Issue #1655, opened 2018-06-14
google_service_account_key broken in 1.14.0 - regression bug


primeroz:
Hi there,

In version 1.14.0 there _seems to be_ a regression bug where the google_service_account_key resource does not work anymore.

According to the [docs](https://www.terraform.io/docs/providers/google/r/google_service_account_key.html) the resource does not expect a key for project. It is , i assume, going to inherit it from the keyring it belongs to 

Starting in version 1.14.0 there seem to be a validation error during apply 

If i don't have a project key specified in the resource i get:
```
Error: Error applying plan:

1 error(s) occurred:

* google_service_account_key.test-bug: 1 error(s) occurred:

* google_service_account_key.test-bug: project: required field is not set
``` 
If i do add the project to the resource i get a vlidation error during plan
```
Error: google_service_account_key.test-bug: : invalid or unknown key: project
```

In version 1.13.0 this work as expected.

Interesting enough if i add the project key to the provider it then works also in version 1.14.0 

### Terraform Version
```
~/tmp/tf-google-bug >>> terraform -v
Terraform v0.11.7
+ provider.google v1.14.0
```
### Affected Resource(s)
- google_service_account_key

### Terraform Configuration Files
```hcl
variable "project_id" {}
variable "region" {}

provider "google" {
  #version = "= 1.13.0"
  version     = "= 1.14.0"
  region      = "${var.region}"
  credentials = "${file("~/.config/gcloud/terraform-account.json")}"
}

resource "google_service_account" "test-bug" {
  account_id   = "test-bug"
  display_name = "test-bug"
  project      = "${var.project_id}"
}

resource "google_service_account_key" "test-bug" {
  service_account_id = "${google_service_account.test-bug.id}"
  #project            = "${var.project_id}"
  public_key_type    = "TYPE_X509_PEM_FILE"
  private_key_type   = "TYPE_GOOGLE_CREDENTIALS_FILE"
}
```

### Debug Output
https://paste.debian.net/hidden/def5b763/

### Expected Behavior
google_service_account_key is created

### Actual Behavior
Terraform fails to run

### Steps to Reproduce
1. Copy example HCL Configuration from above 
2. `terraform init`
3.`terraform apply -var project_id=YOURPROJECT -var region=europe-west1`

petemounce:
Confirmed by me.

vishen:
[...]

PR #1664, opened 2018-06-15
service_account_key: regression fix for v1.14

vishen:
Commit 8f31fec introduced a bug for the 'service_account_key' resource
where it required a project be set either in the provider or in the
resource for 'service_account_key', but a project isn't required if the
service account is a service account fully qualified name or a service
account email.

This PR relaxes the requirement that a project needs to be set for the
'service_account_key' resource, 'service_account' datasource and
'service_account_key' datasource, but will error if we try to build a
fully qualified name from a service account id when no project can be
found.

This also cleans up 'serviceAccountFQN' so it is slightly easier to
follow and return an error if there is no project but we need one to
build the service account fully qualified name.

Fixes: #1655

### DIFF
--- a/google/data_source_google_service_account.go
+++ b/google/data_source_google_service_account.go
@@ -43,15 +43,10 @@ func dataSourceGoogleServiceAccount() *schema.Resource {
 func dataSourceGoogleServiceAccountRead(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	// Get the project from the resource or fallback to the project
-	// in the provider configuration
-	project, err := getProject(d, config)
+	serviceAccountName, err := serviceAccountFQN(d.Get("account_id").(string), d, config)
 	if err != nil {
 		return err
 	}
-
-	// Get the service account as a fully qualified name
-	serviceAccountName := serviceAccountFQN(d.Get("account_id").(string), project)
 
 	sa, err := config.clientIAM.Projects.ServiceAccounts.Get(serviceAccountName).Do()
 	if err != nil {
--- a/google/utils.go
+++ b/google/utils.go
@@ -362,17 +362,27 @@ func lockedCall(lockKey string, f func() error) error {
 }
 
 // serviceAccountFQN will attempt to generate the fully qualified name in the format of:
-// "projects/(-|<project_name>)/serviceAccounts/<service_account_id>@<project_name>.iam.gserviceaccount.com"
-func serviceAccountFQN(serviceAccount, project string) string {
-	// If the service account id isn't already the fully qualified name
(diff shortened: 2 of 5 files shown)


google/data_source_google_service_account.go
```diff
@@ -44,15 +44,15 @@ func dataSourceGoogleServiceAccountRead(d *schema.ResourceData, meta interface{}
 	config := meta.(*Config)
 
 	// Get the project from the resource or fallback to the project
-	// in the provider configuration
-	project, err := getProject(d, config)
+	// in the provider configuration. If neither exist use an empty project
+	// when building the service account FQN
+	project, _ := getProject(d, config)
```
danawillow: Generally I'm not too into the idea of ignoring errors. What if `serviceAccountFQN` took `d` and `config` as parameters, and then it could do the getProject call at the time where there does actually need to be a project set?

vishen: That seems much better, I'll do that!

vishen: @danawillow I have made the requested changes, and it looks much cleaner now, thanks :)

google/data_source_google_service_account.go
```diff
@@ -43,16 +43,11 @@ func dataSourceGoogleServiceAccount() *schema.Resource {
 func dataSourceGoogleServiceAccountRead(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	// Get the project from the resource or fallback to the project
-	// in the provider configuration
-	project, err := getProject(d, config)
+	serviceAccountName, err := serviceAccountFQN(d.Get("service_account_id").(string), d, config)
```
danawillow: Looking at the diff, I think this one needs to be just `account_id`

vishen: Fixed.

google/utils.go
```diff
@@ -362,17 +362,26 @@ func lockedCall(lockKey string, f func() error) error {
...
-		// If the service account id is an email
-		if strings.Contains(serviceAccount, "@") {
-			serviceAccount = "projects/-/serviceAccounts/" + serviceAccount
-		} else {
-			// If the service account id doesn't contain the email, build it
-			serviceAccount = fmt.Sprintf("projects/-/serviceAccounts/%s@%s.iam.gserviceaccount.com", serviceAccount, project)
-		}
+// "projects/(-|<project>)/serviceAccounts/<service_account_id>@<project>.iam.gserviceaccount.com"
+// A project is required if we are trying to build the FQN from a service account id and
+// and error will be returned in this case
```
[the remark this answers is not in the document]
vishen: Done.

PR #1665, opened 2018-06-15
container_cluster: added disk_type node configuration

vishen:
Added node config 'disk_type' which can either be 'pd-standard' or
'pd-ssd', if left blank 'pd-standard' will be the default used by google
cloud.

Attempted fix for https://github.com/terraform-providers/terraform-provider-google/issues/1656

### DIFF
--- a/google/node_config.go
+++ b/google/node_config.go
@@ -1,11 +1,12 @@
 package google
 
 import (
+	"strconv"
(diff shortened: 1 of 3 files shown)


vishen:
Yup, sounds good to me! I am not in any rush to get this in :)

PR #1666, opened 2018-06-17
Allow functions to be updated

dustinblackman:
## Overview

Adds the ability to update Google Functions archive URLs, instead of forcing the resource to be deleted and recreated resulting in downtime of the function.

### DIFF
--- a/google/resource_cloudfunctions_function.go
+++ b/google/resource_cloudfunctions_function.go
@@ -129,7 +129,7 @@ func resourceCloudFunctionsFunction() *schema.Resource {
 			"source_archive_object": {
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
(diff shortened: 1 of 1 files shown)


google/resource_cloudfunctions_function.go
```diff
@@ -129,7 +129,7 @@ func resourceCloudFunctionsFunction() *schema.Resource {
 			"source_archive_object": {
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
+				ForceNew: false,
```
paddycarver: Because the empty value for a boolean is `false`, we can just delete this line entirely.

dustinblackman:
Sorry for the delay. Fixed in https://github.com/terraform-providers/terraform-provider-google/pull/1781. Thanks!

PR #1670, opened 2018-06-19
Add diff suppress for empty/default pod_security_policy_config

### LINKED ISSUE #1667: Unable to set pod_security_policy_config to 0
Hey, 
We set pod_security_policy_config to 1 recently, but then decided to put it back to 0, but we now get this in every single plan (the apply works, but seems to never actually be set).

```
  ~ module.kubernetes.google_container_cluster.cluster
      pod_security_policy_config.#: "1" => "0"
```

Any ideas?

emilymye:
Fixes #1667

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -389,6 +389,7 @@ func resourceContainerCluster() *schema.Resource {
 						},
 					},
 				},
+				DiffSuppressFunc: podSecurityPolicyCfgSuppress,
 			},
 
 			"project": {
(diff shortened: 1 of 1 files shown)


emilymye:
accidently deleted the remote branch herp

google/resource_container_cluster.go
```diff
@@ -1555,3 +1556,17 @@ func masterAuthClientCertCfgSuppress(k, old, new string, r *schema.ResourceData)
 
 	return strings.HasSuffix(k, ".issue_client_certificate") && old == "" && new == "true"
 }
+
+func podSecurityPolicyCfgSuppress(k, old, new string, r *schema.ResourceData) bool {
+	if strings.HasSuffix(k, "pod_security_policy_config.#") && old == "1" && new == "0" {
```
ndmckinley: The `HasSuffix` doesn't make much sense if the next step is to `GetOk` an unprefixed `pod_security_policy_config` - should it just be '=='?

emilymye: I wasn't sure if k actually starts with pod_security_policy_config or has resource-specific naming prefix (i.e. google_container_cluster.my_cluster.pod_security_policy_config.#). Since it wasn't the same value necessarily as the keys for r, I thought I'd err on the cautious side but I can change it.

PR #1671, opened 2018-06-19
Fix Bigtable docs for instance/cluster [skip ci]

mbrukman:
An instance is an abstract container of clusters, it's the cluster that
has the nodes and holds the data, so the number of nodes and location
apply to the cluster.


emilymye:
Thanks for submitting!

PR #1677, opened 2018-06-19
Add project variable for GCS service account datasource

bcadiot:
Hi,

This PR is a small improvement on #1110 allowing to change the project used by this datasource.
It is required when the provider project is not the same as the bucket.

Here is an example : 
```hcl
resource "google_project" "project" {
  name = "test-project"
  project_id = "test-project"
  org_id     = "1234567"
}

data "google_storage_project_service_account" "gcs_account" {
  project = "${google_project.project.project_id}"
}

// Bucket and PubSub topic definitions
// ...
```
I can provide more explanation if needed.

Best,
Bastien


emilymye:
Thanks for submitting!

almorelle:
Nice ! Thanks @bcadiot

Issue #1676, opened 2018-06-19
Importing resources with a project ID that differs to the provider project


roobert:
Hi,

As requested by @paddycarver I'm opening this issue to discuss a feature request regarding cross-gcp-project importing of resources.

It's is currently possible to create resource definitions which cannot be easily imported, specifically, when specifying the `project` parameter on a resource.

For example:

```
provider "google" {
  project     = "test0"
  region      = "europe-west1"
  version     = "~> 1.9"
}

terraform {
  backend "gcs" {
    bucket  = "terraform-state-test0"
  }
}

"resource" "google_dns_managed_zone" "example-com" {
  "name" = "example-com"
  "dns_name" = "example.com."
  "project" = "test1"
}
```

Although this config is valid and can be used to create/update/delete the managed zone for `example-com` in `test1`, if the zone exists already, it is not possible to import the resource into the state file using `terraform import`.

There are a couple of possibilities here:

1) as suggested by @duckfez - create a secondary google provider with an alias such as `google-test1` and then use `terraform import -provider google-test1`. This could potentially work and seems like the provider agnostic solution but is somewhat clumsy.

2) update the terraform resources import templates to support including the project name.

e.g:
```
# instead of:
terraform import google_dns_managed_zone.prod prod-zone

# optionally specify the project:
terraform import google_dns_managed_zone.prod prod-zone test1
```

This solution would be the ideal solution, however I'm not sure if there are any standards or limitations on how `terraform import` should function regarding permitting additional parameters and allowing sensible defaults.

If limitations do exist and it's necessary to update all of the `terraform import` templates to include the `project` parameter, is it feasible and would that be an acceptable change?
```
# resource template which always specifies project since no default can be assumed:
terraform import google_dns_managed_zone.prod test1/prod-zone
```

If anyone can think of any other solutions I'd be interested in hearing them.

roobert:
I've just noticed that some resources (specifically, the global_address resource) supports specifying the project: https://www.terraform.io/docs/providers/google/r/compute_global_address.html#import

So it looks like this would be the best method to support this feature?

paddycarver:
Hey @roobert! Thanks for opening this bug, I appreciate it. As you mentioned, Global Addresses (and other resources) have a template of how we usually resolve this, and I think we'd like to follow that approach here, for consistency.

I see you've opened a PR for this. Thanks! I'll go ahead and review the code and we can hopefully get this fixed ASAP.

[...]

PR #1679, opened 2018-06-19
[WIP] google_dns_managed_zone - project parameter for resource importer

roobert:
Please see #1676


sjungwirth:
Does it make sense to fix this issue for `google_dns_record_set` at the same time? https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_dns_record_set.go#L308

PR #1680, opened 2018-06-19
Add documentation for client certificate authorization setting

jhyelton:
Adding documentation for client certificate auth settings that was added in #1396.


emilymye:
Thank you!

PR #1681, opened 2018-06-20
Allow recreation of recently deleted project and org custom roles

### LINKED ISSUE #1668: Terraform doesn't undelete Custom roles on apply
Hi,

I'm facing an issue with the google cloud provider, If i have custom Roles and when i run terraform destroy, the role goes into a deleted mode and stays so for 60 days, if i run terraform apply again it fails because it tries to create the role again, is it possible to make the provider check if a role exist already and undelete it if it does instead of trying to create it again ?

emilymye:
* Undelete-update recently soft-deleted roles instead of giving error

Fixes #1668

### DIFF
--- a/google/resource_google_organization_iam_custom_role.go
+++ b/google/resource_google_organization_iam_custom_role.go
@@ -50,9 +50,10 @@ func resourceGoogleOrganizationIamCustomRole() *schema.Resource {
 				Optional: true,
 			},
 			"deleted": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				Default:  false,
+				Type:       schema.TypeBool,
+				Optional:   true,
+				Default:    false,
+				Deprecated: `deleted will be converted to a computed-only field soon - if you want to delete this role, please use destroy`,
 			},
 		},
 	}
@@ -62,39 +63,47 @@ func resourceGoogleOrganizationIamCustomRoleCreate(d *schema.ResourceData, meta 
 	config := meta.(*Config)
 
 	if d.Get("deleted").(bool) {
-		return fmt.Errorf("Cannot create a custom organization role with a deleted state. `deleted` field should be false.")
+		return fmt.Errorf("cannot create a custom organization role with a deleted state. `deleted` field should be false.")
 	}
 
 	org := d.Get("org_id").(string)
 	roleId := fmt.Sprintf("organizations/%s/roles/%s", org, d.Get("role_id").(string))
+	orgId := fmt.Sprintf("organizations/%s", org)
+
+	// Look for role with given ID.
+	// If it exists in deleted state, update to match "created" role state
+	// If it exists and and is enabled, return error - we should not try to recreate.
 	r, err := config.clientIAM.Organizations.Roles.Get(roleId).Do()
 	if err == nil {
 		if r.Deleted {
-			// Roles have soft deletes - creating a role with the same name
-			// as a recently deleted role must instead be undelete/update.
+			// This role was soft-deleted; update to match new state.
 			d.SetId(r.Name)
-			return resourceGoogleOrganizationIamCustomRoleUpdate(d, meta)
+			if err := resourceGoogleOrganizationIamCustomRoleUpdate(d, meta); err != nil {
+				// If update failed, make sure it wasn't actually added to state.
+				d.SetId("")
+				return err
+			}
+		} else {
+			// If a role with same name exists and is enabled, just return error
+			return fmt.Errorf("Custom project role %s already exists and must be imported", roleId)
 		}
-		// If old role with same name exists, just return error
-		return fmt.Errorf("Custom project role %s already exists and must be imported", roleId)
+	} else {
+		// If no role was found, actually create a new role.
+		role, err := config.clientIAM.Organizations.Roles.Create(orgId, &iam.CreateRoleRequest{
+			RoleId: d.Get("role_id").(string),
+			Role: &iam.Role{
+				Title:               d.Get("title").(string),
+				Description:         d.Get("description").(string),
+				Stage:               d.Get("stage").(string),
+				IncludedPermissions: convertStringSet(d.Get("permissions").(*schema.Set)),
+			},
+		}).Do()
+		if err != nil {
+			return fmt.Errorf("Error creating the custom organization role %s: %s", d.Get("title").(string), err)
+		}
+
+		d.SetId(role.Name)
 	}
-
-	// If no role is found, actually create a new role.
-	role, err := config.clientIAM.Organizations.Roles.Create("organizations/"+d.Get("org_id").(string), &iam.CreateRoleRequest{
-		RoleId: d.Get("role_id").(string),
-		Role: &iam.Role{
-			Title:               d.Get("title").(string),
-			Description:         d.Get("description").(string),
-			Stage:               d.Get("stage").(string),
-			IncludedPermissions: convertStringSet(d.Get("permissions").(*schema.Set)),
-		},
-	}).Do()
-
-	if err != nil {
-		return fmt.Errorf("Error creating the custom organization role %s: %s", d.Get("title").(string), err)
-	}
-
-	d.SetId(role.Name)
 
 	return resourceGoogleOrganizationIamCustomRoleRead(d, meta)
 }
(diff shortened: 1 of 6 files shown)


google/resource_google_organization_iam_custom_role.go
```diff
@@ -118,12 +133,23 @@ func resourceGoogleOrganizationIamCustomRoleUpdate(d *schema.ResourceData, meta
 			if err := resourceGoogleOrganizationIamCustomRoleDelete(d, meta); err != nil {
 				return err
 			}
-		} else {
+			d.SetPartial("deleted")
+		}
+	}
+
+	// If role is not deleted, make sure it exists and undelete if needed.
```
danawillow: Because this gets called on create, the next bit about d.HasChange will actually return true and send a patch request. It's not a huge deal, but it's still an extra API call that we don't really need. If it's not hard, I think the Undelete should probably just get called straight from Create.

paddycarver:
The context of this one scares me, but on the assumption that we can update roles _as if_ they were just now created (i.e., all user-specifiable fields can be updated), I can't see why not.

Projects, on the other hand (I'm not sure if those are intended, but as they have a similar soft-delete, I just want to be explicit about this) I'd argue against taking this approach with, because there are a lot of things about a project that would need to be cleaned up, and some of them _can't_ be cleaned up, so we'd be saying the user has a fresh project when they, in fact, do not.

Anyways, just wanted to call that subtlety out. I'll take a pass at reviewing the code now :)

google/resource_google_organization_iam_custom_role_test.go
```diff
@@ -78,6 +78,35 @@ func TestAccOrganizationIamCustomRole_undelete(t *testing.T) {
...
+	t.Parallel()
+
+	org := getTestOrgFromEnv(t)
+	roleId := "tfIamCustomRole" + acctest.RandString(10)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGoogleOrganizationIamCustomRoleDestroy,
+		Steps: []resource.TestStep{
```
paddycarver: This is going to sound really nitpicky, but could we verify that recreating it gets us the results we expect? I'd just hate for it to do the deletion regularly, only for us to find out its failing to set the other properties, or something.

google/resource_google_project_iam_custom_role.go
```diff
@@ -124,12 +138,23 @@ func resourceGoogleProjectIamCustomRoleUpdate(d *schema.ResourceData, meta inter
 			if err := resourceGoogleProjectIamCustomRoleDelete(d, meta); err != nil {
 				return err
 			}
-		} else {
+			d.SetPartial("deleted")
+		}
+	}
+
+	// If role is not deleted, make sure it exists and undelete if needed.
```
paddycarver: Same comment Dana pointed out about the unnecessary PATCH applies here, I think.

google/resource_google_project_iam_custom_role_test.go
```diff
@@ -71,6 +71,33 @@ func TestAccProjectIamCustomRole_undelete(t *testing.T) {
...
 
+func TestAccProjectIamCustomRole_createAfterDestroy(t *testing.T) {
+	t.Parallel()
+
+	roleId := "tfIamCustomRole" + acctest.RandString(10)
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGoogleProjectIamCustomRoleDestroy,
+		Steps: []resource.TestStep{
```
paddycarver: Same thing here about verifying the resource has the expected values.

google/resource_google_organization_iam_custom_role.go
emilymye: I might be confused - isn't that what we want? If this gets called on create, what we want is the 1. role.undelete 2. role.patch. If we just call undelete then we still have to call patch to essentially reset the previously deleted role to whatever value we are trying to create.

google/resource_google_organization_iam_custom_role_test.go
emilymye: Done!

emilymye:
Looking at the IAM custom roles docs, it seems like there are two stages of deletion for custom roles:

**Stage 1, 0-7 days after deletion**: The role can be undeleted but otherwise cannot be used.
**Stage 2, 7-37 days after deletion**: The role cannot be undeleted but the role ID cannot be used to make a new role. I'm not sure whether calling `roles.get()` during this stage will return anything.

There are a couple of bugs with the current implementation. 

Currently roles can be deleted through either setting `role.deleted=true` or `terraform destroy`. Problems with this:

- If the user runs `terraform destroy` and `terraform apply` within 37 days without changing the custom role ID, this will error as shown in issue #1668
- If the user makes change in config `deleted=true`, the role is deleted in GCP. `role.get()` returns a value for at least 7 days, no changes can be made to this value other than `role.undelete()`. That means running `terraform apply` with any other changed attributes will fail because it attempts to patch a deleted role. Running `terraform destroy` will fail because it attempts to redelete a deleted object.  

IMO `deleted` should be a computed value - I don't think deleted roles should be treated as existing in terraform state at any point. This is also a fairly large breaking change so we probably can't make this change right away. 

What I think this PR should do is:
- If Role in Stage 1: When user runs `terraform apply`, `undelete` and `patch` to have `role/$ROLE_ID` be the expected value from Terraform state. This is regardless of whether the `terraform plan` is running ResourceCreate (e.g. apply after destroy) or ResourceUpdate (if the state has or was refreshed and found role `deleted=true`).
- If Role in Stage 2:  If user runs `terraform apply`, return an error. We can't do anything about this, so the user pretty much has to abandon that role id. If user runs `terraform destroy`, 

In either case, I'll need to add a check to ResourceDelete so `terraform destroy` just ignores roles if the role has 'deleted=true' or if `role.get()` returns nothing. I'm not sure what reading the role returns at during stage 2 - I'll need to find someone to ask in IAM or find a role that has been deleted for > 7 days but < 37 days, neither of which give timely answers.

**tl;dr**: @paddycarver @danawillow:  I think the current solution should run undelete-patch for either create or update, and we should make `deleted` Computed. Also, I need to add a check for deletion.

danawillow:
@paddycarver @morgante why is project deletion relevant here? This PR is just about custom roles.

morgante:
@danawillow I parsed the title as "Allow recreation of recently deleted (projects) and (org custom roles)" but now realize it's "Allow recreation of recently deleted (project and org) custom roles" so retract my comments. :)

paddycarver:
Parsed the title similarly to @morgante, and am also wary of setting precedent. :/

I'm still parsing and processing @emilymye's awesome comment, to try and wrap my brain about this, then I'll take another review pass.

emilymye:
@paddycarver @danawillow I made some new changes that I think (think??) will not cause breakage and actually handle all errors. I added a deprecation message for deleted (as advised by Dana) and tried to make the Create/Update logic easier to follow, but pls re-review. 

Also, I don't know if I have access to the team city still but can someone run the org tests for me? I need to set that up -__-

project test run:
TF_ACC=1 go test ./google -v --run="TestAccProjectIamCustomRole" -timeout 120m
=== RUN   TestAccProjectIamCustomRole_import
=== PAUSE TestAccProjectIamCustomRole_import
=== RUN   TestAccProjectIamCustomRole_basic
=== PAUSE TestAccProjectIamCustomRole_basic
=== RUN   TestAccProjectIamCustomRole_undelete
=== PAUSE TestAccProjectIamCustomRole_undelete
=== RUN   TestAccProjectIamCustomRole_createAfterDestroy
=== PAUSE TestAccProjectIamCustomRole_createAfterDestroy
=== CONT  TestAccProjectIamCustomRole_import
=== CONT  TestAccProjectIamCustomRole_undelete
=== CONT  TestAccProjectIamCustomRole_createAfterDestroy
=== CONT  TestAccProjectIamCustomRole_basic
--- PASS: TestAccProjectIamCustomRole_import (2.24s)
--- PASS: TestAccProjectIamCustomRole_basic (3.54s)
--- PASS: TestAccProjectIamCustomRole_undelete (3.84s)
--- PASS: TestAccProjectIamCustomRole_createAfterDestroy (4.14s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	4.313s

paddycarver:
I'll take a look and run the tests tonight, hopefully. In the meantime, I also made sure you have access to the CI server. Feel free to drop an email to paddy@hashicorp if you're having trouble, and I can give you more info.

google/resource_google_organization_iam_custom_role.go
```diff
@@ -62,25 +63,48 @@ func resourceGoogleOrganizationIamCustomRoleCreate(d *schema.ResourceData, meta
...
+			if err := resourceGoogleOrganizationIamCustomRoleUpdate(d, meta); err != nil {
+				// If update failed, make sure it wasn't actually added to state.
+				d.SetId("")
+				return err
+			}
+		} else {
+			// If a role with same name exists and is enabled, just return error
+			return fmt.Errorf("Custom project role %s already exists and must be imported", roleId)
+		}
+	} else {
```
danawillow: Do you want to check that the error was actually a 404?

emilymye:
WHOOPS I forgot about this PR

Updates made, also added a warning to docs so custom roles might be less confusing in general. This is a real pickle

$ make testacc TEST=./google TESTARGS='--run="TestAccProjectIamCustomRole"'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google -v --run="TestAccProjectIamCustomRole" -timeout 120m
=== RUN   TestAccProjectIamCustomRole_import
=== PAUSE TestAccProjectIamCustomRole_import
=== RUN   TestAccProjectIamCustomRole_basic
=== PAUSE TestAccProjectIamCustomRole_basic
=== RUN   TestAccProjectIamCustomRole_undelete
=== PAUSE TestAccProjectIamCustomRole_undelete
=== RUN   TestAccProjectIamCustomRole_createAfterDestroy
=== PAUSE TestAccProjectIamCustomRole_createAfterDestroy
=== CONT  TestAccProjectIamCustomRole_import
=== CONT  TestAccProjectIamCustomRole_createAfterDestroy
=== CONT  TestAccProjectIamCustomRole_undelete
=== CONT  TestAccProjectIamCustomRole_basic
--- PASS: TestAccProjectIamCustomRole_import (2.78s)
--- PASS: TestAccProjectIamCustomRole_basic (3.79s)
--- PASS: TestAccProjectIamCustomRole_createAfterDestroy (4.19s)
--- PASS: TestAccProjectIamCustomRole_undelete (4.49s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	4.672s

PR #1684, opened 2018-06-20
Reject disk shrinkage on SQL instances

### LINKED ISSUE #1612: Google TF provider should not attempt decreasing Cloud SQL instance disk size
```
$ terraform version
Terraform v0.11.7
+ provider.google v1.10.0
```
As of now, TF returns an error that it failed to apply disk size reduction on a Cloud SQL instance. But Google TF provider should not even attempt decreasing Cloud SQL instance disk size, as such operation is not supported by Google.

ndmckinley:
Fixes #1612.

Tested with a local example, which did require ForceNew when the disk shrank but updated an increased disk size without trouble.  Running cloud tests now.

### DIFF
--- a/google/resource_sql_database_instance.go
+++ b/google/resource_sql_database_instance.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/terraform/helper/customdiff"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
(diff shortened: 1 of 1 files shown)


ndmckinley:
All pass, except
```
--- FAIL: TestAccSqlDatabaseInstance_basic3 (641.74s)

                * google_sql_database_instance.instance: 1 error(s) occurred:

                * google_sql_database_instance.instance: Error, failed to delete default 'root'@'*' user, but the database was created successfully: 
```

Which I think is a flake, but I'll rerun to be sure.

ndmckinley:
Yep, it passed on the second try.  :)  This is ready to merge.

PR #1687, opened 2018-06-21
Update to private clusters

johnsw87:
Private clusters maintain a public IP address for kubectl to communicate with the master, the workers are internal IP addresses only.

### DIFF
--- a/website/docs/r/container_cluster.html.markdown
+++ b/website/docs/r/container_cluster.html.markdown
@@ -164,7 +164,7 @@ output "cluster_ca_certificate" {
 
 * `private_cluster` - (Optional, [Beta](/docs/providers/google/index.html#beta-features)) If true, a
     [private cluster](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters) will be created, which makes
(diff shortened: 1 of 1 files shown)


website/docs/r/container_cluster.html.markdown
```diff
@@ -164,7 +164,7 @@ output "cluster_ca_certificate" {
 
 * `private_cluster` - (Optional, [Beta](/docs/providers/google/index.html#beta-features)) If true, a
     [private cluster](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters) will be created, which makes
-    the master inaccessible from the public internet and nodes do not get public IP addresses either. It is mandatory to specify
+    the worker nodes inaccessible from the public internet and nodes do not get public IP addresses either. It is mandatory to specify
```
paddycarver: This seems to imply that the nodes are inaccessible, and nodes do not get public IP addresses either.

Could we just remove everything from `, which makes` to `and nodes` and just go with something like `meaning nodes do not get public IP addresses`?

PR #1692, opened 2018-06-22
Fix the perma-diff in storage bucket ACLs

paddycarver:
Add a CustomDiff function to storage bucket ACLs that will ignore a diff
if the config and state have the same role_entities, even if they're in
a different order.

Fixes #1525 and #892.

### DIFF
--- a/google/resource_storage_bucket_acl.go
+++ b/google/resource_storage_bucket_acl.go
@@ -13,10 +13,11 @@ import (
 
 func resourceStorageBucketAcl() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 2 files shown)


tmshn:
Oh Terraform has a option `CustomizeDiff`...! `DiffSuppressFunc` was all I know.
This looks better solution than #1525 🤓 👍

paddycarver:
It's a relatively new API. I actually like #1525 better as a solution, and would like to leave it open for merging as part of 2.0.0, but with backwards compatibility concerns, `CustomizeDiff` may be a better for-the-moment solution.

PR #1695, opened 2018-06-22
Feature/bigquery require partition filter

Malet:
Add the `require_partition_filter` option for bigquery tables.


paddycarver:
Before we can merge this in, we'd also need an acceptance test that exercises the new behavior, so we can verify it works (and keeps working over time).

ndmckinley:
I'm closing since there's been no movement on this in a while - please feel free to reopen and update the PR at any time, ping us if we don't see it!

PR #1717, opened 2018-06-28
Implement 'licenses' field in compute_image resource

Mierdin:
This PR implements the `licenses` field for GCP images in this provider (already exists in the Google API, just needed to be plumbed through).

Terraform users can now use this provider to create images for [nested virtualization](https://cloud.google.com/compute/docs/instances/enable-nested-virtualization-vm-instances), as well as other use cases for setting licenses on an image.

Closes https://github.com/terraform-providers/terraform-provider-google/issues/1045

### DIFF
--- a/google/resource_compute_image.go
+++ b/google/resource_compute_image.go
@@ -106,6 +106,13 @@ func resourceComputeImage() *schema.Resource {
 				Set:      schema.HashString,
 			},
 
+			"licenses": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
 			"label_fingerprint": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
@@ -156,6 +163,11 @@ func resourceComputeImageCreate(d *schema.ResourceData, meta interface{}) error 
 
 	if _, ok := d.GetOk("labels"); ok {
 		image.Labels = expandLabels(d)
+	}
+
+	// Load up the licenses for this image if specified
+	if _, ok := d.GetOk("licenses"); ok {
+		image.Licenses = licenses(d)
 	}
 
 	// Read create timeout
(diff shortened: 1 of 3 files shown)


google/resource_compute_image_test.go
```diff
@@ -29,6 +29,7 @@ func TestAccComputeImage_basic(t *testing.T) {
 					testAccCheckComputeImageFamily(&image, "family-test"),
 					testAccCheckComputeImageContainsLabel(&image, "my-label", "my-label-value"),
 					testAccCheckComputeImageContainsLabel(&image, "empty-label", ""),
+					testAccCheckComputeImageContainsLicense(&image, "https://www.googleapis.com/compute/v1/projects/vm-options/global/licenses/enable-vmx"),
```
paddycarver: I think I'd rather see this in its own test, just so we can verify that creating an instance _without_ a license _also_ works.

google/resource_compute_image.go
```diff
@@ -106,6 +106,13 @@ func resourceComputeImage() *schema.Resource {
 				Set:      schema.HashString,
 			},
 
+			"licenses": &schema.Schema{
+				Type:     schema.TypeList,
```
paddycarver: Does order matter for licenses? Is it reasonable to expect someone to try and reference each individual license using interpolations?

Mierdin: Neither the [CLI](https://cloud.google.com/sdk/gcloud/reference/compute/images/create) or [API](https://godoc.org/google.golang.org/api/compute/v1#Image) docs seem to give an indication that order matters. Does your previous comment about using a Set instead of a List impact this?

Mierdin: Ah, [this is nice](https://www.terraform.io/docs/extend/schemas/schema-types.html#typeset). So Google isn't expecting the order to be meaningful, but I'd say it might still be intuitive to keep this as a TypeList in case the user wants to be able to refer to it using interpolations. (I'm assuming they won't be able to do this if it's an unordered `TypeSet`). So not necessarily for compatibility reasons, but for UX reasons, might be useful to keep it as-is. Let me know if you disagree, I'm not married to it.

google/resource_compute_image_test.go
Mierdin: Yup you're right, this was lazy. Updated in d7c1bf9

Mierdin:
Thanks @paddycarver!

paddycarver:
Thank you! Apologies for the wait. I'm not sure how I feel about list vs set still, _but_ we have a path forward if we decide we want to change it in the future, so let's ship it. :D

PR #1731, opened 2018-07-04
Update compute_ssl_certificate.html.markdown

nyurik:
It was very confusing for `name_prefix` to be deprecated and undeprecated later.
Adding an example of using random_id in addition to name_prefix

### DIFF
--- a/website/docs/r/compute_ssl_certificate.html.markdown
+++ b/website/docs/r/compute_ssl_certificate.html.markdown
@@ -33,6 +33,12 @@ resource "google_compute_ssl_certificate" "default" {
 resource "random_id" "certificate" {
   byte_length = 4
   prefix      = "my-certificate-"
+
+  # For security, do not expose raw certificate values in the output
+  keepers {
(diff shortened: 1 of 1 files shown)


website/docs/r/compute_ssl_certificate.html.markdown
```diff
@@ -22,6 +22,29 @@ resource "google_compute_ssl_certificate" "default" {
...
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+# You may also want to control name generation explicitly:
+
+resource "random_id" "certificate" {
+  byte_length = 4
+  prefix      = "my-certificate-"
```
paddycarver: This would actually not work as intended, I believe--I think we need to use [keepers](https://www.terraform.io/docs/providers/random/index.html#resource-quot-keepers-quot-) for this to generate a new random ID.

nyurik: @paddycarver thx for catching it, fixed!  (or at least I hope i did :) )

PR #1733, opened 2018-07-05
Increase max number of master_authorized_networks to 20, per GKE docs.

rtoma:
See https://cloud.google.com/kubernetes-engine/docs/how-to/authorized-networks#setting_up_an_authorized_network


rtoma:
FYI - I have requested GCP to increase the max number of master authorized networks via https://issuetracker.google.com/issues/111289536

PR #1737, opened 2018-07-06
Add nextPageToken to fields argument against ServiceUsage api

### LINKED ISSUE #1736: google_project_services resource does not page through ServiceUsage API response
### Terraform Version
0.11.7

Google provider version: 1.15

### Affected Resource(s)
- google_project_services

### Debug Output
I am not going to provide a full debug output because it contains too much proprietary content, and the issue is really narrowed down to a single line, which I will paste here:

```
2018-07-06T16:24:20.804Z [DEBUG] plugin.terraform-provider-google_v1.15.0_x4: 2018/07/06 16:24:20 [DEBUG] Google API Request Details:
2018-07-06T16:24:20.804Z [DEBUG] plugin.terraform-provider-google_v1.15.0_x4: ---[ REQUEST ]---------------------------------------
2018-07-06T16:24:20.804Z [DEBUG] plugin.terraform-provider-google_v1.15.0_x4: GET /v1beta1/projects/<project-name>/services?alt=json&fields=services%2Fname&filter=state%3AENABLED HTTP/1.1
2018-07-06T16:24:20.804Z [DEBUG] plugin.terraform-provider-google_v1.15.0_x4: Host: serviceusage.googleapis.com
2018-07-06T16:24:20.804Z [DEBUG] plugin.terraform-provider-google_v1.15.0_x4: User-Agent: google-api-go-client/0.5 Terraform/0.11.7 (+https://www.terraform.io)
```

The request contained in that log message improperly sets the `filter` parameter in a way that discards the `nextPageToken` returned by the API, so Terraform does not even know whether there are additional response pages.  Ideally it would either retrieve the next page token and page through the responses, or it would set the pageSize parameter in a way that ensures that all enabled APIs will be returned (ie set pageSize to some number larger than the total number of APIs available).

### Expected Behavior
Terraform should have placed an API request that did not discard the nextPageToken, and it should have paged through the responses.

### Actual Behavior
Terraform placed an API request that filters the nextPageToken out of the response, so it does not even know whether there is a second page (or beyond) of results

### Steps to Reproduce
1. Create a google_project_services resource with more than 30 or 40 APIs enabled (it seems that the number of items on the first response page depends on the number of enabled APIs.  When we have had 38 APIs enabled, the API returned 30 / page; when we had 46, it returned 40 / page).
2. `terraform apply`

mchalek:
As far as I can tell from making cURL requests to the serviceusage API, and from guessing about what `Pages` does in the google API go client, I suspect this may fix #1736.

However I have no real way of building or testing this.  Just figured it might be helpful to push this to help get more eyes on it, because this problem is a major annoyance to me and my team right now.


paddycarver:
Hey @mchalek, appreciate the PR! As you mentioned you don't have a way of building or testing this, do you mind if I take over the PR and push some acceptance tests and verify that this change resolves the issue? If you'd rather do it yourself, that's great! I'd be happy to point you to some resources. I just don't want to come in and commandeer your PR without at least asking. :)

mchalek:
hi @paddycarver ,

I don't mind at all!  I just wanted to do whatever I could to speed this along.  At this point I suspect that the fastest route to getting this fixed is to have you commandeer the PR, so please do!

thanks,

Kevin

chrisfowles:
I've built and tested this change locally and can confirm it resolves the issue for me. Thanks @mchalek - was pulling my hair out.

mchalek:
Awesome, thanks @chrisfowles for confirming that it works.

@paddycarver assuming your tests confirm that it works, do you have an ETA on when this may be available, e.g. in a new published bugfix version of the google provider?  Like Chris, I have been pulling my hair out due to this issue, but I'm locked into running Terraform on internal infra that I can't easily modify, so I may be stuck waiting until the release is available in some official form.

paddycarver:
Assuming I can get the tests written up and they work (this is on my shortlist of Important Things To Do ASAP) before our next release, it'd be included in that. I don't think we've decided on a release date for it yet, but I believe we're due for one, so it'll probably be ~soon. Checking the release history can give a basic idea of our cadence, and I can't think of a good reason why this release should have a longer delay than any other release.

paddycarver:
I finally got the test for this sorted. Apologies for the delay; curating a list of 50+ services that actually worked with our tests took a bit of doing, just because I kept getting 403 errors for services, but not knowing which of the 50+ services triggered the error. :/ So I had to do some trial and error there. But! I think this should be ready now.

chrisfowles:
Hey, sorry guys. I was traveling today and didn't get a chance to get back to you.

Thanks for your efforts

mchalek:
hey @paddycarver that's great news!  Thanks for putting in the effort to get the tests sorted out.  Sorry I couldn't give you clearer direction on the set of services that cause the problem but I'm glad you were eventually able to reproduce it.

paddycarver:
It may be easier for the test to set a smaller page size, but unless I'm missing something, that would mean we'd need to reduce the page size for production users, too. Given that we have the list of services now, I don't really see a reason to require users to make more requests in production, but I'll fully admit that I could be missing something. In short, I'm game for decreasing the page size, but I'm not sure what advantage that would bring us, and it seems to have a clear disadvantage.

paddycarver:
Definitely, and I'll drop a comment that the point of having that many services is to get us to over 50, which will trigger pagination. Good call.

mchalek:
hooray!  thanks @paddycarver and everybody else!

PR #1753, opened 2018-07-10
prereqs for regional disks

danawillow:
In testing an upcoming `google_compute_region_disk` resource, I had to make these changes. Checking them in separately so that when the magician runs, these changes will already be a part of TF.

### DIFF
--- a/google/disk_type.go
+++ b/google/disk_type.go
@@ -1,12 +1,23 @@
 package google
 
 import (
+	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
 )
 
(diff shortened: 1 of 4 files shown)


google/disk_type.go
```diff
@@ -13,3 +14,13 @@ func readDiskType(c *Config, zone *compute.Zone, project, name string) (*compute
 		return nil, err
 	}
 }
+
+// readDiskType finds the disk type with the given name.
```
ndmckinley: readRegionDiskType, maybe?

danawillow: Good catch, done.

PR #1764, opened 2018-07-12
Be more specific about number of nodes in regional cluster

tpoindessous:
- remove Beta for regional cluster.
- explains how total number of nodes is calculated.

Thanks !

### DIFF
--- a/website/docs/r/container_cluster.html.markdown
+++ b/website/docs/r/container_cluster.html.markdown
@@ -75,9 +75,10 @@ output "cluster_ca_certificate" {
     in `initial_node_count` should be created in. Only one of `zone` and `region`
     may be set. If neither zone nor region are set, the provider zone is used.
 
-* `region` (Optional, [Beta](/docs/providers/google/index.html#beta-features))
+* `region` (Optional)
     The region to create the cluster in, for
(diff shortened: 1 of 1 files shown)


website/docs/r/container_cluster.html.markdown
```diff
@@ -75,9 +75,11 @@ output "cluster_ca_certificate" {
     in `initial_node_count` should be created in. Only one of `zone` and `region`
     may be set. If neither zone nor region are set, the provider zone is used.
 
-* `region` (Optional, [Beta](/docs/providers/google/index.html#beta-features))
+* `region` (Optional)
     The region to create the cluster in, for
     [Regional Clusters](https://cloud.google.com/kubernetes-engine/docs/concepts/multi-zone-and-regional-clusters#regional).
+    In a Regional Cluster, the number of nodes specified in `initial_node_count` is 
+    created in all zones of the region.
```
danawillow: Is this actually true? https://cloud.google.com/kubernetes-engine/docs/concepts/multi-zone-and-regional-clusters#overview says it's three zones.

tpoindessous:
Yes, sorry I worked only with regions with three zones. And documentation
does note that is three zones by default but it’s « overrided »

I will correct my PR

Thanks !

Le mer. 18 juil. 2018 à 22:40, Dana Hoffman <notifications@github.com> a
écrit :

paddycarver:
Thanks!

PR #1765, opened 2018-07-13
Sometimes we run tests on overloaded machines.

ndmckinley:
When we run a test on an overloaded machine, sometimes it needs more than 500 milliseconds to get 2 retries, making this test flake if the CPU is too busy.  :(


danawillow:
Merging as per @ndmckinley's request

PR #1786, opened 2018-07-17
make additional zones block removable

### LINKED ISSUE #1710: google_container_cluster: removing additional_zones block doesn't update cluster
Hi there,

When removing the 'additional_zones' block from a google_container_cluster resource, no changes are applied to the cluster. If you set it to [], the cluster is updated.

### Terraform Version
Google provider 0.15.0

### Affected Resource(s)
google_container_cluster

### Expected Behavior
It should remove the additional zones that were configured

### Actual Behavior
Nothing

### Steps to Reproduce
- create cluster in europe-west4-b and add this block:
```hcl
additional_zones = [
  "europe-west4-a",
]
```
- terraform apply to modify the cluster to run in the addtional zone as well.
- now remove the whole `additional_zone` block
- run terraform again, no updates will be done
- when you add back this code, the cluster is updated.
```hcl
additional_zones = [
]
```

danawillow:
This was originally set to be computed because the list returned from GCP contains the original zone, but since we remove that from the list it doesn't actually need to be computed. Fixes #1710.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -111,7 +111,6 @@ func resourceContainerCluster() *schema.Resource {
 			"additional_zones": {
 				Type:     schema.TypeSet,
 				Optional: true,
-				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 


danawillow:
Closing. The field _does_ need to be computed because it gets set automatically by the server for regional clusters.

PR #1795, opened 2018-07-19
don't call full read when updating billing acct

### LINKED ISSUE #1791: GAE instance creation silently fails on google_project resource but reports complete
### Terraform Version
Terraform v0.11.7

### Affected Resource(s)
- google_project

```hcl
// set the provider
provider "google" {
  credentials = "${file("account.json")}"
}

// create the project
resource "google_project" "terraform_project_poc" {
  name = "${var.project}"
  project_id = "${var.project}"
  billing_account = "XXXXXX-XXXXXX-XXXXXX"
  folder_id  = "folders/99999999999"
  app_engine {
    location_id = "us-east4"
  }
}
```

### Debug Output
https://gist.github.com/blzysh/7fbdf1747179b8dbe16deacbb8789c4c

### Expected Behavior
Based on the apply output I would expect GAE instance to be created:
```
google_project.terraform_project_poc: Creating...
  app_engine.#:                     "" => "1"
  app_engine.0.auth_domain:         "" => "<computed>"
  app_engine.0.code_bucket:         "" => "<computed>"
  app_engine.0.default_bucket:      "" => "<computed>"
  app_engine.0.default_hostname:    "" => "<computed>"
  app_engine.0.feature_settings.#:  "" => "<computed>"
  app_engine.0.gcr_domain:          "" => "<computed>"
  app_engine.0.location_id:         "" => "us-east4"
  app_engine.0.name:                "" => "<computed>"
  app_engine.0.serving_status:      "" => "<computed>"
  app_engine.0.url_dispatch_rule.#: "" => "<computed>"
  auto_create_network:              "" => "true"
  billing_account:                  "" => "XXXXXX-XXXXXX-XXXXXX"
  folder_id:                        "" => "99999999999"
  name:                             "" => "devops-tf11-sb"
  number:                           "" => "<computed>"
  org_id:                           "" => "<computed>"
  policy_data:                      "" => "<computed>"
  policy_etag:                      "" => "<computed>"
  project_id:                       "" => "devops-tf11-sb"
  skip_delete:                      "" => "<computed>"
google_project.terraform_project_poc: Still creating... (10s elapsed)
google_project.terraform_project_poc: Creation complete after 14s (ID: devops-tf11-sb)

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```
### Actual Behavior
Project created but if I go into the UI there is no GAE instance

### Steps to Reproduce
1. `terraform apply`
2. Check UI no GAE instance
3. `terraform apply` 
4: Note the change
```

  ~ google_project.terraform_project_poc
      app_engine.#:                     "0" => "1"
      app_engine.0.feature_settings.#:  "" => <computed>
      app_engine.0.location_id:         "" => "us-east4"
      app_engine.0.url_dispatch_rule.#: "" => <computed>
```
5: Check UI and GAE instance is there

### Important Factoids
Looks to be a timing thing, on second run it creates the instance.

danawillow:
I also fixed the time.sleep call to use 3 seconds, which I think was the original intent (vs 3 nanoseconds, boy is that an unintuitive result)

Fixes #1791.

### DIFF
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -588,12 +588,12 @@ func forceDeleteComputeNetwork(projectId, networkName string, config *Config) er
 func updateProjectBillingAccount(d *schema.ResourceData, config *Config) error {
 	pid := d.Id()
 	name := d.Get("billing_account").(string)
-	ba := cloudbilling.ProjectBillingInfo{}
+	ba := &cloudbilling.ProjectBillingInfo{}
 	// If we're unlinking an existing billing account, an empty request does that, not an empty-string billing account.
 	if name != "" {
 		ba.BillingAccountName = "billingAccounts/" + name
 	}
-	_, err := config.clientBilling.Projects.UpdateBillingInfo(prefixedProject(pid), &ba).Do()
+	_, err := config.clientBilling.Projects.UpdateBillingInfo(prefixedProject(pid), ba).Do()
 	if err != nil {
 		d.Set("billing_account", "")
 		if _err, ok := err.(*googleapi.Error); ok {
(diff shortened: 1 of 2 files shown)


danawillow:
Oh, should probably explain why this is the fix. So when we called the full read() function before, that does all the d.Sets based on what the API returns. Since app engine hasn't been set up at that point, it does a d.Set on the app engine block to be empty. Then, we attempt to set up app engine but the block is now empty, so it doesn't get created.

PR #1829, opened 2018-07-31
Add k8s support to shared vpc service project.

### LINKED ISSUE #1819: google_compute_shared_vpc_host_project support to enable Kubernetes engine
The resource for attaching a service project to a host project does not allow you to set the option to enable Kubernetes Engine access. This can be accomplished on GCP console.

### Affected Resource(s)
- google_compute_shared_vpc_service_project

```hcl
resource "google_compute_shared_vpc_service_project" "project" {
  host_project    = "${google_project.shared_vpc.project_id}"
  service_project = "${google_project.project.project_id}"

  depends_on = [
    "google_compute_shared_vpc_host_project.host_project",
  ]
}
```

ndmckinley:
This PR adds support for kubernetes subnetwork access to `google_compute_shared_vpc_service_project`.  The APIs aren't very related, but the cloud console allows it.  See here: https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#enabling_shared_vpc_and_granting_roles.

Fixes #1819, but let me say I'm not very attached to this as an interface - suggestions very welcome.

### DIFF
--- a/google/resource_compute_shared_vpc_service_project.go
+++ b/google/resource_compute_shared_vpc_service_project.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	cloudresourcemanager "google.golang.org/api/cloudresourcemanager/v1"
 	"google.golang.org/api/compute/v1"
 
 	"log"
@@ -32,6 +33,14 @@ func resourceComputeSharedVpcServiceProject() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+			"kubernetes_subnetwork_access": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


google/resource_compute_shared_vpc_service_project.go
```diff
@@ -85,6 +121,38 @@ func resourceComputeSharedVpcServiceProjectRead(d *schema.ResourceData, meta int
 		return nil
 	}
 
+	// We don't want to iterate through all subnetworks, checking to see which
+	// ones have k8s access.  In that sense, this resource isn't 'authoritative'.
```
danawillow: Why not?

ndmckinley: Haven't thought about it much - it's just that it's not the way that the buttons work on the cloud console.  Do you think we ought to?  That would be fine by me!

danawillow: How do the buttons work? It seems to me like you just click on the ones you want, and leave the ones you don't want unselected. So I guess you could argue either way that that's how the buttons work. I was thinking that if we did iterate through all of them, then import would work, right? You'd have to add a migration fn so we don't end up with another situation where people upgraded and saw diffs like in app engine, but that's not terrible.

website/docs/r/compute_shared_vpc_service_project.html.markdown
```diff
@@ -37,6 +37,14 @@ The following arguments are expected:
 
 * `service_project` - (Required) The ID of the project that will serve as a Shared VPC service project.
 
+- - -
+
+* `kubernetes_subnetwork_access` - (Optional) A list of strings.  Each string is the self link (or partial self-link)
```
danawillow: It probably makes sense to add something about this conflicting with the `google_compute_subnetwork_iam_*` resources (likewise in their documentation)

google/utils.go
```diff
@@ -57,6 +58,54 @@ func getRegionFromInstanceState(is *terraform.InstanceState, config *Config) (st
...
+		}
+	}
+	parts := strings.Split(reference, "/")
+	regionRegexp, _ := regexp.Compile("([a-z]*-[a-z]*[0-9])")
+	if len(parts) == 2 {
+		// This is arguable.  We're working on giving people the maximum
+		// possible number of ways to specify things - they might want
+		// project/resource, or project/region/resource, or region/resource.
+		// So we're checking if the thing before the first slash looks like a
+		// region - that is, foo-bar1.  Some projects probably look like regions,
```
danawillow: Does it makes sense to just call the regions api to check?

website/docs/r/compute_shared_vpc_service_project.html.markdown
morgante: @danawillow Does this mean you cannot use google_compute_subnetwork_iam_* resources with this field? I think that would be a major problem and hope we can consider a different approach. Subnet IAM applies to far more than just Kubernetes, so it seems like this would prevent users from sharing subnets to Kubernetes and GCE (for example).

morgante:
@ndmckinley @danawillow I'm not sure this needs to be implemented in the provider. IMO it potentially adds more complexity/risks than necessary when the same can be easily accomplished in a module.

We have [support](https://github.com/terraform-google-modules/terraform-google-project-factory/pull/7/files) for this in the project factory and I'd be happy to break it out into an independent module if there's demand.

website/docs/r/compute_shared_vpc_service_project.html.markdown
ndmckinley: `subnetwork_iam_member` would be fine - `iam_binding` would conflict.  Do you want me to discard the PR?  I don't have any ideas for another approach - the actual API is just IAM bindings, so anything that claims to be authoritative over subnetwork IAM bindings _will_ conflict, no way around it.

ndmckinley: Ping @morgante.

morgante: @ndmckinley Up to you. Personally, I'm not sure this needs to be in the provider. We don't handle normal subnet sharing in the provider either.

PR #1830, opened 2018-07-31
Added Environment Variables configuration to Cloud Functions

hpidcock:
This PR adds environment_variables configuration field to the cloud function resource. Additionally the vendor pkg has been updated for the cloud functions api.

It is considered a Beta feature by Google's own documentation, but is included in the v1 API documentation. See https://cloud.google.com/functions/docs/reference/rest/v1/projects.locations.functions#CloudFunction

Open to implementing it differently. It is currently an almost verbatim copy of the "labels" field.

### DIFF
--- a/google/resource_cloudfunctions_function_test.go
+++ b/google/resource_cloudfunctions_function_test.go
@@ -66,6 +66,8 @@ func TestAccCloudFunctionsFunction_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(funcResourceName,
 						"trigger_http", "true"),
 					testAccCloudFunctionsFunctionHasLabel("my-label", "my-label-value", &function),
+					testAccCloudFunctionsFunctionHasEnvironmentVariable("TEST_ENV_VARIABLE",
+						"test-env-variable-value", &function),
 				),
 			},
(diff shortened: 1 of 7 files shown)


google/resource_cloudfunctions_function_test.go
```diff
@@ -347,6 +353,21 @@ func testAccCloudFunctionsFunctionHasLabel(key, value string,
...
+	return func(s *terraform.State) error {
+		val, ok := function.EnvironmentVariables[key]
+		if !ok {
+			return fmt.Errorf("Environment Variable with key %s not found", key)
+		}
+
+		if val != value {
+			return fmt.Errorf("Environment Variable value did not match for key %s: expected %s but found %s", key, value, val)
+		}
+		return nil
```
ndmckinley: You can refactor this to use the assignment in the if statement, e.g.

`if val, ok := function.EnvironmentVariables[key]; ok {
  if val != value {
    return fmt.Errorf(...)
  }
} else {
  return fmt.Errorf(...)
}

hpidcock:
Thank-you for the feedback. I've applied the changes requested.

PR #1837, opened 2018-08-02
resource/google_container_cluster: Expose credentials as kubeconfig

radeksimko:
This is to bring GKE resource [in line with AKS](https://www.terraform.io/docs/providers/azurerm/r/kubernetes_cluster.html#kube_config_raw), with two caveats.

The provider is responsible for generating the config, as opposed to the API (like in AKS' case).

I'm aware that bringing the baggage of 73k LOC of dependencies is not pleasant and I spent some time trying to strip out what's not necessary, but eventually landed on conclusion that I'd basically have to re-implement big part of the client library & apimachinery 😑

I was expecting the client library to be lightweight and built to avoid vendoring other K8S projects, it apparently [still depends on apimachinery](https://github.com/kubernetes/client-go/blob/v8.0.0/tools/clientcmd/api/latest/latest.go#L20-L23) 😞 

The special "encoding" helpers buried in those libraries actually do much more than just convert [structs](https://github.com/kubernetes/kubernetes/blob/v1.11.1/staging/src/k8s.io/client-go/tools/clientcmd/api/types.go#L29) to YAML.

It is possible that I missed something obvious and there's different (better & cleaner) approach?


ndmckinley:
@danawillow Sorry to kick this your way, but I know you're the containers expert and I don't really have an opinion about the big revendor, and thought you might.

danawillow:
Sorry for taking so long to get back to this. I've heard rumor that go modules will be a thing at some point, which will remove the need for large vendorings like this. I'd like to hold off on merging this until then, if that's all right.

magic7s:
Will this be merged? I could use a way to get kubeconfig without gcloud sdk.

ndmckinley:
@radeksimko I believe Go modules are enabled / supported now - are you still interested in merging this?  If you update it, we can take another look.

radeksimko:
@ndmckinley I just rebased & updated the PR. I suppose the meat of the PR (last commit) will need to be PR'd into https://github.com/GoogleCloudPlatform/magic-modules and the dependency PRs can stay?

ndmckinley:
Awesome, thank you!  Riley is the container_cluster person, so I'll let him review!

magic7s:
@rileykarson installing gcloud is a heavy install, and we are using this in a CI/CD process, so we want to keep it light (kubectl + kubeconfig). Also using kubeconfig to enable helm authentication and app deployments via helm.

radeksimko:
I can't remember if this PR was originally driven by anything else other than making GKE aligned with AKS. It is indeed questionable if what AKS does is a good practice which is actually worth replicating in GKE and elsewhere - I'm not K8S expert by any mean, in fact quite the opposite - I haven't touched K8S for years now, so I can't judge that objectively.

> What was the precedent for adding this field in AKS? Is it provided directly by the API, or added by Terraform?

It is provided directly by the API, as mentioned in my initial post.

I don't have strong feelings about this - I can understand arguments on both sides and I'm ok with you closing this, or asking for reimplementation as a data source.

rileykarson:
Sorry, I thought I'd closed this out before. If we add this as a feature, I'd rather we implement it as a datasource so that it's not part of the GKE resource proper. I think that OpenID authentication should also be a prerequisite, since it's GKE's preferred auth method. 

I filed https://github.com/terraform-providers/terraform-provider-google/issues/5621 as a proper followup

PR #1844, opened 2018-08-07
Fix cloud function sourceArchiveUrl incorrect interpretation.

hpidcock:
This PR fixes Cloud Function resource's sourceArchiveUrl being incorrectly parsed. Since object names can contain forward-slash, GCS url gs://bucket/a/b/c was being parsed as having an object name of "a" not "a/b/c".


ndmckinley:
Great.  Running the unit tests.

hpidcock:
Awesome, thank-you.

PR #1854, opened 2018-08-08
Link to the detailed test guide in project readme

rileykarson:
Right now, we have a detailed guide to running acceptance tests in [CONTRIBUTING.MD](https://github.com/terraform-providers/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#tests) while this file has the one-size-fits-all guideline of running `make test` and `make testacc` with no arguments. 

Instead, we can link to the detailed, provider specific guide so that we properly surface the required arguments/environment variables for `make testacc`, and don't incorrectly introduce `make test` as the primary means of testing the Google provider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,16 +62,4 @@ $ $GOPATH/bin/terraform-provider-google
 ...
 ```
 
-In order to test the provider, you can simply run `make test`.
-
-```sh
-$ make test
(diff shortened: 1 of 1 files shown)


danawillow:
Good call, thanks!

PR #1860, opened 2018-08-10
Add import support for org, folder, billing logging sinks

### LINKED ISSUE #1494: Support for importing google_logging_organization_sink and google_logging_folder_sink
Please add support for importing [google_logging_organization_sink](https://www.terraform.io/docs/providers/google/r/logging_organization_sink.html) and [google_logging_folder_sink](https://www.terraform.io/docs/providers/google/r/logging_folder_sink.html) resources.

Currently I'm getting error:
```
terraform import google_logging_organization_sink.org_logs_sink_resource_type_api resource_type-api
google_logging_organization_sink.org_logs_sink_resource_type_api: Importing from ID "resource_type-api"...

Error: google_logging_organization_sink.org_logs_sink_resource_type_api (import id: resource_type-api): import google_logging_organization_sink.org_logs_sink_resource_type_api (id: resource_type-api): resource google_logging_organization_sink doesn't support import
```

I get the similar error during importing google_logging_folder_sink.

Used versions:
```
Terraform v0.11.7
+ provider.google v1.12.0
```

rileykarson:
Fixes #1494.

* Add import support for `google_logging_organization_sink`, `google_logging_folder_sink`, `google_logging_billing_account_sink`.

Using `StateFunc` over `DiffSuppressFunc` should only affect tests; for some reason `TestAccLoggingFolderSink_folderAcceptsFullFolderPath` expected a `folder` value of `folders/{{id}}` vs expecting `{{id}}` when only `DiffSuppressFunc` was used, when in real use `DiffSuppressFunc` should be sufficient.

### DIFF
--- a/google/resource_logging_folder_sink.go
+++ b/google/resource_logging_folder_sink.go
@@ -2,6 +2,7 @@ package google
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
 )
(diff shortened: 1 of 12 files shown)


google/resource_logging_folder_sink.go
```diff
@@ -13,12 +14,17 @@ func resourceLoggingFolderSink() *schema.Resource {
...
 	}
 	schm.Schema["folder"] = &schema.Schema{
-		Type:             schema.TypeString,
-		Required:         true,
-		ForceNew:         true,
-		DiffSuppressFunc: optionalPrefixSuppress("folders/"),
+		Type:     schema.TypeString,
+		Required: true,
+		ForceNew: true,
+		StateFunc: func(v interface{}) string {
```
ndmckinley: Is a statefunc better than a diff suppress or flattener for this?

rileykarson: Oops! Just saw this comment now. Like I said in the PR description, this is just some weirdness with   one of the tests - it wasn't following the diff suppress and wrongly errored out comparing `folder/1234` to `1234`, but it played nicely with the StateFunc so I used that instead.

paddycarver: (I think this is a function of the importstateverify tests doing a direct comparison of what's in state, meaning diff suppress functions aren't used, but I could be wrong.)

PR #1863, opened 2018-08-11
Add supported locations of cloudfunction

tomykaira:
Cloudfunction facility is now supported in 4 locations.
This pull-request updates the list to catch up with it.


ndmckinley:
Thanks for your contribution!

PR #1876, opened 2018-08-14
Add support for bucket encryption

### LINKED ISSUE #1828: Add support for GCS bucket encryption parameter
When creating or patching GCS bucket, `encryption` configuration block with `defaultKmsKeyName` parameter is now supported by GCS. This allows for configuration of custom KMS key to encrypt objects inserted into this bucket (Google calls this "Customer-Managed Encryption Keys"). See https://cloud.google.com/storage/docs/json_api/v1/buckets/insert and https://cloud.google.com/storage/docs/encryption/using-customer-managed-keys for reference

### Terraform Version
```
Terraform v0.11.7
+ provider.google v1.16.2
```

### Affected Resource(s)
- google_storage_bucket

### Terraform code
Add support for encryption block, which would look like:
```
resource "google_storage_bucket" "my-bucket" {
  name     = "my-bucket"

  encryption {
    defaultKmsKeyName = "my-key"
  }
}
```

### References
- GoogleCloudPlatform/magic-modules/issues/376  - I've create issue with Magci Modules, although I believe this resource does not use generated code.

ndmckinley:
Fixes #1828.

### DIFF
--- a/google/resource_kms_crypto_key.go
+++ b/google/resource_kms_crypto_key.go
@@ -39,6 +39,10 @@ func resourceKmsCryptoKey() *schema.Resource {
 				Optional:     true,
 				ValidateFunc: validateKmsCryptoKeyRotationPeriod,
 			},
+			"self_link": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -163,6 +167,7 @@ func resourceKmsCryptoKeyRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("key_ring", cryptoKeyId.KeyRingId.terraformId())
 	d.Set("name", cryptoKeyId.Name)
 	d.Set("rotation_period", cryptoKey.RotationPeriod)
+	d.Set("self_link", cryptoKeyId.cryptoKeyId())
 
 	d.SetId(cryptoKeyId.cryptoKeyId())
 
--- a/google/resource_kms_crypto_key.go
+++ b/google/resource_kms_crypto_key.go
@@ -167,7 +167,7 @@ func resourceKmsCryptoKeyRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("key_ring", cryptoKeyId.KeyRingId.terraformId())
 	d.Set("name", cryptoKeyId.Name)
 	d.Set("rotation_period", cryptoKey.RotationPeriod)
-	d.Set("self_link", cryptoKeyId.cryptoKeyId())
+	d.Set("self_link", cryptoKey.Name)
 
 	d.SetId(cryptoKeyId.cryptoKeyId())
 
--- a/google/resource_storage_bucket_test.go
+++ b/google/resource_storage_bucket_test.go
@@ -547,6 +547,33 @@ func TestAccStorageBucket_cors(t *testing.T) {
 	}
 }
 
+func TestAccStorageBucket_encryption(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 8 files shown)


google/resource_storage_bucket_test.go
```diff
@@ -902,6 +929,44 @@ resource "google_storage_bucket" "bucket" {
...
+resource "google_kms_crypto_key" "crypto_key" {
+	name            = "%s"
+	key_ring        = "${google_kms_key_ring.key_ring.id}"
+	rotation_period = "1000000s"
+}
+
+resource "google_storage_bucket" "bucket" {
+	name = "%s"
+	encryption {
+		default_kms_key_name = "projects/${google_project.acceptance.name}/locations/us/keyRings/${google_kms_key_ring.key_ring.name}/cryptoKeys/${google_kms_crypto_key.crypto_key.name}"
```
danawillow: Oh this is sad, but I guess it'll do for now. Maybe in the future we should add a `self_link` or `full_name` or something like that to the kms resources.

ndmckinley: I thought about doing that, but it doesn't have one in the API... do you think that's still a good idea?  I'm happy to do it - I hate this solution a little bit.

danawillow: Yeah, probably. If we don't, people might realize that the `id` field has the value in the format we want, and I don't want people depending on that. It _is_ in the API under `name`, we've just co-opted the name to mean the name instead of the relative path.

sethvargo: FWIW, the KMS team uses these terms:

- `ThingID` - the full resource ID including the projects/...
- `ThingName` - just the things name

e.g.

- `CryptoKeyID` - projects/my-project/locations/global/keyRings/my-keyring/cryptoKeys/my-crypto-key
- `CryptoKeyName` - my-crypto-key

danawillow: Not in the API: https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings
https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys

ndmckinley: Fixed!

google/resource_kms_crypto_key.go
```diff
@@ -163,6 +167,7 @@ func resourceKmsCryptoKeyRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("key_ring", cryptoKeyId.KeyRingId.terraformId())
 	d.Set("name", cryptoKeyId.Name)
 	d.Set("rotation_period", cryptoKey.RotationPeriod)
+	d.Set("self_link", cryptoKeyId.cryptoKeyId())
```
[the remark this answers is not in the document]
ndmckinley: Hm, I would consider that to no longer be the `cryptoKeyId.cryptoKeyId()` if we ever changed it again.  But sure, always good to be safe.  I'm on it - checking if the tests still pass.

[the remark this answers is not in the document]
ndmckinley: Yep, it works!  Done.

ndmckinley:
Yep, good idea, done.

ndmckinley:
Sure.  Also changed it in the tests - running them now.

ndmckinley:
... One of the tests is flaking ... I'll go ahead and try to fix that, too.

ndmckinley:
Okay, that should do it, Once and For All on those flaky tests.  Submitting once Travis is ready.

PR #1878, opened 2018-08-15
Remove duplicate ProjectIDRegex variable

michaelharo:


danawillow:
Looks like unit tests didn't pass for this, can you take a look?

michaelharo:
I don't have the time or background to work on this.

danawillow:
@ndmckinley, assigning you as a reviewer since I made some changes. The docs definitely disagree with this regex, but I also can't be bothered to find a better one (and this comes from the autogenerated GCP client so it's at least somewhat consistent)

ndmckinley:
That works for me.  Bummer when the client and docs disagree - did you file a bug or should I?

danawillow:
I haven't looked into whether it's a docs bug or a client library bug, I'm happy to let you do it :)

I know the docs are at least somewhat wrong because they don't include the domain-scoped project ids, but I'm not sure about long ids (and this regex also accepts ids which start with a number, which feels wrong to me but might be a necessary consequence of allowing domain-scoping )

Issue #1729, opened 2018-07-03
debian-9 should be used instead of debian-8 in tests and docs


danawillow:
debian-8 has reached its [end-of-life](https://www.debian.org/security/2018/dsa-4205). It probably won't break anything if we keep things as-is, but since these are examples that our users might use, we should try to keep them more realistic.

paddycarver:
I've started tackling this, and switching from hardcoded image names to the compute_image data source in our tests, to prevent future breaks when images are deprecated. Just commenting to claim this so we don't duplicate effort.

PR #1879, opened 2018-08-15
Switch from debian-8 to debian-9

paddycarver:
Update our tests to use debian-8 instead of debian-9, mostly addressing #1729. I took the opportunity to sub in the data source instead of hardcoding names, too.

PR to fix documentation and examples incoming, as well.

### DIFF
--- a/google/data_source_google_compute_instance_group_test.go
+++ b/google/data_source_google_compute_instance_group_test.go
@@ -194,8 +194,8 @@ func testAccCheckDataSourceGoogleComputeInstanceGroup(dataSourceName string) res
 func testAccCheckDataSourceGoogleComputeInstanceGroupConfig() string {
 	return fmt.Sprintf(`
 data "google_compute_image" "my_image" {
-       family  = "debian-9"
-       project = "debian-cloud"
+  family  = "debian-9"
+  project = "debian-cloud"
 }
 
 resource "google_compute_instance" "test" {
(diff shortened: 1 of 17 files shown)


google/data_source_google_compute_instance_group_test.go
```diff
@@ -193,14 +193,19 @@ func testAccCheckDataSourceGoogleComputeInstanceGroup(dataSourceName string) res
 
 func testAccCheckDataSourceGoogleComputeInstanceGroupConfig() string {
 	return fmt.Sprintf(`
+data "google_compute_image" "my_image" {
+       family  = "debian-9"
```
danawillow: Would you mind fixing the indentation so it's consistent with the other resources in this file?

google/data_source_google_compute_instance_group_test.go
```diff
@@ -193,14 +193,19 @@ func testAccCheckDataSourceGoogleComputeInstanceGroup(dataSourceName string) res
...
+
 resource "google_compute_instance" "test" {
   name         = "tf-test-%s"
   machine_type = "n1-standard-1"
   zone         = "us-central1-a"
 
   boot_disk {
     initialize_params {
-      image = "debian-cloud/debian-8"
+      image = "${data.google_compute_image.my_image.self_link}"
```
danawillow: I'm fine with using the data source, but I'm not totally sure why it's a better idea than just saying `image = "debian-cloud/debian-9"` (which feels more realistic with what our users are likely to be doing). Can you explain further?

paddycarver: I was trying to use the data source because then the disk image can't change without it being recorded in Terraform. For instance templates and groups especially, this seems like the intended usage. I was trying to standardise, perhaps foolishly. For instances and disks I could see switching back, and don't mind doing that, if that's what you'd prefer.

paddycarver:
Indentation should be fixed now, sorry. Tabs v spaces was invisible to me in GitHub.

PR #1884, opened 2018-08-16
Add support for binary authorization in GKE

danawillow:
This is the small binauth change, big one coming soon.

Since we already have precedent for top-leveling nested objects that have single boolean fields (legacyAbac), I opted to go the same route with this one.

### DIFF
--- a/website/docs/r/container_cluster.html.markdown
+++ b/website/docs/r/container_cluster.html.markdown
@@ -93,7 +93,7 @@ output "cluster_ca_certificate" {
 * `description` - (Optional) Description of the cluster.
 
 * `enable_binary_authorization` - (Optional) Enable Binary Authorization for this cluster.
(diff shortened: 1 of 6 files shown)


website/docs/r/container_cluster.html.markdown
```diff
@@ -92,6 +92,9 @@ output "cluster_ca_certificate" {
 
 * `description` - (Optional) Description of the cluster.
 
+* `enable_binary_authorization` - (Optional) Enable Binary Authorization for this cluster.
+    If enabled, all container images will be validated by Google Binauthz.
```
ndmckinley: `Binauthz`?  Is that a typo?

danawillow:
Nah, that's what they call it. But I expanded it out to Binary Authorization anyway because it's probably a better idea.

PR #1900, opened 2018-08-17
Handle service account key 403 on nonexistent resource requested

### LINKED ISSUE #1891: Refreshing state before destroy fails when gcp service account key not found.
### Terraform Version
v0.11.8 with google provider v1.16.2

### Affected Resource(s)
- google_service_account_key

### Expected Behavior
Since we are trying to destroy this resource, it should not fail when the resource does not exist.

The resource uses the `handleNotFoundError` function but this only does a [check for a 404](https://github.com/terraform-providers/terraform-provider-google/issues/1891). For this resource, the googleapi is actually returning a `403 Forbidden` for the missing resource. I'm not sure if there are scenarios where we want to validly return a 403 failure for a resource you obviously created earlier with terraform, or if, for this resource, we could do a check for that api error as well and continue along.

### Actual Behavior
```
* google_service_account_key.opsman_service_account_key: 1 error(s) occurred:

* google_service_account_key.opsman_service_account_key: google_service_account_key.opsman_service_account_key: Error reading Service Account Key "projects/cf-releng/serviceAccounts/leftovers@cf-releng.iam.gserviceaccount.com/keys/x": googleapi: Error 403: Permission iam.serviceAccountKeys.get is required to perform this operation on service account key projects/cf-releng/serviceAccounts/leftovers@cf-releng.iam.gserviceaccount.com/keys/x., forbidden
```

### Steps to Reproduce
1. `terraform apply`
2. The key gets deleted
3. `terraform destroy`

ndmckinley:
Fixes #1891.

### DIFF
--- a/google/resource_google_service_account_key.go
+++ b/google/resource_google_service_account_key.go
@@ -147,6 +147,8 @@ func resourceGoogleServiceAccountKeyRead(d *schema.ResourceData, meta interface{
 				log.Printf("[DEBUG] Got a 403 error trying to read service account key %s, assuming it's gone.", d.Id())
 				d.SetId("")
 				return nil
+			} else {
+				return err
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


danawillow:
Hi @tsadoklevi, no, I don't think so.

PR #1904, opened 2018-08-19
Fix panic on empty list for authorized masters' `cidr_blocks`

vidbina:
Upon providing

```HCL
master_authorized_networks_config {
  cidr_blocks = []
}
```

the logic asserts that `[nil]` conforms to `[]interface{}` which will cause a run-time panic.

Initially, I modified the test cases to walk through a case in which `cidr_blocks` is set to an empty list. For this, I had to modify the test to accept the argument `emptyValue` which contains the string value which is interpolated into the HCL snippet under test only when the `cidrs` array is empty. Since `emptyValue` is set to an empty string for the original test cases, they exhibit the exact same behavior as they did prior to the changes I'm introducing here i.e.: when `cidrs` is empty they interpolate an empty string into the part of the HCL snippet where the `cidr_blocks` element would otherwise be presented, subsequently producing

```HCL
master_authorized_networks_config {
}
```

which is why the empty list scenario wasn't originally tested.

In a Terraform configuration where `cidr_blocks` is populated through a variable of the list type which can't just be set to an empty string, the variable `gke_authorized_cidr_blocks` declared as

```HCL
variable "gke_authorized_cidr_blocks" {
  type = "list",
  default = [{cidr_block="0.0.0.0/0", display_name = "Looking to get hacked"}]
}
```

and referenced as

```HCL
master_authorized_networks_config {
  cidr_blocks = "${vars.gke_authorized_cidr_blocks}"
}
```

would not be settable to `""` without violating the validations on the HCL list type, thus one may easily find oneself in a scenario where the panic is bound to happen.

Furthermore, I've modified `flattenMasterAuthorizedNetworksConfig` to be able to deal with the idea that CidrBlocks may now be an empty array. Instead of returning `nil` in that case, the logic simply allows the function to continue and return a map with `cidr_blocks` set to an empty array.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -1543,9 +1543,6 @@ func flattenMasterAuthorizedNetworksConfig(c *containerBeta.MasterAuthorizedNetw
 	if c == nil {
 		return nil
 	}
-	if len(c.CidrBlocks) == 0 {
-		return nil
-	}
 	result := make(map[string]interface{})
 	if c.Enabled {
 		cidrBlocks := make([]interface{}, 0, len(c.CidrBlocks))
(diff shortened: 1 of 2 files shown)


emilymye:
Thanks for submitting!

PR #1906, opened 2018-08-20
Add a datasource for google_compute_instance

rileykarson:
Add a datasource for GCE VM Instances (google_compute_instance).

---

As many deprecated/useless fields are removed as possible; a few are remaining with the "Removed" tag and are undocumented; they required editing the `google_compute_instance` resource (unless we wanted to copy a *lot* of almost identical read code) & would have bloated the PR which is already fairly large. If we want to remove them for, it should probably be done as a separate refactor.

The test covers a slight modified `_basic` config, and has attribute tests to smoke test a few nested fields as well as coverage of all the top-level ones. It's not much, but most of the code is shared with `google_compute_instance` anyways & no other data sources had that many tests. I can gladly add more, just let me know what you'd like more test coverage on!


rileykarson:
Travis will fail because of `make fmt`, ignore that failure. Sending a PR to fix right after this.

Updated google_compute_instance datasource to use the resources' schema. I kept the custom read and corresponding docs - we can't read in the state so a lot of deprecated fields just plain won't work. Some fields will "exist" to Terraform but are undocumented and have no value as a result.

google/data_source_google_compute_instance.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if instance.LabelFingerprint != "" {
+		d.Set("label_fingerprint", instance.LabelFingerprint)
+	}
+
+	attachedDisks := []map[string]interface{}{}
+	scratchDisks := []map[string]interface{}{}
+	for _, disk := range instance.Disks {
+		if disk.Boot {
+			d.Set("boot_disk", flattenBootDisk(d, disk, config))
```
paddycarver: Can we check the error here, just because it's a complex type?

google/data_source_google_compute_instance.go
```diff
@@ -0,0 +1,140 @@
...
+	// Remove nils from map in case there were disks in the config that were not present on read;
+	// i.e. a disk was detached out of band
+	ads := []map[string]interface{}{}
+	for _, d := range attachedDisks {
+		if d != nil {
+			ads = append(ads, d)
+		}
+	}
+
+	d.Set("service_account", flattenServiceAccounts(instance.ServiceAccounts))
```
paddycarver: Same with all these, can we check errors on the complex types being set?

website/docs/d/datasource_compute_instance.html.markdown
```diff
@@ -0,0 +1,187 @@
+---
+layout: "google"
+page_title: "Google: google_compute_instance"
+sidebar_current: "docs-google-datasource-compute-instance-x"
+description: |-
+  Get a VM instance within GCE.
```
paddycarver: Could we expand this a little? I think it's a bit more descriptive to say something like "information about" or something, just to be a little more explicit.

website/docs/d/datasource_compute_instance.html.markdown
```diff
@@ -0,0 +1,187 @@
...
+	name = "primary-application-server"
+	zone = "us-central1-a"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the instance.
```
paddycarver: Could we allow retrieval by self_link, as well? I think that'll be a highly requested use case.

google/data_source_google_compute_instance.go
rileykarson: Done

google/data_source_google_compute_instance.go
rileykarson: & done

website/docs/d/datasource_compute_instance.html.markdown
rileykarson: Done

website/docs/d/datasource_compute_instance.html.markdown
rileykarson: Sure!

To do this I took the logic from google_compute_region_instance_group's data source and parameterised it for consistency between the two. The test for that data source will fail but it was failing before and fails in the same way. Sorry for the scope of changes as a result of touching both resources here in this already large PR!

PR #1911, opened 2018-08-21
Naive move from debian 8 to debian 9.

ndmckinley:
I upgraded us to debian 9.  I also took the opportunity to homogenize all our `debian-8-jessie-v201X-XX-XX` into today's up-to-date `debian-9` image, not for any particular reason, just cause it'll be easier next time if it's all homogenous.

This fixes more than half of our broken tests - 🤦‍♂️.

All the image tests pass, and the Instance tests (all ~15 of them) are passing right now.

### DIFF
--- a/google/data_source_google_compute_instance_group_test.go
+++ b/google/data_source_google_compute_instance_group_test.go
@@ -200,7 +200,7 @@ resource "google_compute_instance" "test" {
 
   boot_disk {
     initialize_params {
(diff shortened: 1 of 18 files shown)


danawillow:
fyi #1879

ndmckinley:
Ah, damn, I should have searched first.  Does this have the same indentation issues?  Let me check on that.

PR #1919, opened 2018-08-22
Add docs about image families and instance templates.

paddycarver:
Recommend using our data source instead of hardcoding the family into a
template.

### DIFF
--- a/website/docs/r/compute_instance_template.html.markdown
+++ b/website/docs/r/compute_instance_template.html.markdown
@@ -108,6 +108,61 @@ With this setup Terraform generates a unique name for your Instance
 Template and can then update the Instance Group manager without conflict before
 destroying the previous Instance Template.
 
+## Deploying the Latest Image
+
+A common way to use instance templates and managed instance groups is to deploy the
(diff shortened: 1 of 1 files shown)


paddycarver:
What it looks like on the site:

screen shot 2018-08-22 at 1 48 28 am

website/docs/r/compute_instance_template.html.markdown
```diff
@@ -108,6 +108,61 @@ With this setup Terraform generates a unique name for your Instance
...
+```tf
+resource "google_compute_instance_template" "instance_template" {
+  name_prefix  = "instance-template-"
+  machine_type = "n1-standard-1"
+  region       = "us-central1"
+
+  // boot disk
+  disk {
+    initialize_params {
+      image = "debian-cloud/debian-9"
```
morgante: @paddycarver @danawillow As of #1916, I believe this particular config would create a permadiff. See my config here: https://github.com/terraform-providers/terraform-provider-google/pull/1916#issuecomment-418558489

Issue #1597, opened 2018-06-05
Configuring google_container_cluster Alias IPs without explicit secondary ranges?


reidab:
Looking at [Google's docs for GKE Alias IPs](https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips#secondary_ranges) it seems there are various levels of control possible for secondary IP ranges:

1. [Creating a cluster with Alias IPs](https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips#create): Enable IP aliases and let GKE manage both the cluster IPs and the secondary IP blocks automatically
2. [Creating a cluster with specific IP ranges](https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips#example_specific_ranges): Set the cluster IP ranges manually and let GKE manage the secondary IP blocks automatically
3. [Creating a cluster with existing secondary ranges](https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips#example_existing_ranges): Explicitly set the named secondary ranges

I'd prefer to let GKE handle all the IP range allocation, but it seems that the Terraform provider currently requires named secondary ranges. Is this necessary?

Specifying an empty block for `ip_allocation_policy` passes the plan phase but fails to apply with:

```
* google_container_cluster.primary: clusters using IP aliases must specify secondary ranges
```

### Terraform Version

Terraform v0.11.7
+ provider.google v1.13.0

### Affected Resource(s)

- google_container_cluster

shinzui:
Is this something that's going to be fixed soon? I need Alias IPs to use memorystore, but I would like to avoid managing a network and subnetwork for my kubernetes pods and services.

geekflyer:
@danawillow @rosbo  @paddycarver @ndmckinley 

Could you please prioritize supporting this feature soon? Alias IP enabled clusters is GA and apparently the recommended way to create GKE clusters these days. It's unfortunate that the terraform provider only supports it halfway. Currently we have to always create clusters via gcloud first and then import them into terraform because of this issue...

danawillow:
Hey all, can I confirm for you that what you want from an API perspective would be to be able to set all the non-deprecated fields listed in https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.IPAllocationPolicy? If we continue to not allow setting `useIpAliases` explicitly, would that be all right or do you need the ability to set that to false at any point?

danawillow:
[...]

PR #1921, opened 2018-08-22
Addition of create_subnetwork and other fields relevant for Alias IPs

ndmckinley:
This fixes https://github.com/terraform-providers/terraform-provider-google/issues/1597.  I'm still running the unit tests, but since I'm technically out on vacation and my tethering could stop working any minute, I'm going ahead and including Dana on it.  :)  Dana: review it only if the tests on ci pass!  Otherwise it can wait.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -42,6 +42,10 @@ var (
 			},
 		},
 	}
+
+	ipAllocationSubnetFields    = []string{"ip_allocation_policy.0.create_subnetwork", "ip_allocation_policy.0.subnetwork_name"}
+	ipAllocationCidrBlockFields = []string{"ip_allocation_policy.0.cluster_ipv4_cidr_block", "ip_allocation_policy.0.services_ipv4_cidr_block"}
+	ipAllocationRangeFields     = []string{"ip_allocation_policy.0.cluster_secondary_range_name", "ip_allocation_policy.0.services_secondary_range_name"}
 )
 
 func resourceContainerCluster() *schema.Resource {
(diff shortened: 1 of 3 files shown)


danawillow:
Three tests failed, two of which had been failing before and one which is definitely related:
```
------- Stdout: -------
=== RUN   TestAccContainerCluster_sharedVpc
--- FAIL: TestAccContainerCluster_sharedVpc (274.04s)
		
		* google_container_cluster.shared_vpc_cluster: 1 error(s) occurred:
		
		* google_container_cluster.shared_vpc_cluster: googleapi: Error 400: Using a shared network is only supported with IP aliases., badRequest
FAIL
```

Investigating a bit because I'd like this change to go in fairly soon. So here's the request that's getting sent:
```
{"cluster":{"binaryAuthorization":{"enabled":false},"initialNodeCount":1,"ipAllocationPolicy":{"clusterSecondaryRangeName":"pods","servicesSecondaryRangeName":"services"},"legacyAbac":{"enabled":false},"name":"cluster-test-vi2jya6lx9","network":"projects/tf-xpntest-9x7wdsi3uc-host/global/networks/test-eggvh0tbko","nodeConfig":{"oauthScopes":["https://www.googleapis.com/auth/devstorage.read_only","https://www.googleapis.com/auth/logging.write","https://www.googleapis.com/auth/monitoring","https://www.googleapis.com/auth/service.management.readonly","https://www.googleapis.com/auth/servicecontrol","https://www.googleapis.com/auth/trace.append"]},"subnetwork":"projects/tf-xpntest-9x7wdsi3uc-host/regions/us-central1/subnetworks/test-7ltede4c6s"}}
```
It definitely isn't setting ipAliases to true, which would make sense because it's not in the config. However, this PR didn't actually remove the line where we always set it to true: https://github.com/terraform-providers/terraform-provider-google/blob/6bb3fdedf08762e9bb3f37489debfa0ac991de9e/google/resource_container_cluster.go#L1428 so I have no idea why it isn't being set to true. Either way, this is going to need a little extra work because we were previously always setting it as true- we might want to make the default true, or something else so we don't accidentally break people.

danawillow:
Ok cool, so the reason it didn't set useIpAliases to true is just because we set it to false later. Cool.

What I think we actually want to do here is leave useIpAliases alone, but add the other fields in `ipAllocationPolicy`.

danawillow:
Paddy, requesting your review since I made a bunch of changes since Nathan's original PR.

One of the things I did was changing our existing IP Aliases test to be a bit simpler- we no longer check the failure cases. One of them I removed because we removed the error, and the other I removed because it's testing that the GKE API returns a specific error when you give it bad input, which I don't think is what we really should be aiming to test in Terraform (I want to test _terraform_'s behavior, not GKE's). Let me know if that sounds right to you too!

ndmckinley:
@danawillow FWIW, your changes LGTM, so if the final PR LGTYou, I think we can submit it.

PR #1931, opened 2018-08-22
#644 Add Support For BigQuery Access Control

### LINKED ISSUE #644: Support BigQuery access control
Currently, there are no support of dataset's `access` field.
https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#access

Authorized views can also be defined using `access[].view`

chrismorgan-hb:
I went ahead and added separate tests, leaving the "basic" tests to prove that the existing way (no explicit access block) continues to work.

Fixes #644

This is my first non-toy Go code and my first contribution to this project, so feedback is most welcome.

### DIFF
--- a/google/resource_bigquery_dataset.go
+++ b/google/resource_bigquery_dataset.go
@@ -122,33 +122,28 @@ func resourceBigQueryDataset() *schema.Resource {
 						"role": &schema.Schema{
 							Type:         schema.TypeString,
 							Optional:     true,
-							ValidateFunc: validateRole,
+							ValidateFunc: validation.StringInSlice([]string{"OWNER", "WRITER", "READER"}, false),
 						},
 						"domain": &schema.Schema{
-							Type:          schema.TypeString,
-							Optional:      true,
-							ConflictsWith: []string{"access.group_by_email", "access.special_group", "access.user_by_email", "access.view"},
+							Type:     schema.TypeString,
+							Optional: true,
 						},
 						"group_by_email": &schema.Schema{
-							Type:          schema.TypeString,
-							Optional:      true,
-							ConflictsWith: []string{"access.domain", "access.special_group", "access.user_by_email", "access.view"},
+							Type:     schema.TypeString,
+							Optional: true,
 						},
 						"special_group": &schema.Schema{
-							Type:          schema.TypeString,
-							Optional:      true,
-							ConflictsWith: []string{"access.domain", "access.group_by_email", "access.user_by_email", "access.view"},
+							Type:     schema.TypeString,
+							Optional: true,
 						},
 						"user_by_email": &schema.Schema{
-							Type:          schema.TypeString,
-							Optional:      true,
-							ConflictsWith: []string{"access.domain", "access.group_by_email", "access.special_group", "access.view"},
+							Type:     schema.TypeString,
+							Optional: true,
 						},
 						"view": &schema.Schema{
-							Type:          schema.TypeMap,
-							Optional:      true,
-							ConflictsWith: []string{"access.domain", "access.group_by_email", "access.special_group", "access.user_by_email"},
-							Elem:          schema.TypeString,
+							Type:     schema.TypeMap,
+							Optional: true,
+							Elem:     schema.TypeString,
 						},
 					},
 				},
@@ -230,21 +225,21 @@ func resourceDataset(d *schema.ResourceData, meta interface{}) (*bigquery.Datase
 		access := []*bigquery.DatasetAccess{}
 		for _, m := range v.([]interface{}) {
 			da := bigquery.DatasetAccess{}
-			msi := m.(map[string]interface{})
-			da.Role = msi["role"].(string)
-			if val, ok := msi["domain"]; ok {
+			accessMap := m.(map[string]interface{})
+			da.Role = accessMap["role"].(string)
+			if val, ok := accessMap["domain"]; ok {
 				da.Domain = val.(string)
 			}
-			if val, ok := msi["group_by_email"]; ok {
+			if val, ok := accessMap["group_by_email"]; ok {
 				da.GroupByEmail = val.(string)
 			}
-			if val, ok := msi["special_group"]; ok {
+			if val, ok := accessMap["special_group"]; ok {
 				da.SpecialGroup = val.(string)
 			}
-			if val, ok := msi["user_by_email"]; ok {
+			if val, ok := accessMap["user_by_email"]; ok {
 				da.UserByEmail = val.(string)
 			}
-			if val, ok := msi["view"]; ok {
+			if val, ok := accessMap["view"]; ok {
 				vm := val.(map[string]interface{})
 				if len(vm) > 0 {
 					view := bigquery.TableReference{}
@@ -391,41 +386,21 @@ func parseBigQueryDatasetId(id string) (*bigQueryDatasetId, error) {
 	return nil, fmt.Errorf("Invalid BigQuery dataset specifier. Expecting {project}:{dataset-id}, got %s", id)
 }
 
-func validateRole(value interface{}, _ string) (ws []string, errors []error) {
-	role := value.(string)
-	switch role {
-	case "READER", "WRITER", "OWNER":
-	default:
-		errors = append(errors, fmt.Errorf("Invalid role: %s", role))
-	}
-	return
-}
-
 func flattenAccess(a []*bigquery.DatasetAccess) []map[string]interface{} {
 	access := make([]map[string]interface{}, 0, len(a))
 	for _, da := range a {
-		var ai map[string]interface{}
-		ai = make(map[string]interface{})
-		if da.Role != "" {
-			ai["role"] = da.Role
-		}
-		if da.Domain != "" {
-			ai["domain"] = da.Domain
-		}
-		if da.GroupByEmail != "" {
-			ai["group_by_email"] = da.GroupByEmail
-		}
-		if da.SpecialGroup != "" {
-			ai["special_group"] = da.SpecialGroup
-		}
-		if da.UserByEmail != "" {
-			ai["user_by_email"] = da.UserByEmail
-		}
+		ai := make(map[string]interface{})
+		ai["role"] = da.Role
+		ai["domain"] = da.Domain
+		ai["group_by_email"] = da.GroupByEmail
+		ai["special_group"] = da.SpecialGroup
+		ai["user_by_email"] = da.UserByEmail
 		if da.View != nil {
-			view := make(map[string]string)
-			view["project_id"] = da.View.ProjectId
-			view["dataset_id"] = da.View.DatasetId
-			view["table_id"] = da.View.TableId
+			view := map[string]string{
+				"project_id": da.View.ProjectId,
+				"dataset_id": da.View.DatasetId,
+				"table_id":   da.View.TableId,
+			}
 			ai["view"] = view
 		}
 		access = append(access, ai)
--- a/google/resource_bigquery_dataset.go
+++ b/google/resource_bigquery_dataset.go
@@ -141,9 +141,25 @@ func resourceBigQueryDataset() *schema.Resource {
 							Optional: true,
 						},
 						"view": &schema.Schema{
-							Type:     schema.TypeMap,
+							Type:     schema.TypeList,
 							Optional: true,
-							Elem:     schema.TypeString,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"project_id": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+									"dataset_id": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+									"table_id": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+								},
+							},
 						},
 					},
 				},
@@ -240,19 +256,22 @@ func resourceDataset(d *schema.ResourceData, meta interface{}) (*bigquery.Datase
 				da.UserByEmail = val.(string)
 			}
 			if val, ok := accessMap["view"]; ok {
-				vm := val.(map[string]interface{})
-				if len(vm) > 0 {
-					view := bigquery.TableReference{}
-					if dsId, ok := vm["dataset_id"]; ok {
-						view.DatasetId = dsId.(string)
+				views := val.([]interface{})
+				if len(views) > 0 {
+					vm := views[0].(map[string]interface{})
+					if len(vm) > 0 {
+						view := bigquery.TableReference{}
+						if dsId, ok := vm["dataset_id"]; ok {
+							view.DatasetId = dsId.(string)
+						}
+						if pId, ok := vm["project_id"]; ok {
+							view.ProjectId = pId.(string)
+						}
+						if tId, ok := vm["table_id"]; ok {
+							view.TableId = tId.(string)
+						}
+						da.View = &view
 					}
-					if pId, ok := vm["project_id"]; ok {
-						view.ProjectId = pId.(string)
-					}
-					if tId, ok := vm["table_id"]; ok {
-						view.TableId = tId.(string)
-					}
-					da.View = &view
 				}
 			}
 			access = append(access, &da)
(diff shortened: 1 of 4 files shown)


google/resource_bigquery_dataset.go
```diff
@@ -108,6 +108,52 @@ func resourceBigQueryDataset() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"role": &schema.Schema{
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validateRole,
+						},
+						"domain": &schema.Schema{
+							Type:          schema.TypeString,
+							Optional:      true,
+							ConflictsWith: []string{"access.group_by_email", "access.special_group", "access.user_by_email", "access.view"},
```
danawillow: Sadly this won't work. `ConflictsWith` needs to point to an exact address of a field, and since this is a list, there's no way to wildcard all indexes. If you want a check like this, you'll just have to do it at apply-time (inside the create fn)

google/resource_bigquery_dataset.go
```diff
@@ -180,6 +226,45 @@ func resourceDataset(d *schema.ResourceData, meta interface{}) (*bigquery.Datase
 		dataset.Labels = labels
 	}
 
+	if v, ok := d.GetOk("access"); ok {
+		access := []*bigquery.DatasetAccess{}
+		for _, m := range v.([]interface{}) {
+			da := bigquery.DatasetAccess{}
+			msi := m.(map[string]interface{})
+			da.Role = msi["role"].(string)
+			if val, ok := msi["domain"]; ok {
```
danawillow: I would need to confirm 100% since I don't remember which cases this applies to and which it doesn't, but I _think_ the map will actually have all values in it, even if they weren't set, which would mean you could do:
```
da := bigquery.DatasetAccess{
  Role: msi["role"].(string),
  Domain: msi["domain}.(string),
  [etc]
}
```

google/resource_bigquery_dataset.go
```diff
@@ -180,6 +226,45 @@ func resourceDataset(d *schema.ResourceData, meta interface{}) (*bigquery.Datase
 		dataset.Labels = labels
 	}
 
+	if v, ok := d.GetOk("access"); ok {
+		access := []*bigquery.DatasetAccess{}
+		for _, m := range v.([]interface{}) {
+			da := bigquery.DatasetAccess{}
+			msi := m.(map[string]interface{})
```
danawillow: does msi just stand for map[string]interface? if so maybe a sliiiightly more descriptive var name could be used

google/resource_bigquery_dataset.go
```diff
@@ -108,6 +108,52 @@ func resourceBigQueryDataset() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				// Computed because if unset, BQ adds 4 entries automatically
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"role": &schema.Schema{
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validateRole,
```
danawillow: We already have a validateFn that you can use for enums: `validation.StringInSlice([]string{[values here]}, false),`

google/resource_bigquery_dataset.go
```diff
@@ -304,3 +390,45 @@ func parseBigQueryDatasetId(id string) (*bigQueryDatasetId, error) {
...
+		errors = append(errors, fmt.Errorf("Invalid role: %s", role))
+	}
+	return
+}
+
+func flattenAccess(a []*bigquery.DatasetAccess) []map[string]interface{} {
+	access := make([]map[string]interface{}, 0, len(a))
+	for _, da := range a {
+		var ai map[string]interface{}
+		ai = make(map[string]interface{})
```
danawillow: you shouldn't need to do this in two lines, you can just do
```
ai := make(map[string]interface{})
```

google/resource_bigquery_dataset.go
```diff
@@ -304,3 +390,45 @@ func parseBigQueryDatasetId(id string) (*bigQueryDatasetId, error) {
...
+		if da.Role != "" {
```
danawillow: unset and empty are equivalent, so you can get rid of these if statements

google/resource_bigquery_dataset.go
```diff
@@ -304,3 +390,45 @@ func parseBigQueryDatasetId(id string) (*bigQueryDatasetId, error) {
...
+		}
+		if da.SpecialGroup != "" {
+			ai["special_group"] = da.SpecialGroup
+		}
+		if da.UserByEmail != "" {
+			ai["user_by_email"] = da.UserByEmail
+		}
+		if da.View != nil {
+			view := make(map[string]string)
+			view["project_id"] = da.View.ProjectId
```
danawillow: You can also do this a bit more succinctly:
```
view := map[string]string{
  "project_id": da.View.ProjectId,
  [etc]
}
```

google/resource_bigquery_dataset_test.go
```diff
@@ -112,3 +159,110 @@ resource "google_bigquery_dataset" "test" {
...
+func testAccBigQueryDatasetWithOneAccess(datasetID string) string {
+	return fmt.Sprintf(`
+resource "google_bigquery_dataset" "access_test" {
+  dataset_id                  = "%s"
+  friendly_name               = "foo"
+  description                 = "This is a foo description"
+  location                    = "EU"
+  default_table_expiration_ms = 3600000
+
+  access = [
```
danawillow: this can just be
```
access {
  role = [...]
  user_by_email = [...]
}
```

google/resource_bigquery_dataset_test.go
```diff
@@ -112,3 +159,110 @@ resource "google_bigquery_dataset" "test" {
   }
 }`, datasetID)
 }
+
+func testAccBigQueryDatasetWithOneAccess(datasetID string) string {
+	return fmt.Sprintf(`
+resource "google_bigquery_dataset" "access_test" {
+  dataset_id                  = "%s"
+  friendly_name               = "foo"
+  description                 = "This is a foo description"
```
danawillow: I'm fine with keeping these here, but I would also be fine with the test only setting required fields + access-related stuff

google/resource_bigquery_dataset_test.go
```diff
@@ -112,3 +159,110 @@ resource "google_bigquery_dataset" "test" {
...
+func testAccBigQueryDatasetWithTwoAccess(datasetID string) string {
+	return fmt.Sprintf(`
+resource "google_bigquery_dataset" "access_test" {
+  dataset_id                  = "%s"
+  friendly_name               = "foo"
+  description                 = "This is a foo description"
+  location                    = "EU"
+  default_table_expiration_ms = 3600000
+
+  access = [
```
danawillow: likewise, this would just be
```
access {
  ...
}
access {
  ...
}
```

google/resource_bigquery_dataset_test.go
```diff
@@ -112,3 +159,110 @@ resource "google_bigquery_dataset" "test" {
...
+    {
+	role          = "OWNER"
+	user_by_email = "Joe@example.com"
+    },
+    {
+	view = {
+		project_id = "%s"
```
danawillow: could this be `project_id = "${google_bigquery_dataset.other_dataset.project}"`? would that make sense semantically?

website/docs/r/bigquery_dataset.html.markdown
```diff
@@ -73,6 +84,9 @@ The following arguments are supported:
 
 * `labels` - (Optional) A mapping of labels to assign to the resource.
 
+* `access` - (Optional) An array of objects that define dataset access for
```
danawillow: Can you add a "structure is documented below" statement and then docs for each of the attributes? search for that string in our existing docs and you'll find a bunch of examples.

chrismorgan-hb:
Friendly ping @danawillow (now that you're back from vacation and I'm sure your inbox is swamped).

google/resource_bigquery_dataset.go
danawillow: Really? You're already inside a block that checks whether access is set once you've gotten to this point.

google/resource_bigquery_dataset.go
```diff
@@ -304,3 +385,25 @@ func parseBigQueryDatasetId(id string) (*bigQueryDatasetId, error) {
 
 	return nil, fmt.Errorf("Invalid BigQuery dataset specifier. Expecting {project}:{dataset-id}, got %s", id)
 }
+
+func flattenAccess(a []*bigquery.DatasetAccess) []map[string]interface{} {
+	access := make([]map[string]interface{}, 0, len(a))
+	for _, da := range a {
+		ai := make(map[string]interface{})
```
danawillow: nit: this section can also be done in a single block:
```
ai := map[string]interface{}{
  "role": da.Role,
  ..etc...
}
```

google/resource_bigquery_dataset.go
```diff
@@ -228,6 +308,7 @@ func resourceBigQueryDatasetRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("project", id.Project)
 	d.Set("etag", res.Etag)
 	d.Set("labels", res.Labels)
+	d.Set("access", flattenAccess(res.Access))
```
danawillow: for setting values that aren't primitives, we're trying to get in the habit of checking the error returned from d.Set (we've had some confusing bugs in the past that would have shown up much sooner if we had error-checked). something like:
```
if err := d.Set("access", flattenAccess(res.Access)); err != nil {
  return err
}
```

google/resource_bigquery_dataset_test.go
```diff
@@ -112,3 +159,97 @@ resource "google_bigquery_dataset" "test" {
   }
 }`, datasetID)
 }
+
+func testAccBigQueryDatasetWithOneAccess(datasetID string) string {
+	return fmt.Sprintf(`
+resource "google_bigquery_dataset" "access_test" {
+  dataset_id                  = "%s"
```
danawillow: nit: since this isn't trying to align with anything else, you can move the equals sign way back over to the left

google/resource_bigquery_dataset_test.go
```diff
@@ -112,3 +159,97 @@ resource "google_bigquery_dataset" "test" {
...
+
+  access {
+    role          = "OWNER"
+    user_by_email = "Joe@example.com"
+  }
+  access {
+    role	      = "READER"
+    domain	      = "example.com"
```
danawillow: nit: looks like a tab snuck in here

google/resource_bigquery_dataset_test.go
```diff
@@ -112,3 +159,97 @@ resource "google_bigquery_dataset" "test" {
...
+resource "google_bigquery_table" "table_with_view" {
+  table_id   = "%s"
+  dataset_id = "${google_bigquery_dataset.other_dataset.dataset_id}"
+
+  time_partitioning {
+    type = "DAY"
+  }
+
+  view {
+        query = "SELECT state FROM [lookerdata:cdc.project_tycho_reports]"
```
danawillow: nit: # of spaces consistency with the rest of the resource

google/resource_bigquery_dataset_test.go
```diff
@@ -112,3 +159,97 @@ resource "google_bigquery_dataset" "test" {
...
+  }
+
+  labels {
+    env                         = "foo"
+    default_table_expiration_ms = 3600000
+  }
+}`, datasetID)
+}
+
+func getBigQueryTableWithView(datasetID, tableID string) string {
```
danawillow: Is this func only used in the one place? If so, may as well just inline it

google/resource_bigquery_dataset_test.go
```diff
@@ -112,3 +159,97 @@ resource "google_bigquery_dataset" "test" {
...
+  friendly_name               = "foo"
+  description                 = "This is a foo description"
+  location                    = "EU"
+  default_table_expiration_ms = 3600000
+
+  access = [
+    {
+	role          = "OWNER"
+	user_by_email = "Joe@example.com"
+    },
```
danawillow: Can you look at this section? The formatting seems to have gotten a bit wonky.

website/docs/r/bigquery_dataset.html.markdown
```diff
@@ -26,6 +26,17 @@ resource "google_bigquery_dataset" "default" {
   labels {
     env = "default"
   }
+
+  access = [
```
danawillow: this should be:
```
access {
  role   = "READER"
  domain = "example.com"
}
access {
  role           = "WRITER"
  group_by_email = "writers@example.com"
}
```

There's nothing wrong with the other way, this is just more canonical.

website/docs/r/bigquery_dataset.html.markdown
```diff
@@ -73,6 +84,41 @@ The following arguments are supported:
...
 
+* `access` - (Optional) An array of objects that define dataset access for
+    one or more entities. Structure is documented below.
+
+The `access` block supports:
+* `role` - (Required unless `view` is set) Describes the rights granted to
+    the user specified by the other member of the access object. The following
+    string values are supported: `READER`, `WRITER`, `OWNER`.
+
+* `domain` - (Pick one) A domain to grant access to.
```
danawillow: I'm not quite sure what "Pick one" means in this context, do you mean Optional?

google/resource_bigquery_dataset.go
```diff
@@ -108,6 +108,47 @@ func resourceBigQueryDataset() *schema.Resource {
...
+						"special_group": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"user_by_email": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"view": &schema.Schema{
+							Type:     schema.TypeMap,
```
danawillow: This has specific keys that have to be in the map, correct? If so, then this should be another nested object (list with MaxSize: 1 and a defined schema) so we don't allow users to set keys in the map that aren't valid options.

danawillow:
Oh also! Fun GH trick: if you write fixes #644 in the body of your PR, it'll close the issue automatically: https://help.github.com/articles/closing-issues-using-keywords/

google/resource_bigquery_dataset_test.go
chrismorgan-hb: Done.

google/resource_bigquery_dataset_test.go
chrismorgan-hb: Done.

google/resource_bigquery_dataset_test.go
chrismorgan-hb: Done.

website/docs/r/bigquery_dataset.html.markdown
chrismorgan-hb: I was mimic-ing the [BQ API documentation](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#access), it basically means you have to pick exactly one of `domain`, `group_by_email`, `special_group`, `user_by_email`, or `view`. If there's a way to be more clear, I'm happy to change it.

website/docs/r/bigquery_dataset.html.markdown
chrismorgan-hb: Done.

website/docs/r/bigquery_dataset.html.markdown
danawillow: Ah, got it. I'd just try to match the rest of the Terraform documentation, where we just say Required or Optional there. I think that Pick One itself is pretty redundant, right? There isn't a way to specify multiple of the same field, so you don't really have a choice but to pick one (or zero, but that's covered in Required vs Optional).

chrismorgan-hb: It's a little more nuanced because these are keys in a map, not values that can be taken by a single field. I updated the docs to reflect that (while marking these optional). Let me know what you think.

danawillow:
Yup that looks good to me, thanks! I'll go ahead and merge this.

PR #1933, opened 2018-08-23
Use differentiated issue templates.

paddycarver:
I largely stole these from github.com/terraform-providers/terraform-provider-aws, but with a few changes:

* I called out how we assign users to issues, to make it more transparent to users.
* I added a "success story" issue that will be automatically closed and labeled, to keep track of those stories.
* I added a `[issue-type:TYPE]` string to each, so hashibot can detect and act on them.


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,77 @@
...
+
+<!--- Please leave this line, it helps our automation: [issue-type:bug-report] --->
+<!--- Please keep this note for the community --->
+
+### Community Note
+
+* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
+* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
+* If you are interested in working on this issue or have submitted a pull request, please leave a comment
+* If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.
```
rileykarson: What about using "If this issue"/"If an issue"? "If the issue" seems a little awkward to me especially when we refer to it as "it" later.

paddycarver:
Changes made, merging in!

PR #1944, opened 2018-08-26
Importing  google_dns_managed_zone with project

rwaldvogel:
importing Cloud DNS managed zones from a different project.

Any comments are welcome. Esp. on doing acceptance tests with a different project are very welcome.

- [x] Code
- [x] Tests
- [x] Acceptance tests
 -[x] Documentation

### DIFF
--- a/website/docs/r/dns_managed_zone.markdown
+++ b/website/docs/r/dns_managed_zone.markdown
@@ -53,3 +53,9 @@ DNS managed zones can be imported using the `name`, e.g.
 ```
 $ terraform import google_dns_managed_zone.prod prod-zone
 ```
+
+or by using `project-id` and `name` e.g.
+
+```
+$ terraform import google_dns_managed_zone.foo project/managedZones/foo-zone
+```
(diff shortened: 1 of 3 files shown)


website/docs/r/dns_managed_zone.markdown
```diff
@@ -53,3 +53,9 @@ DNS managed zones can be imported using the `name`, e.g.
 ```
 $ terraform import google_dns_managed_zone.prod prod-zone
 ```
+
+or by using `project-id` and `name` e.g.
+
+```
+$ terraform import google_dns_managed_zone.foo project/managedZones/foo-zone
```
paddycarver: nitpick: could we maybe use `project-id` instead of `project`, to make it clear that `project` isn't a hardcoded string, and to match the description?

rwaldvogel: Sorry for the delay. 
Sure. I documented all import formats like on compute disk using project-id

PR #1955, opened 2018-08-28
Add .editorconfig configured for Go files.

rileykarson:
By using an EditorConfig file we can make sure that we have consistent editor settings regardless of what editor is being used (well, as long as they support EditorConfig 🙂).

This will primarily affect Jetbrains editors since they support it automatically; Github should also change from 8 to 4 tab-width following the config. Other editors can download plugins to enforce the settings.

Let me know if you all think this is a good idea or not! Personally I have had IntelliJ fool me as to what kind of space characters it was using across languages too many times.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,15 @@
-[*.go]
+root=true
+
+[*]
 charset=utf-8
 end_of_line=lf
 insert_final_newline=false
+trim_trailing_whitespace=true
+
+[*.go]
 indent_style=tab
 tab_width=4
 
+[*.tf]
+indent_style=space
+indent_size=2
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+[*.go]
+charset=utf-8
+end_of_line=lf
+insert_final_newline=false
+indent_style=tab
+tab_width=4
+


morgante:
I would like this. 👍

paddycarver:
So it seems there's no strong precedent for this, which is both good (there's nothing saying we can't!) but also bad (there's nothing saying we should!)

My take on this is still that it's probably not worthwhile, because:

* It opens the door for other people to argue their specific config files should be added, and I'm not keen on managing people dotfiles in the repository 😛 
* It encodes personal preference (tabwidth, etc.) into the repo, which encourages bikeshedding over what the "blessed" preference should be. (Though, you could argue go fmt already opened that door.)
* I just don't see the benefit to having it managed and implicitly "blessed" by the repo as opposed to managing it outside the repo.

I'll admit though that none of these are very strong, and I wouldn't mind being outvoted on this. I'm a very weak "against", I think. I'd love to hear more about what benefits this would bring, though, as that could definitely bring me 'round to the "for" side.

randomcamel:
How would this affect a contributor using an EditorConfig-aware editor but doesn't want to use these preferences?

rileykarson:
@randomcamel: IntelliJ (and presumably other Jetbrains IDEs) allow users to ignore the EditorConfig file in their settings (so at the global or project level). The only other one that automatically follows the format is Visual Studio/Visual Studio Mac/MonoDevelop, which I have not tested it with. All other editors require a plugin to conform to EditorConfig rules & those plugins should presumably allow opt-out.

@paddycarver / @chrisst 
`go fmt` (and `terraform fmt`) absolutely bake a large part of formatting into the language! Using an EditorConfig would assist them by letting code already conform as it's written instead of then needing to run a tool - I've run `make build` and then had to immediately run `make fmt` countless times because of silly formatting issues that my editor should be able do a better job at.

In my opinion, the breadth of support differentiates EditorConfig from other dotfiles/config files; it supports every editor I have ever used through plugins (well... except nano!) and has automatic support for Jetbrains family and VS family (sadly not VSCode) editors. What's really cool about that to me is that as I contribute to different projects, even if they share a language, I don't have to change my settings if the projects differ.

Most of the values aren't what I would consider personal preference, and it looks like it's only tab width really? If tab width is contentious, we could easily remove that and keep the non-contentious rules - for example, the explicit codification of when to use spaces or tabs, whether to include a final newline, and whether to trim whitespace. I don't have strong feelings on specific tab widths so I didn't consider that they could be controversial. My apologies!

Also about bikeshedding - the way I see it is that if we have such a file, we contain bikeshedding to that file and discussions about that file. Instead of letting it leak into pull requests and code like it could otherwise - but given, so far, hasn't.

And @ your third point Paddy, I guess I have the opposite view - OSS repositories often have implicit formatting standards that are only going to come up at pull request time. Rules that aren't being managed in the repo are tribal knowledge that infrequent contributors often don't have. By codifying them in the repo, we make explicit what formatting we expect for what files/filetypes, and reduce the burden of knowledge for a new contributor (especially if they never even have to know because their IDE/editor picks it up automatically 🙂 ).

chrisst:
Tabwith:
To me this is a personal preference and display only so there isn't a need to add this to the repo. 

Non-display stuff (newlines, whitespace, etc):
As long as they adhere to what would have happened in `go fmt` or `terraform fmt` then I think it's ok to add them. The only downside I could see is people adding their personal preferences to their .config and accidentally checking it in.

Taking a step back:
If the root issue is that we are catching formatting things at PR time instead of earlier, perhaps we could take other steps such as running fmt as part of the default build. Or making Travis check formatting automatically so formatting doesn't need to be commented on by humans.

Unrelated:
I use an IDE (vscode) with plugins that run `go fmt` and `go import` every time I hit save and it's been pretty great.

tl;dr: I feel that tabwidth should be removed. I'm willing to go either way on the rest.

rileykarson:
Closing this because the consensus was generally against adding one.

Issue #1585, opened 2018-06-04
Feature Request: Attach disks to a running VM


opdude:
It would be great if it were possible to attach disks to an already running GCP instance for reference the AWS provider has this with `aws_volume_attachment`.

abdelsource:
This is a nice feature to overcome a limitation in Terraform where it's not possible to pass a map of info to resources when they are being created.

PR #1957, opened 2018-08-28
Attached disk resource

chrisst:
This is a new resource for attaching disks to instances 

Fixes https://github.com/terraform-providers/terraform-provider-google/issues/1585


google/resource_compute_attached_disk.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func resourceAttachedDiskRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+	d.Set("project", project)
```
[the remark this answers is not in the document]
chrisst: Ah yes. I was parsing the `zone` but treating it pretty different from project. I could either add a `getProjectFromSelfLink` like I did for `zone` or just allow the user to specify / override if they choose.

I'm leaning towards parsing the project link just like I did with zone because if a user specified a different project/zone than what was set on the compute instance it could cause strange behavior. If you think that's over thinking things I'm also happy to just add them both as optional.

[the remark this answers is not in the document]
rileykarson: Total drive by comment, but: why not set the zone and project from the instance we `Get` from the API? That way users don't need to repeat themselves when the values exist in the self link of the instance, and we still find the correct values.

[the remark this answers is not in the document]
chrisst: I think the thing that's making this weird is if the user is only referencing the `name` when running `Create`. In which case we have to look for the values specified more globally. Otherwise I could just grab all the data from the self link. TBH... we don't even really need to set the zone/project in state at all now that I moved to using full/relative links to the compute instance instead of just `name`. I'll throw up a quick commit and see if it's an improvement.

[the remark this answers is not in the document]
rileykarson: Oh yup - there should always be a zone and project to make the GET to at Read time set in both Create and Update!

chrisst:
@paddycarver I think I've covered all the things we talked about and also added a few more configurable properties that I found along the way.

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,218 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccAttachedDiskContainsManyDisks(instanceName, count),
+				),
+			},
+		},
+	})
+
+}
+
+func testAccAttachedDiskIsNowDetached(instanceName, diskName string) resource.TestCheckFunc {
```
danawillow: For all of the checking functions, let's call them `testCheckAttachedDisk[whatever]`

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+}
+
+func testAttachedDiskResourceAttachment() string {
+	return fmt.Sprintf(`
+resource "google_compute_attached_disk" "test" {
+	disk = "${google_compute_disk.test1.self_link}"
+	instance = "${google_compute_instance.test.self_link}"
+}
+	`)
```
danawillow: nit: I usually line the final \` up with either the beginning of the line or directly after the last `}`, though it's totally personal preference and up to you if you prefer it here.

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func testAttachedDiskResourceAttachmentDefaults() string {
```
danawillow: it looks like these are explicitly setting things to be not their defaults, right? i usually use the word "full" to describe tests where i set all the things.

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,218 @@
...
+	return fmt.Sprintf(`
+resource "google_compute_attached_disk" "test" {
+	disk = "${google_compute_disk.test1.self_link}"
```
danawillow: it's not a huge deal but it usually looks better when all the `=` are aligned within a section (that's what terraform fmt does on .tf files)

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,218 @@
...
+	mode = "READ_ONLY"
+	auto_delete = false
+	device_name = "test-device-name"
+}
+	`)
+}
+
+func testAttachedDiskResource(diskName, instanceName string) string {
+	return fmt.Sprintf(`
+resource "google_compute_disk" "test1" {
```
danawillow: I don't really care, but just letting you know that you can totally name the disk and instance resources the same thing

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,218 @@
...
+	size = 10
+}
+
+resource "google_compute_instance" "test" {
+	name         = "%s"
+	machine_type = "f1-micro"
+	zone         = "us-central1-a"
+	
+	lifecycle {
+			ignore_changes = [
```
danawillow: this looks to be indented an extra tab

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,218 @@
...
+				"attached_disk"
+			]
+		}
+	
+	boot_disk {
+		initialize_params {
+		image = "debian-cloud/debian-9"
```
danawillow: and this not enough tabs

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,218 @@
...
+	count = %d
+}
+
+resource "google_compute_instance" "test" {
+	name         = "%s"
+	machine_type = "f1-micro"
+	zone         = "us-central1-a"
+
+	lifecycle {
+			ignore_changes = [
```
danawillow: indentation is wonky around here too

google/resource_compute_attached_disk.go
```diff
@@ -0,0 +1,282 @@
...
+				ValidateFunc: validation.StringInSlice([]string{"READ_ONLY", "READ_WRITE"}, false),
+			},
+		},
+	}
+}
+
+func resourceAttachedDiskCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProjectForAttachedDisk(d, config)
```
danawillow: I feel like you can do this entire next section with existing helpers, something like:
```
fv, err := parseZonalFieldValue("instances", d.Get("instance"), "project", "zone")
if err != nil {
  return err
}

...

op, err := config.clientCompute.Instances.AttachDisk(fv.Project, fv.Zone, fv.Name, &attachedDisk).Do()
```

website/docs/r/compute_attached_disk.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+layout: "google"
+page_title: "Google: google_compute_attached_disk"
+sidebar_current: "docs-google-compute-attached-disk"
+description: |-
+  Resource that allows attaching existing persistent disks to compute instances.
+---
+
+# google\_compute\_attached\_disk
+
+Persistent disks can be attached to a compute instance using [the `attach_disk`
```
danawillow: typo: attached_disk

website/docs/r/compute_attached_disk.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+section within the compute instance configuration](https://www.terraform.io/docs/providers/google/r/compute_instance.html#attached_disk).
+However there may be situations where managing the attached disks via the compute
+instance config isn't preferable or possible. For example: attaching dynamic
```
danawillow: I'm slightly anti-example just because it becomes obsolete in tf 0.12. If you'd like it to remain though, then can you rephrase into a complete sentence? (I'd just do "preferable or possible, such as attaching")

website/docs/r/compute_attached_disk.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+
+The following arguments are supported:
+
+
+* `instance` -
+  (Required)
+  `name` or `self_link` of the compute instance that the disk will be attached to.
+  If the `self_link` is provided then `zone` and `project` are extracted from the
+  self link. If only the name is used then `zone` and `project` must be defined
+  at a global level.
```
danawillow: s/at a global level/as properties on the resource or provider/

website/docs/r/compute_attached_disk.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+* `disk` -
+  (Required)
+  `name` or `self_link` of the disk that will be attached.
+
+
+- - -
+
+* `project` -
+  (Optional)
+  The project that the referenced compute instance is a part of. If `instance` is referenced by it's
```
danawillow: s/it's/its

website/docs/r/compute_attached_disk.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+  `self_link` the project defined in the link will take precedence.
+
+* `zone` -
+  (Optional)
+  The zone that the referenced compute instance is located within. If `instance` is referenced by it's
+  `self_link` the zone defined in the link will take precedence.
+
+* `auto_delete` -
```
danawillow: Hmm, should we let people set this? If a user is configuring their attachment explicitly, then auto-delete feels strange (since the disk would be deleted automatically but this resource would remain)

website/docs/r/compute_attached_disk.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+  (Optional)
+  Specifies a unique device name of your choice that is
+	reflected into the /dev/disk/by-id/google-* tree of a Linux operating
+	system running within the instance. This name can be used to
+	reference the device for mounting, resizing, and so on, from within
+	the instance.
+
+	If not specified, the server chooses a default device name to apply
+	to this disk, in the form persistent-disks-x, where x is a number
+	assigned by Google Compute Engine. This field is only applicable for
```
danawillow: you can remove the last sentence here since we're only dealing with persistent disks

website/docs/r/compute_attached_disk.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+
+This resource provides the following
+[Timeouts](/docs/configuration/resources.html#timeouts) configuration options:
+
+- `create` - Default is 5 minutes.
+- `delete` - Default is 5 minutes.
+
+## Import
+
+Attached Disk is composed of a compute instance and disk and it's `id` is the
```
danawillow: honestly I'm not sure this sentence really adds anything extra, the example is what really helps. (if you want to keep it though, then s/it's/its)

google/resource_compute_attached_disk.go
```diff
@@ -0,0 +1,282 @@
...
+	if err != nil {
+		return err
+	}
+	d.Set("disk", diskPath)
+
+	return nil
+}
+
+func resourceAttachedDiskDelete(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("ZOMG destroying")
```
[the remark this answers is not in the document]
chrisst: oh dear

google/resource_compute_attached_disk.go
```diff
@@ -0,0 +1,282 @@
...
+
+	zone, err := GetPathVariableFromSelfLink(attachedInstance, "zone")
+	if err == nil {
+		return zone, nil
+	}
+
+	// If zone can't be inferred from the compute instance self link, fall back to project
+	zone, err = getZone(d, c)
+	if err != nil {
+		return "", fmt.Errorf("%s to inherit from the attached instance compute use `self_link` instead of `name`", err)
```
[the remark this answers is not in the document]
chrisst: It's layering on top of the [error thrown by the getZone method](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/regional_utils.go#L34). So it would read: "Cannot determine zone: set in this resource, or set provider-level zone. to inherit from the attached instance compute use `self_link` instead of `name`"

I dropped the punctuation as suggested in https://github.com/golang/go/wiki/CodeReviewComments#error-strings but a comma could probably go a long way. Also since we already use full sentences in our error strings I could just change it into a separate sentence. Thoughts?

google/resource_compute_attached_disk.go
```diff
@@ -0,0 +1,219 @@
...
+		log.Printf("[WARN] Referenced disk wasn't found attached to this compute instance. Removing from state.")
+		d.SetId("")
+		return nil
+	}
+
+	d.Set("device_name", ad.DeviceName)
+	d.Set("mode", ad.Mode)
+
+	// Force the referenced resources to a self-link in state because it's more specific then name.
+	instancePath, err := getRelativePath(instance.SelfLink)
```
danawillow: This is _probably_ fine. Some resources are annoying and insist that you give them a full URL instead of a relative one, so we generally use ConvertSelfLinkToV1 so we store the full URL. I doubt people are going to use this for interpolation so it doesn't really matter either way in this case, more of just a "hey letting you know that func exists too"

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,213 @@
...
+	t.Parallel()
+
+	diskName := acctest.RandomWithPrefix("tf-test-disk")
+	instanceName := acctest.RandomWithPrefix("tf-test-inst")
+	importID := fmt.Sprintf("%s/us-central1-a/%s:%s", getTestProjectFromEnv(), instanceName, diskName)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: nil,
```
danawillow: You should be able to add an actual CheckDestroy here instead of doing the second TestStep to detach it. If that's not the case, then it would be nice to have a comment below explaining why.

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+func testAttachedDiskResource(diskName, instanceName string) string {
+	return fmt.Sprintf(`
+resource "google_compute_disk" "test1" {
+	name = "%s"
+	zone = "us-central1-a"
+	size = 10
+	}
+	
+	resource "google_compute_instance" "test" {
```
danawillow: I'm seeing a bunch of indentation weirdness throughout the test, can you double check all of them?

chrisst: Ugh after the last round I copy/pasted every test data block and ran it through terraform fmt, but looks like VScode was being helpful and replacing all my spaces with tabs and reformatting :(

chrisst:
Formatting should be cleaner now.

google/resource_compute_attached_disk.go
```diff
@@ -0,0 +1,219 @@
...
+
+func resourceAttachedDiskCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	zv, err := parseZonalFieldValue("instances", d.Get("instance").(string), "project", "zone", d, config, false)
+	if err != nil {
+		return err
+	}
+
+	instanceName := GetResourceNameFromSelfLink(zv.Name)
```
danawillow: ah sorry what I meant here is that zv.Name _is_ the resource name, so you don't need to call GetResourceNameFromSelfLink at all

google/resource_compute_attached_disk.go
```diff
@@ -0,0 +1,219 @@
...
+	config := meta.(*Config)
+
+	zv, err := parseZonalFieldValue("instances", d.Get("instance").(string), "project", "zone", d, config, false)
+	if err != nil {
+		return err
+	}
+	d.Set("project", zv.Project)
+	d.Set("zone", zv.Zone)
+
+	instanceName := GetResourceNameFromSelfLink(d.Get("instance").(string))
```
danawillow: same thing from before for here

google/resource_compute_attached_disk_test.go
```diff
@@ -0,0 +1,221 @@
...
+	t.Parallel()
+
+	diskName := acctest.RandomWithPrefix("tf-test")
+	instanceName := acctest.RandomWithPrefix("tf-test")
+	importID := fmt.Sprintf("%s/us-central1-a/%s:%s", getTestProjectFromEnv(), instanceName, diskName)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		// Check destroy isn't a good test here, see comment on testCheckAttachedDiskIsNowDetached
```
danawillow: can you add this comment into the other tests too?

chrisst:
ok, formatting should actually be better this time around. No more auto-formatting from vscode.

PR #1958, opened 2018-08-28
updating examples in documentation for debian-9

chrisst:
Making sure that the examples in the docs don't fail due to old debian images that GCP doesn't support any more.


chrisst:
removed auto-gen files, thanks!

PR #1969, opened 2018-08-30
Fix Instance Group Manager rolling updates

paddycarver:
Reverts terraform-providers/terraform-provider-google#1762

Now that release happened, let's try to get this one unblocked.

Fix instance group managers to respect rolling update settings.


paddycarver:
screen shot 2018-08-31 at 9 53 13 am

danawillow:
Talking to the IGM team, they seemed to think we can use PATCH for everything. I noticed though that the docs mention [setInstanceTemplate doing some things to the updatePolicy](https://cloud.google.com/compute/docs/instance-groups/updating-managed-instance-groups#the_versions_field_and_the_setinstancetemplate_method), which was the original bug report. I wonder whether it's possible to reorder things in such a way that we can still use patch?

At some point (maybe in a different PR, definitely soon) we should revisit the list of update strategies. I'll follow up with you offline about this.

PR #1992, opened 2018-09-05
make fmt

rileykarson:
This should fail travis?


rileykarson:
For posterity:

These `make fmt` changes are because of a gofmt change between Go 1.10 and Go 1.11. If you encounter them, try downgrading your Go version to fix them!

Personally, I used these commands with Homebrew on MacOS.

```
brew install go@1.10
```
Then set these environment variables in my .bashrc:
```
export GOROOT=~/homebrew/opt/go@1.10/libexec/bin
export PATH="~/homebrew/opt/go@1.10/bin:$PATH"
```

PR #1995, opened 2018-09-06
Fix perma-diff on instance templates.

paddycarver:
When using instance templates, if you use one of our image shorthands at
the moment, you'll get a perma-diff. This is because the config gets
resolved to another format before it is set in state, and so we need to
set that other format in state.

Unfortunately, resolving images requires network access, so we have to
do this with CustomizeDiff. CustomizeDiff was having trouble (I think?
More on this below) on setting a field as not ForceNew once the field
was already set, so I moved the decision for whether a field was
ForceNew or not into CustomizeDiff. I also resolved the old and new
images, and if they were the same, cleared the diff.

Unfortunately, you can't actually clear a field on a sub-block right
now. You have to clear top-level fields only. So this will currently
throw an error. I opened hashicorp/terraform#18795 to fix that. Once
that's merged, and we vendor it here, this patch fixes the problem.

If hashicorp/terraform#18795 doesn't get merged, the next best
workaround is to keep track of _all_ the fields under `disk` with a diff
in our CustomizeDiff, check whether they've all changed or not, and if
they've all changed, clear the changes on `disk`, which I _think_ will
resolve the issue. That's just a massive pain, unfortunately.

### DIFF
--- a/google/resource_compute_instance_template.go
+++ b/google/resource_compute_instance_template.go
@@ -20,6 +20,7 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 		SchemaVersion: 1,
+		CustomizeDiff: resourceComputeInstanceTemplateCustomizeDiff,
 		MigrateState:  resourceComputeInstanceTemplateMigrateState,
 
 		// A compute instance template is more or less a subset of a compute
@@ -99,10 +100,9 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 						},
 
 						"source_image": &schema.Schema{
-							Type:             schema.TypeString,
-							Optional:         true,
-							DiffSuppressFunc: compareSelfLinkRelativePaths,
-							ForceNew:         true,
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
 						},
 
 						"interface": &schema.Schema{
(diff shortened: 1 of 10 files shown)


danawillow:
Does this fix https://github.com/terraform-providers/terraform-provider-google/issues/2067?

paddycarver:
I don't believe so. This will fix a permadiff, #2067 requires a bit of logic/parsing in the flattenDisks helper for `instance_template`. I'll update that tomorrow.

vendor/vendor.json
```diff
@@ -665,10 +665,10 @@
 			"versionExact": "v0.11.2"
 		},
 		{
-			"checksumSHA1": "JHxGzmxcIS8NyLX9pGhK5beIra4=",
+			"checksumSHA1": "OOwTGBTHcUmQTPBdyscTMkjApbI=",
 			"path": "github.com/hashicorp/terraform/helper/schema",
-			"revision": "41e50bd32a8825a84535e353c3674af8ce799161",
-			"revisionTime": "2018-04-10T16:50:42Z",
+			"revision": "35d82b055591e9d47a254e68754216d8849ba67a",
+			"revisionTime": "2018-09-26T21:21:28Z",
```
danawillow: It's probably not actually a big deal at all, but this has a versionExact and a revision time that is definitely a few versions ahead of the versionExact. Any idea how that happened?

google/resource_compute_instance_template.go
```diff
@@ -420,6 +420,48 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 	}
 }
 
+func resourceComputeInstanceTemplateCustomizeDiff(diff *schema.ResourceDiff, meta interface{}) error {
```
danawillow: nit: I'd call this `resourceComputeInstanceTemplateSourceImageCustomizeDiff`, since presumably if we added more fields to customize we would want to do a customdiff.All to group them.

(also lol, every so often i think that maybe our resources should be like, objects, so we don't have to have these ridiculously long globally-unique function names)

google/resource_compute_instance_template.go
```diff
@@ -420,6 +420,48 @@ func resourceComputeInstanceTemplate() *schema.Resource {
...
+			}
+			newResolved, err := resolveImage(config, project, new.(string))
+			if err != nil {
+				return err
+			}
+			if oldResolved != newResolved {
+				err = diff.ForceNew(key)
+				if err != nil {
+					return err
+				}
```
danawillow: shouldn't there be a continue here? otherwise the diff will get cleared on line 456

vendor/vendor.json
paddycarver: If I had to guess, it's a result of govendor updating the revision and revisionTime, but not the version or versionExact. I've gone ahead and removed the version and versionExact.

google/resource_compute_instance_template.go
paddycarver: Good catch.

google/resource_compute_instance_template.go
paddycarver: Fair! I've been doing massive monolithic CustomizeDiff functions, but I think customdiff.All--which I just learned about!--is a better approach.

And yeah, long names are amazing. -_-

paddycarver:
I mentioned this in the commit, but the missing `continue` that @danawillow pointed out was actually masking that this didn't really fix the problem. Because `resolveImage` can return multiple formats, and only returns a self_link if a self_link is passed in, it was diffing more often than not against a self_link supplied in the config. The missing `continue` just concealed this.

I added the `resolvedImageSelfLink` helper function to turn the output from `resolveImage` into a self_link, which fixes the problem. It also has the knock-on effect of fixing #2067 at the same time, which is nice. I decided to add another function instead of just modifying `resolveImage`
to always return a self_link because I don't want to deal with the backwards compatibility problems of changing how we're storing a bunch of things in state this close to 1.19.0. And honestly, we should
probably just be storing the self_link always, _anyways_.

paddycarver:
Oh, you're right. We'd still need to use resolvedImageSelfLink here: https://github.com/terraform-providers/terraform-provider-google/blob/27bd76d9d40d0149d75a1489f83eba119a3b01d5/google/resource_compute_instance_template.go#L640

But this at least sets up a small PR to fix that?

paddycarver:
> Can you file an issue about this to make sure it's something that we can make better in 2.0.0?

#2151

Going to go ahead and merge this now.

Issue #1803, opened 2018-07-23
Feature request: Support for labels in google_dns_managed_zone


toredash:
Requesting that google_dns_managed_zone gets supports for labels. Support is present in the Cloud DNS API already. Reading from the API docs, I'm not entirely sure that the `patch` call supports modifying an existing resource, but `create` supports labels.
 

### Terraform Version
(not relevant)
Terraform v0.11.7
+ provider.google v1.16.2

https://www.terraform.io/docs/providers/google/d/dns_managed_zone.html

### Affected Resource(s)
- google_dns_managed_zone

### Terraform Configuration Files
Suggestion on how it could look:
```
resource "google_dns_managed_zone" "example-com" {
  name        = "example-com"
  dns_name    = "example.com."
  description = "Production DNS zone"
  labels = {
    cost-center = "marketing",
    env = "production"
  }
}
```

### Expected Behavior
Mentioned resource should support labels: https://cloud.google.com/dns/api/v1beta2/managedZones
https://cloud.google.com/dns/api/v1beta2/managedZones/create
https://cloud.google.com/dns/api/v1beta2/managedZones/patch

### Actual Behavior
No support for labels in google_dns_managed_zone

PR #2009, opened 2018-09-07
Add labels functionality to managed DNS

chrisst:
Fixes https://github.com/terraform-providers/terraform-provider-google/issues/1803

### DIFF
--- a/google/resource_dns_managed_zone.go
+++ b/google/resource_dns_managed_zone.go
@@ -53,6 +53,12 @@ func resourceDnsManagedZone() *schema.Resource {
 				Computed: true,
 				ForceNew: true,
 			},
+
+			"labels": &schema.Schema{
+				Type:     schema.TypeMap,
+				Optional: true,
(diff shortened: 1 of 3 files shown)


google/resource_dns_managed_zone.go
```diff
@@ -72,7 +78,18 @@ func resourceDnsManagedZoneCreate(d *schema.ResourceData, meta interface{}) erro
 		Description: d.Get("description").(string),
 	}
 
+	if clabels, ok := d.GetOk("labels"); ok {
+		labels := map[string]string{}
```
danawillow: we actually have a util called `expandLabels` already, so you may as well just use that

chrisst:
docs updated, thanks!

chrisst:
I'm in the habit of amending commits partly if they are small updates but mostly because I'm in a different habit of forgetting to squash/merge :P I'll try and comment on PR's if I don't add separate commits.

PR #2018, opened 2018-09-10
Make it allowable to add 'network_interface' as workaround for element function

### LINKED ISSUE #1805: google_compute_instance_from_template does not expose network info until VM exists
### Terraform Version
```
$ terraform -v
Terraform v0.11.7
+ provider.google v1.16.0
```

### Affected Resource(s)
- google_compute_instance_from_template

```hcl
resource "google_dns_managed_zone" "sample" {
  name        = "sample"
  dns_name    = "sample.network."
  description = "dns zone for bug reproduction purposes"
}

resource "google_compute_instance_template" "gce-minimal" {
  name        = "gce-minimal"
  description = "minimal image for low-storage, low-memory needs"

  instance_description = "description assigned to basic server instances"
  machine_type         = "n1-standard-2"
  can_ip_forward       = false

  scheduling {
    automatic_restart   = true
  }

  disk {
    source_image = "centos-7"
    auto_delete  = true
    boot         = true
    disk_size_gb = "25"
    disk_type    = "pd-ssd"
  }

  network_interface {
    network = "default"
  }

  service_account {
    scopes = ["userinfo-email", "compute-ro", "storage-ro"]
  }
}

resource google_compute_instance_from_template "sampleXX" {
  name = "sample${format("%02d", count.index + 1)}"
  count = 2
  zone = "us-west1-a"

  source_instance_template = "${google_compute_instance_template.gce-minimal.self_link}"
}

resource "google_dns_record_set" "sampleXX" {
  name  = "sample${format("%02d", count.index + 1)}.${google_dns_managed_zone.sample.dns_name}"
  count = 2
  type  = "A"
  ttl   = 300

  managed_zone = "${google_dns_managed_zone.sample.name}"

  rrdatas = ["${element(google_compute_instance_from_template.sampleXX.*.network_interface.0.address, count.index)}"]
}
```

### Debug Output
https://gist.github.com/elfchief/0aaea3b489a1e50b89909ecfa2015a1b

### Expected Behavior
VMs should be created and then DNS entries should be created for them.

### Actual Behavior
```
Error: Error running plan: 1 error(s) occurred:

* google_dns_record_set.sampleXX: 2 error(s) occurred:

* google_dns_record_set.sampleXX[0]: Resource 'google_compute_instance_from_template.sampleXX' does not have attribute 'network_interface.0.address' for variable 'google_compute_instance_from_template.sampleXX.*.network_interface.0.address'
* google_dns_record_set.sampleXX[1]: Resource 'google_compute_instance_from_template.sampleXX' does not have attribute 'network_interface.0.address' for variable 'google_compute_instance_from_template.sampleXX.*.network_interface.0.address'
```
...if I comment out the google_dns_record_set stanza, run terraform to create the VMs, and then add the google_dns_record_set stanza back, things work as expected.

### Steps to Reproduce
1. set up a 'google' provider with appropriate credentials
2. create config file as above, execute `terraform plan` 

### Important Factoids
None.

### References
None.

ndmckinley:
The `element` function, evaluated in an output or dependent resource, will fail if the thing it's referencing is a list item, and the list is only computed.  This is unfortunate in examples like #1805 - this is a somewhat silly workaround for this that I expect to be obviated when HCL2 is released.

Fixes #1805 - sort of.  :)

### DIFF
--- a/google/resource_compute_instance_from_template.go
+++ b/google/resource_compute_instance_from_template.go
@@ -28,9 +28,14 @@ func resourceComputeInstanceFromTemplate() *schema.Resource {
 func computeInstanceFromTemplateSchema() map[string]*schema.Schema {
 	s := resourceComputeInstance().Schema
 
-	for _, field := range []string{"boot_disk", "machine_type", "network_interface"} {
+	for _, field := range []string{"boot_disk", "machine_type"} {
 		s[field].Required = false
 	}
+
+	s["network_interface"].Required = false
+	// NOT because you should set it - it won't do anything -
+	// but as a workaround for https://github.com/terraform-providers/terraform-provider-google/issues/1805.
+	s["network_interface"].Optional = true
 
 	// Remove deprecated/removed fields that are never d.Set. We can't
 	// programatically remove all of them, because some of them still have d.Set


ndmckinley:
Hmm... I _think_ it has revealed a bug in Terraform, not the resource.  But you could also say that making them `Optional` + `Computed` instead of just `Computed` would be a value-add - you could override values in the template that way, which is worth something.

What do you think?

ndmckinley:
Oh, easy answer - the docs already say you can do that, ha, so it *is* a bug in this resource.  :)  Pushing change.

danawillow:
The PR title should probably be updated now

Issue #1846, opened 2018-08-07
terraform import google_container_node_pool not working. It is not able to find the resource ID of the node-pool.


santinoncs:
### Terraform Version
Terraform v0.11.7
+ provider.google v1.16.2

### Affected Resource(s)
- google_container_node_pool

### Terraform Configuration Files
```hcl
resource "google_container_node_pool" "autoscaling" {
  name       = "autoscaling"
  zone       = "europe-west1-b"
  cluster    = "${google_container_cluster.kube-1.name}"
  region     = "europe-west1"
  node_count = 0
  autoscaling {
      min_node_count = 0
      max_node_count = 20
  }

  node_config {
    preemptible  = true
    machine_type = "n1-standard-2"

    oauth_scopes = [
      "compute-rw",
      "storage-ro",
      "logging-write",
      "monitoring",
    ]
  }
}

```

### Command executed to import

```
terraform import module.gke.google_container_node_pool.autoscaling kubertonic-st/europe-west1-b/kube-1/autoscaling
```

### Output

```
$ terraform import module.gke.google_container_node_pool.autoscaling kubertonic-st/europe-west1-b/kube-1/autoscaling
module.gke.google_container_node_pool.autoscaling: Importing from ID "kubertonic-st/europe-west1-b/kube-1/autoscaling"...
module.gke.google_container_node_pool.autoscaling: Import complete!
  Imported google_container_node_pool (ID: kubertonic-st/europe-west1-b/kube-1/autoscaling)
module.gke.google_container_node_pool.autoscaling: Refreshing state... (ID: kubertonic-st/europe-west1-b/kube-1/autoscaling)

Error: module.gke.google_container_node_pool.autoscaling (import id: kubertonic-st/europe-west1-b/kube-1/autoscaling): 1 error(s) occurred:

* import module.gke.google_container_node_pool.autoscaling result: kubertonic-st/europe-west1-b/kube-1/autoscaling: import module.gke.google_container_node_pool.autoscaling (id: kubertonic-st/europe-west1-b/kube-1/autoscaling): Terraform detected a resource with this ID doesn't
exist. Please verify the ID is correct. You cannot import non-existent
resources using Terraform import.

$
```

### Debug Output

```
[...]

PR #2035, opened 2018-09-12
Set correct nodepool Id using 4 part importer

jamielennox:
The ID that is created for a nodepool is 3 parts in the form of
<location>/<cluster>/<nodepool>. When we use the newer 4 part import
format the ID becomes that string. Then when checking the existence of
the nodepool the name is incorrectly parsed out of the ID.

When using the 4 part import form we need to set the correct ID so that
terraform can parse the name from the ID.

Closes: #1846

### DIFF
--- a/google/resource_container_node_pool.go
+++ b/google/resource_container_node_pool.go
@@ -420,6 +420,9 @@ func resourceContainerNodePoolStateImporter(d *schema.ResourceData, meta interfa
 
 		d.Set("cluster", parts[2])
 		d.Set("name", parts[3])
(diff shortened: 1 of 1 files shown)


jamielennox:
Fixes the problem for me. I can't see any tests that address the importer, if you can point me to an example of how I should test this I can add some.

ndmckinley:
Looks good.  We do have tests for that - I'll run 'em on the CI.  :)

ndmckinley:
They work, merging.

PR #2037, opened 2018-09-12
compute_network_peering works with regional networks too

trthomps:


trthomps:
I guess even with regional networks you address them with global, my bad.

Issue #632, opened 2017-10-25
Add service account e-mail to google_sql_database_instance returned attributes


patrykmikula:
Currently there is no option to reach google_sql_database_instance service account e-mail address.
Service account e-mail address is required to grant write permission to GCP storage bucket in order to export SQL data from instance there (SQL dump/CSV).

'google_storage_bucket_acl' role_entity supports SQL instance service account email:
 
role_entity = [
    "WRITER:user-XXXXXXYYYYYZZZZZ@speckle-umbrella-11.iam.gserviceaccount.com",
  ]

but SQL instance e-mail string is unreachable for terraform. 
As a consequence export command
`gcloud sql instances export DB_NAME gs://BUCKET/DIR/FILE.sql` causes  `Exporting Cloud SQL instance...
..failed.
ERROR: (gcloud.sql.instances.export) ERROR_RDBMS`
and `Access denied for account XXXXXXYYYYYZZZZZ@speckle-umbrella-11.iam.gserviceaccount.com (permission issue?)` visible in SQL instance logs as there is no way to automatically grant RW permission to bucket for each new SQL instance.

WiseLin1125:
Hi there, I have encountered the same issue, and my question is where does the account like `user-XXXXXXYYYYYZZZZZ@speckle-umbrella-11.iam.gserviceaccount.com",` come from?

And I already created a service account as `Project Owner` access rights, in this case, my service account have the biggest privilege right?

But when I import csv from GCS to Cloud SQL(mysql) I still can't import successfully unless I add such as `user-XXXXXXYYYYYZZZZZ@speckle-umbrella-11.iam.gserviceaccount.com",`  into my GCS bucket.

Is there any update about this issue?

rtindru:
Bumping this question >> I have the same issue as @patrykmikula . For our pipeline to work, I need to grant Object Viewer permissions on GCS bucket to the service account generated by terraform for my CloudSQL instance. It appears that the CloudSQL instance attributes do not expose the service account inside terraform.

jpaulgs:
I encountered the same issue as @patrykmikula... it's been 10 months since it was first reported is it likely to be fixed?  

importing the service account isn't ideal as I really hate manual steps when creating new environments...

gorilych:
w/a is to use "null_resource" and "local_file", though this requires gcloud to be installed along with terraform. In below code gcloud uses another service account for auth, but you can use other methods.

```
[...]

PR #2040, opened 2018-09-12
Service account email for SQL instances

chrisst:
Fixes https://github.com/terraform-providers/terraform-provider-google/issues/632


chrisst:
Done

PR #2043, opened 2018-09-13
Fix sql_db test resource name that could start with a #.

rileykarson:


google/resource_sql_database_test.go
```diff
@@ -71,8 +71,8 @@ func TestAccSqlDatabase_update(t *testing.T) {
 
 	var database sqladmin.Database
 
-	instance_name := acctest.RandString(10)
-	database_name := acctest.RandString(10)
+	instance_name := fmt.Sprintf("sqldatabasetest%s", acctest.RandString(10))
```
danawillow: Would `acctest.RandomWithPrefix("sqldatabasetest")` work, or are hyphens not allowed?

PR #2047, opened 2018-09-13
change delay in region instance group data source test

danawillow:
We had a failure in the last CI run while waiting for this resource to be created, so let's increase the timeout.

```
------- Stdout: -------
=== RUN   TestAccDataSourceRegionInstanceGroup
--- FAIL: TestAccDataSourceRegionInstanceGroup (631.53s)
		
		* google_compute_region_instance_group_manager.foo: 1 error(s) occurred:
		
		* google_compute_region_instance_group_manager.foo: timeout while waiting for state to become 'created' (last state: 'creating', timeout: 5m0s)
	testing.go:579: Error destroying resource! WARNING: Dangling resources
		may exist. The full state and error is shown below.
		
		Error: Error refreshing: 1 error(s) occurred:
		
		* google_compute_region_instance_group_manager.foo: 1 error(s) occurred:
		
		* google_compute_region_instance_group_manager.foo: google_compute_region_instance_group_manager.foo: timeout while waiting for state to become 'created' (last state: 'creating', timeout: 5m0s)
		
		State: 
FAIL
```

### DIFF
--- a/google/data_source_google_compute_region_instance_group_test.go
+++ b/google/data_source_google_compute_region_instance_group_test.go
@@ -78,7 +78,7 @@ resource "google_compute_region_instance_group_manager" "foo" {
 
 	auto_healing_policies {
 		health_check = "${google_compute_health_check.autohealing.self_link}"
-		initial_delay_sec = 600
+		initial_delay_sec = 10
 	}
 }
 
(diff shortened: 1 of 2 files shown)


rileykarson:
I don't think this is the fix here, since none of the other autohealing tests have caught anything. 600s is weirdly high compared to our normal 10 though.

Silly question but looking at the config aren't new-style health checks and target pools incompatible?

danawillow:
Where are you getting 10s from? I think even our fastest resources still take more than 10 seconds to create.

rileykarson:
Oh oops I accidentally forgot a sentence. The only thing weird I can see other than the health check in the config is the auto_healing.initial_delay_sec which is weirdly high.

danawillow:
Ohhh yeah that's almost certainly it. Changed to do that.

PR #2065, opened 2018-09-17
Introduce a new Google provider landing page, complementary pages.

rileykarson:
WANT_LGTM=ALL. Feel free to remove yourself as a reviewer if you don't have anything to add, or to add yourself if you have feelings and I haven't added you!

I've added a bunch of information to the provider page and split it in 3.

* The primary/index page contains non-technical information about being a provider user. This means things like how we work on features/bugs, when to file a bug or feature, and some information about Beta resources. It collects information from the repo README, issue templates, and a couple other places. My intention here is to have all that information available w/o looking at our GitHub repo.
	* I kept the Beta Features header unchanged because we still have a few resources linking to it.
* The Getting Started page is intended to introduce a new user to Google Provider concepts like self links, projects, regions & zones and GCP credentials while also pushing users towards using service accounts. My hope is that it is a little easier to follow than being thrown at the provider reference - a user who reads Getting Started should be ready to start using the provider w/o the reference.
	* The intended audience is users who already know how Terraform works but are not necessarily familiar with GCP. Let me know if you think I should make it more/less advanced!
* The provider reference is mostly just the contents of the old index page. I specifically call out not having a provider block because I think that's a fun piece of information.

Please review with any nitpicks you have, especially about tone & grammar, or if you think we could present the content in a better way!

### DIFF
--- a/website/docs/getting_started.html.markdown
+++ b/website/docs/getting_started.html.markdown
@@ -31,7 +31,7 @@ provider "google" {
 }
 ```
 
-* The `project` field should include your personal project id. The `project`
+* The `project` field should be your personal project id. The `project`
 indicates the default GCP project all of your resources will be created in.
 Most Terraform resources will have a `project` field.
 * The `region` and `zone` are [locations](https://cloud.google.com/compute/docs/regions-zones/global-regional-zonal-resources)
@@ -55,18 +55,17 @@ identifies the provider for Terraform, `compute` indicates the GCP product
 family, and `instance` is the resource name.
 
 Google provider resources will generally, although not always, be named after
-the name used in the `gcloud`/the REST API. For example, a VM instance is called
+the name used in `gcloud`/the REST API. For example, a VM instance is called
 [`instance` in the API](https://cloud.google.com/compute/docs/reference/rest/v1/instances).
 Most resource field names will also correspond 1:1 with their `gcloud`/REST API
 names.
 
 If you look at the [`google_compute_instance documentation`](/docs/providers/google/r/compute_instance.html),
 you'll see that `project` and `zone` (VM instances are a zonal resource) are
-listed as optional. You can omit them from a Terraform config and the provider
-defaults will be used. If any of these fields is added to a config, it will be
-used instead of the default.
+listed as optional. When present in a resource's config block, these values will
+be used. If omitted, the provider defaults will be used instead.
 
-Try adding the following to your config file:
+Add the following to your config file:
 
 ```hcl
 resource "google_compute_instance" "vm_instance" {
@@ -92,11 +91,11 @@ resource "google_compute_instance" "vm_instance" {
 credentials. If you want to try out provisioning your VM instance before
 continuing, follow the instructions in the "Adding credentials" section below.
 
-## Connecting to a VPC network
+## Linking GCP resources
 
 Like this VM instance, nearly every GCP resource will have a `name` field. They
-are used as a short way to identify resources, and their name in the Cloud
-Console will be the one defined in the `name` field.
+are used as a short way to identify resources, and a resource's display name in
+the Cloud Console will be the one defined in the `name` field.
 
 When linking resources in a Terraform config though, you'll primarily want to
 use a different field, the `self_link` of a resource. Like `name`, nearly every
@@ -106,16 +105,17 @@ resource has a `self_link`. They look like:
 {{API base url}}/projects/{{your project}}/{{location type}}/{{location}}/{{resouce type}}/{{name}}
 ```
 
-For a real example, let's look at what the instance `self_link` will be assuming
-your project is named `foo`:
+For example, the instance defined earlier in a project named `foo` will have
+the `self_link`:
 
 ```
 https://www.googleapis.com/compute/v1/projects/foo/zones/us-central1-c/instances/terraform-instance
 ```
 
 A resource's `self_link` is a unique reference to that resource. When
-connecting two resources in Terraform, you can use Terraform interpolation to
-avoid typing all that out! Let's use a `google_compute_network` to demonstrate.
+linking two resources in Terraform, you can use Terraform interpolation to
+avoid typing out the self link! Let's use a `google_compute_network` to
+demonstrate.
 
 Add this block to your config:
 
@@ -127,8 +127,8 @@ resource "google_compute_network" "vpc_network" {
 ```
 
 This will create [VPC network resource](/docs/providers/google/r/compute_network.html)
-with a subnetwork (due to auto_create_subnetworks) in each region. Then, change
-the network of the `google_compute_instance`.
+with a subnetwork in each region. Then, change the network of the
+`google_compute_instance` from the `"default"` network to the new network.
 
 ```diff
 network_interface {
@@ -140,10 +140,10 @@ network_interface {
 
 This means that when we create the VM instance, it will use
 `"terraform-network"` instead of the default VPC network for the project. If you
-run `terraform plan`, you should see that `"terraform-instance"` depends on
+run `terraform plan`, you will see that `"terraform-instance"` depends on
 `"terraform-network"`.
 
-You're ready to run `terraform apply`! Except for one last thing...
+You can now run `terraform apply`! Except for one last thing...
 
 ~> You haven't added GCP API credentials yet.
 
@@ -169,7 +169,7 @@ export GOOGLE_CLOUD_KEYFILE_JSON={{path}}
 `bashrc` to store your credentials across sessions!
 
 ## Provisioning your resources
-By now, your config should look something like:
+By now, your config will look something like:
 
 ```hcl
 provider "google" {
@@ -210,8 +210,7 @@ terraform apply
 ```
 
 Congratulations! You've gotten started using the Google provider and provisioned
-a virtual machine on Google Cloud Platform. Along the way, you've learned
-important concepts such as:
+a virtual machine on Google Cloud Platform. The key concepts unique to GCP are:
 
 * How a `project` contains resources
     * and how to use a default `project` in your provider
--- a/website/docs/getting_started.html.markdown
+++ b/website/docs/getting_started.html.markdown
@@ -127,7 +127,7 @@ resource "google_compute_network" "vpc_network" {
 ```
 
 This will create [VPC network resource](/docs/providers/google/r/compute_network.html)
-with a subnetwork in each region. Then, change the network of the
+with a subnetwork in each region. Next, change the network of the
 `google_compute_instance` from the `"default"` network to the new network.
 
 ```diff
@@ -153,7 +153,7 @@ with Terraform is to use a [GCP service account](https://cloud.google.com/docs/a
 a "robot account" that can be granted a limited set of IAM permissions.
 
 From [the service account key page in the Cloud Console](https://pantheon.corp.google.com/apis/credentials/serviceaccountkey)
-choose an existing account, or create a new one. Then, download the JSON key
+choose an existing account, or create a new one. Next, download the JSON key
 file. Name it something you can remember, and store it somewhere secure on your
 machine.
 
--- a/website/docs/getting_started.html.markdown
+++ b/website/docs/getting_started.html.markdown
@@ -10,7 +10,7 @@ description: |-
 
 ## Before you begin
 
-* You'll want to have created an account and project in the [Google Cloud Console](https://console.cloud.google.com/)
+* Create a project in the [Google Cloud Console](https://console.cloud.google.com/)
 and set up billing on that project. Any examples in this guide will be part of
 the [GCP "always free" tier](https://cloud.google.com/free/).
 * [Install Terraform](https://www.terraform.io/intro/getting-started/install.html)
@@ -20,7 +20,7 @@ provider.
 
 ## Configuring the Provider
 
-To configure the provider, first create a Terraform config file. Inside, you'll
+First create a Terraform config file named `"main.tf"`. Inside, you'll
 want to include the following configuration:
 
 ```hcl
@@ -34,7 +34,7 @@ provider "google" {
 * The `project` field should include your personal project id. The `project`
 indicates the default GCP project all of your resources will be created in.
 Most Terraform resources will have a `project` field.
-* The`region` and `zone` are [locations](https://cloud.google.com/compute/docs/regions-zones/global-regional-zonal-resources)
+* The `region` and `zone` are [locations](https://cloud.google.com/compute/docs/regions-zones/global-regional-zonal-resources)
 for your resources to be created in.
     * The `region` will be used to choose the default location for regional
     resources. Regional resources are spread across several zones.
@@ -55,8 +55,10 @@ identifies the provider for Terraform, `compute` indicates the GCP product
 family, and `instance` is the resource name.
 
 Google provider resources will generally, although not always, be named after
-the name used in the REST API. For example, a VM instance is called [`instance` in the API](https://cloud.google.com/compute/docs/reference/rest/v1/instances).
-Most resource field names will also correspond 1:1 with their REST API names.
+the name used in the `gcloud`/the REST API. For example, a VM instance is called
+[`instance` in the API](https://cloud.google.com/compute/docs/reference/rest/v1/instances).
+Most resource field names will also correspond 1:1 with their `gcloud`/REST API
+names.
 
 If you look at the [`google_compute_instance documentation`](/docs/providers/google/r/compute_instance.html),
 you'll see that `project` and `zone` (VM instances are a zonal resource) are
@@ -64,7 +66,7 @@ listed as optional. You can omit them from a Terraform config and the provider
 defaults will be used. If any of these fields is added to a config, it will be
 used instead of the default.
 
-Try adding the following to your state file:
+Try adding the following to your config file:
 
 ```hcl
 resource "google_compute_instance" "vm_instance" {
@@ -87,7 +89,7 @@ resource "google_compute_instance" "vm_instance" {
 ```
 
 ~> Note: Don't use `terraform apply` quite yet! You still need to add GCP
-credentials below. If you want to try out provisioning your VM instance before
+credentials. If you want to try out provisioning your VM instance before
 continuing, follow the instructions in the "Adding credentials" section below.
 
 ## Connecting to a VPC network
@@ -157,10 +159,10 @@ file. Name it something you can remember, and store it somewhere secure on your
 machine.
 
 You supply the key to Terraform using the environment variable
-`GOOGLE_APPLICATION_CREDENTIALS`, setting the value to the location of the file.
+`GOOGLE_CLOUD_KEYFILE_JSON`, setting the value to the location of the file.
 
 ```bash
-export GOOGLE_APPLICATION_CREDENTIALS={{path}}
+export GOOGLE_CLOUD_KEYFILE_JSON={{path}}
 ```
 
 -> Remember to add this line to a startup file such as `bash_profile` or
--- a/website/docs/getting_started.html.markdown
+++ b/website/docs/getting_started.html.markdown
@@ -143,9 +143,8 @@ This means that when we create the VM instance, it will use
 run `terraform plan`, you will see that `"terraform-instance"` depends on
 `"terraform-network"`.
 
-You can now run `terraform apply`! Except for one last thing...
-
-~> You haven't added GCP API credentials yet.
+Your configuration is complete. Before you can run `terraform apply` though,
+Terraform needs GCP credentials.
 
 ## Adding credentials
 In order to make requests against the GCP API, you need to authenticate to prove
--- a/website/docs/getting_started.html.markdown
+++ b/website/docs/getting_started.html.markdown
@@ -102,7 +102,7 @@ use a different field, the `self_link` of a resource. Like `name`, nearly every
 resource has a `self_link`. They look like:
 
 ```
-{{API base url}}/projects/{{your project}}/{{location type}}/{{location}}/{{resouce type}}/{{name}}
+{{API base url}}/projects/{{your project}}/{{location type}}/{{location}}/{{resource type}}/{{name}}
 ```
 
 For example, the instance defined earlier in a project named `foo` will have
--- a/website/docs/getting_started.html.markdown
+++ b/website/docs/getting_started.html.markdown
@@ -224,3 +224,8 @@ Run `terraform destroy` to tear down your resources.
 
 Afterwards, check out the [provider reference](/docs/providers/google/provider_reference.html) for more details on configuring
 the provider block (including how you can eliminate it entirely!).
+
+You can also check out the [GCP Community tutorials](https://cloud.google.com/community/tutorials/)
+such as:
+* [Managing GCP Projects with Terraform](https://cloud.google.com/community/tutorials/managing-gcp-projects-with-terraform)
+* [Modular Load Balancing with Terraform](https://cloud.google.com/community/tutorials/modular-load-balancing-with-terraform)
--- a/website/docs/getting_started.html.markdown
+++ b/website/docs/getting_started.html.markdown
@@ -225,5 +225,6 @@ the provider block (including how you can eliminate it entirely!).
 
 You can also check out the [GCP Community tutorials](https://cloud.google.com/community/tutorials/)
 such as:
+* [Getting started with Terraform on Google Cloud Platform](https://cloud.google.com/community/tutorials/getting-started-on-gcp-with-terraform)
 * [Managing GCP Projects with Terraform](https://cloud.google.com/community/tutorials/managing-gcp-projects-with-terraform)
 * [Modular Load Balancing with Terraform](https://cloud.google.com/community/tutorials/modular-load-balancing-with-terraform)
--- a/website/docs/getting_started.html.markdown
+++ b/website/docs/getting_started.html.markdown
@@ -1,0 +1,224 @@
+---
+layout: "google"
+page_title: "Getting Started with the Google provider"
+sidebar_current: "docs-google-provider-getting-started"
+description: |-
+  Getting started with the Google Cloud Platform provider
+---
+
+# Getting Started with the Google Provider
+
+## Before you begin
+
+* You'll want to have created an account and project in the [Google Cloud Console](https://console.cloud.google.com/)
+and set up billing on that project. Any examples in this guide will be part of
+the [GCP "always free" tier](https://cloud.google.com/free/).
+* [Install Terraform](https://www.terraform.io/intro/getting-started/install.html)
+and read the Terraform getting started guide that follows. This guide will
+assume basic proficiency with Terraform - it is an introduction to the Google
+provider.
+
+## Configuring the Provider
+
+To configure the provider, first create a Terraform config file. Inside, you'll
+want to include the following configuration:
+
+```hcl
+provider "google" {
+  project = "{{YOUR GCP PROJECT}}"
+  region  = "us-central1"
+  zone    = "us-central1-c"
+}
+```
+
+* The `project` field should include your personal project id. The `project`
+indicates the default GCP project all of your resources will be created in.
+Most Terraform resources will have a `project` field.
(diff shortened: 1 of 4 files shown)


website/docs/index.html.markdown
```diff
@@ -1,108 +1,88 @@
...
-    * `CLOUDSDK_COMPUTE_ZONE`
-
-
-## Beta Features
-
-Some Google Provider resources contain Beta features; Beta GCP Features have no
+### Beta Features
+Some resources contain Beta features; Beta GCP Features have no
 deprecation policy, and no SLA, but are otherwise considered to be feature-complete
 with only minor outstanding issues after their Alpha period. Beta is when a GCP feature
```
danawillow: grammar nit (I know this is in the unchanged part): this sentence starts in singular and ends in plural, let's make it plural all the way through (beta is when gcp features are publicly announced)

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,224 @@
...
+sidebar_current: "docs-google-provider-getting-started"
+description: |-
+  Getting started with the Google Cloud Platform provider
+---
+
+# Getting Started with the Google Provider
+
+## Before you begin
+
+* You'll want to have created an account and project in the [Google Cloud Console](https://console.cloud.google.com/)
```
danawillow: This is a teeny bit wordy, and I think it's not super important to say to create an account. maybe just something like "Create a project in [...]"

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,224 @@
...
+and set up billing on that project. Any examples in this guide will be part of
+the [GCP "always free" tier](https://cloud.google.com/free/).
+* [Install Terraform](https://www.terraform.io/intro/getting-started/install.html)
+and read the Terraform getting started guide that follows. This guide will
+assume basic proficiency with Terraform - it is an introduction to the Google
+provider.
+
+## Configuring the Provider
+
+To configure the provider, first create a Terraform config file. Inside, you'll
```
danawillow: Since this is in a section called "configuring the provider", we can probably remove "to configure the provider"

danawillow: just to make it super obvious what's going on (since people might read this without being super fluent in tf), say to name the file main.tf

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,224 @@
...
+  project = "{{YOUR GCP PROJECT}}"
+  region  = "us-central1"
+  zone    = "us-central1-c"
+}
+```
+
+* The `project` field should include your personal project id. The `project`
+indicates the default GCP project all of your resources will be created in.
+Most Terraform resources will have a `project` field.
+* The`region` and `zone` are [locations](https://cloud.google.com/compute/docs/regions-zones/global-regional-zonal-resources)
```
danawillow: nit: space between the and region

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,224 @@
...
+the name used in the REST API. For example, a VM instance is called [`instance` in the API](https://cloud.google.com/compute/docs/reference/rest/v1/instances).
+Most resource field names will also correspond 1:1 with their REST API names.
+
+If you look at the [`google_compute_instance documentation`](/docs/providers/google/r/compute_instance.html),
+you'll see that `project` and `zone` (VM instances are a zonal resource) are
+listed as optional. You can omit them from a Terraform config and the provider
+defaults will be used. If any of these fields is added to a config, it will be
+used instead of the default.
+
+Try adding the following to your state file:
```
danawillow: config, not state

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,224 @@
...
+Instances created in zones outside the US are not part of the always free tier
+and could incur charges.
+
+## Creating a VM instance
+A [Google Compute Engine VM instance](https://cloud.google.com/compute/docs/instances/) is
+named `google_compute_instance` in Terraform. The `google` part of the name
+identifies the provider for Terraform, `compute` indicates the GCP product
+family, and `instance` is the resource name.
+
+Google provider resources will generally, although not always, be named after
```
danawillow: Maybe mention something about that name also being used for gcloud, since people will probably be familiar with that and not with the rest api

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,224 @@
...
+    # A default network is created for all GCP projects
+    network       = "default"
+    access_config = {
+    }
+  }
+}
+```
+
+~> Note: Don't use `terraform apply` quite yet! You still need to add GCP
+credentials below. If you want to try out provisioning your VM instance before
```
danawillow: nit: rm 'below' here

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,224 @@
...
+that it's you making the request. The preferred method of provisioning resources
+with Terraform is to use a [GCP service account](https://cloud.google.com/docs/authentication/getting-started),
+a "robot account" that can be granted a limited set of IAM permissions.
+
+From [the service account key page in the Cloud Console](https://pantheon.corp.google.com/apis/credentials/serviceaccountkey)
+choose an existing account, or create a new one. Then, download the JSON key
+file. Name it something you can remember, and store it somewhere secure on your
+machine.
+
+You supply the key to Terraform using the environment variable
```
danawillow: This is for default credentials, which I'm a bit nervous about telling people to do since I'd assume that it would get overridden if you ended up doing a `gcloud auth application-default login` for some reason. I'd rather tell people to use the explicit env variables that we provide (`GOOGLE_CREDENTIALS` being the preferred) and then to use `export GOOGLE_CREDENTIALS=$(cat {{path}})`

website/docs/provider_reference.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+layout: "google"
+page_title: "google provider reference"
+sidebar_current: "docs-google-provider-reference"
+description: |-
+  The Google provider is used to configure your GCP project, location, and creds
+---
+
+# `google` provider reference
+The `google` provider block is used to configure default values for your GCP
+project, location (`zone` and `region`), and add your credentials credentials.
```
danawillow: credentials credentials

danawillow: project and location

website/docs/provider_reference.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+  If you are running Terraform from a GCE instance, see [Creating and Enabling
+  Service Accounts for Instances][gce-service-account] for details.
+
+  On your computer, if you have made your identity available as the
+  Application Default Credentials by running [`gcloud auth application-default
+  login`][gcloud adc], the provider will use your identity.
+
+  -> [Service accounts][service accounts] are the recommended way
+  to manage GCP credentials. [GCE metadata] is also acceptable, although can
+  only be used in limited scenarios. The `gcloud` method is not guaranteed to work
```
danawillow: limited scenarios being what?

website/docs/provider_reference.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+* `project` - (Optional) The ID of the project to apply any resources to.  This
+  can also be specified using any of the following environment variables (listed
+  in order of precedence):
+
+    * `GOOGLE_PROJECT`
+    * `GOOGLE_CLOUD_PROJECT`
+    * `GCLOUD_PROJECT`
+    * `CLOUDSDK_CORE_PROJECT`
+
+    -> `GOOGLE_CLOUD_PROJECT` is the recommended environment variable to use if
```
danawillow: is there a reason behind this recommendation or did you just decide it was a good idea to have an opinion?

website/docs/index.html.markdown
```diff
@@ -1,108 +1,88 @@
...
 ---
 
-# Google Cloud Provider
+# Google Cloud Platform Provider
 
-The Google Cloud provider is used to interact with
-[Google Cloud services](https://cloud.google.com/). The provider needs
-to be configured with the proper credentials before it can be used.
+The Google provider is used to configure your [Google Cloud Platform](https://cloud.google.com/) infrastructure. 
+See the [Getting Started](/docs/providers/google/getting_started.html) page for an introduction to using the provider. See the [provider reference](/docs/providers/google/provider_reference.html) for advanced
```
danawillow: I wouldn't really call it advanced details, since it's pretty essential information. If someone's coming in with a solid knowledge of terraform and gcp, that's going to be where they'd want to start (vs the getting started page, which is great, but more of a tutorial)

website/docs/index.html.markdown
```diff
@@ -1,108 +1,88 @@
...
+details on authenticating or otherwise configuring the provider.
```
danawillow: I do kind of think that we should have some sort of super bare bones example here so people can get an idea of what they're looking at without having to click on a link

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,224 @@
...
+* How a `project` contains resources
+    * and how to use a default `project` in your provider
+* What a resource being global, regional, or zonal means on GCP
+    * and how to specify a default `region` and `zone`
+* How GCP uses `name` and `self_link` to identify resources
+* How to add GCP service account credentials to Terraform
+
+Run `terraform destroy` to tear down your resources.
+
+Afterwards, check out the [provider reference](/docs/providers/google/provider_reference.html) for more details on configuring
```
danawillow: Oh we should also maybe link to some of the other google-published tutorials here:
https://cloud.google.com/community/tutorials/managing-gcp-projects-with-terraform
https://cloud.google.com/community/tutorials/modular-load-balancing-with-terraform

and maybe something about the module registry too somewhere:
https://registry.terraform.io/browse?provider=google

website/docs/getting_started.html.markdown
rileykarson: Oh yup I totally thought I had used that env var myself but I was wrong I used `GOOGLE_CLOUD_KEYFILE_JSON`. Updated the credentials definition to say paths also work.

website/docs/provider_reference.html.markdown
rileykarson: I decided it was a good idea to have an opinion because I do not like looking @  like 4 different options. Some of our 1P [Google] tools that interact with Terraform use that var so I chose it.

website/docs/getting_started.html.markdown
rileykarson: Done! I linked the module registry on the main page & the tutorials as followups to getting started for now. I'd like to explore expanding guides/modules at some point.

website/docs/index.html.markdown
```diff
@@ -1,108 +1,103 @@
...
-  ~> **Warning:** The gcloud method is not guaranteed to work for all APIs, and
-  [service accounts] or [GCE metadata] should be used if possible.
+Community input is appreciated on outstanding issues! We love to hear what use
+cases you have for new features, and want to provide the best possible
+experience for you using the Google provider.
 
-* `project` - (Optional) The ID of the project to apply any resources to.  This
-  can also be specified using any of the following environment variables (listed
-  in order of precedence):
+If you can't find an existing issue for a bug or feature request you have
```
chrisst: I would break this into 2 different sentences, this doesn't seem to flow well right now as a bulleted list.

website/docs/index.html.markdown
```diff
@@ -1,108 +1,103 @@
...
-    * `CLOUDSDK_COMPUTE_REGION`
+The provider maintainers will often use the assignee field on an issue to mark
+who is working on it.
 
-* `zone` - (Optional) The zone to operate under, if not specified by a given resource.
-  This can also be specified using any of the following environment variables (listed in order of
-  precedence):
+* An issue assigned to an individual maintainer indicates that maintainer is working
+on the issue
+* An issue assigned to the `modular-magician` indicates the feature is being
```
chrisst: Should we encourage them to visit/engage with the upstream PR instead?

website/docs/index.html.markdown
```diff
@@ -1,108 +1,103 @@
...
+* An issue assigned to `hashibot` indicates a member of the community has taken on
+the issue!
 
-    * `GOOGLE_ZONE`
-    * `GCLOUD_ZONE`
-    * `CLOUDSDK_COMPUTE_ZONE`
+## Contributing
 
+If you'd like to help extend the Google provider, we gladly accept community
+contributions! Check out the [provider GitHub repo README](https://github.com/terraform-providers/terraform-provider-google)
```
chrisst: "[provider GitHub repo README]" is a bit verbose given the user is already reading GCP specific docs. I would simplify to [Github README]

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,231 @@
...
+and could incur charges.
+
+## Creating a VM instance
+A [Google Compute Engine VM instance](https://cloud.google.com/compute/docs/instances/) is
+named `google_compute_instance` in Terraform. The `google` part of the name
+identifies the provider for Terraform, `compute` indicates the GCP product
+family, and `instance` is the resource name.
+
+Google provider resources will generally, although not always, be named after
+the name used in the `gcloud`/the REST API. For example, a VM instance is called
```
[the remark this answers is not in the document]
rileykarson: Oh yup there is an extra "the" there.

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,231 @@
...
+    * and how to specify a default `region` and `zone`
+* How GCP uses `name` and `self_link` to identify resources
+* How to add GCP service account credentials to Terraform
+
+Run `terraform destroy` to tear down your resources.
+
+Afterwards, check out the [provider reference](/docs/providers/google/provider_reference.html) for more details on configuring
+the provider block (including how you can eliminate it entirely!).
+
+You can also check out the [GCP Community tutorials](https://cloud.google.com/community/tutorials/)
```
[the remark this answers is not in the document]
rileykarson: 👍 feel free to PR it in once it goes live!

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,231 @@
...
+
+Google provider resources will generally, although not always, be named after
+the name used in the `gcloud`/the REST API. For example, a VM instance is called
+[`instance` in the API](https://cloud.google.com/compute/docs/reference/rest/v1/instances).
+Most resource field names will also correspond 1:1 with their `gcloud`/REST API
+names.
+
+If you look at the [`google_compute_instance documentation`](/docs/providers/google/r/compute_instance.html),
+you'll see that `project` and `zone` (VM instances are a zonal resource) are
+listed as optional. You can omit them from a Terraform config and the provider
```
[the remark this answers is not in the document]
rileykarson: Reworded

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,231 @@
...
+Try adding the following to your config file:
+
+```hcl
+resource "google_compute_instance" "vm_instance" {
+  name         = "terraform-instance"
+  machine_type = "f1-micro"
+
+  boot_disk {
+    initialize_params {
+      image = "debian-cloud/debian-9"
```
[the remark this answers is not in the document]
rileykarson: I'm gonna leave this out so we keep the config simple. We will have to update it later though.

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,231 @@
...
+```
+
+~> Note: Don't use `terraform apply` quite yet! You still need to add GCP
+credentials. If you want to try out provisioning your VM instance before
+continuing, follow the instructions in the "Adding credentials" section below.
+
+## Connecting to a VPC network
+
+Like this VM instance, nearly every GCP resource will have a `name` field. They
+are used as a short way to identify resources, and their name in the Cloud
```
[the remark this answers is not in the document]
rileykarson: Reworded.

website/docs/index.html.markdown
rileykarson: Reworded

website/docs/index.html.markdown
rileykarson: Imo they should still discuss on the Terraform side. I just want it to be clear that the Magician being assigned means we are working on it! (That and I mostly stole that section from @paddycarver's community note 🙂 )

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,231 @@
...
+
+For a real example, let's look at what the instance `self_link` will be assuming
+your project is named `foo`:
+
+```
+https://www.googleapis.com/compute/v1/projects/foo/zones/us-central1-c/instances/terraform-instance
+```
+
+A resource's `self_link` is a unique reference to that resource. When
+connecting two resources in Terraform, you can use Terraform interpolation to
```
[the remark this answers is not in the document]
rileykarson: I think we definitely want to call out interpolating self links explicitly here since that's what makes Terraform figure out how to make the resources in the right order. Unless I'm misunderstanding your comment?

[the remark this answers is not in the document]
chrisst: I'm thinking that any reference of resources from another resource (by name, id, self_link, etc) is pretty standard terraform and doesn't need much explaining here. IIRC the DAG building stuff is well documented in a few videos / posts. The fact that we use self_link's referentially instead of id's is the important part and it doesn't matter what the self_link actually is.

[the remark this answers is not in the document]
rileykarson: Oh sure! I guess it doesn't matter too much. I'm going to keep it b/c my personal preference would be see the format. Being told "there are these unique references to resources" and then not seeing their shape would make me immediately look up their shape.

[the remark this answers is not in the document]
rileykarson: I'm just going to leave `id` out because the less references to it the better imo.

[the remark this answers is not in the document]
chrisst: I was interpreting this as a document for people who have experience in other terraform providers and how do we teach them what is unique to GCP. All the other providers use `id` pretty commonly, so I think an explicit directive to _not_ use `id` is valuable to a user coming from Azure/Aws/etc.

This is all based on my assumption that we are actively discouraging usage of the id field. 

examples from other docs:

```
...
resource "azurerm_network_interface" "myterraformnic" {
    name                      = "myNIC"
    location                  = "eastus"
    resource_group_name       = "${azurerm_resource_group.myterraformgroup.name}"
    network_security_group_id = "${azurerm_network_security_group.myterraformnsg.id}"

    ip_configuration {
        name                          = "myNicConfiguration"
        subnet_id                     = "${azurerm_subnet.myterraformsubnet.id}"
        private_ip_address_allocation = "dynamic"
        public_ip_address_id          = "${azurerm_public_ip.myterraformpublicip.id}"
    }
}
```

[Terraform's getting started doc](https://www.terraform.io/intro/getting-started/dependencies.html):
```
resource "aws_eip" "ip" {
  instance = "${aws_instance.example.id}"
}
```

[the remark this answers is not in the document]
rileykarson: Sure, but some users will also read this as a first time user. They should not see a reference to `id`. I just don't think that we should mention undocumented fields that conveniently exist and also work because they are almost always `self_link`s in an introduction to the provider.

The [eip docs](https://www.terraform.io/docs/providers/aws/r/eip.html#attributes-reference) document an `id` field. And the `instance` field [says it accepts an id](https://www.terraform.io/docs/providers/aws/r/eip.html#attributes-reference). Our resources reference `self_link` and `name`, so the intent of this documentation is to encourage `self_link`s are used over `name` (because if anything they use is ever cross project or region etc users will not be surprised by their provider defaults if they use self links).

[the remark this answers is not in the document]
chrisst: I don't think we need to bike shed too much further on this. I'm happy to discuss more IRL or if other people have opinions to let majority decide.

I will propose that if we have to keep reminding people _not_ to use id in issues then it would be good to document it somewhere so we have a resource to point people to.

website/docs/index.html.markdown
```diff
@@ -1,108 +1,103 @@
...
-  If you are running Terraform from a GCE instance, see [Creating and Enabling
-  Service Accounts for Instances][gce-service-account] for details.
+## Features and Bug Requests
 
-  On your computer, if you have made your identity available as the
-  Application Default Credentials by running [`gcloud auth application-default
-  login`][gcloud adc], the provider will use your identity.
+The Google provider's bugs and feature requests can be found in the [GitHub repo issues](https://github.com/terraform-providers/terraform-provider-google/issues).
+Please avoid "me too" or "+1" comments. Instead, use a thumbs up [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
+on enhancement requests. Provider maintainers will often prioritise work based on the
```
chrisst: prioritize 🇺🇸

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,230 @@
...
++  network = "${google_compute_network.vpc_network.self_link}"
+  access_config = {
+```
+
+This means that when we create the VM instance, it will use
+`"terraform-network"` instead of the default VPC network for the project. If you
+run `terraform plan`, you will see that `"terraform-instance"` depends on
+`"terraform-network"`.
+
+You can now run `terraform apply`! Except for one last thing...
```
chrisst: If you wanted to keep the tone less conversational (drier?) I might change to "Before you can run terraform apply you must add credentials"

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,229 @@
...
+Like this VM instance, nearly every GCP resource will have a `name` field. They
+are used as a short way to identify resources, and a resource's display name in
+the Cloud Console will be the one defined in the `name` field.
+
+When linking resources in a Terraform config though, you'll primarily want to
+use a different field, the `self_link` of a resource. Like `name`, nearly every
+resource has a `self_link`. They look like:
+
+```
+{{API base url}}/projects/{{your project}}/{{location type}}/{{location}}/{{resouce type}}/{{name}}
```
paddycarver: typo: resouce

website/docs/provider_reference.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+}
+```
+
+## Configuration Reference
+
+The following keys can be used to configure the provider.
+
+* `credentials` - (Optional) The path or contents of a file that contains your
+  service account private key in JSON format. You can download your existing
+  [Google Cloud service account file]
```
paddycarver: I think this is a link formatting typo

website/docs/provider_reference.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+  Credentials can also be specified using any of the following environment
+  variables (listed in order of precedence):
+
+    * `GOOGLE_CREDENTIALS`
+    * `GOOGLE_CLOUD_KEYFILE_JSON`
+    * `GCLOUD_KEYFILE_JSON`
+
+  The [`GOOGLE_APPLICATION_CREDENTIALS`][adc]
+  environment variable can also contain the path of a file to obtain credentials
+  from.
```
paddycarver: This should probably mention it's specifically for application default credentials, which are different from service account credentials, I believe.

website/docs/provider_reference.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+  Service Accounts for Instances][gce-service-account] for details.
+
+  On your computer, if you have made your identity available as the
+  Application Default Credentials by running [`gcloud auth application-default
+  login`][gcloud adc], the provider will use your identity.
+
+  -> [Service accounts][service accounts] are the recommended way
+  to manage GCP credentials. [GCE metadata] is also acceptable, although it can
+  only be used when running Terraform from within [certain GCP resources](https://cloud.google.com/docs/authentication/production#obtaining_credentials_on_compute_engine_kubernetes_engine_app_engine_flexible_environment_and_cloud_functions).
+  The `gcloud` method is not guaranteed to work for all APIs.
```
paddycarver: There seems to be some mixing of terms between Application Default Credentials and `gcloud` here. Should we use a unified way to refer to them?

website/docs/provider_reference.html.markdown
paddycarver: Bikeshed: could we use `GOOGLE_PROJECT` or `GCLOUD_PROJECT` as the suggested default, as that is mirrored in the other environment variable names, where as `GOOGLE_CLOUD_REGION` and `GOOGLE_CLOUD_ZONE` aren't supported, making it an inconsistent prefix?

website/docs/provider_reference.html.markdown
rileykarson: Based on https://cloud.google.com/docs/authentication/production#obtaining_and_providing_service_account_credentials_manually they look the same?

paddycarver: I stand corrected, ignore me!

website/docs/getting_started.html.markdown
```diff
@@ -0,0 +1,229 @@
...
+
+Run `terraform destroy` to tear down your resources.
+
+Afterwards, check out the [provider reference](/docs/providers/google/provider_reference.html) for more details on configuring
+the provider block (including how you can eliminate it entirely!).
+
+You can also check out the [GCP Community tutorials](https://cloud.google.com/community/tutorials/)
+such as:
+* [Managing GCP Projects with Terraform](https://cloud.google.com/community/tutorials/managing-gcp-projects-with-terraform)
+* [Modular Load Balancing with Terraform](https://cloud.google.com/community/tutorials/modular-load-balancing-with-terraform)
```
chrisst: woo published: https://cloud.google.com/community/tutorials/getting-started-on-gcp-with-terraform
want to include in this pr?

website/docs/index.html.markdown
```diff
@@ -1,108 +1,103 @@
...
+See the [provider reference](/docs/providers/google/provider_reference.html)
+for more details on authenticating or otherwise configuring the provider.
+
+Interested in the provider's latest features, or want to make sure you're up to date?
+Check out the [changelog](https://github.com/terraform-providers/terraform-provider-google/blob/master/CHANGELOG.md)
+for version information and release notes.
 
-The following keys can be used to configure the provider.
+Take advantage of [Modules](https://www.terraform.io/docs/modules/index.html)
+to simplify your config by using the [Module Registry for GCP modules](https://registry.terraform.io/browse?provider=google)
```
danawillow: s/using/browsing maybe? also period at the end of the sentence

website/docs/provider_reference.html.markdown
```diff
@@ -0,0 +1,96 @@
...
+  Service Accounts for Instances][gce-service-account] for details.
+
+  On your computer, if you have made your identity available as the
+  Application Default Credentials by running [`gcloud auth application-default
+  login`][gcloud adc], the provider will use your identity.
+
+  -> [Service accounts][service accounts] are the recommended way
+  to manage GCP credentials. [GCE metadata] is also acceptable, although it can
+  only be used when running Terraform from within [certain GCP resources](https://cloud.google.com/docs/authentication/production#obtaining_credentials_on_compute_engine_kubernetes_engine_app_engine_flexible_environment_and_cloud_functions).
+  Application Default Credentials, such as those obtained through the `gcloud`
```
danawillow: It's not that Application Default Credentials are not guaranteed, it's that _gcloud_ credentials are not guaranteed, and I want to make sure we aren't confusing the two thing.

Application Default Credentials just means the credentials that GCP APIs can find without you having to explicitly pass them in (https://cloud.google.com/docs/authentication/production#providing_credentials_to_your_application). It encompasses GCE service accounts, App Engine, gcloud credentials, and anything that sets the `GOOGLE_APPLICATION_CREDENTIALS` environment variable.

gcloud credentials are ones granted from gcloud specifically. _These_ are the ones not guaranteed to work for all APIs, because the project that grants these credentials needs to explicitly enable the API in question.

I'm not sure whether our users have a concept of ADC vs gcloud credentials, so I think we don't necessarily need to use either term. Just saying that credentials obtained through `gcloud` are not guaranteed to work for all APIs, and so using a service account is recommended is probably sufficient.

PR #2069, opened 2018-09-18
Replace "Google Apps domain" with "G Suite domain"

spesnova:
## What does this PR do?
Replace "Google Apps domain" with "G Suite domain (primary, instead of alias")

## Motivations
Google Apps is a former name.

Also, I mention primary domain and domain alias. It seems I have to use the primary domain. Otherwise, GCP replaces the field to the primary one. And every time Terraform plan returns a diff about it.

I used domain alias for `members` arg in `google_project_iam_member` resource. `terraform apply` was succeeded. However, when I run modify other resources, I got a diff about the `google_project_iam_member` resource.


spesnova:
@paddycarver 

Updated. Changed only under `website/*`.  Could you check it again?

Issue #1774, opened 2018-07-16
New Resource Request: Filestore


Chupaka:
Hi there,

Cloud Filestore is in Beta now, can we have it in Terraform, please? :)

https://cloud.google.com/filestore/
https://cloud.google.com/filestore/docs/reference/rest/

rolandkool:
+1

TFaga:
+1 as well. I'd like to take a crack at this and make a PR, if still available and possible.

danawillow:
Hey @TFaga, we're generally trying to have new resources these days be generated by [Magic Modules](https://github.com/GoogleCloudPlatform/magic-modules). If you're interested in submitting a PR there, you're more than welcome to! Otherwise, unless it's an urgent business need, it makes our (the terraform-provider-google maintainers) lives easier if we can avoid new handwritten resources.

Chupaka:
Is there any progress on this?.. @TFaga

danawillow:
I think @ndmckinley is also planning on looking at this.

ndmckinley:
I am!  :)  It'll be the next thing I work on.

Chupaka:
@ndmckinley sorry for bothering you, but is there any ETA?..

ndmckinley:
Hi!  I'm about to head out on a 2 week vacation, unfortunately, so it likely won't be until early September.

ryancurrah:
@ndmckinley woot have a good vacay! Looking forward to filestore in tf

rojopolis:
> Hi! I'm about to head out on a 2 week vacation, unfortunately, so it likely won't be until early September.

We're pushing the limit of what could be considered early in September.  Is there any progress to report?

ndmckinley:
:)  Unfortunately, Filestore still isn't available in the golang SDK - that's the last step.  Everything else _ought_ to work once that comes out.

rojopolis:
Thanks @ndmckinley.  I put [this module](https://registry.terraform.io/modules/rojopolis/filestore/google/0.1.0) together until we can get a proper resource type.

ndmckinley:
I am _impressed_.  I promise I'm working on it on my end, but your module should earn some kind of "best workaround" award.  :)

ndmckinley:
Great news, the SDK came together and I'll be back to getting this out the door very soon.

ndmckinley:
Hooray!  This will be in the next release.  :)

PR #2074, opened 2018-09-18
Add Filestore client to config.

ndmckinley:
Another step for #1774.


ndmckinley:
Cool, enough tests pass that I'm persuaded this doesn't break anything.  :)

PR #2038, opened 2018-09-12
Add support for max pods per node to node pool.

ndmckinley:
Fixes #2023.

PR #2077, opened 2018-09-19
Set computed flag for 'max_pods_per_node'

vtorhonen:
After upgrading `terraform-provider-google` to version 1.18 users will notice an unnecessary Terraform plan difference in existing node pool configurations. This is caused by a new configuration parameter `max_pods_per_node` which was introduced in #2038. At the moment it expects end users to explicitly define the parameter value in the configuration or it expects to find it from the existing state. Since neither of these will be found the provider will fallback to default value of 0.

Example:

```
$ terraform init -upgrade
...
- Downloading plugin for provider "google" (1.18.0)...
...

$ terraform plan
  ~ google_container_node_pool.ingress
      max_pods_per_node: "110" => "0"
```

This PR fixes the issue by letting the Google Container API to decide what the value is for existing clusters, which is 110 in this case. If the value is not defined in the Terraform configuration then it won't be stored in the state file either.

### DIFF
--- a/google/resource_container_node_pool.go
+++ b/google/resource_container_node_pool.go
@@ -87,6 +87,7 @@ var schemaNodePool = map[string]*schema.Schema{
 	"max_pods_per_node": &schema.Schema{
 		Type:     schema.TypeInt,
 		Optional: true,
+		Computed: true,
 	},
 
 	"initial_node_count": &schema.Schema{


ndmckinley:
Interesting - might that have been set in the console, outside of Terraform?  I ask because we're not seeing that behavior in our unit tests.  Still, probably a good idea to have it as `Computed`.

vtorhonen:
@ndmckinley at least in my case all node pools were created in Terraform with version 1.17 of the provider. Also I had a completely new environment from early this week. I noticed the plan difference after upgrading the provider, so I guess it should be reproducible if someone wants to try that.

Thanks for merging!

PR #2090, opened 2018-09-20
Check for project deletion when refreshing/deleting project_service.

### LINKED ISSUE #1292: Deleting activated services on a project fails
I'm trying to delete a module which both created a project and activated some services on that project, but it fails because the project gets created first.

### Terraform Version
```
Terraform v0.11.5
+ provider.external v1.0.0
+ provider.google v1.8.0
+ provider.gsuite (unversioned)
+ provider.null v1.0.0
+ provider.random v1.1.0
```

### Affected Resource(s)
- google_project_service

```hcl
resource "google_project" "folder" {
  count           = "${var.folder_id != false ? 1 : 0}"
  name            = "${var.name}"
  project_id      = "${local.final_project_id}"
  folder_id       = "${var.folder_id}"
  billing_account = "${var.billing_account}"
}

locals {
  project_id       = "${element(coalescelist(google_project.folder.*.project_id, google_project.org.*.project_id), 0)}"
}

resource "google_project_service" "project_services" {
  count = "${length(var.activate_apis)}"

  project = "${local.project_id}"
  service = "${element(var.activate_apis,count.index)}"
}
```

### Debug Output
Ping me privately if this is needed.

### Expected Behavior
The services resource is "deleted" before the project (not that it actually needs to do something).

### Actual Behavior
The service enters a zombie state where it constantly tries to get removed even though the project isn't even active any more.

Whenever I try to apply afterwards it's unable to fetch the state:

```
* module.application-xxx-xxx-xxx.module.project-xxx-xxx.google_project_service.project_services: google_project_service.project_services: googleapi: Error 404: Project 'marketo-xxx-xxx-xxx' not found or permission denied., notFound
```

paddycarver:
When reading or deleting the google_project_service resource, first
retrieve the project. If it's not found, or if it's pending deletion,
remove the service, as it's functionally disabled.

Also, add a test to confirm the behaviour.

This should resolve #1292.

### DIFF
--- a/google/resource_google_project_service.go
+++ b/google/resource_google_project_service.go
@@ -67,6 +67,16 @@ func resourceGoogleProjectServiceRead(d *schema.ResourceData, meta interface{}) 
 		return err
 	}
 
+	project, err := config.clientResourceManager.Projects.Get(id.project).Do()
+	if err != nil {
+		return handleNotFoundError(err, d, id.project)
+	}
+	if project.LifecycleState == "DELETE_REQUESTED" {
+		log.Printf("[WARN] Removing %s from state, its project is deleted", id.terraformId())
+		d.SetId("")
+		return nil
+	}
+
 	services, err := getApiServices(id.project, config, map[string]struct{}{})
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


google/resource_google_project_service.go
```diff
@@ -67,6 +67,16 @@ func resourceGoogleProjectServiceRead(d *schema.ResourceData, meta interface{})
 		return err
 	}
 
+	project, err := config.clientResourceManager.Projects.Get(id.project).Do()
+	if err != nil {
+		return handleNotFoundError(err, d, id.project)
+	}
+	if project.LifecycleState == "DELETE_REQUESTED" {
+		log.Printf("[WARN] Removing %s from state, its project is deleted", id.terraformId())
```
chrisst: Having this in delete makes lots of sense, but having it in READ makes me slightly uncomfortable.

Issue #2099, opened 2018-09-22
cloudbuild_trigger missing (ignoredFiles, includedFiles)


sourcec0de:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description
The cloudbuild API supports filtering builds based on files that were changed during a commit.
https://cloud.google.com/cloud-build/docs/api/reference/rest/v1/projects.triggers#BuildTrigger

`includedFiles[]` and `ignoredFiles[]` are arrays of strings

### New or Affected Resource(s)

* google_cloudbuild_trigger

### Potential Terraform Configuration

```tf
resource "google_cloudbuild_trigger" "cloudbuild" {
  project = "allegedly-emerging-tiger"

  ignored_files = [
    "path/from/root/dont-track/**"
  ]

  included_files = [
    "path/from/root/**"
  ]

  trigger_template = {
    branch_name = "master"
    project     = "routinely-included-fawn"
    repo_name   = "super-hello-universe"
  }
}
```

### References

* https://cloud.google.com/cloud-build/docs/api/reference/rest/v1/projects.triggers#BuildTrigger
* https://godoc.org/google.golang.org/api/cloudbuild/v1#BuildTrigger

sourcec0de:
I submitted a pull request to add support for this. Hopefully others find this useful

paulvalla:
Also, this pull request might need to sit in the [terraform-provider-google-beta](https://github.com/terraform-providers/terraform-provider-google-beta) repo since build triggers are still in BETA.

Thank you for your work by the way.

sourcec0de:
@paulvalla Thank you for the feedback and the review. Glad to contribute.

[...]

PR #2100, opened 2018-09-22
added ignored_files and included_files to google_cloudbuild_trigger

sourcec0de:
Please see  https://github.com/terraform-providers/terraform-provider-google/issues/2099 for more details on why this is needed.


chrisst:
We are currently generating both the beta and the ga versions of the google provider from a single codebase. So unfortunately this PR will need to be made against the upstream version of these files: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/provider/terraform/resources/resource_cloudbuild_build_trigger.go This terraform resource file has just been copied upstream so it should be a similar change.

Would you mind creating the PR against the magic modules repo and then @ me and I'll review?

sourcec0de:
@chrisst that's not a problem. I'll have some free time this weekend. I'll move it over on Saturday. 👍

ndmckinley:
Closing in favor of #2553.

PR #2101, opened 2018-09-23
Make schema for args of Cloud Build steps a list

nownabe:
Because strings.Split cannot handle string arguments including space.

For example, "git commit -m \"New Commit\""

Thanks.

cf. https://godoc.org/google.golang.org/api/cloudbuild/v1#BuildStep

### DIFF
--- a/google/resource_cloudbuild_build_trigger.go
+++ b/google/resource_cloudbuild_build_trigger.go
@@ -68,9 +68,10 @@ func resourceCloudBuildTrigger() *schema.Resource {
 										ForceNew: true,
 									},
 									"args": &schema.Schema{
-										Type:     schema.TypeString,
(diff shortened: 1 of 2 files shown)


paddycarver:
Hi! Thanks for the PR. I see what you're going for here, but sadly, we're unable to merge this change until 2.0.0, as it's a breaking change. If you're comfortable waiting for that, that's awesome. If not, an interim solution may be to look into a more flexible parsing approach for the string that's able to take quotes into consideration.

nownabe:
Hi @paddycarver ! Thank you for your reply! :smile: 

I understood. I would like to think about another parsing approach. I have a branch to change another point of cloudbuild trigger, so I'll make a PR of this problem after that.

PR #2137, opened 2018-10-01
Refactor google_sql_database_instance to use Expand functions.

rileykarson:
Refactor `google_sql_database_instance` to use Expand functions in Create. Update will start using them when we make the resource authoritative, this change does not modify the resource functionality.

### DIFF
--- a/google/resource_sql_database_instance.go
+++ b/google/resource_sql_database_instance.go
@@ -455,7 +455,7 @@ func resourceSqlDatabaseInstanceCreate(d *schema.ResourceData, meta interface{})
 	instance := &sqladmin.DatabaseInstance{
 		Name:                 name,
 		Region:               region,
-		Settings:             expandSettings(d.Get("settings").([]interface{}), !isFirstGen(d)),
+		Settings:             expandSqlDatabaseInstanceSettings(d.Get("settings").([]interface{}), !isFirstGen(d)),
 		DatabaseVersion:      d.Get("database_version").(string),
 		MasterInstanceName:   d.Get("master_instance_name").(string),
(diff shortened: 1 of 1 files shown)


google/resource_sql_database_instance.go
```diff
@@ -442,260 +442,36 @@ func resourceSqlDatabaseInstanceCreate(d *schema.ResourceData, meta interface{})
...
-			}
-
-			if vp, okp := _replicaConfiguration["ca_certificate"]; okp {
-				mySqlReplicaConfiguration.CaCertificate = vp.(string)
-			}
-
-			if vp, okp := _replicaConfiguration["client_certificate"]; okp {
-				mySqlReplicaConfiguration.ClientCertificate = vp.(string)
-			}
+	d.Set("name", name)
```
danawillow: This is unusual! Do you know why we do this?

google/resource_sql_database_instance.go
```diff
@@ -745,6 +521,151 @@ func resourceSqlDatabaseInstanceCreate(d *schema.ResourceData, meta interface{})
 	return nil
 }
 
+func expandSettings(configured []interface{}, secondGen bool) *sqladmin.Settings {
```
danawillow: this should be expandSqlDatabaseInstanceSettings since funcs are package-scoped

google/resource_sql_database_instance.go
rileykarson: Originally we had a `d.Set` under the else block, but I pulled it out because this block mirrors a `name_prefix` block such as [in SslCertificate](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_ssl_certificate.go#L319). I don't think the d.Set is necessary at all b/c of the Read, but I figured we can converge on a single shape for automatically generated names.

PR #2144, opened 2018-10-02
Add a documentation entry for CNAME records

roshan:
CNAME records must only have a single rrdatas entry corresponding to the
intended Canonical Name of the record.


roshan:
Thank you!

Issue #2118, opened 2018-09-26
Proposal: move app_engine sub-block to its own resource


paddycarver:
This is a proposal issue to gather feedback and discussion about our contentious `app_engine` design. Currently, App Engine applications are created by adding an `app_engine` sub-block to the `google_project` resource. This has a few benefits:

* It clearly illustrates the tight coupling that App Engine applications have to GCP projects.
* It allows you to delete App Engine applications.

But it comes with drawbacks, too:

* Deleting App Engine applications involves destroying and recreating your entire project.
* Project names aren't immediately available for reuse, so recreating your project probably won't work the way you intend it to, unless you plan for it ahead of time.
* People don't _think_ of App Engine apps as part of the project resource, which leads to surprises when they have an App Engine app and don't include it in their project in Terraform.

What we've heard from feedback is that people don't feel the pros outweigh the cons. As we look into version 2.0.0 of the provider, and have the opportunity to make some changes that we otherwise couldn't, we're reevaluating this decision and seeking input.

My current thought for a design is a `google_app_engine_app` resource, with the same fields as the current block in `google_project`. However, because there's no API to delete App Engine applications, any plan that would involve deleting the app--including `terraform destroy`--would throw an error. To actually delete the app, you'd need to set a special field--possibly called `ack_noop_destroy`, or some other name we'll decide on--on the `google_app_engine_app` resource, acknowledging that you understand that just because Terraform _says_ it's deleted does not mean it's actually deleted. This would allow us to get around the problem of Terraform doing something other than what it says in `plan`, which is an important property of Terraform.

Does this design work for people? Do people have any feedback on this? We'd love any thoughts and experience reports that might influence the design.

Related: #1561 #1973 #1728 #638

morgante:
I've encountered a lot of issues with the current `app_engine` block implementation and support making it less error-prone/surprising.

How would we deal with people including multiple app_engine configs for a single project? That's something which should be caught *before* it hits the API IMO.

> However, because there's no API to delete App Engine applications, any plan that would involve deleting the app--including terraform destroy--would throw an error.

Let's make sure this error is very descriptive and points to a clear doc explaining the tight coupling and how to get around it.

paddycarver:
> How would we deal with people including multiple app_engine configs for a single project? That's something which should be caught before it hits the API IMO.

[...]

PR #2147, opened 2018-10-02
Add google_app_engine_application resource.

paddycarver:
Deprecate the app_engine sub-block of google_project, and create a
google_app_engine_application resource instead. Also, add some tests for
its behaviour, as well as some documentation for it.

Note that this is largely an implementation of the ideas discussed in
 #2118, except we're not using CustomizeDiff to reject deletions without
our special flag set, because CustomizeDiff apparently doesn't run on
Delete. Who knew? This leaves us rejecting the deletion at apply time,
which is less than ideal, but the only other option I see is to silently
not delete the resource, and that's... not ideal, either.

This also stops the app_engine sub-block on google_project from forcing
new when it's removed, and sets it to computed, so users can safely move
from using the sub-block to using the resource without state surgery or
deleting their entire project. This does mean it's impossible to delete
an App Engine application from a sub-block now, but seeing as that was
the same situation before, and we just papered over it by making the
project recreate itself in that situation, and people Were Not Fans of
that, I'm considering that an acceptable casualty.

### DIFF
--- a/google/resource_app_engine_application.go
+++ b/google/resource_app_engine_application.go
@@ -1,0 +1,285 @@
+package google
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	appengine "google.golang.org/api/appengine/v1"
+)
+
+func resourceAppEngineApplication() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAppEngineApplicationCreate,
+		Read:   resourceAppEngineApplicationRead,
+		Update: resourceAppEngineApplicationUpdate,
+		Delete: resourceAppEngineApplicationDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"project": &schema.Schema{
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateProjectID(),
+			},
+			"auth_domain": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"location_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"northamerica-northeast1",
+					"us-central",
+					"us-west2",
+					"us-east1",
+					"us-east4",
+					"southamerica-east1",
+					"europe-west",
+					"europe-west2",
+					"europe-west3",
+					"asia-northeast1",
+					"asia-south1",
+					"australia-southeast1",
+				}, false),
+			},
+			"serving_status": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"UNSPECIFIED",
+					"SERVING",
+					"USER_DISABLED",
+					"SYSTEM_DISABLED",
+				}, false),
+				Computed: true,
+			},
+			"feature_settings": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem:     appEngineApplicationFeatureSettingsResource(),
+			},
+			"ack_delete_noop": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"url_dispatch_rule": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     appEngineApplicationURLDispatchRuleResource(),
+			},
+			"code_bucket": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"default_hostname": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"default_bucket": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"gcr_domain": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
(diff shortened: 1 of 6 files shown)


google/resource_app_engine_application.go
```diff
@@ -0,0 +1,278 @@
...
+			},
+			"gcr_domain": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func appEngineApplicationURLDispatchRuleResource() *schema.Resource {
```
danawillow: nit: any reason these are funcs and not vars?

google/resource_app_engine_application.go
```diff
@@ -0,0 +1,278 @@
...
+
+	return resourceAppEngineApplicationRead(d, meta)
+}
+
+func resourceAppEngineApplicationRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	pid := d.Id()
+
+	app, err := config.clientAppEngine.Apps.Get(pid).Do()
+	if err != nil && !isGoogleApiErrorWithCode(err, 404) {
```
danawillow: I think since it's its own resource now, you can do a regular `handleNotFoundError` call

google/resource_app_engine_application.go
```diff
@@ -0,0 +1,278 @@
...
+	if waitErr != nil {
+		return waitErr
+	}
+	log.Printf("[DEBUG] Updated App Engine App")
+
+	return resourceAppEngineApplicationRead(d, meta)
+}
+
+func resourceAppEngineApplicationDelete(d *schema.ResourceData, meta interface{}) error {
+	log.Println("[DEBUG] App Engine applications cannot be destroyed once created. The project must be deleted to delete the application.")
```
danawillow: it doesn't really matter since from my memory log levels only apply to tf core and not the providers, but if they _did_ make a difference, i would say this should be a WARNING like we do for KmsKeyRing

google/resource_app_engine_application.go
```diff
@@ -0,0 +1,278 @@
...
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+	app, err := expandAppEngineApplication(d, project)
+	if err != nil {
+		return err
+	}
+	app.Id = project
```
danawillow: this line can be removed, it's handled in the expander

google/resource_app_engine_application.go
```diff
@@ -0,0 +1,278 @@
...
+	result.FeatureSettings = featureSettings
+	return result, nil
+}
+
+func expandAppEngineApplicationFeatureSettings(d *schema.ResourceData) (*appengine.FeatureSettings, error) {
+	blocks := d.Get("feature_settings").([]interface{})
+	if len(blocks) < 1 {
+		return nil, nil
+	}
+	if len(blocks) > 1 {
```
danawillow: this is already a plan-time check, so no need to check it again

google/resource_app_engine_application.go
```diff
@@ -0,0 +1,278 @@
...
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateProjectID(),
+			},
+			"auth_domain": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"location_id": &schema.Schema{
```
danawillow: Is this updatable? It's not in the updateMask if so

website/docs/r/app_engine_application.html.markdown
```diff
@@ -0,0 +1,72 @@
+---
+layout: "google"
+page_title: "Google: google_app_engine_application"
+sidebar_current: "docs-google-app-engine-application"
+description: |-
+ Allows management of an App Engine application.
+---
+
+# google\_app_engine_application
```
danawillow: I would assume we either need slashes before all underscores or before no underscores

website/docs/r/app_engine_application.html.markdown
```diff
@@ -0,0 +1,72 @@
...
+~> App Engine applications cannot be deleted once they're created; you have to delete the
+   entire project to delete the application. Terraform will report the application has been
+   successfully deleted; this is a limitation of Terraform, and will go away in the future.
+   Terraform is not able to delete App Engine applications.
+
+## Example Usage
+
+```hcl
+resource "google_project" "my_project" {
+  name = "My Project"
```
danawillow: nit: alignment

website/docs/r/app_engine_application.html.markdown
```diff
@@ -0,0 +1,72 @@
...
+  project_id = "your-project-id"
+  org_id     = "1234567"
+}
+
+resource "google_app_engine_application" "app" {
+  project         = "${google_project.my_project.project_id}"
+  location_id     = "us-central'
```
danawillow: nit: too many spaces after location_id

website/docs/r/app_engine_application.html.markdown
```diff
@@ -0,0 +1,72 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `location_id` - (Required) The [location](https://cloud.google.com/appengine/docs/locations)
+   to serve the app from.
+
+* `auth_domain` - (Optional) The domain to authenticate users with when using App Engine's User API.
+
+* `serving_status` - (Optional) The serving status of the app. Note that this can't be updated at the moment.
```
danawillow: the code makes it look like it can, is this sentence wrong or should the code be updated?

google/resource_app_engine_application.go
paddycarver: I think it's conventional/how I learned how to do it? If I had to guess, it's to keep important mutable state out of the global scope (the Schema is a pointer, someone mutating it will have a big effect on everyone else, probably unintentionally) but I could be very wrong about that.

google/resource_app_engine_application.go
paddycarver: Nope. I guess I'll make it ForceNew, even though that will cause a problem if people do update it? I'm not 100% sure what a good user experience is in this case. I guess I could customize diff to reject this, at least.

paddycarver:
@danawillow ready for round two, I think!

paddycarver:
> Our migration story is going to be import I assume, right?

Yup!

1. Import the app.
2. Add the app to your config.
3. Remove the app_engine sub-block from google_project. Removing it won't ForceNew now.

Issue #2143, opened 2018-10-01
Deprecate things that should be deprecated


danawillow:
* [x] Beta fields
* [x] Instance Group Manager
* [x] `resource_compute_instance.network_interface.access_config.assigned_nat_ip`
* [x] cloudfunctions_function doesn't match the API in an important way (we have trigger_topic and trigger_bucket instead of the more expressive event_trigger, which has more options) and doesn't match the API in a couple less important ways (trigger_http as a string instead of https_trigger as an object with just a string in it; source_archive_bucket and source_archive_object instead of source_archive_url). We should fix all of them.
* [x] bigtable_instance is doing something funky with clusters and hiding them. We should deprecate cluster_id, and expose a cluster sub-block, maybe even a cluster resource.

And probably some more things!

paddycarver:
* [ ] cloudfunctions_function doesn't match the API in an important way (we have trigger_topic and trigger_bucket instead of the more expressive event_trigger, which has more options) and doesn't match the API in a couple less important ways (trigger_http as a string instead of https_trigger as an object with just a string in it; source_archive_bucket and source_archive_object instead of source_archive_url). We should fix all of them.

PR #2155, opened 2018-10-02
Deprecate assigned_nat_ip in compute_instance family

rileykarson:
Part of #2143


paddycarver:
Does this take into account the rationale for the change (hashicorp/terraform#3841)?

paddycarver:
So we did some digging.

* hashicorp/terraform#3841 got this added so that nat_ips could be removed.
* #639 removed that ability by making nat_ip Computed, because nobody understood why we did it in the first place
* That was months ago, nobody seems to be missing the ability to remove nat_ips, so I'm comfortable with dropping the extra field we have for no reason now.

PR #2156, opened 2018-10-03
Deprecate beta fields for (R)IGM.

### LINKED ISSUE #1971: Proposal: deprecate managed instance groups' instance_template field in favor of versions
Right now, we have an `instance_template` field on `google_compute_instance_group_manager`, to specify the template you want to use with the instance group.

However, we also have a `version` sub-block that allows for canary deploys, and more.

It would be fine to have both of them, except the IGM API fills in a default `version` if you don't specify one, meaning we need to make the blocks optional and computed. Which limits our ability to do things like detect drift, and complicates the code for rolling deploys significantly.

This issue is meant as a way to just gather feedback and hear opinions on whether we should deprecate the `instance_template` field, and guide everyone over to using `version` for 2.0.0 of the provider.

I'd love to hear more feedback from other maintainers, as well as from users.

[EDIT] Versions are a beta feature. This is predicated on them achieving GA status prior to 2.0.0; I don't want to push everyone towards a beta API if we can avoid it. So assume for the moment this will only happen if the changes make it to GA in time.

paddycarver:
Update (Regional) Instance Group Managers to deprecate the beta fields
(rolling_update_policy, versions, and auto_healing_policy). Also, rename
RESTART to REPLACE for Instance Group Manager's update_strategy, to be a
bit clearer. This required a DiffSuppress to make sure RESTART and
REPLACE are considered equivalent. In 2.0.0, we should remove RESTART
and drop the DiffSuppressFunc. Sadly, we have no way to do a deprecation
message for RESTART. We'll have to put it in the CHANGELOG. I've already
removed it from the website.

Fixes #1971, replaces #1969.


paddycarver:
So `update_strategy` (AFAIK) predates `update_policy`, and is a Terraform-specific construct. It basically offers a way to have Terraform force the recreation of every instance in a managed instance group when the instance_template changes.

It should not survive once `update_policy` is available in GA, as it becomes pointless.

> Why NONE and REPLACE? Wouldn't OPPORTUNISTIC and PROACTIVE describe them more accurately and in line with the future/beta API? This is a bigger change than just replace => restart of course, but imo would make the field easier to understand and would make it clearer how to translate to update_policy.type in the future.

Honestly, it's to minimize the size of the step people need to take now, given we can't actively deprecate values. I'm not wed to it, but RESTART was actively disingenuous about what it was doing, whereas OPPORTUNISTIC/PROACTIVE feels more a matter of aesthetics than correctness. From what I understand, we're not going to be keeping this alive very long, so I'm hesitant to sink too much effort into moving people onto better values, when the entire field will go away.

> Building on that, what would the downside be to just using update_policy.type with our current behaviour? (along with using a subset of versions in config early! The idea being that the migration to the updater will be seamless) When the feature goes GA, behaviour will change a little because of fun new features like max surge and max unavailable, but I don't think that would be a negative change for any use case nor a breaking change.

It'd mean deprecating update_strategy now, and honestly, I'm just wary of having two varied behaviours with the same name. It's easier to keep track of which is which when we have distinct names for them.

I can see the attraction of deprecating the instance_template now and using version, and deprecating update_strategy so we can use update_policy.type, because it would mean we don't have to carry update_strategy and instance_template through 2.0.0 as deprecated fields, but I think the subtle differences in what they do, the fact that they were already smushed into a resource together with confusing (and broken) behaviour, and the confusion of the deprecation will lead to headaches as people try to sort out exactly what each field does and what they can use. In my opinion, it's simpler to separate it into the "old behaviour" that was never labeled beta, and the "new behaviour" that was always beta and is now in the beta provider.

google/resource_compute_instance_group_manager.go
```diff
@@ -38,9 +38,10 @@ func resourceComputeInstanceGroupManager() *schema.Resource {
 			},
 
 			"version": &schema.Schema{
-				Type:     schema.TypeList,
-				Optional: true,
-				Computed: true,
+				Type:       schema.TypeList,
+				Optional:   true,
+				Computed:   true,
+				Deprecated: "Use the instance_group_manager resource in the google-beta provider instead. See https://terraform.io/docs/providers/google/provider-versions.html for more details.",
```
rileykarson: What do you think about prefixing these fields with a message that the field is beta? "This field is beta, use the instance_group_manager resource..."

Reading the message as it stands suggests that the resource itself is beta to me.

Issue #1672, opened 2018-06-19
Add support for Cloud Bigtable replication


mbrukman:
Currently, support for Cloud Bigtable assumes a 1:1 correspondence between instances and clusters, which was true up until recently, when we launched support for replication (see [blog post](https://cloudplatform.googleblog.com/2018/05/regional-replication-for-cloud-bigtable.html)).

In [the docs](https://github.com/terraform-providers/terraform-provider-google/blob/7e64b3f1c1124f612f784800f0a168a47ca877c5/website/docs/r/bigtable_instance.html.markdown), an instance and a cluster are configured together, and assumes that there is exactly 1 of each `instance` and `cluster` created together:

```hcl
resource "google_bigtable_instance" "instance" {
  name         = "tf-instance"
  cluster_id   = "tf-instance-cluster"
  zone         = "us-central1-b"
  num_nodes    = 3
  storage_type = "HDD"
}
```

What we need to do is enable configuring an instance (an abstract container) together with a list of clusters, each of which has a unique cluster id (so it could actually be a map as well).

In fact, the list of clusters is how it's implemented in the [`CreateInstanceRequest` proto](https://github.com/googleapis/googleapis/blob/master/google/bigtable/admin/v2/bigtable_instance_admin.proto#L187-L207) (note: ignore the comment about "exactly 1 cluster" per instance; that will be fixed soon).

emilymye:
To document, this requires: 
- Updating bigtable go lib to v2
- Changing `google_bigtable_instance` schema to have cluster schema block (possibly a breaking change)

rackonnoiter:
What is the recommended method of updating the bigtable go library? Manually copy it over from [the repo](https://github.com/GoogleCloudPlatform/google-cloud-go/tree/master/bigtable) into the vendor directory?

paddycarver:
@mbrukman can you confirm that the note in the [API docs](https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/projects.instances/create) for creating an instance about `clusters` ("Currently exactly one cluster must be specified.") is incorrect? If so, does the API request not care about the number of clusters at all, or does it need 1 _or more_? And is there any difference between creating the clusters with the instance and creating the clusters using the [clusters API](https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/projects.instances.clusters/create)?

I'd love any clarification so we can get this design right.

mbrukman:
@paddycarver wrote:

[...]

PR #2161, opened 2018-10-03
Move Bigtable config to cluster block.

paddycarver:
Match the bigtable_instance resource to the API structure, deprecating
the inlined config parameters. This is a complicated deprecation,
because we don't want to ForceNew when updating to use the new
undeprecated fields, but we want to ForceNew if they change.

This is achieved using a CustomizeDiff.

### DIFF
--- a/google/resource_bigtable_instance.go
+++ b/google/resource_bigtable_instance.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/hashicorp/terraform/helper/customdiff"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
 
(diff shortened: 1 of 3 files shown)


paddycarver:
> Can you add a test with multiple clusters?
>
> Does using a default zone make sense in the multiple-cluster world? Can we make a Cloud Bigtable with 2 clusters in the same zone? I don't think that makes sense and they should be in multiple zones, so the API probably won't allow it.

I have no idea. Since this wasn't _supposed_ to be enhancement work, just deprecation, how would you feel about me just setting `MaxItems` to 1 for now, and we can sort through the enhancements at a later date? At this point, I just want to get the config in the right shape, I don't really care about supporting multiple clusters _today_. We could do that work some other time.

paddycarver:
I'd be down for removing `getZone` in 2.0.0, but we can't do that _today_.

rileykarson:
Oh yup! That's what I meant by that*

paddycarver:
Cool, I already have #1672 in the 2.0.0 milestone, so we've got an issue to track that enhancement.

PR #2162, opened 2018-10-04
Update compute_vpn_tunnel.html.markdown

Jochen2584:
I think it makes more sense to create a route for the remote network instead of the peer_ip network


rileykarson:
Sure! This resource is autogenerated by [Magic Modules](https://github.com/GoogleCloudPlatform/magic-modules), so any changes made to it will actually need to be made against that repo. If you'd like the GitHub attribution for the change, you can make it by editing [this file](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/templates/terraform/examples/vpn_tunnel_basic.tf.erb#L54).

Alternatively, I can also make the change upstream for you - just let me know if you'd like me to do so.

PR #2178, opened 2018-10-05
Remove beta-features header

rileykarson:
Remove the `beta-features` header now that we have the provider versions page.

Also looks like RIGM doesn't have the same docs as IGM. I added the beta disclaimer to RIGM's beta fields.

Are we missing a deprecation on `update_strategy` = `ROLLING_UPDATE` on purpose? I didn't touch it since I'm not sure what the plan to document rigm/igm differences come 2.0.0 is.

### DIFF
--- a/website/docs/r/container_cluster.html.markdown
+++ b/website/docs/r/container_cluster.html.markdown
@@ -181,7 +181,7 @@ output "cluster_ca_certificate" {
     This property is in beta, and should be used with the terraform-provider-google-beta provider.
     See [Provider Versions](https://terraform.io/docs/provider/google/provider_versions.html) for more details on beta fields.
 
-* `privat_cluster_config` - (Optional) A set of options for creating
+* `private_cluster_config` - (Optional) A set of options for creating
     a private cluster. Structure is documented below.
     This property is in beta, and should be used with the terraform-provider-google-beta provider.
     See [Provider Versions](https://terraform.io/docs/provider/google/provider_versions.html) for more details on beta fields.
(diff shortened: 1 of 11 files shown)


paddycarver:
A lot of links still exist to that header I think.

rileykarson:
That live in the docs, or elsewhere? Afaik I got all of them in there based on what "find in path" told me.

paddycarver:
I need to learn to read before commenting, sorry.

paddycarver:
> Are we missing a deprecation on update_strategy = ROLLING_UPDATE on purpose? I didn't touch it since I'm not sure what the plan to document rigm/igm differences come 2.0.0 is.

May be worth noting, but the field isn't deprecated, just the value. We just need to be clear about it.

I'm also not 100% sure what the best way to document the differences is. I think `update_strategy (GA only)` and `update_policy (Beta only)` may have to do :/

website/docs/r/container_cluster.html.markdown
```diff
@@ -175,18 +175,18 @@ output "cluster_ca_certificate" {
...
 
-* `pod_security_policy_config` - (Optional, [Beta](/docs/providers/google/index.html#beta-features)) Configuration for the
+* `pod_security_policy_config` - (Optional) Configuration for the
     [PodSecurityPolicy](https://cloud.google.com/kubernetes-engine/docs/how-to/pod-security-policies) feature.
     Structure is documented below.
     This property is in beta, and should be used with the terraform-provider-google-beta provider.
     See [Provider Versions](https://terraform.io/docs/provider/google/provider_versions.html) for more details on beta fields.
 
-* `privat_cluster_config` - (Optional, [Beta](/docs/providers/google/index.html#beta-features)) A set of options for creating
+* `privat_cluster_config` - (Optional) A set of options for creating
```
danawillow: while you're here, can you fix the typo on `privat`?

Issue #1947, opened 2018-08-27
GCE instances being created with automatic_restart set to OFF


Susurrus19:
### Terraform Version
Terraform v0.11.8
+ provider.google v1.16.2

### Affected Resource(s)
- google_compute_instance

Affects multiple resources.

### Terraform Configuration Files
```hcl
data "google_compute_image" "memcache_image" {
  project = "terraform-sandbox"
  family  = "${ var.base_disk_image }"
}

data "google_compute_subnetwork" "primary" {
  name    = "primary"
  project = "sharedvpc-prod"
}

resource "google_compute_instance" "memcache-deploy" {
  count            = "${ var.size }"
  name             = "${ var.keyspace }${ format("%03d", count.index + 1) }"
  machine_type     = "${ var.machine_type }"
  min_cpu_platform = "${ var.min_cpu_platform }"
  zone             = "${ count.index % 2 == 0 ? lookup(var.region_zone, 0) : lookup(var.region_zone, 1) }"

  tags = "${ var.tags }"

  boot_disk {
    initialize_params {
      size  = "${ var.disk_size }"
      type  = "pd-standard"
      image = "${ data.google_compute_image.memcache_image.self_link }"
    }
  }

  allow_stopping_for_update = "true"

  lifecycle {
    ignore_changes = ["boot_disk"]
  }

  network_interface {
    subnetwork         = "${ data.google_compute_subnetwork.primary.self_link }"
    access_config      = []
  }
}
```

### Debug Output
Not sure how to do this

### Expected Behavior
Newly created instances are instantiated with "Automatic Restart: On (true)"

### Actual Behavior
Newly created instances are coming up with "Automatic Restart: Off (false)"

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Nothing to mention, mostly vanilla Terraform

### References
None

paddycarver:
You can obtain debug logs by [setting some environment variables](https://www.terraform.io/docs/internals/debugging.html).

I'd love to see some debug logs for this, because 1.16.2 has the default set to `true`. So it's very bizarre that they're getting set to `false`, if you're not specifying any value.

Chupaka:
Huh... It's indeed off (we use Private TFE). I'll try to check locally later...

Chupaka:
With super-minimal config:

```
resource "google_compute_instance" "test" {
  name         = "test-autorestart"
  machine_type = "g1-small"

  boot_disk {
    initialize_params {
[...]

PR #2182, opened 2018-10-07
Default to automatic restart on compute instance to true.

beenanner:
The default for compute instances created in the UI for automatic_restart is true so matching with terraform instance creation to mirror console creation by users.


Chupaka:
Well, if it does what it should be, then it fixes https://github.com/terraform-providers/terraform-provider-google/issues/1947 :)

beenanner:
:-) yep would be cool, but I'll let someone speak up if this does indeed help.

shaharmor:
Any update on this?

danawillow:
Closing this since the default is already true (you can see it two lines below the one that was added in this PR)

Issue #1167, opened 2018-03-08
Make all resources authoritative


paddycarver:
Early versions of the Google provider had resources that tried to respect changes made outside of Terraform, instead of overwriting them. Later versions introduced an `authoritative` field that would turn this behavior off, overwriting changes outside Terraform. Both of these behaviors should be removed, so all resources just overwrite changes outside of Terraform, as that is how Terraform is meant to function. This issue is to collect the list of resources with the previous behaviors so that we can track updating them.

danawillow:
Found another! `google_compute_instance.metadata` is not authoritative (this one might be tricky since I assume metadata may get added out of band by gcp itself (this should be verified though))

rileykarson:
- [ ] maybe `google_cloudiot_registry`
- [ ] `google_compute_project_metadata`
- [ ] `google_storage_object_acl.role_entity`

rileykarson:
- [ ] `google_storage_default_object_acl`

PR #2187, opened 2018-10-08
[Website] Make resource_storage_object_acl authoritative.

rileykarson:
Moved to https://github.com/GoogleCloudPlatform/magic-modules/pull/585

Note: This doesn't include import even though it's possible now - we can add it as a separate change, because it will take an `id` migration to support new-style imports and this PR is already large enough.

`resource_storage_object_acl` now authoritatively manages a storage object's access control list, setting it to a "`private`" ACL when removed. We likely want to block merging this on [adding ObjectAccessControl from MM](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/storage/api.yaml#L321) so that users can use Terraform to non-authoritatively manage their object ACLs. Users can also now move between `predefined_acl` and explicit `role_entity` pairs without recreates.

Instead of requiring users to explicitly define a role/entity pair for the object owner granting an OWNER role every time, Terraform implicitly adds the re pair using a `CustomizeDiff`.

Part of #1167, these changes (specifically `getRoleEntityChange`) will be mirrored to `google_storage_default_object_acl.role_entity`.

### DIFF
--- a/website/docs/r/storage_object_acl.html.markdown
+++ b/website/docs/r/storage_object_acl.html.markdown
@@ -8,10 +8,17 @@ description: |-
 
 # google\_storage\_object\_acl
 
-Creates a new object ACL in Google cloud storage service (GCS). For more information see 
+Authoritatively manages the access control list (ACL) for an object in a Google
+Cloud Storage (GCS) bucket. Removing a `google_storage_object_acl` sets the
+acl to the `private` [predefined ACL](https://cloud.google.com/storage/docs/access-control#predefined-acl).
+
+For more information see 
 [the official documentation](https://cloud.google.com/storage/docs/access-control/lists) 
 and 
 [API](https://cloud.google.com/storage/docs/json_api/v1/objectAccessControls).
+
+-> Want fine-grained control over object ACLs? Use `google_storage_object_access_control` to control individual
+role entity pairs.
 
 ## Example Usage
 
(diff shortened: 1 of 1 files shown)


google/resource_storage_object_acl.go
```diff
@@ -56,196 +102,247 @@ func resourceStorageObjectAclCreate(d *schema.ResourceData, meta interface{}) er
...
+		if err != nil {
+			return fmt.Errorf("Error updating object %s in %s: %v", object, bucket, err)
+		}
 
-	if len(predefined_acl) > 0 {
-		if len(role_entity) > 0 {
-			return fmt.Errorf("Error, you cannot specify both " +
-				"\"predefined_acl\" and \"role_entity\"")
+		return resourceStorageObjectAclRead(d, meta)
+	} else if reMap := d.Get("role_entity").(*schema.Set); len(reMap.List()) > 0 {
```
danawillow: nit: you can just do reMap.Len()

google/resource_storage_object_acl.go
```diff
@@ -56,196 +102,247 @@ func resourceStorageObjectAclCreate(d *schema.ResourceData, meta interface{}) er
...
-					bucket, object, objectAccessControl).Do()
-			}
+	return roleEntities, nil
+}
 
-			// Now we only store the keys that have to be removed
-			delete(old_re_map, pair.Entity)
+// Creates 3 lists of changes we need to make to go from one set of entities to another- which entities need to be created, update, and deleted
+// Not resource specific
+func getRoleEntityChange(old []interface{}, new []interface{}, owner string) ([]*RoleEntity, []*RoleEntity, []*RoleEntity, error) {
```
danawillow: if you make old and new []strings, I think you can get rid of some code that's converting them from strings -> interfaces and back

google/resource_storage_object_acl.go
```diff
@@ -56,196 +102,247 @@ func resourceStorageObjectAclCreate(d *schema.ResourceData, meta interface{}) er
...
+	newEntitiesUsed := make(map[string]struct{})
+	for _, v := range new {
+		res, err := getRoleEntityPair(v.(string))
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("New state has malformed Role/Entity pair: %v", err)
```
danawillow: This check happens a few times around this file- could it be a validatefunc instead?

google/resource_storage_object_acl.go
```diff
@@ -56,196 +102,247 @@ func resourceStorageObjectAclCreate(d *schema.ResourceData, meta interface{}) er
...
+		if _, ok := oldEntitiesUsed[res.Entity]; ok {
+			remove = append(remove, res)
+		}
+	}
 
-		log.Printf("[DEBUG]: removing entity %s", entity)
+	return create, update, remove, nil
+}
+
+// Takes in 3 lists of changes for a Storage Object and makes those changes, ignoring the bucket owner entity - they can't be modified
```
danawillow: technically this function doesn't ignore the bucket owner entity- by the time it reaches here, it's already been ignored.

google/resource_storage_object_acl.go
rileykarson: `ValidateFunc` doesn't work on list types unfortunately. We could probably move the validation to the start of Create/Update, but I'd rather do that separately  and rework `getRoleEntityPair` entirely for all 3-4 resources that use it if we do.

danawillow: I think it works on the elements inside the list, just not on the list as a whole.

rileykarson:
Changes live in https://github.com/GoogleCloudPlatform/magic-modules/pull/585 now

PR #2191, opened 2018-10-09
Fix incorrect link in index.html.markdown

Chupaka:


rileykarson:
Thanks for the catch! It looks like we mixed up `/providers/` with `/provider/` in several places, so I've made #2196 (and https://github.com/GoogleCloudPlatform/magic-modules/pull/538) to catch them all, including your commit in the history for completeness.

PR #2193, opened 2018-10-09
Fix Provider Versions link target

sportebois:
Fixes the missing `www.` in the google provider version link from https://www.terraform.io/docs/providers/google/index.html
Issue ref: [#19028](https://github.com/hashicorp/terraform/issues/19028)

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -8,7 +8,7 @@ description: |-
 
 # Google Cloud Platform Provider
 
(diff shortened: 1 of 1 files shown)


rileykarson:
Thanks for the catch! It looks like we mixed up /providers/ with /provider/ in several places, so I've made #2196 (and GoogleCloudPlatform/magic-modules#538) to catch them all, including your commit in the history for completeness.

PR #2210, opened 2018-10-09
WIP Remove deprecated raw_key field from tests

chrisst:
Fixes test failures which would be caused by https://github.com/terraform-providers/terraform-provider-google/pull/2209


chrisst:
Meh, I'm just going to push this straight to the Magician PR. Tests will fail here without the upstream changes.

Issue #1203, opened 2018-03-15
Remove deprecated attributes


paddycarver:
When we release version 2.0.0, we should make sure to remove any fields and attributes that are deprecated. This issue exists to track those fields and attributes, so we can ensure they're all removed for 2.0.0.

* [x] `google_compute_instance.network_ip`
* [x] `google_compute_instance.create_timeout`
* [x] `google_project_iam_policy.authoritative`
* [x] `google_project_iam_policy.disable_project`
* [x] `google_dataproc_cluster.delete_autogen_bucket`
* [x] `google_compute_image.create_timeout`
* [x] `google_service_account.policy_data`
* [x] `google_compute_disk.disk_encryption_key_raw`
* [x] `google_compute_disk.disk_encryption_key_sha256`
* [x] `google_cloudfunctions_function` (trigger_bucket, trigger_topic)
* [x] `google_bigtable_instance` (cluster convenience fields)
* [x] un-remove `google_compute_network.ipv4_range`
* [x] `google_project.policy_etag`, `google_project.policy_data`
* [x] [stray compute_instance docs](https://github.com/terraform-providers/terraform-provider-google/blob/51e63bfff2d2acba78bdbb35227669b820a4d61e/website/docs/r/compute_instance.html.markdown#L291)
* [x] `google_compute_instance_template.address`
* [x] `google_compute_instance.assigned_nat_ip`
* [x] `google_container_node_pool.name_prefix`
* [x] `google_project.app_engine`
* [x] `google_cloudfunctions_function.retry_on_failure`
* [x] `google_container_cluster.private_cluster`
* [x] `google_endpoints_service.protoc_output`
* [x] `data_source.google_service_account_key.service_account_id`
* [x] `google_organization_iam_custom_role.deleted`
* [x] `google_project_iam_custom_role.deleted` - may only be deprecated at 1.20.0 but that's fine
* [x] `google_compute_snapshot` disk_encryption_key, snapshot_encryption_key
* [x] `google_compute_region_instance_group_manager.update_strategy` at ga (does not exist at beta)
* [x] `google_compute_instance_*.address`

Beta:
* [x] `resourceContainerNodePool.maxPodsPerNode`
* [x] `google_container_cluster` `enable_binary_authorization`, `enable_tpu`, `pod_security_policy_config`, `master_ipv4_cidr_block`
* [x] `iam_compute_subnetwork` resource
* [x] `google_compute_global_forwarding_rule.labels`
[...]

PR #2219, opened 2018-10-10
Remove address and create_timeout from Instance

chrisst:


chrisst:
2 fields for https://github.com/terraform-providers/terraform-provider-google/issues/1203

PR #2224, opened 2018-10-10
Don't allow destroying backend buckets in dataproc

chrisst:
Removing the delete_autogen_bucket property
Buckets will not be deleted when a dataproc resource is destroyed


chrisst:
For https://github.com/terraform-providers/terraform-provider-google/issues/1203

Issue #2220, opened 2018-10-10
Panic on google_sql interface {} is nil, not map[string]interface {}


meyskens:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment
* If an issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to "hashibot", a community member has claimed the issue already.

### Terraform Version

```
Terraform v0.11.8
+ provider.google v1.19.0
+ provider.google-beta v1.19.0
```

### Affected Resource(s)

* google_sql_user
* google_sql_database_instance

### Terraform Configuration Files

```tf
module "lookout_staging_cloudsql" {
  source      = "../../modules/google-cloud/sql" 
  project = "srcd-public-development"
  name = "lookout-staging"
  machine_type = "db-f1-micro"
  user_name = "maartje"
  user_host = "*"
  user_password = "<redacted>"
  db_name = "lookout"
}
```

```tf
resource "google_sql_database_instance" "primary" {
  name             = "${var.name}-primary"
  region           = "${var.region}"
  project          = "${var.project}"
  database_version = "POSTGRES_9_6"

  settings {
    tier              = "${var.machine_type}"
    disk_type         = "${var.disk_type}"
    disk_size         = "${var.disk_size}"
    disk_autoresize   = "${var.disk_autoresize}"
    activation_policy = "ALWAYS"
    availability_type = "REGIONAL"
    replication_type  = "${var.replication_type}"

    backup_configuration {
      enabled = true
    }
  }
}

resource "google_sql_database_instance" "secondary" {
  name                 = "${var.name}-secondary"
  region               = "${var.region}"
  project              = "${var.project}"
  database_version     = "POSTGRES_9_6"
  master_instance_name = "${google_sql_database_instance.primary.name}"

  settings {
    tier              = "${var.machine_type}"
    disk_type         = "${var.disk_type}"
    disk_size         = "${var.disk_size}"
    disk_autoresize   = "${var.disk_autoresize}"
[...]

PR #2229, opened 2018-10-11
Guard against nil empty typelists.

rileykarson:
PR is against master as we can roll this into `1.19.1` to fix some panics. It will reach `2.0.0` on the next merge & I'll PR against tpgb after submitting this one. Also can be fixed in MM as well - I'll work on a PR for that now.

It turns out that an empty `schema.TypeList` holds a `nil` and not a `map[string]interface{}`, leading to panics when we immediately tried to cast the first element. While this only fixes a panic for lists with no `Required` elements, I don't see the harm in guarding those ones too, so I fixed every occurrence I caught in a quick grep.

Fixes https://github.com/terraform-providers/terraform-provider-google/issues/2220

Note that behaviour still isn't great - we'll see a diff from `"0" => "1"`. Configs with an empty block like this are generally but not always wrong (`google_compute_instance.network_interface.#.access_config` is correct), but having a permadiff is better than a crash here.

### DIFF
--- a/google/resource_cloudfunctions_function.go
+++ b/google/resource_cloudfunctions_function.go
@@ -586,28 +586,25 @@ func resourceCloudFunctionsDestroy(d *schema.ResourceData, meta interface{}) err
 }
 
 func expandEventTrigger(configured []interface{}, project string) *cloudfunctions.EventTrigger {
-	if len(configured) == 0 {
+	if len(configured) == 0 || configured[0] == nil {
 		return nil
 	}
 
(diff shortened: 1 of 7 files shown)


rileykarson:
Seeing an error in a test that I think is unrelated but confirming - I'll delete this comment when I'm done.

rileykarson:
Experienced #2230 when I ran out of quota in backend service tests, but it's present on master too and I haven't seen it interactively.

PR #2241, opened 2018-10-11
Change deprecated fields back to 'Removed'

chrisst:
To provide a better user experience we're going to mark the fields as removed
so that users will get a more accurate error message. We will follow up a while
later to actually remove them from schema.


google/resource_compute_instance.go
```diff
@@ -280,6 +288,13 @@ func resourceComputeInstance() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"create_timeout": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  4,
```
rileykarson: I think we want to remove defaults. I'm not sure if they do anything, but let's not take the chance.

google/resource_dataproc_cluster.go
```diff
@@ -88,6 +88,16 @@ func resourceDataprocCluster() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 
+						"delete_autogen_bucket": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+							Removed: "autogenerated buckets are shared by all clusters in the same region, " +
```
rileykarson: This message can probably be updated? We don't need to explain why this is a bad idea anymore, we just need to state you can't.

PR #2275, opened 2018-10-17
Create a link to storage_bucket_acl documentation

tpoindessous:
Create a link to storage_bucket_acl ressource documentation page.

It's easier than scrolling and scrolling or copy text and search it on the page.

Thank you !

### DIFF
--- a/website/docs/r/storage_bucket.html.markdown
+++ b/website/docs/r/storage_bucket.html.markdown
@@ -10,7 +10,7 @@ description: |-
 
 Creates a new bucket in Google cloud storage service (GCS).
 Once a bucket has been created, its location can't be changed.
-[ACLs](https://cloud.google.com/storage/docs/access-control/lists) can be applied using the `google_storage_bucket_acl` resource.
+[ACLs](https://cloud.google.com/storage/docs/access-control/lists) can be applied using the [`google_storage_bucket_acl` resource](./storage_bucket_acl.html).
 For more information see
 [the official documentation](https://cloud.google.com/storage/docs/overview)
 and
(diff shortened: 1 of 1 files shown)


tpoindessous:
Tell me if making link to other pages of ressources is allowed and is wanted by the project.

I find it very useful and I'm very ready to submit another PR, way larger, with a lot of links.

Thanks !

website/docs/r/storage_bucket.html.markdown
```diff
@@ -10,7 +10,7 @@ description: |-
 
 Creates a new bucket in Google cloud storage service (GCS).
 Once a bucket has been created, its location can't be changed.
-[ACLs](https://cloud.google.com/storage/docs/access-control/lists) can be applied using the `google_storage_bucket_acl` resource.
+[ACLs](https://cloud.google.com/storage/docs/access-control/lists) can be applied using the [`google_storage_bucket_acl` resource](./storage_bucket_acl.html).
```
rileykarson: Hmm - it looks like we're kind of all over the place in how we format links to other places in the docs. I (admittedly pretty arbitrarily) favour the ones where we qualify it starting from `/docs/` such as the link to `google_compute_region_backend_service` in [`google_compute_backend_service`](https://github.com/terraform-providers/terraform-provider-google/blob/2.0.0/website/docs/r/compute_backend_service.html.markdown) (apologies - I don't think I can link to line numbers in markdown). Would it be much trouble to edit this change (and future PRs) to use links with that form?

tpoindessous: Thanks ! It will be easier to link to datasource from resource or to resource from datasource.

PR #2282, opened 2018-10-17
add underscores as a valid character for 'name'

sudoforge:
Pretty straightforward. The underscore (`_`) is a valid character in the upstream API, but the regex used in the property's validation doesn't include it. This causes valid `google_spanner_database` names to error out.


sudoforge:
Thanks for the heads up, @emilymye. I'll recreate it over there.

PR #2287, opened 2018-10-19
Add defaultPartitionExpirationMs field to google_bigquery_dataset resource

amatellanes:
Add [`defaultPartitionExpirationMs`](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets) to [google_bigquery_dataset](https://www.terraform.io/docs/providers/google/r/bigquery_dataset.html) resource.


rileykarson:
Huh! Looks like changing the base on this and #2288 added some other commit. Weird! I didn't think that would be a history-altering action, my apologies.

PR #2290, opened 2018-10-19
Add the sql ssl (client) cert resource

benjvi:
Was trying to connect to a SQL instance and felt myself in the need of this resource. I believe this also would complete #635 ?

### DIFF
--- a/google/resource_sql_ssl_cert.go
+++ b/google/resource_sql_ssl_cert.go
@@ -17,32 +17,16 @@ func resourceSqlSslCert() *schema.Resource {
 		SchemaVersion: 1,
 
 		Schema: map[string]*schema.Schema{
-			"instance": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
-			},
-
 			"common_name": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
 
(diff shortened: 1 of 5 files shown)


google/resource_sql_ssl_cert.go
```diff
@@ -0,0 +1,177 @@
...
+func resourceSqlSslCert() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceSqlSslCertCreate,
+		Read:   resourceSqlSslCertRead,
+		Delete: resourceSqlSslCertDelete,
+
+		SchemaVersion: 1,
+
+		Schema: map[string]*schema.Schema{
+			"instance": &schema.Schema{
```
ndmckinley: We have a particular ordering of schema elements we like to maintain.  First, all `Required` fields, then, all `Optional`s (including `Optional` / `Computed`), then all `Computed`.  Within each group, we like to sort alphabetically.

ndmckinley:
Great, the tests pass.  Go ahead and make that simple ordering change, and that'll work for me!  :)

benjvi:
Cool! Just pushed the ordering changes

ndmckinley:
The merge process for this one is going to look a little goofy - sorry for the delay - the issue is to do with how we're generating most files in this repo.  I'm trying to invent a process that will handle the complication and right now, I don't need you to do anything else.  I hope I'll have this in today or tomorrow.

PR #2311, opened 2018-10-23
update contributing since beta split

chrisst:
I will copy this into the beta provider as well once accepted.


.github/CONTRIBUTING.md
```diff
@@ -8,6 +8,16 @@ The following are certain Google Provider-specific things to be aware of when co
 
 We aim to make the Google Provider a good steward of Go practices. See https://github.com/golang/go/wiki/CodeReviewComments for common Go mistakes that you should attempt to avoid.
 
+## Autogenerated Resources
+
+We maintain 2 different versions of the google terraform provider; the [GA (General Audience) provider](https://github.com/terraform-providers/terraform-provider-google) and the [Beta provider](https://github.com/terraform-providers/terraform-provider-google). We are using code generation tool called Magic Modules that uses the same code base to generate both of these providers. Some Terraform resources are defined in `api.yaml` files [located in magic modules](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/products) and are fully generated, whereas some resources are hand written and located in [the provider/terraform/ folder in magic modules](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/provider/terraform/resources). Generated resources will have a prominent header at the top of the file identifying them.
```
paddycarver: It looks like the link for the beta provider is wrong. We should probably link to the Magic Modules repo when introducing it, as well. We should probably also note that the handwritten files are only _temporarily_ handwritten while we convert them to yaml files.

.github/CONTRIBUTING.md
```diff
@@ -8,6 +8,16 @@ The following are certain Google Provider-specific things to be aware of when co
...
+
+For more details on Magic Modules please visit [the readme](https://github.com/GoogleCloudPlatform/magic-modules). For feature requests or bugs regarding those resources, please continue to file [issues](https://github.com/terraform-providers/terraform-provider-google/issues) here. PRs changing those resources will not be accepted.
```
paddycarver: Instead of asking them to file issues "here", let's be specific and say the "terraform-provider-google repositoriy's issue tracker".

.github/CONTRIBUTING.md
rileykarson:
```suggestion
There are two versions of the Google provider; the [`google` provider](https://github.com/terraform-providers/terraform-provider-google)
and the [`google-beta` provider](https://github.com/terraform-providers/terraform-provider-google-beta).
The `google` provider supports GA ([general availability](https://cloud.google.com/terms/launch-stages))
features, and `google-beta` supports Beta features.

Both providers are generated by our code generation tool, [Magic Modules](https://github.com/GoogleCloudPlatform/magic-modules).
Magic Modules uses a shared code base to generate both providers.
Some Terraform resources are autogenerated. these resources are defined by yaml files called `api.yaml` located in the directory corresponding to their product under the
[Magic Modules `products` directory](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/products).
Autogenerated resources have a prominent header at the top of the file
identifying them.

Handwritten resources have been moved to Magic Modules as well, and are located in
[the `third_party` directory](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/third_party/terraform).
Some handwritten resources are the same in both providers; they are stored in
Magic Modules as `.go` files.

Other resources ("compiled" files) are different in each of `google` and `google-beta`;
they'll have the filetype `.go.erb`, and we switch what code is generated based on which
provider is being generated in that run of Magic Modules. 
```

I had some suggestions around reflowing + content, but I could only make a single suggestion because the content was all on one line; feel free to take all or to take out content piecemeal.

.github/CONTRIBUTING.md
```diff
@@ -8,6 +8,16 @@ The following are certain Google Provider-specific things to be aware of when co
...
+
+## Beta vs GA providers
+
+Fields that are only available in beta versions of the google cloud api will need to be added only to the beta provider and excluded from the GA provider. When adding beta features or resources you will need to use templating to exclude them from generating into the ga version. Look for `*.erb` files in [resources](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/provider/terraform/resources) for examples.
```
rileykarson: s/google cloud api/Google Cloud Platform API/

rileykarson: nit: My preference is to refer to the beta provider as `google-beta` and the GA provider as the `google` provider when referring to them.

This means that a sentence like

```
only to the beta provider and excluded from the GA provider
```

becomes:

```
only to the `google-beta` provider, and excluded from the `google` provider
```

rileykarson: This link is outdated after the `third_party` move

chrisst:
History got all jacked up on this branch, moved to https://github.com/terraform-providers/terraform-provider-google/pull/2457/files

PR #2355, opened 2018-10-29
Remove obsolete beta files.

rileykarson:
Remove the files we skip now with https://github.com/GoogleCloudPlatform/magic-modules/pull/631


rileykarson:
`filestore_operation` and the handwritten tests will be blanked out by https://github.com/GoogleCloudPlatform/magic-modules/pull/631. The only tests being removed are generated ones.

Added the missing resources - I must have lost them while working out the binauth issues.

rileykarson:
Also website should keep beta things, shouldn't it? Since we publish from tpg.

PR #2385, opened 2018-11-01
Fix regional cluster example

dkoston:
Update regional cluster example to point cluster to google_container_cluster.regional.name rather than google_container_cluster.primary.name


ndmckinley:
This isn't a problem for me, but can you explain the difference?

dkoston:
It’s just an incorrect reference. primary is defined in the example above for non-regional and regional is defined in this example. If you copied and pasted the code and ran “terraform plan” it would error that primary is not defined or worse if you copied all the examples, you’d be referencing the non-regional cluster when you thought you were referencing the regional one.

ndmckinley:
Oh shoot!  Thanks.  :)  You're right!

PR #2419, opened 2018-11-07
Add documentation on resource.google_container_cluster

Phylu:
When a google_container_cluster is created using terraform the oauth_scope needs to be set to allow authentication of the VM service account on the google services. In addition, the default service account will need the IAM permissions for doing so.

The default service account will have this permissions if the project is created and the service is enabled using the Google Cloud Console. If those are done using terraform, the default account will not have this permissions.

I added some documentation regarding this issue, as it took me quite some time to figure out what was missing.

I hope, I placed this addition at the correct spot in the documentation.

### DIFF
--- a/website/docs/r/container_cluster.html.markdown
+++ b/website/docs/r/container_cluster.html.markdown
@@ -362,6 +362,12 @@ The `node_config` block supports:
   * `monitoring` (`https://www.googleapis.com/auth/monitoring`),
     if `monitoring_service` points to Google
 
+  In addition to the `oauth_scopes`, the `service_account` which is used by the Node VMs
+  needs to have the corresponding IAM permissions to make use of these services. To enable
+  the logging and monitoring this means setting at least the [roles/logging.logWriter](https://cloud.google.com/iam/docs/understanding-roles#stackdriver_logging_roles)
+  and [roles/monitoring.metricWriter](https://cloud.google.com/iam/docs/understanding-roles#stackdriver_monitoring_roles)
+  roles.
+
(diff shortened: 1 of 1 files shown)


website/docs/r/container_cluster.html.markdown
```diff
@@ -362,6 +362,12 @@ The `node_config` block supports:
   * `monitoring` (`https://www.googleapis.com/auth/monitoring`),
     if `monitoring_service` points to Google
 
+  In addition to the `oauth_scopes`, the `service_account` which is used by the Node VMs
```
rileykarson: Correct me if I'm wrong (I'm not too familiar with `oauth_scopes`), but this seems like it has more to do with `service_account`. If we move this block to service account, we could go from:

```markdown
 * `service_account` - (Optional) The service account to be used by the Node VMs.
     If not specified, the "default" service account is used.
```

```markdown
 * `service_account` - (Optional) The service account to be used by the Node VMs.
If not specified, the "default" service account is used.
The service account being used needs the [roles/logging.logWriter](https://cloud.google.com/iam/docs/understanding-roles#stackdriver_logging_roles)
and [roles/monitoring.metricWriter](https://cloud.google.com/iam/docs/understanding-roles#stackdriver_monitoring_roles)
roles.

-> Projects that enable the API with Terraform may need these roles added manually. Projects that enable the API in the Cloud Console should have them added automatically.
```

Phylu: Makes sense to move it there. However it should highlight, what these permissions are needed for. I will update the pull request accordingly.

PR #2457, opened 2018-11-12
Update contributing.md after beta split

chrisst:


.github/CONTRIBUTING.md
```diff
@@ -43,17 +55,13 @@ GOOGLE_ORG
 GOOGLE_BILLING_ACCOUNT
 ```
 
-The only region we support running tests in right now is `us-central1` - some products that are tested here are only available in a few regions, and the only region that all products are available in is `us-central1`.
```
rileykarson: Looks like we lose this + another change, do you mind copying them back in?

PR #2480, opened 2018-11-15
support firestore cloud function triggers

matthewrj:
This PR adds support for firestore event triggers for cloud functions.

The docs for firestore events can be found here: https://cloud.google.com/functions/docs/calling/cloud-firestore

### DIFF
--- a/google/resource_cloudfunctions_function.go
+++ b/google/resource_cloudfunctions_function.go
@@ -530,17 +530,15 @@ func expandEventTrigger(configured []interface{}, project string) *cloudfunction
 		shape = "projects/%s/buckets/%s"
 	case strings.HasPrefix(eventType, "google.pubsub.topic."):
 		shape = "projects/%s/topics/%s"
-	case strings.HasPrefix(eventType, "google.firestore.document."):
-		shape = "projects/%s/databases/(default)/documents/%s"
 	// Legacy style triggers
 	case strings.HasPrefix(eventType, "providers/cloud.storage/eventTypes/"):
 		shape = "projects/%s/buckets/%s"
 	case strings.HasPrefix(eventType, "providers/cloud.pubsub/eventTypes/"):
 		shape = "projects/%s/topics/%s"
 	case strings.HasPrefix(eventType, "providers/cloud.firestore/eventTypes/"):
+		// Firestore doesn't not yet support multiple databases, so "(default)" is assumed.
+		// https://cloud.google.com/functions/docs/calling/cloud-firestore#deploying_your_function
 		shape = "projects/%s/databases/(default)/documents/%s"
-	default:
-		return nil
 	}
 
 	return &cloudfunctions.EventTrigger{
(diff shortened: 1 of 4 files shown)


google/resource_cloudfunctions_function.go
```diff
@@ -550,9 +556,30 @@ func flattenEventTrigger(eventTrigger *cloudfunctions.EventTrigger) []map[string
...
+	case strings.HasPrefix(eventTrigger.EventType, "google.firestore.document."):
+		parts := strings.SplitN(eventTrigger.Resource, "/", 6)
+		resource = parts[len(parts)-1]
+		// Legacy style triggers
+	case strings.HasPrefix(eventTrigger.EventType, "providers/cloud.storage/eventTypes/"):
+		resource = GetResourceNameFromSelfLink(eventTrigger.Resource)
+	case strings.HasPrefix(eventTrigger.EventType, "providers/cloud.pubsub/eventTypes/"):
+		resource = GetResourceNameFromSelfLink(eventTrigger.Resource)
+	case strings.HasPrefix(eventTrigger.EventType, "providers/cloud.firestore/eventTypes/"):
+		parts := strings.SplitN(eventTrigger.Resource, "/", 6)
```
rileykarson: I'm assuming the `6` is significant here, and we can't just use `GetResourceNameFromSelfLink` because we need to store a value like `messages/{messageId}`. Is that correct?

If you don't mind, could you comment including an example of a full `parts` value and explain why we need to deviate here?

google/resource_cloudfunctions_function.go
```diff
@@ -550,9 +556,30 @@ func flattenEventTrigger(eventTrigger *cloudfunctions.EventTrigger) []map[string
...
+		resource = parts[len(parts)-1]
+	default:
```
rileykarson: nit: I don't think we'd want to continue having not matched anything., so we should probably omit the default case or return an error if we have one. (both here and above)

google/resource_cloudfunctions_function.go
```diff
@@ -530,11 +530,17 @@ func expandEventTrigger(configured []interface{}, project string) *cloudfunction
 		shape = "projects/%s/buckets/%s"
 	case strings.HasPrefix(eventType, "google.pubsub.topic."):
 		shape = "projects/%s/topics/%s"
+	case strings.HasPrefix(eventType, "google.firestore.document."):
```
rileykarson: Are there any triggers using this shape? I only see the other form [in the docs](https://cloud.google.com/functions/docs/calling/cloud-firestore)

google/resource_cloudfunctions_function.go
```diff
@@ -530,11 +530,17 @@ func expandEventTrigger(configured []interface{}, project string) *cloudfunction
...
+		shape = "projects/%s/databases/(default)/documents/%s"
 	// Legacy style triggers
 	case strings.HasPrefix(eventType, "providers/cloud.storage/eventTypes/"):
 		shape = "projects/%s/buckets/%s"
 	case strings.HasPrefix(eventType, "providers/cloud.pubsub/eventTypes/"):
 		shape = "projects/%s/topics/%s"
+	case strings.HasPrefix(eventType, "providers/cloud.firestore/eventTypes/"):
+		shape = "projects/%s/databases/(default)/documents/%s"
```
rileykarson: DO you mind adding a note that we expect `(default)` here? Linking to https://cloud.google.com/functions/docs/calling/cloud-firestore#deploying_your_function

google/resource_cloudfunctions_function.go
matthewrj: I added a comment detailing the specifics of this

google/resource_cloudfunctions_function.go
matthewrj: I have omitted the default case, which was the original behaviour

google/resource_cloudfunctions_function.go
matthewrj: You're right, I for some reason thought there were a google.firestore.document series of event types. I have removed this case.

google/resource_cloudfunctions_function.go
matthewrj: I have added a comment.

matthewrj:
@rileykarson I've addressed all the points you raised.

PR #2482, opened 2018-11-15
Add support for data source 'google_iam_role'

manuelfelipe:
Add `google_iam_role` data source to be able to fetch and query for GCP IAM roles permissions.

### DIFF
--- a/google/data_source_google_iam_role.go
+++ b/google/data_source_google_iam_role.go
@@ -1,0 +1,48 @@
+package google
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceGoogleIamRole() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGoogleIamRoleRead,
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"title": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"included_permissions": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"stage": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceGoogleIamRoleRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	roleName := d.Get("name").(string)
+	role, err := config.clientIAM.Roles.Get(roleName).Do()
+	if err != nil {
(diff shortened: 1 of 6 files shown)


google/data_source_google_iam_role.go
```diff
@@ -0,0 +1,48 @@
...
+
+func dataSourceGoogleIamRoleRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	roleName := d.Get("name").(string)
+	role, err := config.clientIAM.Roles.Get(roleName).Do()
+	if err != nil {
+		return handleNotFoundError(err, d, fmt.Sprintf("Error reading IAM Role %s: %s", roleName, err))
+	}
+
+	d.SetId(GetResourceNameFromSelfLink(role.Name))
```
rileykarson: nit: We can set `role.name` w/o running it through `GetResourceNameFromSelfLink`.

website/docs/r/datasource_google_iam_role.html.markdown
```diff
@@ -0,0 +1,37 @@
+---
+layout: "google"
+page_title: "Google: google_iam_role"
+sidebar_current: "docs-google-datasource-folder"
```
rileykarson: This should be `docs-google-datasource-iam-role`

rileykarson: You'll also need to add this datasource to `google.erb` so it shows in the sidebar.

website/docs/r/datasource_google_iam_role.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+  value = "${data.google_iam_role.roleinfo.included_permissions}"
+}
+
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` (Required) - The name of the Role to lookup.
```
rileykarson: Can you include the part about the format from here? https://cloud.google.com/iam/reference/rest/v1/roles/get

website/docs/r/datasource_google_iam_role.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `name` - The resource name of the Role.
```
rileykarson: This is redundant with the argument above

website/docs/r/datasource_google_iam_role.html.markdown
manuelfelipe: also just noticed that I put originally under `/r` and not under `/d` where its suppose to be. fixed now

manuelfelipe:
hey @rileykarson, pushed the changes. ready for a second review

website/docs/r/datasource_google_iam_role.html.markdown
rileykarson: Nice catch!

PR #2490, opened 2018-11-16
vendor composer v1beta1

danawillow:
vendoring the beta API into TPG too makes it so we don't have to have a ton of conditionals in the composer resources. We can just use the beta API all the time, but only send beta fields in the beta provider (this is what we're doing already for basically everything else)


paddycarver:
I want to talk about that approach _eventually_ but I don't think the end result makes enough difference for it to be important right this second.

PR #2533, opened 2018-11-27
Adding Stackdriver Monitoring Api

chrisst:


vendor/vendor.json
```diff
@@ -1542,6 +1542,12 @@
 			"revision": "04bb50b6b83d0e72253821af8cf3252d8e866517",
 			"revisionTime": "2018-11-02T01:57:16Z"
 		},
+		{
+			"checksumSHA1": "5IotLW5dbvT/Ba7YS3wN55AQ2k0=",
+			"path": "google.golang.org/api/monitoring/v3",
+			"revision": "0a71a4356c3f4bcbdd16294c78ca2a31fda36cca",
```
paddycarver: Could we use the same revision we have for the rest of the google.golang.org/api repo packages? Either updating the rest of them to this revision, or using `04bb50b6b83d0e72253821af8cf3252d8e866517` for `google.golang.org/api/monitoring/v3`?

chrisst: Done. Out of curiosity - why?

paddycarver: I'm trying to switch to go modules instead of govendor. And for whatever reason, using different commits of the same repo for packages makes life difficult with go modules. I think it has something to do with the entire repo being considered a module, but I'll confess my go modules knowledge isn't strong enough to make that assertion confidently. I do know it has caused problems for people trying to use the tpg repo as a module, though.

Ideally, I'll be switching us to go modules in the near future, and a `go get -u google.golang.org/api/monitoring/v3` will automatically update `google.golang.org/api` for everything. Ideally. I think. I've read like 70 pages of documentation on go modules, and I still remain a little foggy about the specifics of how they behave in various scenarios.

PR #2549, opened 2018-11-30
Specify that attach_disk can be used multiple times

smaslennikov:
I'm actually still largely unclear if that's a proper usage of the `attach_disk` block, as current docs mention it's a "list" of disks - though, that didn't see to work for me, at least judging by the plan terraform spit out.


rileykarson:
For a bit more context, specifying a block multiple times is a list (or a set) of complex objects in Terraform. I wanted to include documentation from core about that, but it isn't covered on the [configuration syntax page](https://www.terraform.io/docs/configuration/syntax.html). Huh.

We tend to prefer spelling that out how you have, or with a phrase like `This can be specified multiple times` instead of calling complex fields a list to avoid confusion, so thanks for this clarifying change!

smaslennikov:
@chrisst @rileykarson thank you for clarifications and merging this in. Cheers!

PR #2608, opened 2018-12-07
Fix monitoring alert policy example

rolandkool:
Enabled is a require parameter so it makes sense to put it in the example.


danawillow:
Hey @rolandkool, `enabled` defaults to true, so it shouldn't be necessary to include in the config. Are you seeing errors when you don't include it?

rolandkool:
Yes, the example from the page doesn't work if you copy/paste it. Complains about the enabled parameter missing.

danawillow:
Ah I see, it was changed in https://github.com/terraform-providers/terraform-provider-google/pull/2222/files.

danawillow:
Closing this since we're not planning on pushing the website again until 2.0.0 is released, which contains the changes to make the field optional. Thanks though!

PR #2614, opened 2018-12-09
project field is needed on google_comp_firewall.

OriishiTakahiro:
Please update the document about google_compute_firewall.
Change state of the project property from "Optional" to "Required".
I got the following error message if execute "terraform apply" without the project property.
```
Error: Error applying plan:

1 error(s) occurred:

* google_compute_firewall.defualt: 1 error(s) occurred:

* google_compute_firewall.defualt: Invalid value for network: project: required field is not set

```
Terraform and google provider version.
```
$ terraform version                
Terraform v0.11.10
+ provider.google v1.19.1
```

### DIFF
--- a/website/docs/r/compute_firewall.html.markdown
+++ b/website/docs/r/compute_firewall.html.markdown
@@ -103,8 +103,7 @@ The following arguments are supported:
 * `allow` -
   (Optional)
   The list of ALLOW rules specified by this firewall. Each rule
-  specifies a 
-  tocol and port-range tuple that describes a permitted
+  specifies a protocol and port-range tuple that describes a permitted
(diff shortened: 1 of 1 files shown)


Chupaka:
But... "If it is not provided, the provider project is used" still means "Optional", not "Required" :)

Any example of the code that doesn't work for you? All our google_compute_firewall resources don't have 'project' property set.

danawillow:
Yup, I don't think this is the correct approach. Basically all of our projects have "project" as an optional field in order to allow users to default to the provider project. However, if you set neither, then it makes sense for the apply to fail. Firewall isn't unique in this. I'm going to go ahead and close this PR, but let me know if you think I'm wrong :)

PR #2709, opened 2018-12-18
Add an example versioned module using google and google-beta

rileykarson:
Writing modules against tpg/tpgb is hard - introduce an example of how to use beta features with both providers in a single module.

### DIFF
--- a/examples/example-versioned-module/ip/main.tf
+++ b/examples/example-versioned-module/ip/main.tf
@@ -1,0 +1,16 @@
+resource "google_compute_address" "ip_address" {
+  # We'll only generate this block if the value of
+  # has_labels is 0! Effectively an if statement.
+  count = "${1 - local.has_labels}"
+
+  name = "${var.name}"
+}
+
+resource "google_compute_address" "ip_address_beta" {
+  # And this block is only present if we have
+  # at least one entry, effectively an elif.
+  count = "${local.has_labels}"
+
+  name   = "${var.name}"
+  labels = "${var.labels}"
+}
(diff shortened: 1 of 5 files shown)


chrisst:
⚠ you're still against 2.0.0

rileykarson:
Thanks! I'd pushed the change so the commit was properly based but had forgotten to click the UI button.

examples/example-versioned-module/ip/main.tf
```diff
@@ -0,0 +1,16 @@
+resource "google_compute_address" "ip_address" {
+  # We'll only generate this block if the value of
+  # has_labels is 0! Effectively an if statement.
+  count = "${1 - local.has_labels}"
+
+  name = "${var.name}"
+}
+
+resource "google_compute_address" "ip_address_beta" {
```
morgante: @rileykarson I'm confused how this works- shouldn't this resource need to specify the beta provider?

rileykarson: See [this example of invoking it](https://github.com/terraform-providers/terraform-provider-google/pull/2709/files#diff-93659ab1ad57330040aa5f8dcc7254f3R10); while normally aliased provider names are namespaced eg `provider = "google.alias"`, we're able to override the `google` provider in the module [with the providers block](https://github.com/terraform-providers/terraform-provider-google/pull/2709/files#diff-93659ab1ad57330040aa5f8dcc7254f3R20).

Using `provider = "google-beta"` explicitly throughout the module is probably what I would advise now, although a mechanism like this works to share a single config w/ both provider versions at the cost of needing to use a `providers` block to use beta features.

morgante: Yeah, `provider = "google-beta"` is what we've settled on as well. Maybe we should update this example?

rileykarson: I think it's worth keeping this as a distinct example of a module that can support either version; see the README: https://github.com/terraform-providers/terraform-provider-google/pull/2709/files#diff-b56d3749183c558a9af3bf22e0ef72b7R10

PR #2772, opened 2018-12-30
Dataproc encryption params added

vktr-brlv:


anton-opsguru:
Folks what's the ETA to get this merged ?

google/resource_dataproc_cluster.go
```diff
@@ -310,6 +310,21 @@ func resourceDataprocCluster() *schema.Resource {
...
 							},
 						},
+						"encryption_config": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Computed: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key_name": {
```
chrisst: Since this field will only support keys that have been stored in google KMS, can you rename to `kms_key_name` please?

google/resource_dataproc_cluster.go
```diff
@@ -310,6 +310,21 @@ func resourceDataprocCluster() *schema.Resource {
 								},
 							},
 						},
+						"encryption_config": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Computed: true,
```
chrisst: I'm not sure this should be `Computed`. We normally only tag a field as `Computed` if it's going to have something coming back from the api when it hasn't originally been passed in the original request or if it's a completely calculated field in the upstream api.

When I tested the GET api on a simple cluster it didn't show anything coming back for `encryption_config`. Since it doesn't appear to be a calculated or output style field I think somebody should explicitly have this in their config if the cluster has encryption on it. I think that `Optional` but not `Computed` is the right choice here.

google/resource_dataproc_cluster.go
```diff
@@ -310,6 +310,19 @@ func resourceDataprocCluster() *schema.Resource {
...
 						},
+						"encryption_config": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"kms_key_name": {
+										Type:     schema.TypeString,
+										Optional: true,
```
chrisst: Sorry I didn't catch on the first pass, but I don't think `kms_key_name` is optional if the `encryption_block` has been defined. 

In HCL it's possible to have a top level field marked as optional, but have nested fields required. This is the equivalent of saying "encryption_config doesn't have to be specified, but if it is then kms_key_name must be present".

vktr-brlv: agree

vktr-brlv:
Hi @chrisst ,
Are we still missing something?

chrisst:
Hey, sorry for the delay. Usually I ask for integration tests to be updated so that the new feature is tested but the new KMS features aren't very easily tested and require [a lot of boiler plate](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_instance_template_test.go#L1426-L1507) to get a single test to run. I'm working on a feature that should make it a lot easier to test KMS integration with new resources that I should have up shortly.

Whether my new test strategy works or not I'll make sure this gets in for our next release.

vktr-brlv:
Do you need any help with that?
We can share working terraform examples with new params if you need.

chrisst:
Possible test implementation: https://github.com/terraform-providers/terraform-provider-google/pull/2824

chrisst:
This was added upstream and the generated pr was merged in  https://github.com/terraform-providers/terraform-provider-google/pull/2840

PR #2785, opened 2019-01-02
script to determine which tests to run on a pr

danawillow:
This is intended to be used in the magician so it can run tests automatically on PRs, but also works as a standalone script.

Example output:
```
$ go run scripts/affectedtests/affectedtests.go -pr 2771
2019/01/02 14:47:18 Found addition: +++ b/google/resource_compute_route.go
2019/01/02 14:47:18 Found addition: +++ b/google/resource_compute_vpn_tunnel_test.go
2019/01/02 14:47:18 PR contains resource files [resource_compute_route.go]
2019/01/02 14:47:19 Full match is "google_compute_route":                   resourceComputeRoute()
2019/01/02 14:47:19 File resource_compute_route.go matches resource google_compute_route
TestAccComputeRoute_basic
TestAccComputeRoute_defaultInternetGateway
TestAccComputeRoute_hopInstance
TestAccComputeRoute_routeBasicExample
TestAccComputeVpnGateway_targetVpnGatewayBasicExample
TestAccComputeVpnTunnel_vpnTunnelBasicExample
```

Once this is merged, I'll open a PR to start syncing the scripts directory between MM and TPG(B).

### DIFF
--- a/scripts/affectedtests/affectedtests.go
+++ b/scripts/affectedtests/affectedtests.go
@@ -1,0 +1,257 @@
+// affectedtests determines, for a given GitHub PR, which acceptance tests it affects.
+//
+// Example usage: git diff HEAD~ > tmp.diff && go run affectedtests.go -diff tmp.diff
+//
+// It is also possible to get the diff from a PR: go run affectedtests.go -pr 2771
+// However, this mode only reads the changed files from the PR and does not (currently)
+// take into account new resources/tests that might have been added in this PR.
+//
+// This script currently only works for changes to resources.
+// It is a TODO to make it work for changes to tests, data sources, and common utilities.
+// It also currently does not pick up tests that use configs from other files.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"regexp"
+	"runtime"
+	"sort"
+	"strings"
+)
+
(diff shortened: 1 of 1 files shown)


scripts/affectedtests/affectedtests.go
```diff
@@ -0,0 +1,257 @@
...
+		fmt.Println("Exactly one of -pr and -diff must be set")
+		flag.Usage()
+		os.Exit(1)
+	}
+
+	_, filename, _, ok := runtime.Caller(0)
+	if !ok {
+		log.Fatal("Could not get current working directory")
+	}
+	tpgDir := filepath.Clean(filepath.Dir(filename) + "/../..") // we're in scripts/affectedtests, so go up two dirs
```
ndmckinley: If it were me, I'd probably just go up until I find a `.git` directory or file.  What's your stance on that?  Advantage: we can move the file without problems.  Disadvantage: doesn't ever work outside a git repo.

danawillow: I changed it to go up until the directory is called "terraform-provider-*", how does that look?

ndmckinley: Also good - although in this instance you have to handle "what if that doesn't happen" - I guess just if path == '/', right?  Another option: you could use the working directory, `os.Getwd()`.

danawillow: Sure, I added a check for path == '/'. The working directory is what directory we're in when we run it, and I'd rather have more of a dependency on where the script is vs where we're running it from.

PR #2804, opened 2019-01-04
Introduce a google_kms_key_ring data source

joestump:
This introduces a `google_kms_key_ring` data source that allows you to pull in KMS key ring information based on location and name.

I've also made a more generic `testAccDataSourceMatchesResourceCheck` function based on the one that was in the `google_project` data source. Due to the way my GCP organization is structured I've also introduced logic to allow `folder_id` to be used instead of `org_id` in my test.

I've tried running acceptance tests locally, but my GCP organization is structure in a way that's preventing me from running them. I'm PR'ing this now in the hopes that acceptance tests will run on the PR. 

In the meantime, I need to figure out how to fix this error, which I'm 99% sure is unrelated to anything in this PR:

```
Error setting billing account "XXXXXX-YYYYYY-ZZZZZZ" for project "projects/terraform-gf2qlqy307": googleapi: Error 400: Precondition check failed., failedPrecondition
```

### DIFF
--- a/google/data_source_google_kms_key_ring_test.go
+++ b/google/data_source_google_kms_key_ring_test.go
@@ -2,6 +2,7 @@ package google
 
 import (
 	"fmt"
+	"os"
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/acctest"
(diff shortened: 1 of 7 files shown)


google/test_utils.go
```diff
@@ -51,3 +55,33 @@ func toBool(attribute string) (bool, error) {
 	}
 	return strconv.ParseBool(attribute)
 }
+
+func testAccDataSourceMatchesResourceCheck(dataSourceName string, resourceName string, attrsToCheck []string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		ds, ok := s.RootModule().Resources[dataSourceName]
+		if !ok {
+			return fmt.Errorf("root module has no resource called %s", dataSourceName)
```
chrisst: Can you reword the error here and on line 68? I think this should say "module has no data source" instead of "resource" and the second error should be explicit that it can't find the "Resource"

google/data_source_google_kms_key_ring_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccDataSourceGoogleKmsKeyRing_basic(t *testing.T) {
+	projectId := "terraform-" + acctest.RandString(10)
+	projectOrg := getTestOrgFromEnv(t)
+	billingAccount := getTestBillingAccountFromEnv(t)
+	keyRingName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	folderId := os.Getenv("GOOGLE_FOLDER_ID")
```
chrisst: We don't currently have a pattern for running tests in many configurations based on environment variables. If we need test 2 different configurations of a resource we have to create 2 separate tests. So as this test is encoded it will only ever run 1 of these paths in our CI system. Can you remove the logic for the folder test as it will be dead code for our CI system and anybody who doesn't have that variable set?

Normally I would encourage 2 separate tests, but in this case I think testing both project configurations is a more of a `google_project` test than a KMS data source test.

chrisst:
First of all, thank you for adding this! Also I added a quick commit to get the tests to pass but thanks for making that method more general purpose. Let me know when you've addressed the comments and I'll get this into our next build.

chrisst:
Duplicated by https://github.com/terraform-providers/terraform-provider-google/pull/2891

PR #2824, opened 2019-01-07
Kms key bootstrap

chrisst:
Add a method that will bootstrap a KMS keyring

It will either grab a well-known key or create a new one in the project
that is running the tests. This will allow reuse of KMS keys without needing
to create/delete them for all of the new resources that are starting to
support KMS integration.

Goal:
Since it looks like there is a strong push to get KMS support on many different resources in the GCP ecosystem I want to making testing those resources easier. This should help to speed up tests, reduce our project quota burden and simplify tests.

### DIFF
--- a/google/provider_test.go
+++ b/google/provider_test.go
@@ -293,9 +293,6 @@ func bootstrapKMSKey(t *testing.T) bootstrappedKMS {
 		t.FailNow()
 	}
 
-	// TODO chrisst - Set up permissions?
-	// TODO chrisst - Check if tests are failing due to inactive keys and build activation logic.
-
 	return bootstrappedKMS{
 		keyRing,
 		cryptoKey,
--- a/google/provider_test.go
+++ b/google/provider_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/terraform-providers/terraform-provider-random/random"
+	"google.golang.org/api/cloudkms/v1"
 )
 
 var testAccProviders map[string]terraform.ResourceProvider
@@ -178,6 +179,10 @@ func getTestRegionFromEnv() string {
 	return multiEnvSearch(regionEnvVars)
 }
 
+func getTestZoneFromEnv() string {
+	return multiEnvSearch(zoneEnvVars)
+}
+
 func getTestOrgFromEnv(t *testing.T) string {
 	skipIfEnvNotSet(t, orgEnvVars...)
 	return multiEnvSearch(orgEnvVars)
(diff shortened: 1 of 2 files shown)


rileykarson:
Super high level question: Why not just require a `CRYPTO_KEY_SELF_LINK` env var set and fail the test if it's not provided? (Or skip, but I prefer fail so we notice it in CI if it's misconfigured) That seems a lot easier than handling creates and gets and stuff.

chrisst:
Summary of IRL conversation: 
I don't think asking people to have to manually bootstrap their GCP environments with a keyring/key is good and using the ENV variable approach doesn't add a significant advantage over this approach.

google/provider_test.go
```diff
@@ -206,3 +212,100 @@ func multiEnvSearch(ks []string) string {
 	}
 	return ""
 }
+
+/**
+* bootstrapKMSkey will return a KMS key that can be used in tests that are
+* testing KMS integration with other resources.
+*
+* This will either return an existing key or create one if it hasn't been created
+* in the project yet. The motivation is because keyrings don't get deleted and we
```
danawillow: This is a really long sentence, can you clean it up a bit?

google/provider_test.go
```diff
@@ -206,3 +212,100 @@ func multiEnvSearch(ks []string) string {
...
+* don't want a linear growth of disabled keyrings in a project but we also don't want
+* to incur the overhead of creating a new project for each test that needs to use
+* a KMS key.
+**/
+var SharedKeyRing = "tftest-shared-keyring-1"
```
danawillow: Why "-1"?

google/provider_test.go
```diff
@@ -206,3 +212,100 @@ func multiEnvSearch(ks []string) string {
...
+			if err != nil {
+				t.Errorf("Unable to bootstrap KMS key. Cannot create keyRing: %s", err)
+			}
+		} else {
+			t.Errorf("Unable to bootstrap KMS key. Cannot retrieve keyRing: %s", err)
+		}
+	}
+
+	if keyRing == nil {
+		t.Errorf("Unable to bootstrap KMS key. keyRing is nil!")
```
danawillow: These two lines can be replaced with a t.Fatalf

google/provider_test.go
chrisst: It's for versioning. I added it for my own testing to make sure that new keys were getting created correctly when the constant changed. I left it because if we need to change it in future we can version bump to 2. I could check if the key regex support a 'v1' syntax.

google/provider_test.go
```diff
@@ -206,3 +212,100 @@ func multiEnvSearch(ks []string) string {
...
+**/
+var SharedKeyRing = "tftest-shared-keyring-1"
+var SharedCyptoKey = "tftest-shared-key-1"
+
+type bootstrappedKMS struct {
+	*cloudkms.KeyRing
+	*cloudkms.CryptoKey
+}
+
+func bootstrapKMSKey(t *testing.T) bootstrappedKMS {
```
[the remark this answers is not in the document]
chrisst: I was thinking it might be better off in a kms specific file or just a 'bootstrapping' file if we decide to use the same approach for project reuse

chrisst:
Duplicated by https://github.com/terraform-providers/terraform-provider-google/pull/2837

PR #2828, opened 2019-01-08
No forced update on ignored/included files for cloud build

naseemkullah:


rileykarson:
Hey @naseemkullah! Since this is an autogenerated file using [Magic Modules](https://github.com/GoogleCloudPlatform/magic-modules), I'll probably supersede this PR when performing some other updates to our definitions there. I expect this will happen before our next release- since I'm not getting to it immediately though, you can remove the `input` lines from both definitions [in this yaml file](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/cloudbuild/api.yaml#L80) in a PR to accomplish this as well.

naseemkullah:
Will do, thanks @rileykarson!

naseemkullah:
https://github.com/GoogleCloudPlatform/magic-modules/pull/1244 created

PR #2829, opened 2019-01-08
Add detail to google_project_iam_member import example

josiahp:
The previous import example for objects of type google_project_iam_member generates an error (foo@example.com is not a valid resource Id). In order to clarify, I've added additional information for the case of importing service accounts.

### DIFF
--- a/website/docs/r/google_project_iam.html.markdown
+++ b/website/docs/r/google_project_iam.html.markdown
@@ -114,6 +114,12 @@ IAM resources can be imported using the `project_id`, role, and account.
 $ terraform import google_project_iam_policy.my_project your-project-id
 
 $ terraform import google_project_iam_binding.my_project "your-project-id roles/viewer"
+```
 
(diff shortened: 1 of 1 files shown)


emilymye:
@josiahp Thanks for contributing a PR - we really appreciate it! There are a couple of reasons why I'm going to close it now, apologies! 

First, we've recently moved all our code into a separate repo, Magic Modules (MM): https://github.com/GoogleCloudPlatform/magic-modules. We're using this repo to autogenerate resources and versioned providers, and so all code changes now need to be done in that repo. We'd have to move this over and close it out under most circumstances.

However, rather than have you do that work, I've just ended up creating [a larger PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/1240) because it turns out this is actually a problem within most of the other IAM member import documentation. I might as well fix it while we're here :) Thanks for bringing this to our attention!

PR #2908, opened 2019-01-22
Update gometalinter for modules, add staticcheck

rileykarson:
This isn't working locally, but it matches the changes made in https://github.com/terraform-providers/terraform-provider-azurerm/pull/2722 so maybe something freaky is going on and it will only work in CI?

Edit: This only works in CI (well, at least doesn't work locally for *me*) but the same is true in the Azure provider. Most of those rules can be turned back on provided we make the corresponding code changes, but let's bring back `unused` for now. I'll dupe over to tpgb on approval.


staticcheck.conf
```diff
@@ -0,0 +1,11 @@
+checks = ["all", "-S1002", "-S1007", "-S1008", "-S1009", "-S1019", "-S1021", "-S1025", "-ST1000", "-ST1003", "-ST1005", "-ST1017", "-SA4006", "-SA4010", -"SA6000", "-SA6005"]
```
dominikh: Typo, `-"SA6000"` should be `"-SA6000"`

rileykarson: Thanks!

rileykarson:
@paddycarver: Re-requesting because this changed a bunch based on reading the AWS / Azure providers' configs as well.

I took the output from https://github.com/terraform-providers/terraform-provider-azurerm/pull/2787 and added it in our Makefile, alongside some other small changes.

I found that modules didn't work well with gometalinter, so I explicitly disabled them on every command. I have `GO111MODULE=on` set locally, and it caused me some... weird dependency issues.

PR #2932, opened 2019-01-25
Fix vendor directory.

ndmckinley:
In go 1.11 and forward until go 1.11.4, checksums were calculated incorrectly if a module contained symlinks.  In go 1.11.4, the go team made a breaking change that requires people using go modules to delete their vendor directory and recreate it.  It also requires that everyone working on a project be either pre-1.11.4 or post 1.11.4.  The magician is on 1.11.5, but the rest of us are probably on 1.11 or 1.11.1.  Whoops!  I updated go and did the go vendor again.  Here's the result.  Make sure to update your go version!

Caused by https://github.com/golang/go/issues/27093.


ndmckinley:
Nice, thanks govendor.  I'll figure it out.

ndmckinley:
Oh, haha, travis is on go 1.11.0.

PR #2934, opened 2019-01-25
Update go versions used in development

rileykarson:
Adding `.x` to Travis means use the latest minor version according to its docs.

I didn't update the "Requirements" header because this is just because of Go Modules; see https://github.com/terraform-providers/terraform-provider-google/pull/2932 for why. Since it's just a development change, it doesn't impact _using_ the provider.

Although, out of curiosity- is that header for development? Or is there a real requirement to have Go installed to use Terraform providers?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
 - docker
 language: go
 go:
-- "1.11"
+- "1.11.x"
 
 install:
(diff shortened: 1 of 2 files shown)


paddycarver:
Go is only required for dev. Providers ship as binaries, so they don't need Go installed to be used.

PR #2950, opened 2019-01-28
feat(google/resource_compute_instance.go): adding Hostname property

afshinm:
Hello maintainers :wave: 

This is my first contribution to Terraform so please bear with me!

I'm sending this PR to add `Hostname` to Google Compute Instance, which was recently added: https://cloud.google.com/compute/docs/instances/custom-hostname-vm

This property is:

  - Optional
  - String
  - Read-only (cannot be updated once the instance is created)
  - Only valid if `[a-z]([-a-z0-9]*[a-z0-9])?`

Please let me know if you want me to add any more tails and/or tests to this PR.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -508,6 +508,13 @@ func resourceComputeInstance() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"hostname": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
 			"cpu_platform": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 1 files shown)


emilymye:
Hi @afshinm! I'm actually working on this exact task in a separate PR (that is not out yet). Because we do all our changes now against [the magic-modules  repo](https://github.com/GoogleCloudPlatform/magic-modules) and our external contributing process is a little complicated because of it, it will probably be less work in the end to just wait for my work to get in :) Thanks for contributing though!

emilymye:
Sorry for the confusion, please see https://github.com/GoogleCloudPlatform/magic-modules/pull/1316 for the upstream PR

PR #2954, opened 2019-01-29
Updated BigQuery dataset locations

evanhfox:
Updated bigquery dataset regional locations validation as per GCP BigQuery documentation: https://cloud.google.com/bigquery/docs/locations


evanhfox:
Hi Riley,

I think removing ValidateFunc all together makes the most sense as keeping
up to date with GCP Service availability is going to be a perpetual
catch-up process.

Evan

rileykarson:
Superseded by #3406

Issue #1719, opened 2018-06-29
google_compute_instance should allow to specify instance state


roman8422:
Hi,

There are several long lasting issues open with the same functionality request for aws_instance.
It would be very useful to have such option for google_compute_instance as well.

### Terraform Version
Terraform v0.11.7

### Affected Resource(s)
- google_compute_instance

### Terraform Configuration Files
```hcl
resource "google_compute_instance" "default" {
  name         = "test"
  machine_type = "f1-micro"
  zone         = "us-central1-a"
# state        = "running"
# state        = "stopped"

  boot_disk {
    initialize_params {
      image = "debian-cloud/debian-8"
    }
  }

  network_interface {
    network = "default"
  }
}
```

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- https://github.com/terraform-providers/terraform-provider-aws/issues/22
- https://github.com/terraform-providers/terraform-provider-aws/issues/190

paddycarver:
Hi, thanks for the feature request! As you noted, this is a common feature request for other providers, and I think the current state of things is best summed up by [Mitchell's response](https://github.com/hashicorp/terraform/issues/1579#issuecomment-95117648) to the feature request for AWS: managing the state of an instance is more complicated than managing the presence of an instance (how do I go from "terminated" to "stopped"? What do we even do with "suspended"?)  I think upstream Terraform was thinking on some of this a bit ago, but I don't know that a decision was ever reached.

hounded:
This is so needed! I tried deploying the new gitlab chart to a cluster running  2 `n1-standard-1` ($50 a month) and the resource hungry beast wouldn't run, it did run on 2 `n1-standard-4` but I can't foot a $300 a month bill for my devops when I'm a solo developer, would be so handy to be able to shutdown node pools and then bring them up when needed.

pondohva:
@paddycarver I'm not really sure why do we have "STOPPED" status in compute client. https://cloud.google.com/compute/docs/instances/instance-life-cycle describes TERMINATED state as state for stopped instance. I've created PR using life cycle info from this doc.

upodroid:
@rileykarson We can close this. It was fixed in #4797

PR #2956, opened 2019-01-29
ability to change instance_state

agadelshin:
this PR add a feature requested in #1719

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -131,7 +131,7 @@ func resourceComputeInstance() *schema.Resource {
 			"instance_state": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				Default:      "running",
+				Default:      "RUNNING",
 				ValidateFunc: validation.StringInSlice([]string{"stopped", "running"}, false),
 			},
 
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -132,7 +132,7 @@ func resourceComputeInstance() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Default:      "RUNNING",
-				ValidateFunc: validation.StringInSlice([]string{"stopped", "running"}, false),
+				ValidateFunc: validation.StringInSlice([]string{"RUNNING", "TERMINATED"}, false),
 			},
 
 			"name": {
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/customdiff"
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/mitchellh/hashstructure"
@@ -125,6 +126,13 @@ func resourceComputeInstance() *schema.Resource {
 			"machine_type": {
 				Type:     schema.TypeString,
 				Required: true,
+			},
+
+			"instance_state": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "running",
+				ValidateFunc: validation.StringInSlice([]string{"stopped", "running"}, false),
 			},
 
 			"name": {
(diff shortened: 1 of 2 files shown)


pondohva:
Should I do something to move this forward?

google/resource_compute_instance.go
```diff
@@ -762,6 +770,15 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("can_ip_forward", instance.CanIpForward)
 	d.Set("machine_type", GetResourceNameFromSelfLink(instance.MachineType))
 
+	// PROVISIONING, STAGING, RUNNING, STOPPING, STOPPED,
+	//	// SUSPENDING, SUSPENDED, and TERMINATED
+	switch instance.Status {
```
rileykarson: Instead of flattening several states into either `RUNNING` or `STOPPED`, let's see if we can approach this from a different direction.

Right now, we use an `Operation` from the GCP API to tell if the instance has been provisioned. That doesn't tell us the status of the logical instance, just of the GCP resource. What if we also waited in `Create` for the instance to be `RUNNING` if the field is set to `RUNNING`, and stop it if it's set to `TERMINATED`.

Then, in `Read` we can wait to enter a terminal value and d.Set the exact value.

Finally, in `Update` if we're in a different state than expected we need to start / stop depending on which terminal state we want to be in. There's some additional nuance around this section; https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_instance.go#L1253

We'll also need to ensure that `TERMINATED ` instances can be deleted.

The intent of this change would be that we're explicitly entering a terminal state in each of the CRUD methods, so Terraform will always be in a well-known place, and we won't need to worry about how we can transition from `PROVISIONING` -> `TERMINATED` for example.

google/resource_compute_instance.go
```diff
@@ -127,6 +128,13 @@ func resourceComputeInstance() *schema.Resource {
 				Required: true,
 			},
 
+			"instance_state": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "running",
+				ValidateFunc: validation.StringInSlice([]string{"stopped", "running"}, false),
```
```suggestion
				ValidateFunc: validation.StringInSlice([]string{"RUNNING", "TERMINATED"}, false),
```

google/resource_compute_instance.go
```diff
@@ -127,6 +128,13 @@ func resourceComputeInstance() *schema.Resource {
 				Required: true,
 			},
 
+			"instance_state": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "running",
```
```suggestion
				Default:      "RUNNING",
```

google/resource_compute_instance.go
```diff
@@ -964,6 +981,55 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
 		d.SetPartial("tags")
 	}
 
+	if d.HasChange("instance_state") {
```
rileykarson: Stated above, but we'll need to watch out for changes to the instance state in this method; https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_instance.go#L1253

google/resource_compute_instance.go
```diff
@@ -964,6 +981,55 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
...
+		if d.Get("instance_state") == "running" {
+			_, err := config.clientCompute.Instances.Start(project,
+				zone, d.Id()).Do()
+			if err != nil {
+				return fmt.Errorf(
+					"Error starting an instance (%s): %s", d.Id(), err)
+			}
+
+			stateConf := &resource.StateChangeConf{
```
rileykarson: We can probably extract this whole chunk into a function like `waitForTerminalState` and wait for either terminal status?

google/resource_compute_instance_test.go
```diff
@@ -1043,6 +1043,37 @@ func TestAccComputeInstance_secondaryAliasIpRange(t *testing.T) {
...
+			{
+				Config: testAccComputeInstance_running(instanceName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeInstanceExists(
+						"google_compute_instance.foobar", &instance),
+					testAccCheckComputeInstanceRunning(&instance),
+				),
+			},
+			{
+				Config: testAccComputeInstance_stopped(instanceName),
```
rileykarson: Good start! We'll need to test a handful of cases though.



- [x] You've covered stopping an existing instance
- [x] And deleting a stopped instance
- [ ] Updating a field on a stopped instance
- [ ] Also updating a field that requires stopping, like `machine_type`
- [ ] And doing that at the same time as stopping + starting for each of those
- [ ] Creating a stopped instance

pondohva: Thank you for useful suggestions and review. I've got you point, looks reasonable. I'm going to start with suggested test cases and will ask you for review!

rileykarson:
Superseded with #4797

Issue #2050, opened 2018-09-13
Add `force_destroy` field for BigQuery dataset


ocervell:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

Currently, we have to write our own custom script to essentially empty bigquery tables before deleting the dataset.

The shell script looks like:
```
PROJECT_ID=$1
DATASET_NAME=$2
for i in $(bq ls $PROJECT_ID:$DATASET_NAME | sed 1,2d | awk "{print \$1}");
do
  bq rm -ft $PROJECT_ID:$DATASET_NAME.$i;
done
```

and the resource definition:

```
# BigQuery dataset
resource "google_bigquery_dataset" "dataset" {
  count      = "${local.destination_type == "bigquery" ? 1 : 0}"
  dataset_id = "${local.destination_name}"
  project    = "${local.destination_project}"

  # Delete all tables in dataset on destroy.
  # This is required because a dataset cannot be deleted if it contains any data.
  provisioner "local-exec" {
    when    = "destroy"
    command = "sh ${path.module}/scripts/delete-bq-tables.sh ${local.destination_project} ${local.destination_name}"
  }

  depends_on = [
    "google_project_service.enable_destination_api",
  ]
}
```

Similarly to the `google_storage_bucket` resource, we should have a `force_destroy` flag that we can set to delete all tables on destroy.

### New or Affected Resource(s)

* google_bigquery_dataset

### Potential Terraform Configuration

```tf
resource "google_bigquery_dataset" "dataset" {
  dataset_id = "${var.dataset_id}"
  project    = "${var.project_id}"
  force_destroy = true
}
```

rileykarson:
Fixed by https://github.com/terraform-providers/terraform-provider-google/pull/2986, `delete_contents_on_destroy` is available in `2.0.0`.

PR #2986, opened 2019-02-02
Add force_destroy option to bigquery dataset

zhilingc:
Added the `force_destroy` option for bigquery datasets, as described in #2050.

### DIFF
--- a/google/resource_bigquery_dataset.go
+++ b/google/resource_bigquery_dataset.go
@@ -184,6 +184,9 @@ func resourceBigQueryDataset() *schema.Resource {
 				},
 			},
 
+			//Delete Contents on Destroy: [Optional] If True, delete all the tables in the dataset.
+			// If False and the dataset contains tables, the request will fail.
+			// Default is False.
 			"delete_contents_on_destroy": {
 				Type:     schema.TypeBool,
 				Optional: true,
--- a/google/resource_bigquery_dataset.go
+++ b/google/resource_bigquery_dataset.go
@@ -188,9 +188,10 @@ func resourceBigQueryDataset() *schema.Resource {
 			// If False and the dataset contains tables, the request will fail.
 			// Default is False.
 			"delete_contents_on_destroy": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				Default:  false,
+				Type:             schema.TypeBool,
+				Optional:         true,
+				Default:          false,
+				DiffSuppressFunc: emptyOrDefaultStringSuppress("false"),
 			},
 
 			// SelfLink: [Output-only] A URL that can be used to access the resource
--- a/google/resource_bigquery_dataset.go
+++ b/google/resource_bigquery_dataset.go
@@ -184,6 +184,12 @@ func resourceBigQueryDataset() *schema.Resource {
 				},
 			},
 
+			"force_delete": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
 			// SelfLink: [Output-only] A URL that can be used to access the resource
 			// again. You can use this URL in Get or Update requests to the
 			// resource.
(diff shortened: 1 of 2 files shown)


google/resource_bigquery_dataset.go
```diff
@@ -184,6 +184,12 @@ func resourceBigQueryDataset() *schema.Resource {
 				},
 			},
 
+			"force_destroy": {
```
rileykarson: I'm leaning slightly towards copying the naming the API uses, `delete_contents` instead of calling the field `force_destroy`. Or maybe `delete_contents_on_destroy`; it's a little verbose, but describes the behaviour well. What do you think?

zhilingc: `delete_contents_on_destroy` sounds great to me actually. I'll make the change.

zhilingc:
@rileykarson 
I've added the items you've listed! For the acceptance tests, I excluded the field from the ImportState verification because it's not actually part of the infrastructure - let me know if this is a wrong assumption and I'll correct it asap

zhilingc:
@rileykarson Added the documentation and the tests. I've excluded the field from the ImportState verification step because I don't think it's part of the import state at all, let me know if this is an incorrect assumption to make.

google/resource_bigquery_dataset_test.go
```diff
@@ -214,6 +272,24 @@ resource "google_bigquery_dataset" "test" {
...
 
+func testAccBigQueryDatasetDeleteContents(datasetID string) string {
+	return fmt.Sprintf(`
+resource "google_bigquery_dataset" "contents_test" {
+  dataset_id                  = "%s"
+  friendly_name               = "foo"
+  description                 = "This is a foo description"
+  location                    = "EU"
+  default_partition_expiration_ms = 3600000
+	default_table_expiration_ms = 3600000
```
rileykarson: nit: this line + below are tabs, but should be spaces for consistency

google/resource_bigquery_dataset_test.go
```diff
@@ -205,7 +263,7 @@ resource "google_bigquery_dataset" "test" {
   description                 = "This is a bar description"
   location                    = "EU"
   default_partition_expiration_ms = 7200000
-  default_table_expiration_ms = 7200000
+	default_table_expiration_ms = 7200000
```
rileykarson: nit: this line is a tab, but should be spaces for consistency with other test configs

google/resource_bigquery_dataset.go
```diff
@@ -184,6 +184,15 @@ func resourceBigQueryDataset() *schema.Resource {
 				},
 			},
 
+			// Delete Contents on Destroy: [Optional] If True, delete all the tables in the dataset.
+			// If False and the dataset contains tables, the request will fail.
+			// Default is False.
+			"delete_contents_on_destroy": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
```
rileykarson: I'm inclined to remove the default; as-is, Terraform will report a diff for every existing user of bigquery dataset, even if it's a no-op. We can fix that by either removing the default (it will actually still default to false, it just won't be recorded in state) & presumably all the `ImportStateVerify` checks but the one where the value is explicitly set.

Alternatively, we could add a `DiffSuppressFunc` either similar to or reusing https://github.com/terraform-providers/terraform-provider-google/blob/69e445bd59301e491e17d3490babc6e2262ebf08/google/utils.go#L154 (Judging by the `SchemaDiffSuppressFunc`, Terraform treats booleans as strings inside that function)

zhilingc: I think I like the latter option better, it makes for more readable code :) Thanks! I didn't know of the existence of this function haha.

google/resource_bigquery_dataset_test.go
zhilingc: whoops. fixed.

zhilingc:
> Looking great! Just one more comment & a couple spacing nitpicks.

Done. Thanks so much for the quick reviews!

zhilingc:
awesome 👍

PR #2996, opened 2019-02-05
Updated google_container_cluster

night-gold:
Added precision on the valid subnet for 'master_ipv4_cidr_block' in 'private_cluster_config' block.


rileykarson:
Sorry for the delay on our end @night-gold- do you have a source for that value needing to be `/28`? It doesn't look like that's reflected on the GKE docs.

night-gold:
I tried it with terraform and it was in the error message.

In the deprecated way of doing it, that has been removed, it was already written that it should be a /28 CIDR for it to work.

mrparkers:
I can also confirm the `/28` limitation for masters in private clusters.  Here are the supporting docs: https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#limitations

inductor:
@rileykarson ping

PR #3017, opened 2019-02-08
Add Timeouts configuration to cloud functions

cyrushanlon:
I've been having an issue where a timeout would occur after 4 minutes when deploying my cloud function, the timeouts are now passed through to the operation wait rather than using the hard coded value.

### DIFF
--- a/google/cloudfunctions_operation.go
+++ b/google/cloudfunctions_operation.go
@@ -18,12 +18,12 @@ func (w *CloudFunctionsOperationWaiter) QueryOp() (interface{}, error) {
 	return w.Service.Operations.Get(w.Op.Name).Do()
 }
 
-func cloudFunctionsOperationWait(service *cloudfunctions.Service, op *cloudfunctions.Operation, activity string) error {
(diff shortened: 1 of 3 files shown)


emilymye:
Hi @cyrushanlon! Thank you for contributing. We've actually recently started editing code in https://github.com/GoogleCloudPlatform/magic-modules, which autogenerates PRs against the terraform-provider-google repo. This allows us to autogenerate more resources and different versions of the provider. 

If you really want contributer credit, you are welcome to move the PR over there -this one is pretty straightforward since this resource is not being autogenerated, so it'd be 1:1 changes to the corresponding files ((as an example,  https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/resources/resource_cloudfunctions_function.go)

However, our external contribution process is a little slow right now and has some extra manual steps that need to be run per push (including getting up-to-date with master), so generally we just end up just doing the necessary changes ourselves. If you don't mind, I can go ahead and make the changes for you, or you can just open the PR against that repo. Please let me know which one you'd prefer!

cyrushanlon:
Ah my bad, thanks for letting me know, I'll move the changes over when I get chance :)

PR #3032, opened 2019-02-11
Update CONTRIBUTING to be more clear

rileykarson:
/cc @rambleraptor


.github/CONTRIBUTING.md
```diff
@@ -31,27 +34,33 @@ If you're developing against Magic Modules, vendoring changes is done automatica
...
 
-To run a specific test, use a command such as:
 ```
 make testacc TEST=./google TESTARGS='-run=TestAccContainerNodePool_basic'
 ```
 
 The `TESTARGS` variable is regexp-like, so multiple tests can be run in parallel by specifying a common substring of those tests (for example, `TestAccContainerNodePool` to run all node pool tests).
 
-To run all tests, you can simply omit the `TESTARGS` argument - but please keep in mind that that is quite a few tests and will take quite a long time and create some fairly expensive resources.  It usually is not advisable to run all tests.
+To run all tests, you can omit the `TESTARGS` argument - but please keep in mind that that is quite a few tests and will take quite a long time and create some fairly expensive resources.  It usually is not advisable to run all tests.
```
rambleraptor: I'd make this paragraph a little scarier and maybe bold 'fairly expensive resources.'

The tests spin up so much stuff and it can cost a lot of money.

rileykarson: I think this is fine because every example includes a `TESTARGS` and anyone reading this will pick up on the fact that it will run _every_ test (and they take hours to run so you'd notice pretty quickly that you did something Not Good)

.github/CONTRIBUTING.md
```diff
@@ -31,27 +34,33 @@ If you're developing against Magic Modules, vendoring changes is done automatica
 ### Running Tests
 
 The following environment variables must be set in order to run tests:
+
 ```
 GOOGLE_PROJECT
-GOOGLE_CREDENTIALS or GOOGLE_USE_DEFAULT_CREDENTIALS
+GOOGLE_CREDENTIALS
```
rambleraptor: This might be obvious to TF users, but can you specify that this should be set to a JSON service account creds file.

rileykarson: The primary place for configuration information is the provider reference and I'm not really interested in homing that information here; https://www.terraform.io/docs/providers/google/provider_reference.html#configuration-reference

This is actually a subset of the fields that work and I'm fairly certain any set of provider env vars actually works, so once someone's read that it will be clear what the values they need are.

Issue #3014, opened 2019-02-07
Switch from gometalinter to golang-cli


rileykarson:

paddycarver:
This has been merged.

PR #3049, opened 2019-02-14
Switch to using golangci-lint instead of go-metalinter

paddycarver:
Following in the footsteps of our friends working on the AWS provider in terraform-providers/terraform-provider-aws#7457 and prompted by [go-metalinter's mainainer's stance](https://twitter.com/alecthomas/status/1091467762625474560), we're switching from go-metalinter to golangci-lint. This PR updates the Travis config, makefile, and lint configs to make this happen.


paddycarver:
Travis failed due to lint issues that it previously didn't catch. I filed https://github.com/GoogleCloudPlatform/magic-modules/pull/1387 to resolve all those issues. When that PR is merged, these lint issues will go away.

paddycarver:
Fixes #3014.

.travis.yml
```diff
@@ -6,7 +6,7 @@ language: go
 go:
 - "1.11.x"
 env:
-  - GO111MODULE=off
+  - GO111MODULE=on
```
rileykarson: Can you see if we're able to safely set `GOFLAGS=-mod=vendor` as well?

paddycarver: I believe it's unnecessary, as the linting tool does it automatically?

rileykarson: Oh not for the linting tool, just because Go will realise we're using vendored modules in that case.

PR #3058, opened 2019-02-14
Documentation note about creating TXT records longer than 255 characters.

booi:
Common issue with 2048-bit DKIM records.


booi:
similar to https://github.com/terraform-providers/terraform-provider-aws/pull/2771

PR #3087, opened 2019-02-20
Fix compute_network import

cblecker:
The import for compute_network was broken (the project wasn't being set on the resource, if the default provider project wasn't set).

This fixes it with a basic parser, so if you specific the project in the object import string, it will import.

This was done manually, as this resource isn't managed by MM.


rileykarson:
Hey @cblecker! I'd had a PR to generate `google_compute_network` in Magic Modules in progress that I sent out superseding this; https://github.com/GoogleCloudPlatform/magic-modules/pull/1489

Regardless, thanks for taking the time to send this PR! Sorry about not picking this up earlier.

cblecker:
@rileykarson That's totally fine! I'm just glad to see this change in -- didn't have time to try and tackle the MM changes myself.

Thank you!

PR #3205, opened 2019-03-08
Update OAuth Scopes

mikebrooks-ovo:
Update the OAuth scopes to remove the insecure default and match googles latest recommendations https://cloud.google.com/kubernetes-engine/docs/how-to/access-scopes


website/docs/r/container_cluster.html.markdown
```diff
@@ -36,9 +36,7 @@ resource "google_container_cluster" "primary" {
   }
 
   node_config {
-    oauth_scopes = [
-      "https://www.googleapis.com/auth/compute",
-      "https://www.googleapis.com/auth/devstorage.read_only",
+    oauth_scopes = [
```
Chupaka: Are those spaces at the end of line necessary?..

mikebrooks-ovo: Re-pushed with whitespace stripped

PR #3239, opened 2019-03-14
Document best practice for service accounts, scopes, and IAM roles

conorgil:
After further research, I realized that what I wrote in https://github.com/terraform-providers/terraform-provider-google/pull/3116 was incorrect. The `scopes` field cannot be left blank.

This PR corrects that error and adds fully functional examples to `comput_instance` and `compute_instance_template` explaining the best practice for assigning permissions to instances using service accounts and IAM roles.

I make sure to link to the supporting documentation where I got my information from.

### DIFF
--- a/website/docs/r/compute_instance.html.markdown
+++ b/website/docs/r/compute_instance.html.markdown
@@ -362,7 +362,7 @@ The `service_account` block supports:
    ```
 
    See
-   [Best practice: service account, scopes, and IAM roles](best-practice-service-account-scopes-and-iam-roles)
+   [Best practice: service account, scopes, and IAM roles](#best-practice-service-account-scopes-and-iam-roles)
    for a full example.
 
 The `scheduling` block supports:
--- a/website/docs/r/compute_instance.html.markdown
+++ b/website/docs/r/compute_instance.html.markdown
@@ -55,10 +55,10 @@ resource "google_compute_instance" "default" {
 ```
 
 ## Best practice: service account, scopes, and IAM roles
-
-The [service accounts documentation](https://cloud.google.com/compute/docs/access/service-accounts#accesscopesiam)
-explains that access scopes are the legacy method of specifying
-permissions for an instance. Current
+The
+[Service accounts - Access scopes](https://cloud.google.com/compute/docs/access/service-accounts#accesscopesiam)
+documentation explains that access scopes are the legacy
+method of specifying permissions for an instance. Current
 [best practice](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#best_practices)
 is to assign the instance a service account, which is
 granted IAM roles for only the resources that it needs.
@@ -67,7 +67,10 @@ You cannot set only IAM roles on the service account and omit
 access scopes when creating an instance. The level of access a
 service account has is determined by a combination of access
 scopes and IAM roles so you must configure both access scopes
-and IAM roles for the service account to work properly.
+and IAM roles for the service account to work properly. For
+more details, see the
+[Service accounts - Permissions](https://cloud.google.com/compute/docs/access/service-accounts#service_account_permissions)
+docs.
 
 When following the best practice of utilizing IAM Roles,
 you should specify the `cloud-platform` scope to allow
--- a/website/docs/r/compute_instance.html.markdown
+++ b/website/docs/r/compute_instance.html.markdown
@@ -54,6 +54,86 @@ resource "google_compute_instance" "default" {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


conorgil:
@rileykarson ping since you merged #3116 and likely have the best context for this.

conorgil:
@chrisst thanks for commenting!

At a minimum, the Terraform docs need to be updated to correct the error that was merged in #3116; the `scopes` field cannot be an empty array when using IAM roles to specify permissions on a service account. Whether that happens in this PR or a new one is TBD.

> The content here is super valuable but the resource documentation isn't the right place for documenting GCP best practices.

I agree that the resource documentation is likely not the right place for detailed information about many best practices, but I think this might be an exception because IAM permissions are intrinsically linked to creating compute instances.

> No examples or comments should violate best practices

The examples in both [compute_instance](https://www.terraform.io/docs/providers/google/r/compute_instance.html) and [compute_instance_template](https://www.terraform.io/docs/providers/google/r/compute_instance_template.html) violate best practice because they only show use of the `scopes` field, which GCP docs clearly state is a legacy approach. The examples should be updated to highlight use of a service account and IAM roles.

> Best practices for using a resource should be minimal and link to GCP documentation

I tried my best to strive for this goal in this PR. Is there something specific that jumps out at you as non-minimal in the example that potentially could be removed? I'm happy to update.

If the title of the example is contentious because it says "best practice", then I'd be happy to change the title to something like "Using with service accounts and IAM roles".

> References to how the GCP products work should do the same

Perhaps my example adds too much pre-text explanation and this is what you think can/should be pared down? I was motivated by the incredibly useful examples in the `compute_instance_template` docs for [Using with Instance Group Manager](https://www.terraform.io/docs/providers/google/r/compute_instance_template.html#using-with-instance-group-manager) and [Deploying the Latest Image](https://www.terraform.io/docs/providers/google/r/compute_instance_template.html#deploying-the-latest-image), which contain a similar amount of background explanation.

> There should be at least 1 example for the most minimal configuration

The example in my PR is a very minimal example for launching a compute instance and controlling its permissions following the best practice of using a Service Account and IAM roles. I could make it smaller by eliminating use of `locals` and inline the IAM role in the `google_project_iam_member` resource.

Do you mean that the example should be the most minimal example possible to convey the concept of how the resource works, but does not necessarily need to be fully functional?

> Other examples should be for very common use patterns or mutually exclusive configurations.

The main reason I think this example should be included here instead of an external blog post or community article is that this is **the** common use pattern; anyone who launches a compute instance in almost any scenario will want to assign it permissions. Figuring out how to utilize multiple Terraform resources to achieve this basic use-case was incredibly frustrating for me, so I expect others have felt the same.


Has any of the above convinced you that the example should be included in these docs?

chrisst:
I think the intent of the examples in instance_template was to document how to use resources together when they require extra information to use together without errors. Information like `create_before_destroy` shows how to use the resources when the most straight forward approach would throw errors. The example of deploying the latest image is a bit blurrier but it's there to document some [choices that were made](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/image.go#L73-L86) in the provider that aren't part of the upstream api.

In your case this isn't an explanation of nuances within the provider code but a good pattern of using GCP. It's a more compelling argument to me for documenting a common usage pattern of compute_instance but I'm not sure how often people are attaching service_accounts to their compute instances rather than managing communication between services in their application logic. I'll see if my team mates have more insight on this and get back to you.

I do think that correcting the examples and the scopes documentation is a good move though. If you'd like to add that in another PR while this is being resolved that would get them in much more quickly.

website/docs/r/compute_instance.html.markdown
```diff
@@ -54,6 +54,89 @@ resource "google_compute_instance" "default" {
...
+permissions of the instance are completely determined
+by the IAM roles of the service account:
+
+```hcl
+resource "google_service_account" "web_server" {
+  // This service account will be assigned to
+  // web server instances and will be granted
+  // IAM roles that define the minimum set of
+  // required permissions.
+  account_id = "web-server"
```
bdronneau: I already encounter [soft delete](https://cloud.google.com/iam/docs/creating-managing-service-accounts#undeleting_a_service_account) in GCP with service when playing with `terraform apply/destroy`

Since this time I use random provider to get new name on `apply/destroy`

rileykarson:
I'm going to close this as won't fix. Along the lines that @chrisst was saying, we want to hit a lot closer to reference documentation than best practices documentation (unless those best practices are specific to Terraform).

Regardless, thanks for your suggestions!

PR #3241, opened 2019-03-14
compute_instance_template: Remove copy/paste typo in docs

conorgil:
The `allow_stopping_for_update` argument is not defined for resource `compute_instance_template`,
so remove this typo in the docs. That field is defined for `compute_instance` resource, so I think it is a copy/paste typo.


danawillow:
Thanks for catching that, @conorgil!

PR #3262, opened 2019-03-17
Add GCP zone to client_config datasource

mt-inside:
The datasource google_client_config currently reads the region from the gcloud config (returning empty string if unset). I need to use the currently configured zone in a system I'm working on; this PR adds that.


mt-inside:
No problem!

PR #3293, opened 2019-03-21
adding new resource google_bigtable_gc_policy

Eraac:
Adding new resource for manage garbage collection into bigtable

```
make testacc TEST=./google TESTARGS='-run=TestAccBigtableGC'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google -v -run=TestAccBigtableGC -timeout 240m -ldflags="-X=github.com/terraform-providers/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccBigtableGCPolicy_basic
=== PAUSE TestAccBigtableGCPolicy_basic
=== RUN   TestAccBigtableGCPolicy_union
=== PAUSE TestAccBigtableGCPolicy_union
=== CONT  TestAccBigtableGCPolicy_basic
=== CONT  TestAccBigtableGCPolicy_union
--- PASS: TestAccBigtableGCPolicy_union (21.21s)
--- PASS: TestAccBigtableGCPolicy_basic (23.85s)
PASS
ok  	github.com/terraform-providers/terraform-provider-google/google	23.902s
```

### DIFF
--- a/google/resource_bigtable_gc_policy.go
+++ b/google/resource_bigtable_gc_policy.go
@@ -148,11 +148,18 @@ func resourceBigtableGCPolicyRead(d *schema.ResourceData, meta interface{}) erro
 	defer c.Close()
 
 	name := d.Get("table").(string)
-	_, err = c.TableInfo(ctx, name)
+	ti, err := c.TableInfo(ctx, name)
 	if err != nil {
 		log.Printf("[WARN] Removing %s because it's gone", name)
 		d.SetId("")
 		return fmt.Errorf("Error retrieving table. Could not find %s in %s. %s", name, instanceName, err)
+	}
+
+	for _, fi := range ti.FamilyInfos {
+		if fi.Name == name {
+			d.SetId(fi.GCPolicy)
+			break
+		}
 	}
 
 	d.Set("project", project)
--- a/google/resource_bigtable_gc_policy.go
+++ b/google/resource_bigtable_gc_policy.go
@@ -50,7 +50,7 @@ func resourceBigtableGCPolicy() *schema.Resource {
 			},
 
 			"max_age": {
-				Type:     schema.TypeMap,
+				Type:     schema.TypeList,
 				Optional: true,
 				ForceNew: true,
 				Elem: &schema.Resource{
@@ -64,7 +64,7 @@ func resourceBigtableGCPolicy() *schema.Resource {
 			},
 
 			"max_version": {
-				Type:     schema.TypeMap,
+				Type:     schema.TypeList,
 				Optional: true,
 				ForceNew: true,
 				Elem: &schema.Resource{
--- a/google/resource_bigtable_gc_policy.go
+++ b/google/resource_bigtable_gc_policy.go
@@ -7,8 +7,8 @@ import (
 	"time"
 
 	"cloud.google.com/go/bigtable"
-	"github.com/hashicorp/terraform/helper/schema"
-	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 )
 
 const (
--- a/google/resource_bigtable_gc_policy.go
+++ b/google/resource_bigtable_gc_policy.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"strconv"
 	"time"
 
 	"cloud.google.com/go/bigtable"
@@ -208,18 +207,16 @@ func generateBigtableGCPolicy(d *schema.ResourceData) (bigtable.GCPolicy, error)
 	}
 
 	if aok {
-		d, err := strconv.Atoi(ma.(map[string]interface{})["days"].(string))
-		if err != nil {
-			return nil, err
-		}
+		l, _ := ma.([]interface{})
+		d, _ := l[0].(map[string]interface{})["days"].(int)
+
 		policies = append(policies, bigtable.MaxAgePolicy(time.Duration(d)*time.Hour*24))
 	}
 
 	if vok {
-		n, err := strconv.Atoi(mv.(map[string]interface{})["number"].(string))
-		if err != nil {
-			return nil, err
-		}
+		l, _ := mv.([]interface{})
+		n, _ := l[0].(map[string]interface{})["number"].(int)
+
 		policies = append(policies, bigtable.MaxVersionsPolicy(n))
 	}
 
--- a/google/resource_bigtable_gc_policy.go
+++ b/google/resource_bigtable_gc_policy.go
@@ -36,7 +36,7 @@ func resourceBigtableGCPolicy() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"family": {
+			"column_family": {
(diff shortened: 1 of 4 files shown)


Eraac:
Actually if you re-create `bigtable_table`, `gc_policy` is not re-apply. Discussion for this kind of issue is still open https://github.com/hashicorp/terraform/issues/8099 but maybe something is possible for "fixing" this ?

gnalsa:
Hi @rileykarson  I am not very familiar with feature process, but this would be a great addition to enable management of bigtable.

rileykarson:
Sorry for the radio silence on my end here! At first glance, your contribution looks great, but I'm wondering if this is something we should add to this provider.

It's come up with a few other APIs- BigQuery, Spanner, etc. where we're not sure whether some features are GCP or "data" level. In general, we want to keep this provider focused on provisioning GCP infrastructure and not performing "data" operations on services- even those exclusive to GCP.

Since this isn't exposed in the [Bigtable Admin API](https://cloud.google.com/bigtable/docs/reference/admin/rest/), I'm inclined to say this falls more on the data side, but I'll bring this up with the rest of my team in our sync later this week.

Eraac:
@rileykarson I understood but is like "lifecycle" ressources, I not agree with "data" level

Furthermore is actually documented [here](https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/projects.instances.tables/modifyColumnFamilies#gcrule) below modifyColumnFamilies

rileykarson:
Understood! We're weighing some considerations to make a decision about whether it falls into the scope of the Google provider or not.

Eraac:
@rileykarson Any news?

google/resource_bigtable_gc_policy.go
```diff
@@ -0,0 +1,227 @@
...
+				ForceNew: true,
+			},
+
+			"table": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"family": {
```
rileykarson: nit: Do you mind renaming this to `column_family`?

website/docs/r/bigtable_gc_policy.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+    family = "name"
+  }
+}
+
+resource "google_bigtable_gc_policy" "policy" {
+  instance_name = "${google_bigtable_instance.instance.name}"
+  table         = "${google_bigtable_table.table.name}"
+  family        = "name"
+  
+  max_age {
```
rileykarson: Do you mind showing an example requiring `UNION` or `INTERSECTION`?

google/resource_bigtable_gc_policy.go
```diff
@@ -0,0 +1,227 @@
...
+
+			"mode": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringInSlice([]string{GCPolicyModeIntersection, GCPolicyModeUnion}, false),
+			},
+
+			"max_age": {
+				Type:     schema.TypeMap,
```
rileykarson:
```suggestion
				Type:     schema.TypeList,
```

rileykarson: Types in Terraform are a little unintuitive, and it has no real way to represent nested objects. `TypeList` is the best approximate.

google/resource_bigtable_gc_policy.go
```diff
@@ -0,0 +1,227 @@
...
+						"days": {
+							Type:     schema.TypeInt,
+							Required: true,
+						},
+					},
+				},
+			},
+
+			"max_version": {
+				Type:     schema.TypeMap,
```
rileykarson:
```suggestion
				Type:     schema.TypeList,
```

google/resource_bigtable_gc_policy.go
```diff
@@ -0,0 +1,227 @@
...
+	for _, i := range table.FamilyInfos {
+		if i.Name == family {
+			d.SetId(i.GCPolicy)
+		}
+	}
+
+	return resourceBigtableGCPolicyRead(d, meta)
+}
+
+func resourceBigtableGCPolicyRead(d *schema.ResourceData, meta interface{}) error {
```
rileykarson: Is it possible to recreate a GCPolicy from the string in FamilyInfo? https://github.com/terraform-providers/terraform-provider-google/blob/ab7c29f09dafc6d836888b3db57b53f148e37727///vendor/cloud.google.com/go/bigtable/admin.go#L231

Drift detection would be nice if it can, but it's not a must-have.

website/docs/r/bigtable_gc_policy.html.markdown
Eraac: I've added example for UNION, it's good for you?

google/resource_bigtable_gc_policy.go
Eraac: It's weird I forget about this, you mean about the SetId("...") to ensure nothing change between two plan/apply? If yes, should be good now

rileykarson: The `id` value won't actually change anything, so not setting it is fine. I'm curious whether we can extract values like `max_age` from the GCPolicy string, so we can set those in state.

Eraac: I see, it's probably possible, but I thinking about all cases, what if Google change something? Or add another policy? I am not even sure order of the policy is guaranteed (can be max_age && max_version / max_version && max_age)

I wonder, what's happen in terraform if you have 2 resources, with the same ID (actually, that can happen if you set same gc policy on two table), should I prefix ID with the column name?

rileykarson: 2 resources can share an id, Terraform only distinguishes between the `id` value being unset and set.

google/resource_bigtable_gc_policy.go
```diff
@@ -0,0 +1,234 @@
...
+	if !aok && !vok {
+		return bigtable.NoGcPolicy(), nil
+	}
+
+	if mode == "" && aok && vok {
+		return nil, fmt.Errorf("If multiple policies are set, mode can't be empty")
+	}
+
+	if aok {
+		d, err := strconv.Atoi(ma.(map[string]interface{})["days"].(string))
```
rileykarson: I think that `ma` needs to be converted to `[]interface{}`, and the `0` element should be a `map[string]interface{}`.

google/resource_bigtable_gc_policy.go
```diff
@@ -0,0 +1,234 @@
...
+		if err != nil {
+			return nil, err
+		}
+		policies = append(policies, bigtable.MaxAgePolicy(time.Duration(d)*time.Hour*24))
+	}
+
+	if vok {
+		n, err := strconv.Atoi(mv.(map[string]interface{})["number"].(string))
```
rileykarson: Same as above.

google/resource_bigtable_gc_policy.go
Eraac: I've change, not sure I should check if slice has at least one element, shouldn't go in this statement if value isn't set in .tf files

Eraac:
@rileykarson  Sure, no problem. I've update the resource file (and test)

PR #3306, opened 2019-03-25
Removed data

mrsiejas:
Could not create "google_service_account_key" until removed "data" from the variable.

### DIFF
--- a/website/docs/d/datasource_google_service_account.html.markdown
+++ b/website/docs/d/datasource_google_service_account.html.markdown
@@ -26,7 +26,7 @@ data "google_service_account" "myaccount" {
 }
 
 resource "google_service_account_key" "mykey" {
-  service_account_id = "${data.google_service_account.myaccount.name}"
+  service_account_id = "${google_service_account.myaccount.name}"
 }
 
 resource "kubernetes_secret" "google-application-credentials" {


Chupaka:
Any details on your configuration? Because it looks wrong :)

mrsiejas:
Sorry for syntax, its myfirst contribution GH. In general I found using Data Sources in tutorial somewhat confusing. I understand it is possible, but not required thereby not required

Here's a snippet worked for me

`resource "google_service_account" "cloudsql_sa" {
  account_id   = "cloudsql"
  display_name = "cloudsql-sa"
  project      = "${var.project}"
}

resource "google_project_iam_binding" "project" {
  project = "${var.project}"
  role    = "roles/cloudsql.client"

  members = [
    "serviceAccount:${google_service_account.cloudsql_sa.account_id}@${var.project}.iam.gserviceaccount.com",
  ]
}

resource "google_service_account_key" "cloudsql_credentials_key" {
  service_account_id = "${google_service_account.cloudsql_sa.name}"
}`

Is using Data Sources is a standard practice in Terraform docs?

Chupaka:
So, you're trying to edit an example for datasource so that it doesn't use the datasource?.. You're looking at wrong place,
https://www.terraform.io/docs/providers/google/d/datasource_google_service_account.html (datasource)
and
https://www.terraform.io/docs/providers/google/r/google_service_account.html (resource)
are different things.

PR #3307, opened 2019-03-25
Add support for CloudSQL source representation instances

maljub01:
Those are metadata-only CloudSQL instances that are used for replication
of external MySQL databases.

See the following documentation links for more details:
- https://cloud.google.com/sql/docs/mysql/replication/external-server
- https://cloud.google.com/sql/docs/mysql/replication/replication-from-external

### DIFF
--- a/google/resource_sql_source_representation_instance.go
+++ b/google/resource_sql_source_representation_instance.go
@@ -1,0 +1,227 @@
+package google
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"google.golang.org/api/googleapi"
+	sqladmin "google.golang.org/api/sqladmin/v1beta4"
+)
+
+func resourceSqlSourceRepresentationInstance() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceSqlSourceRepresentationInstanceCreate,
+		Read:   resourceSqlSourceRepresentationInstanceRead,
+		Delete: resourceSqlSourceRepresentationInstanceDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceSqlSourceRepresentationInstanceImporter,
+		},
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(10 * time.Minute),
+			Delete: schema.DefaultTimeout(10 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"database_version": {
+				Type:     schema.TypeString,
+				Default:  "MYSQL_5_7",
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"host": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"port": {
+				Type:         schema.TypeInt,
+				Default:      3306,
+				Optional:     true,
+				ForceNew:     true,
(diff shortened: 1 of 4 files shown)


google/resource_sql_source_representation_instance.go
```diff
@@ -0,0 +1,227 @@
...
+
+	instance, err := config.clientSqlAdmin.Instances.Get(project, d.Id()).Do()
+	if err != nil {
+		return handleNotFoundError(err, d, fmt.Sprintf("SQL Source Representation Instance %q", d.Get("name").(string)))
+	}
+
+	d.Set("name", instance.Name)
+	d.Set("region", instance.Region)
+	d.Set("database_version", instance.DatabaseVersion)
+	hostPort := strings.Split(instance.OnPremisesConfiguration.HostPort, ":")
```
emilymye: Since instance.OnPremisesConfiguration could be nil, we should probably add a nil check here. Please also add a check on the split string to make sure it was %s:%s

google/resource_sql_source_representation_instance.go
```diff
@@ -0,0 +1,227 @@
...
+	instance := &sqladmin.DatabaseInstance{
+		Name:            name,
+		Region:          region,
+		DatabaseVersion: d.Get("database_version").(string),
+		OnPremisesConfiguration: &sqladmin.OnPremisesConfiguration{
+			HostPort: hostPort,
+		},
+	}
+
+	backoff := time.Second
```
emilymye: Is there a reason you aren't using a retryTime() like in delete/other sql instances?

website/docs/r/sql_source_representation_instance.html.markdown
```diff
@@ -0,0 +1,91 @@
...
+  Creates a new SQL Source Representation Instance, which holds metadata about
+  on-premises MySQL instances. This enables Google CloudSQL MySQL instances to
+  become replicas of on-premises databases.
+
+  Replicas are configured by setting the source representation instance name in
+  their `master_instance_name` value.
+
+---
+
+# google\_sql\_database\_instance
```
emilymye: google\_sql\_source\_representation\_instance

website/docs/r/sql_source_representation_instance.html.markdown
```diff
@@ -0,0 +1,91 @@
...
+page_title: "Google: google_sql_source_representation_instance"
+sidebar_current: "docs-google-sql-source-representation-instance"
+description: |-
+  Creates a new SQL Source Representation Instance, which holds metadata about
+  on-premises MySQL instances. This enables Google CloudSQL MySQL instances to
+  become replicas of on-premises databases.
+
+  Replicas are configured by setting the source representation instance name in
+  their `master_instance_name` value.
+
```
emilymye: Can you add a note here about the difference between a normal instance and a regular instance? Just for clarity, since I was a bit confused. Copy-paste from https://cloud.google.com/sql/docs/mysql/replication/external-server#config-description would be fine

website/docs/r/sql_source_representation_instance.html.markdown
```diff
@@ -0,0 +1,91 @@
...
+---
+
+# google\_sql\_database\_instance
+
+Creates a new source representation Instance for Google CloudSQL.
+
+For more information, see the [official documentation](https://cloud.google.com/sql/),
+or the [JSON API](https://cloud.google.com/sql/docs/admin-api/v1beta4/instances).
```
emilymye: Can you add the links you put in the PR description here as well?

google/resource_sql_source_representation_instance.go
emilymye: For reference: [retryTime()](https://github.com/terraform-providers/terraform-provider-google/blob/1319ddd86de649e62984da8a31ca5643c4ef1787/google/utils.go#L322) which uses [isRetryableError()](https://github.com/terraform-providers/terraform-provider-google/blob/1319ddd86de649e62984da8a31ca5643c4ef1787/google/utils.go#L341)

danawillow:
Since this has had no activity in a while, closing it out. See https://github.com/terraform-providers/terraform-provider-google/issues/5620#issuecomment-589799193 for more context.

PR #3320, opened 2019-03-26
fix markdown link in contributing docs

mrparkers:
the extra `]` is causing the link to render incorrectly


rileykarson:
Thanks!

PR #3331, opened 2019-03-27
Add maintainer information to CONTRIBUTING

rileykarson:
Pulled some stuff out of the teamsite

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -100,12 +100,6 @@ ln -s $GOPATH/bin/terraform-provider-google-beta ~/.terraform.d/plugins/terrafor
 
 ## Reviewing / Merging Code
 
-When reviewing code, follow the guidelines set in the
+When reviewing/merging code, roughly follow the guidelines set in the
 [Maintainer's Etiquette](https://github.com/hashicorp/terraform/blob/master/docs/maintainer-etiquette.md)
-guide. If reviewing another maintainer, it's expected that they'll merge their own code. Otherwise, merge at will once
-you've approved the contribution.
-
-Generally, we use squash merges on the team; it's unlikely you'll be modifying this repo directly regardless, as most development happens through Magic Modules, but if you do you're able to use rebase / traditional merges to preserve commit history.
-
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -95,3 +95,17 @@ time you run `make build`, you can symlink the built binary into the directory:
...
 ln -s $GOPATH/bin/terraform-provider-google-beta ~/.terraform.d/plugins/terraform-provider-google-beta
 ```
+
+# Maintainer-specific information
+
+## Reviewing / Merging Code
+
+When reviewing code, follow the guidelines set in the
+[Maintainer's Etiquette](https://github.com/hashicorp/terraform/blob/master/docs/maintainer-etiquette.md)
+guide. If reviewing another maintainer, it's expected that they'll merge their own code. Otherwise, merge at will once
```
chrisst: This is a repeat of a more detailed version in the maintainer's etiquette and this concise version doesn't make things more clear. I would just defer to the linked version.

.github/CONTRIBUTING.md
```diff
@@ -95,3 +95,17 @@ time you run `make build`, you can symlink the built binary into the directory:
...
+you've approved the contribution.
+
+Generally, we use squash merges on the team; it's unlikely you'll be modifying this repo directly regardless, as most development happens through Magic Modules, but if you do you're able to use rebase / traditional merges to preserve commit history.
```
chrisst: I'm also not sure what this provides over the linked squash/merge policy as well. If the intent was to describe how MM changes are the primary source then rephrasing might be helpful. Or maybe examples of non-MM changes or why changes might be non-mm. Rebasing and merge is an option in the github UI so IMO you'd only need to talk about them to encourage people _not_ to use them.

.github/CONTRIBUTING.md
rileykarson: Huh, I hadn't clicked through to that in a while. I guess our teamsite duplicated this a whole bunch too. Added a caveat that the maintainer guide is pretty old (it was from the pre-provider split days) but that general guidance should be followed.

PR #3347, opened 2019-03-31
transport replaceVars url escape values

### LINKED ISSUE #2895: Some `google_storage_*` resources don't encode object names properly
### Community Note

### Terraform Version

```
Terraform v0.11.11
+ provider.google v1.20.0
```

### Summary

This is a bug report about a bunch of resource that makes use of the GCS API directly because the `google.golang.org/api/storage/v1` package does not support the API calls for some reason.

When using the JSON API that requires the object names in the URL, we have to take care in URL encoding object names that contains `/` for the URL. See https://cloud.google.com/storage/docs/json_api/#encoding

These resources do not seem to do any form of URL encoding in API call.

### Affected Resource(s)

I looked through the source code searching for the string `https://www.googleapis.com/storage/v1` so this might not be exhaustive:

- `data.google_storage_bucket_object`
- `google_storage_default_default_object_access_control`
- `google_storage_object_access_control`

Consider the following example:

```tf
provider "google" {
  region  = "${var.region}"
  project = "${var.project_id}"
}

terraform {
  backend "gcs" {}
}

resource "google_storage_bucket" "test" {
  name     = "${var.bucket_name}"
  location = "${var.region}"

  storage_class = "REGIONAL"
}

resource "google_storage_bucket_object" "object" {
  name    = "foo/bar/baz"
  bucket  = "${google_storage_bucket.test.name}"
  content = "Hello GCP"
}

resource "google_storage_object_access_control" "public_rule" {
  object = "${google_storage_bucket_object.object.name}"
  bucket = "${google_storage_bucket.test.name}"
  role   = "READER"
  entity = "allUsers"
}

variable "bucket_name" {
  description = "Bucket name"
}

variable "project_id" {
  description = "Project ID to deploy the cluster to"
}

variable "region" {
  description = "Default region for GCP"
  default     = "asia-southeast1"
}

```

### Debug Output

https://gist.github.com/lawliet89/0258872b901513386f10a5b1c8f755a0

### Expected Behavior

ACL for object is created.

### Actual Behavior

Failed.

```
* google_storage_object_access_control.public_rule: Error creating ObjectAccessControl: googleapi: got HTTP response code 404 with body: Not Found
```

If you look at the debug log, you can see that the object name in the URL was not escaped properly.

Also, using the `urlencode` interpolation function does not work. It causes a HTTP 400 error where the URL object name differs from the one in the POST body.

odedniv:
Fixes #2895 

According to https://cloud.google.com/storage/docs/json_api/#encoding URI paths parts are supposed to be encoded.

For example in `google_storage_object_access_control` if the object name contains a `/`, the request will fail with:

```
* google_storage_object_access_control.public_rule: Error creating ObjectAccessControl: googleapi: got HTTP response code 404 with body: Not Found
```

To my knowledge, `replaceVars()` is specifically for replacing variables in URLs (as for the name of the parameter "linkTmpl"), so I'm assuming this would work globally.

### DIFF
--- a/google/transport.go
+++ b/google/transport.go
@@ -166,7 +166,7 @@ func buildReplacementFunc(re *regexp.Regexp, d TerraformResourceData, config *Co
 		}
 		v, ok := d.GetOk(m)
 		if ok {
-			return fmt.Sprintf("%v", v)
+			return url.QueryEscape(fmt.Sprintf("%v", v))
 		}
 		return ""
 	}
--- a/google/transport_test.go
+++ b/google/transport_test.go
@@ -126,6 +126,14 @@ func TestReplaceVars(t *testing.T) {
 			},
 			Expected: "projects/project1/zones/zone1/instances/instance1",
 		},
+		"escape": {
+			Template: "b/{{bucket}}/o/{{object}}",
+			SchemaValues: map[string]interface{}{
+				"bucket": "bucket1",
+				"object": "path/to/object",
+			},
+			Expected: "b/bucket1/o/path%2Fto%2Fobject",
+		},
 	}
 
 	for tn, tc := range cases {


odedniv:
It contains conflicts now because 966229afb allowed adding `%` to variable names in order to URL encode them, but I believe it should _always_ URL encode them because `replaceVars` is used specifically for URLs.

odedniv:
Maybe closed by https://github.com/GoogleCloudPlatform/magic-modules/pull/1717 (if I understand the repo structure correctly)

odedniv:
Not relevant as discussed in https://github.com/GoogleCloudPlatform/magic-modules/pull/1717

PR #3348, opened 2019-03-31
storage_object_acl delete before create/update

odedniv:
It's a very specific edge case, but maybe there are more ways to reproduce this.

I have this snippet:

```terraform
resource "google_storage_bucket" "this" {
  name = "test-bucket"
}

resource "google_storage_object" "this" {
  bucket  = "${google_storage_bucket.this.name}"
  name    = "test-object"
  content = "test"
}

resource "google_storage_object_acl" "this" {
  bucket = "${google_storage_object.this.bucket}"
  object = "${google_storage_object.this.name}"

  role_entity = [
    "OWNER:project-owners-${var.google_project_id}",
  ]
}
```

Note that I am using `google_project_id` and _not_ `google_project_number` for the entity of the ACL. Since Google actually works by project number here, this works but saves the entity as the number. This means that on every apply I see something like this:

```
  ~ google_storage_object_acl.this
      role_entity.1899103587: "OWNER:project-owners-<PROJECT NUMBER>" => ""
      role_entity.3063552365: "" => "OWNER:project-owners-<PROJECT ID>"
```

I'm fine with this (it's maybe kind of my fault), but since it deletes _after_ it creates, the final result is having _no_ ACL permission for the project owners.

This PR should solve this.

### DIFF
--- a/google/resource_storage_object_acl.go
+++ b/google/resource_storage_object_acl.go
@@ -301,6 +301,13 @@ func getRoleEntityChange(old []string, new []string, owner string) (create, upda
 
 // Takes in lists of changes to make to a Storage Object's ACL and makes those changes
 func performStorageObjectRoleEntityOperations(create []*RoleEntity, update []*RoleEntity, remove []*RoleEntity, config *Config, bucket string, object string) error {
+	for _, v := range remove {
+		err := config.clientStorage.ObjectAccessControls.Delete(bucket, object, v.Entity).Do()
+		if err != nil {
+			return fmt.Errorf("Error deleting ACL item %s for object %s in %s: %v", v.Entity, object, bucket, err)
+		}
+	}
+
 	for _, v := range create {
 		objectAccessControl := &storage.ObjectAccessControl{
 			Role:   v.Role,
@@ -320,13 +327,6 @@ func performStorageObjectRoleEntityOperations(create []*RoleEntity, update []*Ro
 		_, err := config.clientStorage.ObjectAccessControls.Update(bucket, object, v.Entity, objectAccessControl).Do()
 		if err != nil {
 			return fmt.Errorf("Error updating ACL item %s for object %s in %s: %v", v.Entity, object, bucket, err)
-		}
-	}
-
-	for _, v := range remove {
-		err := config.clientStorage.ObjectAccessControls.Delete(bucket, object, v.Entity).Do()
-		if err != nil {
-			return fmt.Errorf("Error deleting ACL item %s for object %s in %s: %v", v.Entity, object, bucket, err)
 		}
 	}
 


rileykarson:
Hey @odedniv! Sorry for the delay in getting to this.

I think my concern here is that it's easy to accidentally lock an account out after this. Let's say a service account was granted the `OWNER` ACL individually, but it was then removed & granted to a group the account was a member of. The account could lock itself out when it otherwise wouldn't have, & that would leave Terraform in an unrecoverable state without manual intervention.

There's a place for dangerous things being dangerous, such as when configuring authoritative IAM policies, but this is preventable by adding before deleting like the resource does now.

It's unfortunate that the API canonicalises id's to numbers, but this is also preventable in config - [`data.google_project.number`](https://www.terraform.io/docs/providers/google/r/google_project.html) will let you canonicalise from `id` -> `number` in your own config, stopping this diff from occurring.

odedniv:
No worries @rileykarson.

I understand the concern you brought up, and that was exactly the case for me: the current implementation with my example removed the owners group because it deletes after creating.

`storage_object_acl` seem to always leave the creator of the object as an owner (even if not specified explicitly), so if the creator is the deployer of the Terraform configuration (using `google_storage_object`) the deployer (e.g a CI/CD service account) will be able to add the permissions after deleting the others. On another note, GCS does not allow leaving an object without an owner so you should never be completely locked out.

About using `data.google_project.number`: when I used it to define a bucket name Terraform always seemed to define the bucket name as "computed" during planning, and even though it kept the same computed name it kept *recreating* the bucket (which is dangerous in itself). I believe that the values for `data.google_project` are calculated during deploy and so anything using it is considered "computed" during planning and can be recreated (depending on the id of the resource). In this case you would have the same issue described above: The ACL resource will be "recreated" and the "old" resource will be destroyed. I didn't test it though (I just remember it about the bucket).

rileykarson:
> seem to always leave the creator of the object as an owner

Yep! The situation I'm envisioning here is one where the object owner and the Terraform runner aren't the same user; if we delete first, it's fairly easy to enter a state that needs manual reconciliation. This would likely also constitute a breaking change for the provider.

> seemed to define the bucket name as "computed" during planning

This is expected behaviour from Terraform! When using an output value / `Computed` value, Terraform doesn't necessarily know the value until apply time. We know that a project number is stable for a given project ID, but Terraform doesn't. Once it's retrieved the value [at apply time], Terraform is able to tell that there was no change and won't take any action.

I wouldn't expect Terraform to be performing recreates in this sitation; was it your experience that it was actually removing buckets?

odedniv:
> I wouldn't expect Terraform to be performing recreates in this sitation; was it your experience that it was actually removing buckets?

This was my experience, but maybe it was a more specific setup since I cannot reproduce it now. Maybe another force at play when it happened.

odedniv:
Thanks for the responses, very appreciated! Closing the pull request since it's probably more dangerous this way than how it was before the PR.

PR #3354, opened 2019-04-01
vendor iam libraries

chrisst:
vendoring changes for https://github.com/GoogleCloudPlatform/magic-modules/pull/1597


chrisst:
Build process was fixed so we don't need to add vendored code in separate PRs any more.

PR #3384, opened 2019-04-05
Update logging_billing_account_sink.html.markdown

n0npax:
fix issue in docs


n0npax:
I've no idea about which kind of docs are you talking about and I don't want to make commitment. Sorry.

danawillow:
These are the ones I meant:
https://www.terraform.io/docs/providers/google/r/logging_folder_sink.html
https://www.terraform.io/docs/providers/google/r/logging_organization_sink.html
https://www.terraform.io/docs/providers/google/r/logging_project_sink.html

If you don't think you'll get around to them though that's totally fine! I can merge this tomorrow as-is.

n0npax:
Thanks for links. PR updated.

PR #3480, opened 2019-04-23
Add Project Field To Google Container Cluster Resource

abejide001:
the project field is required to run, without this, `terraform apply` throws an error
Screenshot 2019-04-23 at 5 51 47 PM


rileykarson:
Since `project` is a field that every user will need to change, we generally omit it from examples. Most users will use the behaviour where the project is [inferred from the value set in the provider](https://www.terraform.io/docs/providers/google/r/container_cluster.html#project); https://www.terraform.io/docs/providers/google/provider_reference.html#project.

PR #3481, opened 2019-04-23
Add support for creating instances with CMEK

zachberger:
This adds support for creating google_compute_instance with a Customer-managed encryption key.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -3,6 +3,7 @@ package google
 import (
 	"crypto/sha256"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"log"
 	"strings"
@@ -277,8 +278,9 @@ func resourceComputeInstance() *schema.Resource {
 						},
 
 						"kms_key_self_link": {
-							Type:     schema.TypeString,
-							Optional: true,
+							Type:          schema.TypeString,
+							Optional:      true,
+							ConflictsWith: []string{"boot_disk.0.disk_encryption_key_raw"},
 						},
 
 						"disk_encryption_key_sha256": {
@@ -364,12 +366,6 @@ func resourceComputeInstance() *schema.Resource {
 							Optional:  true,
 							ForceNew:  true,
 							Sensitive: true,
-						},
-
-						"kms_key_self_link": {
-							Type:     schema.TypeString,
-							Optional: true,
-							ForceNew: true,
 						},
 
 						"disk_encryption_key_sha256": {
@@ -1384,15 +1380,20 @@ func expandAttachedDisk(diskConfig map[string]interface{}, d *schema.ResourceDat
 		disk.DeviceName = v.(string)
 	}
 
-	if v, ok := diskConfig["disk_encryption_key_raw"]; ok {
+	keyValue, keyOk := diskConfig["disk_encryption_key_raw"]
+	if keyOk {
 		disk.DiskEncryptionKey = &computeBeta.CustomerEncryptionKey{
-			RawKey: v.(string),
+			RawKey: keyValue.(string),
 		}
 	}
 
-	if v, ok := diskConfig["kms_key_self_link"]; ok {
+	kmsValue, kmsOk := diskConfig["kms_key_self_link"]
+	if kmsOk {
+		if keyOk {
+			return nil, errors.New("Only one of kms_key_self_link and disk_encryption_key_raw can be set")
+		}
 		disk.DiskEncryptionKey = &computeBeta.CustomerEncryptionKey{
-			KmsKeyName: v.(string),
+			KmsKeyName: kmsValue.(string),
 		}
 	}
 	return disk, nil
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -871,10 +871,17 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 			}
 			if key := disk.DiskEncryptionKey; key != nil {
 				if inConfig {
-					di["disk_encryption_key_raw"] = d.Get(fmt.Sprintf("attached_disk.%d.disk_encryption_key_raw", adIndex))
+					rawKey := d.Get(fmt.Sprintf("attached_disk.%d.disk_encryption_key_raw", adIndex))
+					if rawKey != "" {
+						di["disk_encryption_key_raw"] = rawKey
+					}
 				}
-				di["kms_key_self_link"] = key.KmsKeyName
-				di["disk_encryption_key_sha256"] = key.Sha256
+				if key.KmsKeyName != "" {
+					di["kms_key_self_link"] = key.KmsKeyName
+				}
+				if key.Sha256 != "" {
+					di["disk_encryption_key_sha256"] = key.Sha256
+				}
 			}
 			// We want the disks to remain in the order we set in the config, so if a disk
 			// is present in the config, make sure it's at the correct index. Otherwise, append it.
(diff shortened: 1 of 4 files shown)


google/resource_compute_instance_test.go
```diff
@@ -282,6 +282,40 @@ func TestAccComputeInstance_diskEncryption(t *testing.T) {
...
+		CheckDestroy: testAccCheckComputeInstanceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccComputeInstance_disks_encryption(bootKmsKeyName, diskNameToEncryptionKey, instanceName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeInstanceExists("google_compute_instance.foobar", &instance),
+					testAccCheckComputeInstanceDiskKmsEncryptionKey("google_compute_instance.foobar", &instance, bootKmsKeyName, diskNameToEncryptionKey),
+				),
+			},
+		},
```
danawillow: can you also add an import step? `computeInstanceImportStep("us-central1-a", instanceName, []string{}),` should do the trick. There's a bunch of examples in other tests in this file. If it ends up being not too much effort, it would also be a big help to add that step to the other encryption key test while you're here.

google/resource_compute_instance_test.go
```diff
@@ -282,6 +282,40 @@ func TestAccComputeInstance_diskEncryption(t *testing.T) {
 	})
 }
 
+func TestAccComputeInstance_kmsDiskEncryption(t *testing.T) {
+	t.Parallel()
+
+	var instance compute.Instance
+	var instanceName = fmt.Sprintf("instance-test-%s", acctest.RandString(10))
+	bootKmsKeyName := "projects/project/locations/us-central1/keyRings/testing-cloud-kms/cryptoKeys/key-0/cryptoKeyVersions/1"
+	diskNameToEncryptionKey := map[string]*compute.CustomerEncryptionKey{
```
danawillow: It looks like the actual config that this calls only looks at the RawKey in this block. I'd recommend creating a completely new test to test KMS keys, especially since we also tend to create new projects for those tests (see https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_kms_crypto_key_test.go)

google/resource_compute_instance.go
```diff
@@ -355,6 +366,12 @@ func resourceComputeInstance() *schema.Resource {
 							Sensitive: true,
 						},
 
+						"kms_key_self_link": {
```
danawillow: This can be removed- nobody can set the "disk" block in configs anymore.

google/resource_compute_instance.go
```diff
@@ -1371,6 +1389,12 @@ func expandAttachedDisk(diskConfig map[string]interface{}, d *schema.ResourceDat
 			RawKey: v.(string),
 		}
 	}
+
+	if v, ok := diskConfig["kms_key_self_link"]; ok {
+		disk.DiskEncryptionKey = &computeBeta.CustomerEncryptionKey{
+			KmsKeyName: v.(string),
```
danawillow: Is it possible for `KmsKeyName` and `RawKey` to both be set at the same time? (my guess would be no, but I haven't checked.) If so, then we'll want to add that functionality to this section of code. If not, then let's explicitly document that (normally we'd also check it plan-time with `ConflictsWith`, but I don't think those work on inner elements of objects in lists).

google/resource_compute_instance.go
```diff
@@ -1506,6 +1530,12 @@ func expandBootDisk(d *schema.ResourceData, config *Config, zone *compute.Zone,
 		}
 	}
 
+	if v, ok := d.GetOk("boot_disk.0.kms_key_self_link"); ok {
```
danawillow: In this case because there's only ever one boot disk, we can add `ConflictsWith` in the schema (so you'd put `ConflictsWith: "[]string{"boot_disk.0.disk_encryption_key_raw"},` in the boot disk kms key section

google/resource_compute_instance.go
zachberger: Only one can be set.

Should we check explicitly with something like the code below, or should we wait for the API to catch it?

```go
if v, ok := diskConfig["disk_encryption_key_raw"]; ok {
	disk.DiskEncryptionKey = &computeBeta.CustomerEncryptionKey{
		RawKey: v.(string),
	}
}

if v, kmsOk := diskConfig["kms_key_self_link"]; kmsOk {
	if ok {
		return nil, errors.New("Only one of kms_key_self_link and disk_encryption_key_raw can be set")
	}
	disk.DiskEncryptionKey = &computeBeta.CustomerEncryptionKey{
		KmsKeyName: v.(string),
	}
}
```
In terms of documentation, is documentation on the website for these values sufficient?

zachberger:
@danawillow I believe I've addressed all comments. Thanks!

google/resource_compute_instance.go
danawillow: My general advice is that if the API error message is good, then let it fall through to the API, otherwise add our own.

And yes, thank you!

zachberger: After a bit more investigation the API accepts both, but ignores the RawKey and uses the KMS key if both are specified. That being said, I don't think it makes sense to allow both to be specified.

google/resource_compute_instance_test.go
```diff
@@ -282,6 +282,41 @@ func TestAccComputeInstance_diskEncryption(t *testing.T) {
...
+		fmt.Sprintf("instance-testd-%s", acctest.RandString(10)): {
+			KmsKeyName: "projects/project/locations/us-central1/keyRings/testing-cloud-kms/cryptoKeys/key-1/cryptoKeyVersions/1",
```
danawillow: The acceptance tests create real resources, which means that their dependencies (in this case, kms key names) also have to correspond to real resources. There's some information about running tests at https://github.com/terraform-providers/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#tests if you'd like to be able to run them yourself.

google/resource_compute_instance.go
danawillow: Cool, I agree with you there. Thanks for checking!

google/resource_compute_instance_test.go
danawillow: In the case of KMS resources specifically, you can use https://github.com/terraform-providers/terraform-provider-google/blob/master/google/bootstrap_utils_test.go#L31. Take a look at https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_disk_test.go#L316 for example usage.

zachberger: Ah, This is why the tests were running so fast 😅`go test` wasn't actually doing anything. I'll work on this.

zachberger:
@danawillow I've been able to run and fix the tests. PTAL. Thanks for all of the guidance.

google/resource_compute_instance.go
```diff
@@ -1501,8 +1541,18 @@ func expandBootDisk(d *schema.ResourceData, config *Config, zone *compute.Zone,
 	}
 
 	if v, ok := d.GetOk("boot_disk.0.disk_encryption_key_raw"); ok {
-		disk.DiskEncryptionKey = &computeBeta.CustomerEncryptionKey{
-			RawKey: v.(string),
+		if v != "" {
```
danawillow: Just curious, did you run into any cases where d.GetOk returned true, but the value was empty?

danawillow: (I ask because that should be impossible, and so the code could be a tiny bit cleaner without the extra `!= ""` checks)

zachberger: Without the check the API was complaining that I was trying to create a disk with a customer-supplied encryption key without specifying a key.

danawillow: Hmm that's surprising. I don't think it's worth either of our time to really dig in though since it seems to work just fine with the checks. Thanks!

PR #3485, opened 2019-04-24
node_count is required

minac:
Without node_count, autoscaling will never kick in because e.g. 0 is not between 2 and 5.

### DIFF
--- a/website/docs/r/container_node_pool.html.markdown
+++ b/website/docs/r/container_node_pool.html.markdown
@@ -134,8 +134,7 @@ type-specific `region` for regional clusters / `zone` for zonal clusters.
 * `node_config` - (Optional) The node configuration of the pool. See
     [google_container_cluster](container_cluster.html) for schema.
 
-* `node_count` - (Optional) The number of nodes per instance group. This field can be used to
-    update the number of nodes per instance group but should not be used alongside `autoscaling`.
+* `node_count` - (Required) The number of nodes per instance group. This field needs to be used otherwise `autoscaling` will never kick in.
 
 * `project` - (Optional) The ID of the project in which to create the node pool. If blank,
     the provider-configured project will be used.


danawillow:
Hi @minac, I'm not convinced this is the correct approach. This variable essentially says "node count should be X, hard stop". If autoscaling is also defined, then it won't work: any time terraform is re-run, it'll try to set the node count back to whatever was defined in the configuration. Instead, I'd recommend using `initial_node_count` to create the node pool with the specified number of nodes, while not enforcing that it remains the same size it was when it was originally created.

Because either `initial_node_count` or `node_count` can be specified, I don't feel comfortable changing the docs to say either are required. However, I'm happy to accept a PR that clarifies when to use one vs the other, especially in relation to autoscaling.

minac:
Hi @danawillow! Using initial_node_count seems like a better alternative. I just wish it was required to avoid the issue described above. If one does not use the field and sets the min of an autoscaling group to 2 and max to 4, the cluster will never create correctly, because it will never reach the minimum.
This was my experience. As you say this may not be a documentation issue, but a module issue, as an autoscaling group should guarantee that the number of nodes is between a min and a max.

danawillow:
Neither field should be required though, because they both have their use cases. `initial_node_count` works better when you want to use autoscaling, and `node_count` works better when you don't. We also used to check that node pools had a count > 0, but people wanted the ability to scale their cluster down (https://github.com/terraform-providers/terraform-provider-google/pull/752).

I think the correct approach is not to say that the field is required, but instead to add a note to the autoscaling docs to say that it only works if the initial node count starts in that range.

minac:
Sounds good

rileykarson:
Based on https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#my-cluster-is-below-minimum--above-maximum-number-of-nodes-but-ca-did-not-fix-that-why, I don't think that your `initial_node_count` needs to be within autoscaler ranges, so I'm going to close this out.

PR #3486, opened 2019-04-24
Fix invalid assumptions about Project ID's in BQ table

KJTsanaktsidis:
The bigquery table resource tries to infer the project and dataset from the table's full name. However, it does so incorrectly as it assumes that the project ID cannot have : or . characters in it, yet both are valid.

This is leading to errors like this for me when I run `terraform plan`:

```
* module.bigquery.google_bigquery_table.connect_product_analytics_table_user_search_filters: 1 error(s) occurred:

* module.bigquery.google_bigquery_table.connect_product_analytics_table_user_search_filters: google_bigquery_table.connect_product_analytics_table_user_search_filters: Invalid BigQuery table specifier. Expecting {project}:{dataset-id}.{table-id}, got outbound.io:outbound:connect_product_analytics_experiment.user_search_filters
```

As you can see, our project ID is `outbound.io:outbound`, so the string-splitting algorithm here doesn't work.

When using this branch, the plan works correctly.

Also - I'm targeting the PR against master, but we're actually still using the 1.20 version of the provider. Is there any way this fix can get backported to the 1.x series as well?

### DIFF
--- a/google/resource_bigquery_table.go
+++ b/google/resource_bigquery_table.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
-	"strings"
+	"regexp"
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/structure"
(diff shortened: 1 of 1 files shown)


KJTsanaktsidis:
Any chance of getting some 👀 on this soon? The fact that this resource is in our statefile breaks planning for our entire team (unless they're running a fork of the provider)

KJTsanaktsidis:
Hey, thanks for this. The main thing blocking us upgrading to a 2.x release is we are using the bigtable resources in a deprecated way that’ll need to be fixed (drawing a proper distinction between instances and clusters). I need to spend some time to have a good look at how yo upgrade it without accidentally deleting our cluster :)

rileykarson:
Fair enough! It's also always an option to use [multiple provider instances](https://www.terraform.io/docs/configuration/providers.html#alias-multiple-provider-instances) at different versions, although I wouldn't recommend that with more than 1-2 resources ahead or behind the rest.

PR #3502, opened 2019-04-26
GNUmakefile: add make dev

DeviaVir:
Hi!

This allows individual contributors to simply run `make dev` and `terraform init` (may need to remove `.terraform/plugins/` if it exists in your terraform directory) to quickly test their changes locally.

Thanks.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -2,10 +2,18 @@ TEST?=$$(go list ./... |grep -v 'vendor')
 WEBSITE_REPO=github.com/hashicorp/terraform-website
 PKG_NAME=google
 
+# Path to Terraform plugins
+PLUGIN_PATH ?= "${HOME}/.terraform.d/plugins"
+
 default: build
 
 build: fmtcheck
 	go install
+
+dev:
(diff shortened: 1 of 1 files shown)


rileykarson:
Hey @DeviaVir! Generally, our guidance for running test versions of the provider locally is to follow the [steps outlined in our CONTRIBUTING file](https://github.com/terraform-providers/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#running-a-local-copy-of-the-provider) to symlink the `make build` copies to the provider discovery dir. Is there something about that approach that doesn't work for you?

DeviaVir:
Hiya @rileykarson. I've seen the `make dev` logic for other providers, which skips another step you need to remember. I think the symlink approach works fine, so feel free to ignore if you're not a fan of this pattern.

rileykarson:
I think I'd prefer to keep a single pattern as the well lit path for now; developing for the Google provider can be complex between having several providers and Magic Modules, so keeping this simple is a priority. Regardless, thanks for the suggestion!

PR #3529, opened 2019-05-01
Vendor new Go GCP API clients

rileykarson:
Used the commands in https://github.com/terraform-providers/terraform-provider-google/pull/3528

Parallel to https://github.com/terraform-providers/terraform-provider-google-beta/pull/650


chrisst:
Looks like you are going to have to update config.go as some of the methods have been deprecated.

rileykarson:
Yep! See https://github.com/GoogleCloudPlatform/magic-modules/pull/1704 which links in to here. We'll need to merge these first since the Magician doesn't vendor anymore.

Issue #3209, opened 2019-03-09
Add support for Shielded VMs


danisla:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

[Shielded VMs](https://cloud.google.com/security/shielded-cloud/shielded-vm) give users the ability to create instances with verifiable integrity to prevent against malware and rootkits.

Sample from compute engine REST API:

```
POST https://www.googleapis.com/compute/v1/projects/my-project/zones/us-west1-a/instances
{
...
  "shieldedVmConfig": {
    "enableSecureBoot": true,
    "enableVtpm": true,
    "enableIntegrityMonitoring": true
  }
}
```

### New or Affected Resource(s)

* google_compute_instance
* google_compute_instance_template

### Potential Terraform Configuration

```tf
resource "google_compute_instance" "default" {
  shielded_vm_config = {
    enable_secure_boot = true
    enable_vtpm = true
    enable_integrity_monitoring = true
  }
}
```

### References

* [Shielded VM documentation](https://cloud.google.com/security/shielded-cloud/shielded-vm)
* [Modifying Shielded VMS](https://cloud.google.com/compute/docs/instances/modifying-shielded-vm)

mlauter:
hi! I'd like to work on this if no one else is.

rileykarson:
Go ahead! I've marked the issue assigned to HashiBot to indicate that you're working on it. (We can only assign users with merge access to issues / PRs, so we mark community members with HashiBot)

mlauter:
hi @rileykarson after i submitted this PR i saw the stuff about magic modules and that most of the activity for these resources is coming from there. just wanted to confirm whether that's how these changes ought to be made instead?

rileykarson:
[...]

PR #3531, opened 2019-05-01
Add support for Shielded VMs to compute_instance and compute_instance_template (#3209)

mlauter:
Adds support for [Shielded VMs](https://cloud.google.com/security/shielded-cloud/shielded-vm) to compute instances and instance templates.

Test output:
```
make testacc TEST=./google TESTARGS='-run=TestAccComputeInstance_shieldedVmConfig'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google -v -run=TestAccComputeInstance_shieldedVmConfig -timeout 240m -ldflags="-X=github.com/terraform-providers/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccComputeInstance_shieldedVmConfig
=== PAUSE TestAccComputeInstance_shieldedVmConfig
=== RUN   TestAccComputeInstance_shieldedVmConfigSkip
=== PAUSE TestAccComputeInstance_shieldedVmConfigSkip
=== CONT  TestAccComputeInstance_shieldedVmConfig
=== CONT  TestAccComputeInstance_shieldedVmConfigSkip
--- PASS: TestAccComputeInstance_shieldedVmConfigSkip (237.78s)
--- PASS: TestAccComputeInstance_shieldedVmConfig (248.79s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 248.853s

make testacc TEST=./google TESTARGS='-run=TestAccComputeInstance_basic'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google -v -run=TestAccComputeInstance_basic -timeout 240m -ldflags="-X=github.com/terraform-providers/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccComputeInstance_basic1
=== PAUSE TestAccComputeInstance_basic1
=== RUN   TestAccComputeInstance_basic2
=== PAUSE TestAccComputeInstance_basic2
=== RUN   TestAccComputeInstance_basic3
=== PAUSE TestAccComputeInstance_basic3
=== RUN   TestAccComputeInstance_basic4
=== PAUSE TestAccComputeInstance_basic4
=== RUN   TestAccComputeInstance_basic5
=== PAUSE TestAccComputeInstance_basic5
=== CONT  TestAccComputeInstance_basic1
=== CONT  TestAccComputeInstance_basic4
=== CONT  TestAccComputeInstance_basic3
=== CONT  TestAccComputeInstance_basic5
=== CONT  TestAccComputeInstance_basic2
--- PASS: TestAccComputeInstance_basic1 (156.77s)
--- PASS: TestAccComputeInstance_basic4 (239.65s)
--- PASS: TestAccComputeInstance_basic3 (271.10s)
--- PASS: TestAccComputeInstance_basic5 (292.26s)
--- PASS: TestAccComputeInstance_basic2 (187.23s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 344.041s

make testacc TEST=./google TESTARGS='-run=TestAccComputeInstanceTemplate_shieldedVmConfig'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google -v -run=TestAccComputeInstanceTemplate_shieldedVmConfig -timeout 240m -ldflags="-X=github.com/terraform-p
roviders/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccComputeInstanceTemplate_shieldedVmConfig
=== PAUSE TestAccComputeInstanceTemplate_shieldedVmConfig
=== RUN   TestAccComputeInstanceTemplate_shieldedVmConfigSkip
=== PAUSE TestAccComputeInstanceTemplate_shieldedVmConfigSkip
=== CONT  TestAccComputeInstanceTemplate_shieldedVmConfig
=== CONT  TestAccComputeInstanceTemplate_shieldedVmConfigSkip
--- PASS: TestAccComputeInstanceTemplate_shieldedVmConfigSkip (13.38s)
--- PASS: TestAccComputeInstanceTemplate_shieldedVmConfig (15.62s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 15.722s

make testacc TEST=./google TESTARGS='-run=TestAccComputeInstanceTemplate_basic'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google -v -run=TestAccComputeInstanceTemplate_basic -timeout 240m -ldflags="-X=github.com/terraform-providers/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccComputeInstanceTemplate_basic
=== PAUSE TestAccComputeInstanceTemplate_basic
=== CONT  TestAccComputeInstanceTemplate_basic
--- PASS: TestAccComputeInstanceTemplate_basic (11.32s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 11.372s

make testacc TEST=./google TESTARGS='-run=TestAccComputeInstanceFromTemplate_basic'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google -v -run=TestAccComputeInstanceFromTemplate_basic -timeout 240m -ldflags="-X=github.com/terraform-providers/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccComputeInstanceFromTemplate_basic
=== PAUSE TestAccComputeInstanceFromTemplate_basic
=== CONT  TestAccComputeInstanceFromTemplate_basic
--- PASS: TestAccComputeInstanceFromTemplate_basic (272.02s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 272.069s
```

### DIFF
--- a/google/compute_instance_helpers.go
+++ b/google/compute_instance_helpers.go
@@ -303,3 +303,29 @@ func resourceInstanceTags(d TerraformResourceData) *computeBeta.Tags {
 
 	return tags
 }
+
+func expandShieldedVmConfigs(d *schema.ResourceData) *computeBeta.ShieldedVmConfig {
+	if _, ok := d.GetOk("shielded_instance_config"); !ok {
+		return nil
+	}
+
+	prefix := "shielded_instance_config.0"
+	return &computeBeta.ShieldedVmConfig{
+		EnableSecureBoot:          d.Get(prefix + ".enable_secure_boot").(bool),
+		EnableVtpm:                d.Get(prefix + ".enable_vtpm").(bool),
+		EnableIntegrityMonitoring: d.Get(prefix + ".enable_integrity_monitoring").(bool),
+		ForceSendFields:           []string{"EnableSecureBoot", "EnableVtpm", "EnableIntegrityMonitoring"},
+	}
+}
+
+func flattenShieldedVmConfig(shieldedVmConfig *computeBeta.ShieldedVmConfig) []map[string]bool {
+	if shieldedVmConfig == nil {
+		return nil
+	}
+
+	return []map[string]bool{{
+		"enable_secure_boot":          shieldedVmConfig.EnableSecureBoot,
+		"enable_vtpm":                 shieldedVmConfig.EnableVtpm,
+		"enable_integrity_monitoring": shieldedVmConfig.EnableIntegrityMonitoring,
+	}}
+}
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -510,6 +510,37 @@ func resourceComputeInstance() *schema.Resource {
 				},
 			},
 
+			"shielded_instance_config": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				// Since this block is used by the API based on which
+				// image being used, the field needs to be marked as Computed.
+				Computed:         true,
+				DiffSuppressFunc: emptyOrDefaultStringSuppress(""),
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enable_secure_boot": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
+
+						"enable_vtpm": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
+						},
+
+						"enable_integrity_monitoring": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
+						},
+					},
+				},
+			},
+
 			"tags": {
 				Type:     schema.TypeSet,
 				Optional: true,
@@ -685,6 +716,7 @@ func expandComputeInstance(project string, zone *compute.Zone, d *schema.Resourc
 		DeletionProtection: d.Get("deletion_protection").(bool),
 		Hostname:           d.Get("hostname").(string),
 		ForceSendFields:    []string{"CanIpForward", "DeletionProtection"},
+		ShieldedVmConfig:   expandShieldedVmConfigs(d),
 	}, nil
 }
 
@@ -904,6 +936,7 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("scratch_disk", scratchDisks)
 	d.Set("scheduling", flattenScheduling(instance.Scheduling))
 	d.Set("guest_accelerator", flattenGuestAccelerators(instance.GuestAccelerators))
+	d.Set("shielded_instance_config", flattenShieldedVmConfig(instance.ShieldedVmConfig))
 	d.Set("cpu_platform", instance.CpuPlatform)
 	d.Set("min_cpu_platform", instance.MinCpuPlatform)
 	d.Set("deletion_protection", instance.DeletionProtection)
@@ -1364,6 +1397,22 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
 		if opErr != nil {
 			return opErr
 		}
+	}
+
+	if d.HasChange("shielded_instance_config") {
+		shieldedVmConfig := expandShieldedVmConfigs(d)
+
+		op, err := config.clientComputeBeta.Instances.UpdateShieldedVmConfig(project, zone, d.Id(), shieldedVmConfig).Do()
+		if err != nil {
+			return fmt.Errorf("Error updating shielded vm config: %s", err)
+		}
+
+		opErr := computeSharedOperationWaitTime(config.clientCompute, op, project, int(d.Timeout(schema.TimeoutUpdate).Minutes()), "shielded vm config update")
+		if opErr != nil {
+			return opErr
+		}
+
+		d.SetPartial("shielded_instance_config")
 	}
 
 	// We made it, disable partial mode
--- a/google/resource_compute_instance_template.go
+++ b/google/resource_compute_instance_template.go
@@ -389,6 +389,41 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 				},
 			},
 
+			"shielded_instance_config": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				ForceNew: true,
+				// Since this block is used by the API based on which
+				// image being used, the field needs to be marked as Computed.
+				Computed:         true,
+				DiffSuppressFunc: emptyOrDefaultStringSuppress(""),
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enable_secure_boot": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+							ForceNew: true,
+						},
+
+						"enable_vtpm": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
+							ForceNew: true,
+						},
+
+						"enable_integrity_monitoring": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
+							ForceNew: true,
+						},
+					},
+				},
+			},
+
 			"guest_accelerator": {
 				Type:     schema.TypeList,
 				Optional: true,
@@ -644,6 +679,7 @@ func resourceComputeInstanceTemplateCreate(d *schema.ResourceData, meta interfac
 		Scheduling:        scheduling,
 		ServiceAccounts:   expandServiceAccounts(d.Get("service_account").([]interface{})),
 		Tags:              resourceInstanceTags(d),
+		ShieldedVmConfig:  expandShieldedVmConfigs(d),
 	}
 
 	if _, ok := d.GetOk("labels"); ok {
@@ -839,6 +875,11 @@ func resourceComputeInstanceTemplateRead(d *schema.ResourceData, meta interface{
 	if instanceTemplate.Properties.GuestAccelerators != nil {
 		if err = d.Set("guest_accelerator", flattenGuestAccelerators(instanceTemplate.Properties.GuestAccelerators)); err != nil {
 			return fmt.Errorf("Error setting guest_accelerator: %s", err)
+		}
+	}
+	if instanceTemplate.Properties.ShieldedVmConfig != nil {
+		if err = d.Set("shielded_instance_config", flattenShieldedVmConfig(instanceTemplate.Properties.ShieldedVmConfig)); err != nil {
+			return fmt.Errorf("Error setting shielded_instance_config: %s", err)
 		}
 	}
 	return nil
--- a/google/resource_compute_instance_template_test.go
+++ b/google/resource_compute_instance_template_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
+	computeBeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
 )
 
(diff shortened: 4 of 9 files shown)


google/resource_compute_instance.go
```diff
@@ -488,6 +488,33 @@ func resourceComputeInstance() *schema.Resource {
 				},
 			},
 
+			"shielded_vm_config": {
+				Type:     schema.TypeList,
```
mlauter: I made this as `TypeList` with `MaxItems: 1` like some other configuration options for this resource. This would make the config a block rather than a map as suggested in the original issue description. I wasn't sure which was preferable here.

google/resource_compute_instance.go
```diff
@@ -488,6 +488,33 @@ func resourceComputeInstance() *schema.Resource {
...
+				MaxItems: 1,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enable_secure_boot": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
```
mlauter: Maybe it would be better to default to true?

mlauter:
Linter failures don't appear to be related to my code changes.

google/resource_compute_instance.go
rileykarson: Terraform maps are only string -> string, and appropriate for user-modifiable KV pairs like `labels`. Since this is representing a nested object, this is the correct Terraform schema.

google/resource_compute_instance.go
rileykarson: Sticking to the API default- I'd imagine `false`- is our best option here imo.

google/resource_compute_instance_template_test.go
```diff
@@ -498,6 +499,53 @@ func TestAccComputeInstanceTemplate_EncryptKMS(t *testing.T) {
...
+func TestAccComputeInstanceTemplate_shieldedVmConfigSkip(t *testing.T) {
+	t.Parallel()
+
+	var instanceTemplate computeBeta.InstanceTemplate
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeInstanceTemplateDestroy,
+		Steps: []resource.TestStep{
```
rileykarson: Do you mind adding an import step to this test like the one above?

google/resource_compute_instance_test.go
```diff
@@ -1063,6 +1064,51 @@ func TestAccComputeInstance_hostname(t *testing.T) {
...
+	t.Parallel()
+
+	var instance computeBeta.Instance
+	instanceName := fmt.Sprintf("terraform-test-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeInstanceDestroy,
+		Steps: []resource.TestStep{
```
rileykarson: Do you mind adding an import step here like in the test above?

google/resource_compute_instance_test.go
```diff
@@ -1063,6 +1064,51 @@ func TestAccComputeInstance_hostname(t *testing.T) {
...
+			{
+				Config: testAccComputeInstance_shieldedVmConfig(instanceName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeInstanceExists("google_compute_instance.foobar", &instance),
+					testAccCheckComputeInstanceHasShieldedVmConfig(&instance, true, true, true),
```
rileykarson: I'm curious what happens if we include the block with two different shapes; mixed false+true values, and an entirely false block. Based on your defaults, we can define:

```
	shielded_vm_config {
		enable_secure_boot          = true
		enable_vtpm                 = true
	}
```

```
	shielded_vm_config {}
```

Do you mind adding instances with definitions like these in the `testAccComputeInstance_shieldedVmConfig` config?

google/resource_compute_instance.go
mlauter: got it, makes sense!

google/resource_compute_instance.go
mlauter: Did some more testing, and it looks like the api defaults are actually
```
secureBoot - false
enableVtpm - true
enableIntegrityMonitoring - true
```

This feels confusing to me and my instinct is to stick with `false` but let me know if that doesn't seem right to you.

google/compute_instance_helpers.go
```diff
@@ -303,3 +303,29 @@ func resourceInstanceTags(d TerraformResourceData) *computeBeta.Tags {
...
+	if _, ok := d.GetOk("shielded_vm_config"); !ok {
+		return nil
+	}
+
+	prefix := "shielded_vm_config.0"
+	return &computeBeta.ShieldedVmConfig{
+		EnableSecureBoot:          d.Get(prefix + ".enable_secure_boot").(bool),
+		EnableVtpm:                d.Get(prefix + ".enable_vtpm").(bool),
+		EnableIntegrityMonitoring: d.Get(prefix + ".enable_integrity_monitoring").(bool),
+		ForceSendFields:           []string{"EnableSecureBoot", "EnableVtpm", "EnableIntegrityMonitoring"},
```
mlauter: Without this empty fields (including `false`) do not get sent

rileykarson: 👍

google/resource_compute_instance.go
```diff
@@ -495,6 +495,33 @@ func resourceComputeInstance() *schema.Resource {
 				},
 			},
 
+			"shielded_vm_config": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
```
```suggestion
				Optional: true,
				// Since this block is used by the API based on which
				// image being used, the field needs to be marked as Computed.
				Computed: true
```

rileykarson: Do you mind adding this, including the comment? While we're trying to avoid uses of `Optional` + `Computed` together in the provider going forward, this is an appropriate place to use it.

Because the API sets values by default when the block is unsent in the request, users will see a diff if they're using a shielded image without this block defined. We can't require them to define it though, because that would count as a breaking change. Computed here means that if the field is undefined in config, Terraform won't try and make a diff.

google/resource_compute_instance.go
rileykarson: I'm inclined to say we should mirror the API defaults; Terraform only computes default values if a block is defined, so we'd see different values when we defined

```hcl
shielded_vm_config {}
```

vs not defining a block at all.

google/resource_compute_instance_template_test.go
```diff
@@ -520,6 +521,136 @@ func TestAccComputeInstanceTemplate_soleTenantNodeAffinities(t *testing.T) {
...
+			{
+				ResourceName:      "google_compute_instance_template.foobar",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccComputeInstanceTemplate_shieldedVmConfigSkip(t *testing.T) {
```
rileykarson: I'm inclined to omit this test since it's nearly identical to https://github.com/terraform-providers/terraform-provider-google/blob/b1e181e09ffc6492d17ccb220f0c03ccd6d39ef3/google/resource_compute_instance_template_test.go#L18. You can add `testAccCheckComputeInstanceTemplateHasShieldedVmConfig` to that test if you'd like, though.

google/resource_compute_instance_template_test.go
```diff
@@ -520,6 +521,136 @@ func TestAccComputeInstanceTemplate_soleTenantNodeAffinities(t *testing.T) {
...
+			{
+				ResourceName:      "google_compute_instance_template.foobar",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccComputeInstanceTemplate_shieldedVmConfig3(t *testing.T) {
```
rileykarson: I think cases 3 and 4 are redundant with 1 and 2, they're testing how Terraform's defaults work more than they're testing novel resource behaviour. Let me know if you think I'm missing something, though.

google/resource_compute_instance_template.go
```diff
@@ -389,6 +389,37 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 				},
 			},
 
+			"shielded_vm_config": {
```
rileykarson: The suggestions for compute instance apply here as well, unless this resource behaves differently in your experience.

website/docs/r/compute_instance.html.markdown
```diff
@@ -134,6 +134,9 @@ The following arguments are supported:
 
 * `tags` - (Optional) A list of tags to attach to the instance.
 
+* `shielded_vm_config` - (Optional) Enable [Shielded VM](https://cloud.google.com/security/shielded-cloud/shielded-vm) on this instance. Shielded VM provides verifiable integrity to prevent against malware and rootkits. Defaults to disabled. Structure is documented below.
```
rileykarson: Nice! I was gonna ask if we could list the images, but you already did.

rileykarson: (also, remember to update the defaults in these docs too)

google/resource_compute_instance_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/terraform/helper/acctest"
```
rileykarson: Same comments about the tests as I had for the instance template one, cases 3 and 4 seem unnecessary and `Skip` is redundant with the `_basic` tests.

google/resource_compute_instance.go
mlauter: that's a really good point

google/resource_compute_instance.go
mlauter: given that, should this field also be added here? https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_instance_from_template.go#L47

google/resource_compute_instance_template_test.go
mlauter: I initially wanted to add the `lacksShieldedVmConfig` check to the `basic` test. That would just make sure that omitting shieldedVmConfig from the instance config would mean the block was absent rather than `shieldedVmConfig {}` which would result in an api error. 

The issue is that `TestAcccomputeInstance(Template)_basic` uses `compute.Instance` and not `compute.BetaInstance`. This actually was a bit of a point of confusion for me. `ShieldedVmConfig` settings _are_ available in the v1 REST api (under the name `ShieldedInstanceConfig`, but neither `ShieldedInstanceConfig` nor `ShieldedVmConfig` are available in the generated go structs for the v1 api. I wasn't sure where that left the provider in terms of including the block in the regular vs beta provider.

google/resource_compute_instance.go
rileykarson: Nope- that's for a specific class of field where having the field defined as null is distinct from the field not being defined. That's not the case here.

google/resource_compute_instance_template_test.go
rileykarson: I'd be happy with it being added to any test that's already using the beta instance, or changing that test to use the beta instance. They're pretty indistinguishable.

I hadn't noticed that the field was renamed from Beta -> GA, how do you feel about using the GA naming in the Terraform schema name for the block? That way we'll be aligned with the finalised naming of the feature.

google/resource_compute_instance_test.go
```diff
@@ -1115,7 +1116,9 @@ func TestAccComputeInstance_secondaryAliasIpRange(t *testing.T) {
 func TestAccComputeInstance_hostname(t *testing.T) {
```
mlauter: slightly weird place, but was less disruptive to convert this test to use computeBeta than the basic* tests in this file

mlauter:
Latest update:
- removes redundant tests
- moves `lacksShieldedVmConfig` check to a pre-existing test in `instance_test` and `instance_template_test`
- renames `shielded_vm_config` to `shielded_instance_config` to match the v1 REST api (but only for the config key, i left the variables names as is to match the google struct name, but happy to revisit)
- Updates defaults to match api defaults (and docs)
- Adds computed property

Thanks so much for being so patient and helpful!

google/resource_compute_instance_test.go
rileykarson: 👍

PR #3540, opened 2019-05-02
IAM Policy on subnets is now GA, moving into non-beta provider

reechar-goog:
This feature is now in the GA V1 compute API: https://cloud.google.com/compute/docs/reference/rest/v1/subnetworks/setIamPolicy

Was promoted from beta to GA in the gcloud sdk as well on 2/12/2019 https://cloud.google.com/sdk/docs/release-notes#compute_engine_9


rileykarson:
Hey @reechar-goog! This file is managed from [Magic Modules](https://github.com/GoogleCloudPlatform/magic-modules), so it's best if we make these changes there instead of in this repo, since we'll need to duplicate them to MM regardless.

Each file you edited will have an equivalent in https://github.com/GoogleCloudPlatform/magic-modules/tree/master/third_party/terraform; if you remove the version switches, MM should generate this same change.

reechar-goog:
Sounds good, I'll make the corresponding change over in Magic Module removing the version switch

Issue #3200, opened 2019-03-07
vertical pod autoscaling beta feature for google_container_cluster


mattnworb:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

[Vertical pod autoscaling](https://cloud.google.com/kubernetes-engine/docs/how-to/vertical-pod-autoscaling) is available in beta now in GKE. Can support for it be added to `terraform-provider-google-beta` as well?

### New or Affected Resource(s)

* google_container_cluster

### References

* https://cloud.google.com/kubernetes-engine/docs/how-to/vertical-pod-autoscaling

deenski:
100% down to help implement this. I could use a little guidance, but would love to get this going.

deenski:
Looking deeper into this, this feature should already either be enabled, or is very close to working. Looking at the v1beta1/container-gen.go and v1beta1/container-api.json ... Everything is in place for this to work. I don't really know where/how to update the resource definition, but will gladly do so if needed to get this feature working with the beta provider.

mattnworb:
@deenski I believe the place to update (or at least one of them) would be https://github.com/GoogleCloudPlatform/magic-modules

rileykarson:
@mattnworb is correct! MM is probably the best place to change this, specifically to https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/resources/resource_container_cluster.go.erb

https://github.com/terraform-providers/terraform-provider-google-beta/commit/d1bf585724cd9854c2476302efa5ebdfedfc580a is a good template, since that change added a similar field; https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/resources/resource_container_cluster.go.erb#L299-L309 shows what we do to omit beta fields from the GA provider for handwritten resources like GKE cluster. This field can probably be omitted entirely, since it's be an entirely new field.

[...]

PR #3584, opened 2019-05-07
Add support for vertical pod autoscaling

tinkerborg:
This implements vertical pod autoscaling for GKE as discussed in https://github.com/terraform-providers/terraform-provider-google/issues/3200

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -1743,12 +1743,10 @@ func expandVerticalPodAutoscaling(configured interface{}) *containerBeta.Vertica
 	if len(l) == 0 {
 		return nil
 	}
-	result := &containerBeta.VerticalPodAutoscaling{}
 	config := l[0].(map[string]interface{})
-	if enabled, ok := config["enabled"]; ok && enabled.(bool) {
-		result.Enabled = true
+	return &containerBeta.VerticalPodAutoscaling{
+		Enabled: config["enabled"].(bool),
 	}
-	return result
 }
 
 func expandPodSecurityPolicyConfig(configured interface{}) *containerBeta.PodSecurityPolicyConfig {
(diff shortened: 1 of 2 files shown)


google/resource_container_cluster.go
```diff
@@ -610,6 +610,21 @@ func resourceContainerCluster() *schema.Resource {
 				},
 			},
 
+			"vertical_pod_autoscaling": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Computed: true,
```
chrisst: We try to avoid setting something as `Computed` unless there's a default that is sent back from the API when nothing is passed. Is vertical_pod_autoscaling part of the API response even if this isn't present and it hasn't been configured upstream?

tinkerborg: Thanks, that was just an artifact of me cribbing off of other definitions. I've removed it.

chrisst:
Thanks. Can you add a test to exercise the new fields? If you want to provide a best effort implementation and I'll make sure it's running correctly.

tinkerborg:
> Thanks. Can you add a test to exercise the new fields? If you want to provide a best effort implementation and I'll make sure it's running correctly.

Done, hopefully I got it somewhat correct!

luis-silva:
@tinkerborg Can you add this to the google provider beta as well? VPA is still under beta and it's needed when bringing a cluster with other beta functionality.
Thanks.

danawillow:
Good catch @luis-silva: this is a beta feature, so it should be in the beta provider instead. Our team (in this case, @chrisst) can take care of adding it to [magic modules](https://github.com/GoogleCloudPlatform/magic-modules) (the source of truth for our terraform resources), but if you want credit for the PR in the terraform provider, it probably makes sense to open this PR against https://github.com/terraform-providers/terraform-provider-google-beta (or directly in magic modules).

chrisst:
As dana said, you can move this PR against the beta provider if you want the changed lines to be attributed to you. Otherwise the upstream PR will be generating the code and you'll still get credit as a Magic Modules contributor.

luis-silva:
@tinkerborg Thanks for the work, I actually started doing the work myself until I stumbled on your PR. Once you have a chance, please re-open the PR against the beta provider so these guys can get it merged quickly.
I just opened a similar PR there for intranode visibility. I currently have to edit the clusters post creation for these two options making it a very slow process. I'm eager to get these two PRs merged to have the full functionality at create time in my clusters, rather than having to use work arounds to edit the cluster post creation.

tinkerborg:
Thanks everyone, I've resubmitted to https://github.com/terraform-providers/terraform-provider-google-beta/pull/749

I'll go ahead and close this one.

PR #3622, opened 2019-05-14
Update container cluster docs subnetwork requirement

adnang:


danawillow:
I think this is only mostly true. If you want to use the default network and specify the subnetwork, I believe you can do that. It's a little bit silly, because you could just not specify either of them and then you get defaults for both, but I think maybe being explicit about it needing to be set if not using the default is probably better.

danawillow:
Closing due to lack of activity

Issue #3650, opened 2019-05-17
Clusters with multiple node pools regularly need re-runs


Stono:
### Affected Resource(s)

* google_container_node_pool

Hey, 
This happens pretty much every time we do a cluster upgrade:

```
Error: Error applying plan:

1 error(s) occurred:

* module.kubernetes.google_container_node_pool.normal-prod-20190401: 1 error(s) occurred:

* google_container_node_pool.normal-prod-20190401: googleapi: Error 400: Operation operation-1558074043943-c4492e58 is currently upgrading cluster delivery-platform. Please wait and try again once it is done., failedPrecondition

```

I just have to wait a couple of minutes and run it again to do the next pool.

Not debilitating, but frustrating :-) 

b/299443045

emilymye:
Answers to the following questions would be helpful:

- What version of the provider are you using?
- Is this specifically on node-pool creation or update?

If you post your debug logs (run with TF_LOG="DEBUG"), we can see exactly what requests are causing the conflict and determine whether you're timing out on retries or if we aren't retrying the operation. 

Is the node pool change being attempted during another change by Terraform? Or just some cluster update outside of Terraform?

Stono:
Apologies it was raised in a bit of a rush.

I'll push through a cluster upgrade with debug enabled on Monday for you.

Latest version of the provider btw and the cluster had 5 pools

Chupaka:
@Stono, a friendly reminder, just in case :)

Stono:
Doing a cluster upgrade today so will try and catch this.  It's just done it in one environment but i didn't set TF_LOG="DEBUG", doh.

rileykarson:
@ndmckinley: #3726 / a successor to #3726 should resolve this, so assigning to you for now. Feel free to hand off to me

Stono:
'ere you go! 
[tf.log](https://github.com/terraform-providers/terraform-provider-google/files/3336092/tf.log)

Stono:
BTW i upgraded to `2.9.1` and got the same thing

Stono:
Still getting this btw:
```
Error: Error applying plan:

1 error occurred:
	* module.kubernetes.google_container_node_pool.normal-20190618: 1 error occurred:
	* google_container_node_pool.normal-20190618: googleapi: Error 400: Operation operation-1564663845524-bf1f7474 is currently upgrading cluster delivery-platform. Please wait and try again once it is done., failedPrecondition

```

[...]

PR #3726, opened 2019-05-27
Allow container cluster to wait for status RUNNING on import.

ndmckinley:
I think this is the simplest possible configuration that will allow Terraform to wait for a container to finish creating.  It won't generalize that well, but it will work fine in Container for now.  If it looks good I can bring it up into Magic Modules.

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -1908,6 +1908,24 @@ func resourceContainerClusterStateImporter(d *schema.ResourceData, meta interfac
 	d.Set("name", clusterName)
 	d.SetId(clusterName)
 
+	if config.WaitOnImport {
+		err := resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
+			name := containerClusterFullName(project, location, clusterName)
+			cluster, err := config.clientContainerBeta.Projects.Locations.Clusters.Get(name).Do()
+			if err != nil {
+				return resource.NonRetryableError(err)
+			}
+			if cluster.Status != "RUNNING" {
+				return resource.RetryableError(fmt.Errorf("Cluster %q has status %q with message %q", d.Get("name"), cluster.Status, cluster.StatusMessage))
+			}
+			return nil
+		})
+
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return []*schema.ResourceData{d}, nil
 }
 
--- a/google/config.go
+++ b/google/config.go
@@ -51,12 +51,13 @@ import (
 // Config is the configuration structure used to instantiate the Google
 // provider.
 type Config struct {
-	Credentials string
-	AccessToken string
-	Project     string
-	Region      string
-	Zone        string
-	Scopes      []string
+	Credentials  string
+	AccessToken  string
+	Project      string
+	Region       string
+	Zone         string
+	Scopes       []string
+	WaitOnImport bool
 
 	client    *http.Client
 	userAgent string
--- a/google/provider.go
+++ b/google/provider.go
@@ -74,6 +74,12 @@ func Provider() terraform.ResourceProvider {
 				Optional: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+
+			"wait_on_import": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 
 		DataSourcesMap: map[string]*schema.Resource{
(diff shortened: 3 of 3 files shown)


chrisst:
@ndmckinley we talked about this in the sync yesterday and wanted to get buy-in from paddy before we move forward.
I think the 1 thing we were hung up on was should this be a toggle-able behavior for all MM resources (plus this and composer) or should we just make it the default behavior for this and any other long running creates.

paddycarver:
A couple things:

* First, making this a global provider options is a bit :/. It would feel better to make this a thing we could include on an import-by-import basis, even if it defaults to on. I know we can't include it in the config, but we could shoe-horn it in as an option that gets added to the ID passed in for the import?
* Second, it looks like we're waiting on import, but is it worth doing this as a wait on read instead of on import? Import will call into read, and will also cover cases where the resource is tainted and `terraform untaint` is used.

chrisst:
I like the concept of "wait on read". On most happy paths the polling will have happened during the `apply` phase and will have reached some sort of terminal state. However I think this would work best if there was a way to recover the `Operation` and check the state of that rather than the cluster's state. I think that way we can use wait on read for partially created resources where we captured a partial state (ignoring taint for this argument) and also for import. It would avoid having to catch all the possible terminal states (RUNNING is only 1). Additionally it would avoid the scenario where a cluster's state was in an intermediate status but the initial Create operation had succeeded (I can't think of a real scenario where this might be the case other than maybe a 'stuck' cluster?".

rileykarson:
> Additionally it would avoid the scenario where a cluster's state was in an intermediate status but the initial Create operation had succeeded

GKE clusters will leave the `RUNNING` state and enter the `UPGRADING` (or similarly named) state in some circumstances, particularly when used with Terraform. Cluster Masters are sized based on the size of your nodes, and our mechanism where we delete a small default node pool and then create fine-grained node pools often leaves clusters to immediately upgrade (really a resize) after creating.

ndmckinley:
It does seem to me like it's a cleaner, smaller change with less chance of affecting users other than the ones who want this behavior to leave it in import.  If it seems like a good idea to put it in read, I'm happy to do that.  I'm a little unsure about the idea of throwing an option into the import id, e.g. `wait_on_read!regular_name_of_resource` or what have you.  I think that since it *is* configuration, it probably belongs in some sort of existing center of configuration, and this is the only one that exists and is accessible at import time.

ndmckinley:
ping!  Just looking for a confirmation that you prefer for this to go in read and to be a switch on the string-typed import id.  I didn't love the way that feels, but said I'd be happy to implement it that way if you're sure.  :)

google/resource_container_cluster.go
```diff
@@ -1908,6 +1908,24 @@ func resourceContainerClusterStateImporter(d *schema.ResourceData, meta interfac
...
 	d.SetId(clusterName)
 
+	if config.WaitOnImport {
+		err := resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
+			name := containerClusterFullName(project, location, clusterName)
+			cluster, err := config.clientContainerBeta.Projects.Locations.Clusters.Get(name).Do()
+			if err != nil {
+				return resource.NonRetryableError(err)
+			}
+			if cluster.Status != "RUNNING" {
```
ndmckinley: A comment from Arpit:

>Sometimes we've seen clusters in Error/Failed state where Get() would work but state is never going to change.
It might be better to fail the import right away.

rileykarson:
Given that out of band `status` changes can make it so Terraform can't update the cluster, I wonder if there are any cases where we don't want to wait until we're in a terminal state so we can make operations against it?

Today, we fail out if the status isn't `RUNNING` but I think it would be reasonable to always wait on `PROVISIONING`, `RECONCILING`, `STOPPING` to become terminal, and error only on `ERROR` / `DEGRADED`.

ndmckinley:
Is that a proposal for causing `Read` to wait for any of those states?  I'm down.  Is that consensus on your side or should I wait for the Thursday meeting to happen before making that change?

rileykarson:
Let me raise it in Slack, I'll post here with the outcome.

ndmckinley:
Any updates?  :stuck_out_tongue_winking_eye:

rileykarson:
Talked it over and we have consensus- if we stuck this in `Read`, `terraform plan` could potentially take a long time. Instead, we can wait in a few locations:

* Post-Create
* Pre-Update
* Pre-Delete
* At the end of the import fn

In Read, we can error out on terminal error states instead of non-`RUNNING` states (as we do now); in those other locations, we'll wait for a terminal state and error out if it's non-`RUNNING` (unless `DEGRADED` is a valid state!)

rileykarson:
Also related: https://github.com/terraform-providers/terraform-provider-google/issues/3650

ndmckinley:
Great, I'll do that then!

ndmckinley:
Superceded by https://github.com/GoogleCloudPlatform/magic-modules/pull/2021 - which I still need to test after it runs through the Magician.

Issue #2115, opened 2018-09-26
google_container_cluster tries to recreate cluster always when used in combination with google_container_node_pool


mpgomez:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment
* If an issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to "hashibot", a community member has claimed the issue already.

### Terraform Version
Terraform v0.11.8
+ provider: google: 1.18

### Affected Resource(s)
* google_container_cluster
* google_container_node_pool

### Terraform Configuration Files

```tf
resource "google_container_cluster" "primary" {
  name               = "${var.cluster_name}"
  # If we want a regional cluster, should we be looking at https://cloud.google.com/kubernetes-engine/docs/concepts/regional-clusters#regional
  #  region = "${var.region}"
  zone               = "${var.main_zone}"
  additional_zones   = "${var.additional_zones}"
  # Node count for every region
  initial_node_count = 1
  project            = "${var.project}"
  remove_default_node_pool = true
  enable_legacy_abac = true

  node_config {
    oauth_scopes = [
      "https://www.googleapis.com/auth/compute",
      "https://www.googleapis.com/auth/devstorage.read_write",
      "https://www.googleapis.com/auth/sqlservice.admin",
      "https://www.googleapis.com/auth/cloud-platform",
      "https://www.googleapis.com/auth/logging.write",
      "https://www.googleapis.com/auth/monitoring",
    ]
  }
  addons_config {
    horizontal_pod_autoscaling {
      disabled = false
    }
  }
}

resource "google_container_node_pool" "nodepool" {
  name               = "${var.cluster_name}nodepool"
  zone               = "${var.main_zone}"
  cluster            = "${google_container_cluster.primary.name}"
  node_count         = "${var.node_count}"

  autoscaling {
    min_node_count = "${var.min_node_count}"
    max_node_count = "${var.max_node_count}"
  }
}
```

### Debug Output
[...]

PR #3733, opened 2019-05-28
container_cluster.html.markdown: flip recommended

flokli:
don't advertise a separately managed node pool as recommended, until
https://github.com/terraform-providers/terraform-provider-google/issues/2115
is fixed.


flokli:
The diff looks larger than it is - I simply moved the separately managed node pool example down, and added a warning there.

rileykarson:
A separately managed node pool is recommended. Replying inline to the original issue.

PR #3745, opened 2019-05-29
Replace access_config argument with block

acmiyaguchi:
```
Error: Unsupported argument

  on main.tf line 21, in resource "google_compute_instance" "vm_instance":
  21:     access_config = {

An argument named "access_config" is not expected here. Did you mean to define
a block of type "access_config"?
```


rileykarson:
Actually, it looks like there's a couple other occurences in the file. Do you mind editing those as well?

acmiyaguchi:
@rileykarson Good catch. This is ready for another look.

Issue #3770, opened 2019-06-03
Expose service ipv4 cidr as attribute


fprimex:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

Please expose the service ipv4 cidr block for the `google_container_cluster` resource. It appears to be in the SDK:

https://github.com/googleapis/google-cloud-go/blob/master/container/container.go#L135-L138

This attribute differs from the value in `ip_allocation_policy`, and can be viewed with `gcloud container clusters describe <cluster-name>`.

### New or Affected Resource(s)

* `google_cluster_container`

### Potential Terraform Configuration

```tf
resource "google_container_cluster" "primary" {
...
}

# Elsewhere

   google_container_cluster.primary.service_ipv4_cidr

```

### References

* https://github.com/googleapis/google-cloud-go/blob/master/container/container.go#L131-L138
* https://cloud.google.com/sdk/gcloud/reference/container/clusters/create#--services-ipv4-cidr

paddycarver:
This was merged as GoogleCloudPlatform/magic-modules#1873. It should be part of our next release.

rileykarson:
Closing this out as it was released.

PR #3772, opened 2019-06-03
Add top-level services_ipv4_cidr to GKE clusters.

paddycarver:
There's a computed, output-only services_ipv4_cidr attribute for
GKE Clusters that we're currently not populating. This PR adds support
for them. See #3770.

```
=== PAUSE TestAccContainerCluster_basic
=== CONT  TestAccContainerCluster_basic
--- PASS: TestAccContainerCluster_basic (460.70s)
PASS
ok      github.com/terraform-providers/terraform-provider-google/google 460.713s
```

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -9,7 +9,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/errwrap"
-	"github.com/hashicorp/go-version"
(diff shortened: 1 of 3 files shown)


paddycarver:
This is supposed to be against GoogleCloudPlatform/magic-modules. Because my brain isn't working.

PR #3778, opened 2019-06-04
Add lifecycle block for node pool resources

NaurisSadovskis:
**What:** Update `google_container_node_pool` documentation to include rolling update of their GKE node pools. 

**Why:** If users have only single node pool configured, this change will prevent users from destroying their only node pool when performing configuration change. Lifecycle blocks are good practice in general, so it makes sense to include it in default example. I believe it should be there by default and users can remove if they choose. 

Existing documentation is already using lifecycle blocks ([compute_ssl_certificate](https://www.terraform.io/docs/providers/google/r/compute_ssl_certificate.html), [compute_instance_template](https://www.terraform.io/docs/providers/google/r/compute_instance_template.html#using-with-instance-group-manager)).

### DIFF
--- a/website/docs/r/container_cluster.html.markdown
+++ b/website/docs/r/container_cluster.html.markdown
@@ -58,6 +58,10 @@ resource "google_container_node_pool" "primary_preemptible_nodes" {
       "https://www.googleapis.com/auth/monitoring",
     ]
   }
+
+  lifecycle {
+    create_before_destroy = true
+  }
 }
 ```
 
--- a/website/docs/r/container_node_pool.html.markdown
+++ b/website/docs/r/container_node_pool.html.markdown
@@ -41,6 +41,10 @@ resource "google_container_node_pool" "primary_preemptible_nodes" {
       "https://www.googleapis.com/auth/monitoring",
     ]
   }
+
+  lifecycle {
+    create_before_destroy = true
+  }
 }
 ```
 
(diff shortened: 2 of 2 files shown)


andyshinn:
Won't this fail with an `alreadyExists` error? I already run into this trying to recreate node pools with `create_before_destroy` when the `random_id` doesn't change:

```
Error: error creating NodePool: googleapi: Error 409: Already exists: projects/default-3aef9459/locations/us-east1/clusters/application/nodePools/api-helpful-ostrich., alreadyExists
```

The only way to safely do this right now is with a `random` provider resource (I'm playing with using `random_pet` and then `taint` the `random_pet` to force recreation of the node pool.

rileykarson:
This PR got lost. Sorry about that! As @andyshinn pointed out, this wouldn't work in practice because the name would conflict with the already-created node pool. It would be possible by omitting the `name` (the node pool resource automatically chooses a name).

I'd expect rolling updates of nodes to involve a series of steps that can't be encapsulated in a `terraform apply`; creating a new pool, cordoning the nodes in K8S, and then scaling down + deleting the old pool. Because of that, I'm not sure I'd like to encourage `create_before_destroy` here.

NaurisSadovskis:
@rileykarson , @andyshinn - no problem. I forgot to follow up myself. As for the lifecycle block, you're absolutely right. It doesn't make sense to add the block when a name does not change.

PR #3804, opened 2019-06-06
Update container_cluster.html.markdown

w-leads:
Example of google_container_cluster resource in the doc didn't work due to lack of `=`


danawillow:
Looks like this got missed, but fixed in https://github.com/GoogleCloudPlatform/magic-modules/pull/1925. Thanks @w-leads!

PR #3852, opened 2019-06-14
capture current hashibot config

appilon:
This is where hashibot configuration is captured for operation against this repo.

Once this PR is merged, as well as the other open PRs to achieve "parity" with what is currently running for the bot in production, I can push the new bot to production. That is to say there is no implication by hitting merge at the moment, but soon this is the place to configure google provider specific bot behaviors. You can also override/disable some of the global behaviors that are defined here:

https://github.com/terraform-providers/.hashibot/blob/master/.hashibot.hcl

Documentation on this system is in progress.


chrisst:
ℹ️ We will need to make sure https://github.com/terraform-providers/terraform-provider-google/pull/3852 gets merged first.

appilon:
@chrisst Looks like you referenced this issue, probably a typo?

chrisst:
ack thanks, updated.

PR #3883, opened 2019-06-19
Updated composer output variables documentation for configs 

michyliao:
Terraform docs misleading - updated docs with proper configurations for Composer's config output variables


danawillow:
Thanks @michyliao!

Issue #3980, opened 2019-07-04
terraform 0.12.x Inappropriate value for attribute "nat_ips" in google_compute_router_nat


scippio:
terraform -v
Terraform v0.12.3
+ provider.google v2.10.0

```
Error: Incorrect attribute value type

  on gke-cluster.tf line 107, in resource "google_compute_router_nat" "egress_nat_production":
 107:   nat_ips                            = ["${google_compute_address.egress_production.*.self_link}"]

Inappropriate value for attribute "nat_ips": element 0: string required.
```

tysen:
Hi @scippio - can you try `nat_ips = "$google_compute_address.egress_production[*].self_link"`?

TheMagoo73:
Thanks @tysen - that's resolved the same issue I'd just run into, perfect timing! Hopefully it will resolve @scippio too.

FakeBuddy:
This parameter should be fixed in documentation https://www.terraform.io/docs/providers/google/r/compute_router_nat.html

tysen:
Docs fixed in https://github.com/GoogleCloudPlatform/magic-modules/pull/2056

dpkirchner:
The error occurs, for me, when using the lines in the updated documentation. The easiest way to reproduce this is by copying the production-like configuration as-is ([at the above-referenced documentation URL](https://www.terraform.io/docs/providers/google/r/compute_router_nat.html)). This is with terraform v0.12.6 and provider.google v2.12.0:

```
Error: Incorrect attribute value type

  on network-us-west2.tf line 32, in resource "google_compute_router_nat" "advanced-nat":
  32:   nat_ips                            = ["${google_compute_address.address[*].self_link}"]

Inappropriate value for attribute "nat_ips": element 0: string required.
```

Replacing line 32 with the string:

```
nat_ips = google_compute_address.address[*].self_link
```

works fine.

tysen:
Thanks @dpkirchner - fixed in https://github.com/GoogleCloudPlatform/magic-modules/pull/2125

PR #4004, opened 2019-07-11
Fix nat_ips syntax in example

TheMagoo73:
Syntax in the previous version didn't work with Terraform 0.12. This update is based on discussion in #3980


Chupaka:
Does this new syntax work with 0.11?

TheMagoo73:
Good call; with the new splat syntax it won't work on 0.11 - the old entry was correct for that version.

danawillow:
I believe this is no longer needed after https://github.com/GoogleCloudPlatform/magic-modules/pull/2125, thanks though @TheMagoo73!

PR #4009, opened 2019-07-12
Fix: setting initial_node_count on imported resource forces full re-c…

damienleger:
…reation of google_container_cluster resource

As stated by the documentation https://www.terraform.io/docs/providers/google/r/container_cluster.html#node_pool to have a separately managed node pool one must set:
  initial_node_count = 1

On imported google_container_cluster resource, node pool is not separately managed (ie. node_config attribute is in the resource). Setting "remove_default_node_pool = true" is working and allow to remove node_config attribute from code of resource to handle pool separately, but adding "initial_node_count = 1" forces recreation of resource whereas it seems to only be useful on the bootstrap of resource by terraform.

This fix allows using the same code for resource bootstraped via terraform or imported ones.

Workaround is to add in resource:
  lifecycle {
    ignore_changes = [initial_node_count, ]
  }

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -290,7 +290,6 @@ func resourceContainerCluster() *schema.Resource {
 			"initial_node_count": {
 				Type:     schema.TypeInt,
 				Optional: true,
-				ForceNew: true,
 			},
 
 			"logging_service": {


damienleger:
The travis CI build is failing on files I had not modify. So what can I do about it?

danawillow:
Hi @damienleger, I don't believe this is the right way to fix the issue. Keeping the property as ForceNew makes more sense with what it is- since it's the _initial_ node count, then it shouldn't be able to be modified- you can only set it on cluster creation time, and no other time, and so changing it should be seen as wanting to create the cluster with a different initial node count.

I think setting the lifecycle block (or leaving initial_node_count empty until you wish to recreate the cluster) is probably the way to go.

I'm going to go ahead and close this PR, but thank you for the contribution!

PR #4026, opened 2019-07-15
add ability to update Bigtable instance cluster node count

### LINKED ISSUE #2521: Add update support for bigtable clusters (such as resizing)
I just changed the number of nodes and this is the diff that is produced:

```
-/+ google_bigtable_instance.prod-instance (new resource required)
      id:            "us-central1-prod" => <computed> (forces new resource)
      cluster_id:    "us-central1-prod-cluster" => "us-central1-prod-cluster"
      display_name:  "us-central1-prod" => <computed>
      instance_type: "PRODUCTION" => "PRODUCTION"
      name:          "us-central1-prod" => "us-central1-prod"
      num_nodes:     "3" => "4" (forces new resource)
      project:       "grafanalabs-global" => <computed>
      storage_type:  "SSD" => "SSD"
      zone:          "us-central1-f" => "us-central1-f"
```

This is dangerous and we're manually scaling the cluster and adding a lifecycle block with `prevent_destroy`. You should be able to resize solely via terraform.

```
$ terraform -v
Terraform v0.11.10
+ provider.google v1.19.1
```

mackenziestarr:
This PR allows the Bigtable instance cluster `num_nodes` to updated without forcing the instance to be re-created. I also tried to support upgrading from a `DEVELOPMENT` to a `PRODUCTION` instance but the bigtable package only exposes that information at instance creation time. Switching to the `v2` admin API would allow this via [Instance.GetType](https://godoc.org/google.golang.org/genproto/googleapis/bigtable/admin/v2#Instance.GetType) but for now I've left the `instance_type` schema as `ForceNew`. All tests under `make testacc TEST=./google TESTARGS='-run=TestAccBigtableInstance*'` pass for me.

fixes #2521

### DIFF
--- a/google/resource_bigtable_instance.go
+++ b/google/resource_bigtable_instance.go
@@ -178,7 +178,6 @@ func resourceBigtableInstanceRead(d *schema.ResourceData, meta interface{}) erro
 	d.Set("project", project)
 
 	clusters := d.Get("cluster").([]interface{})
-
 	clusterState := []map[string]interface{}{}
 	for _, cl := range clusters {
 		cluster := cl.(map[string]interface{})
@@ -194,7 +193,6 @@ func resourceBigtableInstanceRead(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	err = d.Set("cluster", clusterState)
-
 	if err != nil {
 		return fmt.Errorf("Error setting clusters in state: %s", err.Error())
 	}
--- a/google/resource_bigtable_instance.go
+++ b/google/resource_bigtable_instance.go
@@ -15,6 +15,7 @@ func resourceBigtableInstance() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceBigtableInstanceCreate,
 		Read:   resourceBigtableInstanceRead,
+		Update: resourceBigtableInstanceUpdate,
 		Delete: resourceBigtableInstanceDestroy,
 
 		Schema: map[string]*schema.Schema{
@@ -25,10 +26,8 @@ func resourceBigtableInstance() *schema.Resource {
 			},
 
 			"cluster": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Required: true,
-				ForceNew: true,
-				MaxItems: 4,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"cluster_id": {
@@ -44,7 +43,6 @@ func resourceBigtableInstance() *schema.Resource {
 						"num_nodes": {
 							Type:         schema.TypeInt,
 							Optional:     true,
-							ForceNew:     true,
 							ValidateFunc: validation.IntAtLeast(3),
 						},
 						"storage_type": {
@@ -57,11 +55,9 @@ func resourceBigtableInstance() *schema.Resource {
 					},
 				},
 			},
-
 			"display_name": {
 				Type:     schema.TypeString,
 				Optional: true,
-				ForceNew: true,
 				Computed: true,
 			},
 
@@ -137,7 +133,7 @@ func resourceBigtableInstanceCreate(d *schema.ResourceData, meta interface{}) er
 		conf.InstanceType = bigtable.PRODUCTION
 	}
 
-	conf.Clusters = expandBigtableClusters(d.Get("cluster").(*schema.Set).List(), conf.InstanceID)
+	conf.Clusters = expandBigtableClusters(d.Get("cluster").([]interface{}), conf.InstanceID)
 
 	c, err := config.bigtableClientFactory.NewInstanceAdminClient(project)
 	if err != nil {
@@ -181,7 +177,8 @@ func resourceBigtableInstanceRead(d *schema.ResourceData, meta interface{}) erro
 
 	d.Set("project", project)
 
-	clusters := d.Get("cluster").(*schema.Set).List()
+	clusters := d.Get("cluster").([]interface{})
+
 	clusterState := []map[string]interface{}{}
 	for _, cl := range clusters {
 		cluster := cl.(map[string]interface{})
@@ -197,6 +194,7 @@ func resourceBigtableInstanceRead(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	err = d.Set("cluster", clusterState)
+
 	if err != nil {
 		return fmt.Errorf("Error setting clusters in state: %s", err.Error())
 	}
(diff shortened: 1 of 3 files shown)


rileykarson:
Hey @mackenziestarr! Thanks for your contribution. Unfortunately, I believe that changing the type of `cluster` from `TypeSet` to `TypeList` will count as a breaking change (as order will suddenly matter, resulting in diffs for some users), blocking us from applying this change until the `3.0.0` release of the provider.

Assuming the functionality discussed at the end in #2521 is exposed in the client libraries, we'll be able to add this in a future `2.X` series release. If that's the case, we'll supersede this PR with one exposing it there.

mackenziestarr:
Hi @rileykarson thanks for taking a look. I converted that to `TypeList` because of the issue with `TypeSet` elements with `ForceNew` causing the entire resource to be recreated reference here https://github.com/hashicorp/terraform/issues/15420#issue-239299444
> A change in any property in the set acts as if the entire set has changed (and if any of the properties in the set are ForceNew=true, then the entire resource gets re-created)

I see your point about `TypeList` not being ideal because it enforces ordering where there shouldn't be any. Is there a way to get around the bug above with `TypeSet` (possibly using a different structure for the implementation that is transparent to user's existing configs) — it doesn't seem to be a limitation of the Bigtable APIs as described in #2521 but an actual bug / unexpected property of using `TypeSet`

rileykarson:
I'm not aware of a workaround for the issue that wouldn't constitute a breaking change for users (I looked into the same topic a few days ago). I see two possible paths forward:

* Convert this field to a `TypeList` in `3.0.0`, assuming clusters are ordered when returned from the API. We have a limited ability to suppress diffs when they're returned unordered that we've used before, but it isn't transparent- I believe if any field has changed Terraform returns an unintuitive diff.

* The other gets a little gritty. Terraform can't tell which set elements have been modified / removed at update time. When an API only allows us to call `addX`/`modifyX`/`removeX` endpoints on subfields (effectively, this is how BT Cluster behaves) instead of supporting PUT or PATCH on the parent, we can't update the subfield. Terraform _can_ tell the final state of the field, so a PUT or PATCH endpoint or a similar function in a client library would both let us add / remove clusters *and* perform size updates transparently even while we continue to use `TypeSet`.

mackenziestarr:
@rileykarson just to clarify on the second option, how would that work with fields that can't be changed like `cluster_id` (and therefore should be marked in the schema with `ForceNew`)? Would the new API endpoint need to accept that in the `PUT` / `PATCH` payload and throw a validation error?  Thanks for all the great info so far, this is my first contribution.

rileykarson:
Ideally the client lib / API would throw a validation error if we attempted an impossible change, but we can work around it if it doesn't.

This functionality was recently added to the grpc client with https://code-review.googlesource.com/c/gocloud/+/42850. With that implementation, adding/removing clusters isn't possible (which could simplify our client code considerably :slightly_frowning_face: ). Regardless, it means we don't need to own much code in our `Update` fn and can just send our entire end state to the client lib.

Looking at this again, I think that we can safely make this change with approach #1, provided we can perform diff comparisons in a `CustomizeDiff` to avoid a breaking change. We've treated a `TypeList` as unordered in the past in GCE Subnetwork with https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_subnetwork.go#L164, if we add a similar custom diff function here we can do the same.

I believe all we need to do is make the same schema changes you've made, and add the CD to make sure that ordering of the list doesn't matter. `CustomizeDiff` functions can be pretty finicky- let me know if you'd like to take a shot at the implementation. If not, I can take over on implementing it.

mackenziestarr:
hey @rileykarson thanks for all the info I think with the prior art there I can take a shot at implementing a `CustomizeDiff` function for the change, I'll try to be back with an update soon.

rileykarson:
Sounds good! If you're stuck, feel free to reach out here.

mackenziestarr:
Hey @rileykarson when I went to implement this change I noticed this from the [`ResourceDiff` docs](https://godoc.org/github.com/hashicorp/terraform/helper/schema#ResourceDiff)
> All functions in ResourceDiff, save for ForceNew, can only be used on computed fields.

Given that, should I expect to have to modify the schema similar to [this PR](https://github.com/terraform-providers/terraform-provider-google/pull/1787/files#diff-fd74f2b3189f2693caf28225b82b8bc5L27)?

rileykarson:
Yep!

Because of that change, in the customizediff we'll need to verify that the new `#`` for cluster is at least 1 to simulate it being `Required: true` and `MinItems: 1`. You can see that value being pulled into `newCount` in this line: https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_subnetwork.go#L169

mackenziestarr:
Hey @rileykarson I managed to get a passing test for reordering clusters when making changes but I still need to handle the reordering cases where additional clusters are added / removed. Hopefully I'll have that soon!

google/resource_bigtable_instance.go
```diff
@@ -265,3 +312,77 @@ func expandBigtableClusters(clusters []interface{}, instanceID string) []bigtabl
...
+	var count int
+	if oldCount.(int) < newCount.(int) {
+		count = newCount.(int)
+	} else {
+		count = oldCount.(int)
+	}
+
+	// simulate Required:true and MinItems:1
+	if count < 1 {
+		return nil
```
rileykarson: We should error out here, right? That way we don't continue if users have specified no clusters / haven't specified a cluster.

google/resource_bigtable_instance.go
```diff
@@ -265,3 +312,77 @@ func expandBigtableClusters(clusters []interface{}, instanceID string) []bigtabl
 	}
 	return results
 }
+
+func resourceBigtableInstanceClusterReorderTypeList(diff *schema.ResourceDiff, meta interface{}) error {
+	keys := diff.GetChangedKeysPrefix("cluster")
```
rileykarson: I'd imagine we can omit this check and run through this logic every time, unless you have a reason otherwise?

google/resource_bigtable_instance_iam_test.go
```diff
@@ -12,6 +12,7 @@ func TestAccBigtableInstanceIamBinding(t *testing.T) {
 	t.Parallel()
 
 	instance := "tf-bigtable-iam-" + acctest.RandString(10)
+	cluster := "c-" + acctest.RandString(10)
```
rileykarson: Good catch!

google/resource_bigtable_instance.go
mackenziestarr: yeah good call, will return an error here

google/resource_bigtable_instance.go
mackenziestarr: yeah good call, i think i was initially just copying the customize diff example you linked

mackenziestarr:
@rileykarson refactored the code a bit in 376017de467c478f109afa8a90d061f892465b5e to make it cleaner and fixed up the diff with regard to #4156, thanks for the review!

mackenziestarr:
@rileykarson just bumping this, I think it’s ready to be merged

google/resource_bigtable_instance.go
```diff
@@ -57,11 +61,9 @@ func resourceBigtableInstance() *schema.Resource {
 					},
 				},
 			},
-
 			"display_name": {
 				Type:     schema.TypeString,
 				Optional: true,
-				ForceNew: true,
```
rileykarson: Since we haven't added the ability to update this field, we should keep it `ForceNew` right?

google/resource_bigtable_instance.go
```diff
@@ -207,6 +209,51 @@ func resourceBigtableInstanceRead(d *schema.ResourceData, meta interface{}) erro
...
+	}
+
+	c, err := config.bigtableClientFactory.NewInstanceAdminClient(project)
+	if err != nil {
+		return fmt.Errorf("Error starting instance admin client. %s", err)
+	}
+	defer c.Close()
+
+	if d.Get("instance_type").(string) == "DEVELOPMENT" {
+		return resourceBigtableInstanceRead(d, meta)
```
rileykarson: Since `num_nodes` can't be updated for development clusters (or set), we should probably error instead of silently doing nothing. As-is, setting a value for `num_nodes` on an existing development cluster results in a permadiff.

Erroring here would cause an apply-time error. Alternatively, if we add a `CustomizeDiff` step we can error earlier, at plan time. We can  check for these failure conditions:

* the new value of `instance_type` is `DEVELOPMENT` and `new_count` != 1

* the new value of `instance_type` is `DEVELOPMENT` and `cluster.0.num_nodes` != `nil` or `0` (I believe it'll always be 0, but I'm not 100% certain of that.)

google/resource_bigtable_instance.go
mackenziestarr: Oh yeah, good catch this can be fixed in another PR

google/resource_bigtable_instance.go
mackenziestarr: Sounds good to me, will update. Seems like we can rely on `0` being the default based on this https://github.com/hashicorp/terraform/blob/39b6a322e67625ed641695821b25dd718f50ca9a/helper/schema/schema.go#L1837-L1838

mackenziestarr:
@rileykarson no worries, that happens to me as well with github notifications sometimes — just updated with some new test cases to cover the failure conditions you pointed out.

rileykarson:
Adding changelog metadata:

```releasenote
`num_nodes` can now be updated in `google_bigtable_instance`
```

PR #4048, opened 2019-07-17
Use correct hyperlink in google_project_organization_policy data source.

lkastenson:
The hyperlink points to google_project instead of the intended google_project_organization_policy.


slevenick:
Thanks for the fix! I'll make sure this gets to the beta provider as well

PR #4086, opened 2019-07-23
Remove pubsub reference from bigtable_instance_iam

vipulagarwal:
There is a copy-paste error as there are reference of pubsub in the bigtable iam docs


website/docs/r/bigtable_instance_iam.html.markdown
```diff
@@ -18,7 +18,7 @@ Three different resources help you manage IAM policies on bigtable instances. Ea
 
 ~> **Note:** `google_bigtable_instance_iam_binding` resources **can be** used in conjunction with `google_bigtable_instance_iam_member` resources **only if** they do not grant privilege to the same role.
 
-## google\_pubsub\_subscription\_iam\_policy
+## google\_bigtable\_subscription\_iam\_policy
```
danawillow: Oh hold up, can you change where it says subscription to say instance instead?

vipulagarwal: @danawillow I have fixed it now, thanks for the heads up.

PR #4132, opened 2019-07-30
Deleting service default service accounts when creating a new project

gabrielgbim:
Deleting default service accounts after the creation of a google_project.
Didn't change the default behavior, just include a new optional parameter to allow the new behavior.
The new parameter works the same way as `auto_create_network`.

### DIFF
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -51,7 +51,7 @@ func resourceGoogleProject() *schema.Resource {
 			"delete_service_accounts": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  true,
+				Default:  false,
 			},
 			"name": {
 				Type:         schema.TypeString,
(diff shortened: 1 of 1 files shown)


slevenick:
Hey @gabrielgbim 

Thanks for the contribution, and sorry for taking so long to get eyes on it! It looks like there has been some drift in `resource_google_project` since you made your changes, can you update it?

Also, I'm seeing diffs when running the project tests, particularly around importing the `google_project` resource.

Test failure:
```
--- FAIL: TestAccProject_labels (14.60s)
    testing.go:569: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=23) "delete_service_accounts": (string) (len=5) "false"
        }
```

It looks like we are writing `delete_service_accounts: false` in state, but when we import it does not set that. 

I believe that dropping the Default: false will fix this issue, but you will have to also check for existence when checking the value

rileykarson:
Closing this PR since it's got conflicts and is fairly out-of-date. I'll explicitly link to it as a reference in the linked issue.

PR #4149, opened 2019-08-01
Updated google_folder.html

marinkovicpetar:
The page in the first example shows that you should use organization_id with value of 1234567. In the Import example, it's not clear whether organization_id is user, or folder_id is used. API call behind this import command is only accepting folder_id (can be checked when setting TF_LOG to trace and viewing the API call)


website/docs/r/google_folder.html.markdown
```diff
@@ -60,7 +60,7 @@ exported:
 
 ## Import
 
-Folders can be imported using the folder autogenerated `name`, e.g.
+Folders can be imported using the folder autogenerated `name`, and `folder_id`, not `organization_id` e.g.
```
danawillow: This sentence isn't quite right- it seems to imply that the import format has two parts, but it's only one. I think it's better to let the examples speak for themselves in this case, but since it does have two import formats (the `name` or the numerical id that makes up part of the name) we could change it to use a more generic term like id. I don't think we should be mentioning the `organization_id` at all, since there's no reason for it to be relevant here.
```suggestion
Folders can be imported using the folder's id, e.g.
```

Since the number was used above for the organization, I also think it would make sense to just use a different number in this example, that way there's no reason why anyone would think to put the organization in when trying to import a folder.

marinkovicpetar: Yes, you're right, it's best just to mention `folder_id`, the number confused me since it was already mentioned on the top of the page as `organization_id`.

PR #4307, opened 2019-08-21
Add service_networking_connection import documentation

KieranP:
Took me a while to find out how to import an existing service_networking_connection. Had to dive into the source code. Hopefully this docs update helps others.


website/docs/r/service_networking_connection.html.markdown
```diff
@@ -48,3 +48,17 @@ The following arguments are supported:
...
+
+## Import
+
+Service networking connections can be imported using the following accepted format:
+
+```
+$ terraform import google_service_networking_connection.foobar {{network}}:{{service}}
+```
+
+* `network` - Name of VPC network connected with service producers using VPC peering.
```
danawillow: Since `network` and `service` are already defined above, I would just remove them here to keep the docs a bit cleaner

rileykarson:
Done in https://github.com/GoogleCloudPlatform/magic-modules/pull/4041

PR #4315, opened 2019-08-22
Update description text

missinglink:
Please review. I'm not sure if this is the expected meaning but the current text seems to be missing a word.


rileykarson:
Merged as https://github.com/terraform-providers/terraform-provider-google/pull/4565

PR #4333, opened 2019-08-23
Vendor hashicorp/terraform@v0.12.7

rileykarson:
Enables https://github.com/GoogleCloudPlatform/magic-modules/pull/2229


rileykarson:
We'll merge that PR soon after merging this, in order to avoid the deprecated fn error living for long.

PR #4369, opened 2019-08-29
Missing custom route advertise on Cloud Router BGP Peer

reechar-goog:
This feature adds the missing parameters that allow a Cloud Router BGP Peer to have custom advertised routes. The following added arguments: advertise_mode, advertised_groups, and advertised_ip_ranges are available in the Google Cloud Compute API but missing from the terraform provider.


reechar-goog:
Looks like it was handled already by: https://github.com/GoogleCloudPlatform/magic-modules/pull/2271

PR #4415, opened 2019-09-06
Enable http timeout errors to be retried.

spew:
The `isRetryableError` function does handle `url.Error`s. However, those errors are being dropped on line 364 by a call to `errwrap.GetAllType(err, &googleapi.Error{})`.

This change ensures that `url.Error`s are checked for retry-ability.

### DIFF
--- a/google/utils.go
+++ b/google/utils.go
@@ -361,13 +361,24 @@ func retryTimeDuration(retryFunc func() error, duration time.Duration, errorRetr
 		if err == nil {
 			return nil
 		}
-		for _, e := range errwrap.GetAllType(err, &googleapi.Error{}) {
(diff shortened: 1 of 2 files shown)


spew:
Hi @rileykarson this catches all HTTP timeout errors, for example when a service is not responding, TLS handshakes fail, etc.

spew:
For example, errors with this message were NOT being retried: `net/http: request canceled (Client.Timeout exceeded while awaiting headers)`

rileykarson:
Great! That was the one I expected this to resolve as well.

spew:
Have terraform users been complaining about that?

rileykarson:
It's come up occasionally. We tend to see it pop up every few nights in our integration tests, since about 4-5 months ago.

PR #4432, opened 2019-09-09
Document STANDARD storage_class in storage_bucket

umairidris:
STANDARD is the current default but is not found in the list of supported values.


rileykarson:
I'll upstream this to Magic Modules and then merge.

PR #4489, opened 2019-09-18
Log `go list` errors when linting.

paddycarver:
See golangci/golangci-lint#401.


rileykarson:
Closing since the lint step was fixed.

PR #4491, opened 2019-09-18
Additional note: role_id must be camel case

mdeggies:
The role_id's must be camel case, and cannot contain `-` characters, so I added those notes to the definition. I keep seeing people on our team try to use `-`'s, hence the explicit mention.


danawillow:
@tysen when you get a chance can you upstream this?

tysen:
Will do.

PR #4551, opened 2019-09-25
Vendor upgrades required to upgrade to the provider sdk.

ndmckinley:
Changes only to `vendor/`, `go.mod`, and `go.sum`.

I strongly suspect this will break the tests, but we should merge it anyway - a generated PR that fixes the breakage will come soon after.


ndmckinley:
Mm, merge conflicts, delicious.  I'll fix it.

PR #4560, opened 2019-09-26
Ignore services that are known to be auto-enabled by configured services.

ubschmidt2:


google/resource_google_project_services.go
```diff
@@ -48,19 +84,43 @@ func resourceGoogleProjectServices() *schema.Resource {
...
 			},
 			"disable_on_destroy": {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  true,
 			},
+			"ignore_auto_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
```
emilymye: I'd rather not have this default to true, as the current behavior is not this.

ndmckinley:
We're considering deprecating / deleting this resource due to these concerns.  We think that `google_project_service` and https://github.com/terraform-google-modules/terraform-google-project-factory/tree/master/modules/project_services#project-api-activation solve a similar problem without the complexity of maintaining this map ourselves.

What do you think?

ubschmidt2:
Hi Emily, hi Nathan!

thanks so much for your review comments!

> We tend to really try avoiding hardcoding like this because of maintanence burden, especially in more recent development on the provider. I know there's the ignoredProjectServices list that somewhat undercuts my point, but this list will probably change more.

Yes, I fully agree. The dependency info should not be hardcoded here. Unfortunately, there's no public API for retrieving the dependency info... Let me think more about this. I'm exploring alternatives, e.g. just adding an optional attribute for that.

> Can I know more about what you are trying to do? Is there a reason using a variable with `google_project_service` and `count/foreach` does not work? Lately, we've been improving the behavior of that resource so it batches calls, so its much closer in quota usage to services.

I see two very related use cases:
(1) granular control for config owners over which APIs are actually enabled, ideally authoritatively (but need to solve the issue of APIs that are (or should be) enabled out-of-band). Sometimes, APIs are enabled just by browsing the cloud console. For many APIs, IAM bindings get added as part of the activation, in critical GCP projects you want to have tight control over both API activations and IAM bindings.
(2) graceful handling of APIs that are enabled out-of-band, if these APIs are known dependencies of explicitly enabled APIs. Recent example is the bigquery issue. 

> Mostly, I don't know if this solution is ideal. I don't think there's a way to detect a diff if you want to disable only one API that is auto-enabled while still ignoring other auto-enabled services.

I'm also not sure if ideal, but not for the concern you are mentioning. The auto-enabled API will appear as a diff once the API that auto-enables it is removed from the config. It don't need/want what you are describing - I think It's not desirable to disable an API that is auto-enabled by an API that is still configured, otherwise subtle breakages may result.

from Nathan:
> We're considering deprecating / deleting this resource due to these concerns. We think that `google_project_service` and https://github.com/terraform-google-modules/terraform-google-project-factory/tree/master/modules/project_services#project-api-activation solve a similar problem without the complexity of maintaining this map ourselves.
> 
> What do you think?

Yes, in it's current form, google_project_services is insufficient. Existing users should strongly consider moving to google_project_service. The provided module helps a bit to maybe save a few lines. On the downside, IIUIC, there's then no support at all for authoritative management of enabled APIs as this cannot be emulated on top of google_project_service.

Stefan

ndmckinley:
>On the downside, IIUIC, there's then no support at all for authoritative management of enabled APIs as this cannot be emulated on top of google_project_service.

That's correct, and that's intentional - the way the Google APIs are managed really doesn't support authoritative management, and we'd like to drop support for it in the provider.

ubschmidt2:
Thanks, Nathan! This makes sense. I agree. Closing this PR.

PR #4602, opened 2019-10-02
Update google_service_account_key.html.markdown

paulwoelfelcpl:
There is an error in the syntax of `kubernetes_secret`. The `data` element is missing an `=`.

See also https://www.terraform.io/docs/providers/kubernetes/r/secret.html


paulwoelfelcpl:
Is there somebody who can review this MR?

slevenick:
Yep! Sorry for the delay, I'll upstream this to our generator so it gets preserved!

PR #4605, opened 2019-10-02
Enable retries for QueryOp(...) calls

spew:
The CommonRefreshFunc(...) queries the operation and then immediately returns in error if any sort of http timeout occurs when querying. 

```release-note:enhancement
added retries for timeouts while fetching operations
```

### DIFF
--- a/google/common_operation.go
+++ b/google/common_operation.go
@@ -3,6 +3,7 @@ package google
 import (
 	"fmt"
 	"log"
+	"net/url"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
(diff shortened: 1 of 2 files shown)


google/common_operation.go
```diff
@@ -109,6 +110,13 @@ func CommonRefreshFunc(w Waiter) resource.StateRefreshFunc {
 				return op, "done: false", nil
 			}
 
+			for _, e := range getAllTypes(err, &googleapi.Error{}, &url.Error{}) {
+				if isRetryableError(e, nil) {
```
rileykarson: While you're here, do you mind making the 404 condition above a [retry predicate](https://github.com/terraform-providers/terraform-provider-google/blob/cf8fa58279d0b65132bc864448652c54429fc71a/google/error_retry_predicates.go) and passing it in to `isRetryableError`?

The reason we retry those 404's is that [operations don't always exist immediately](https://gist.github.com/rileykarson/cc3823cea8e333bd9f57dff34b0cc83c).

spew:
Updated with conversion of the 404 check to predicate and formatting of test file.

rileykarson:
Sorry for the delay! We were in a short-term freeze for the `3.0.0` release.

PR #4613, opened 2019-10-03
Remove unsupported import command format

mattalexander-pantheon:
When just the topic ID is passed with no project, the result is a error
"Error: project: required field is not set".

### DIFF
--- a/website/docs/r/pubsub_topic.html.markdown
+++ b/website/docs/r/pubsub_topic.html.markdown
@@ -145,12 +145,11 @@ This resource provides the following
 
 ## Import
 
-Topic can be imported using any of these accepted formats:
+Topic can be imported using either of these accepted formats:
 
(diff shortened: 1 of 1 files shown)


megan07:
Hi @mattalexander-pantheon!  Thank you so much for submitting this PR and contributing to the Google Terraform Provider.  I ran a test to double check why this wasn't working, and I was able to import with only the pubsub topic name.  Could you double check that your Google project is configured on your provider?  You can find more information here: https://www.terraform.io/docs/providers/google/provider_reference.html#project-1

Please let me know if you're able to reproduce the issue after double-checking that `project` is set.
Thanks again!

megan07:
I haven't heard back in awhile, so I will close this PR, but if the problem continues, please feel free to open a new PR or issue.  Thanks!

PR #4666, opened 2019-10-12
Fix removal of google_storage_bucket.versioning not applying

RJacksonm1:
Hi y'all,

Spotted this issue not so long ago. Removing a `versioning` block from a `google_storage_bucket` resource does not apply correctly.

Plan output:

```
  # google_storage_bucket.versioning_on will be updated in-place
  ~ resource "google_storage_bucket" "versioning_on" {
        bucket_policy_only = false
        force_destroy      = false
        id                 = "test-bucket-please-ignore-versioning-on"
        labels             = {}
        location           = "US"
        name               = "test-bucket-please-ignore-versioning-on"
        project            = "REDACTED"
        requester_pays     = false
        self_link          = "https://www.googleapis.com/storage/v1/b/test-bucket-please-ignore-versioning-on"
        storage_class      = "STANDARD"
        url                = "gs://test-bucket-please-ignore-versioning-on"

      - versioning {
          - enabled = true -> null
        }
    }

```

Expectation

* Disables versioning on the bucket

Actual

* Sends an empty PATCH request (does not disable versioning)
* Every `terraform plan` shows the above change

----

Having a glance through the code, it looks like `sb.NullFields` is used to guard against this exact problem for other fields, so I've added it to the `versioning` field and tested various transitions with a local build of this provider.

I've checked

* the contents of the API request sent by inspecting the output of `TF_LOG=debug terraform apply`
* the final versioning state of the bucket with `gsutil versioning get gs://test-bucket-please-ignore-versioning-on/`

| Versioning from | To | API request contents | Versioning state |
| ---- | ---- | ---- | ---- |
| (not set) | `true` | `{ "versioning": { "enabled": true } }` | Enabled |
| (not set) | `false` | `{ "versioning": { "enabled": false } }` | Suspended |
| `true` | (not set) | `{ "versioning": null }` | Suspended |
| `true` | `false` | `{ "versioning": { "enabled": false } }` | Suspended |
| `false` | (not set) | `{ "versioning": null }` | Suspended |
| `false` | `true` | `{ "versioning": { "enabled": true } }` | Enabled |

I wasn't sure how to add automated tests for this. If that's necessary, any pointers will be appreciated!

### DIFF
--- a/google/resource_storage_bucket.go
+++ b/google/resource_storage_bucket.go
@@ -436,6 +436,8 @@ func resourceStorageBucketUpdate(d *schema.ResourceData, meta interface{}) error
 	if d.HasChange("versioning") {
 		if v, ok := d.GetOk("versioning"); ok {
 			sb.Versioning = expandBucketVersioning(v)
+		} else {
+			sb.NullFields = append(sb.NullFields, "Versioning")
 		}
 	}
 


danawillow:
Hey @RJacksonm1, sorry for the delayed response, and thanks for the contribution! You can add an  automated test in this file: https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_storage_bucket_test.go. `TestAccStorageBucket_bucketPolicyOnly` is a pretty standard test of an update. I think just one test that goes from `true` to `not set` would be sufficient. We have information on how to run them at https://github.com/terraform-providers/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#tests.

rileykarson:
I'm going to close this as obsolete, given that it's around nine months since there was any activity. If anyone is interested in this functionality, please send another PR or file an issue!

Issue #4324, opened 2019-08-22
Drop support for protocol 4/Terraform 0.11


paddycarver:
## About Terraform Protocols

Terraform operates by using a [core binary](https://github.com/hashicorp/terraform) that [communicates over local RPC with provider plugin binaries](https://github.com/hashicorp/terraform/tree/master/docs/plugin-protocol). In Terraform 0.11, we were using version 4 of this RPC protocol. Terraform 0.12 communicates using version 5 of this RPC protocol.

The RPC compatibility is signaled when publishing a provider, and `terraform init` will only download providers compatible with the protocol versions supported by the version of Terraform core running the command.

## Our Current Setup

Right now, all our provider releases are published signaling their support for version 4 and version 5 of the protocol. This means Terraform 0.11 (speaks protocol 4) can use them, and Terraform 0.12 (speaks protocol 5) can use them. Which is obviously a nice thing.

It also, unfortunately, means we're limited to a lowest common denominator of functionality between the two providers, which is sometimes unpredictable in how it behaves.

For example, we are currently unable to publish alpha or beta versions of provider releases, because Terraform 0.11 has a broken implementation of semver parsing for providers, and won't handle those correctly.

## The Proposal

My proposed fix for this is to publish future versions of the provider advertising support _only_ protocol 5. This would make Terraform 0.11 ignore them when running `terraform init`. In theory, users could download the binaries themselves and put them in a specific location on their hard drive and continue using the new binaries with 0.11, but we'd stop supporting that and some things likely wouldn't work as intended.

## The Blockers

This would be a major breaking change, and would thus need to be bundled into a major release.

Also, providers are currently released to releases.hashicorp.com. Terraform 0.11 downloads the appropriate providers during `terraform init` by iterating through them in reverse, looking for a provider that supports the necessary protocol version. So the more protocol v5 releases we do on releases.hashicorp.com, the slower Terraform 0.11 will be to init. So ideally we'd like to wait until releases are published somewhere other than releases.hashicorp.com.

paddycarver:
There's a blog post on this: https://www.hashicorp.com/blog/deprecating-terraform-0-11-support-in-terraform-providers/

I have one last release process change to make, and this should be done.

PR #4702, opened 2019-10-17
Migrate state for google_bigtable_instance

### LINKED ISSUE #4701: google_bigtable_instance schema change causes refresh to fail
### Community Note

### Terraform Version

```
Terraform v0.11.14 (also v0.12.9)
+ provider.google-beta v2.14.0 (and up)
```

### Affected Resource(s)

* `google_bigtable_instance`

Relevant resource:
```tf
resource "google_bigtable_instance" "test_instance" {
  name = "test-instance"

  cluster {
    cluster_id   = "cluster1"
    zone         = "us-central1-a"
    num_nodes    = 3
    storage_type = "SSD"
  }
}
```

### Debug Output

Debug log from above: https://gist.github.com/nickmoorman/de60a218fcdb200baff333163d4c37e3#file-01-terraform-plan-debug-log

### Panic Output

Interestingly, the explicit crash (with the `...!!! TERRAFORM CRASH !!!...` output) only happens when the `--target` argument is passed to only refresh the `google_bigtable_instance` resource, in which case similar output to above is shown with the more informative error `panic: interface conversion: interface {} is nil, not map[string]interface {}` and the trace to go with it.  Full `crash.log` can be found [here](https://gist.github.com/nickmoorman/de60a218fcdb200baff333163d4c37e3#file-02-terraform-plan-target-crash-log).

### Expected Behavior

`terraform plan` should show that there are no changes when nothing has changed.

### Actual Behavior

Terraform errors (or outright crashes if the `--target` argument is passed to `terraform plan`) on arbitrary resources due to a crash in a plugin goroutine.

### Steps to Reproduce

1. Create a `google_bigtable_instance` resource with version 2.13.0 or earlier of the `google` provider (note: must be done with Terraform 0.11!)
2. Upgrade the provider to 2.14.0 or higher
3. Run `terraform plan`
4. Observe errors

### References

* #4026

nickmoorman:
This PR fixes #4701.  I accomplished this by providing a `MigrateState` function for `google_bigtable_instance` to handle the discrepancy described in the issue.

I added a slew of unit tests that test the pure functionality of the new `MigrateState` function (that is, without going through any RPCs); due to the nature of the bug, two different versions would be required to reproduce with acceptance tests, so I thought manually crafting state and migrating it would be more effective.

In addition to these unit tests, I performed some thorough integration/acceptance testing of my own across many versions of the `google` provider, using [this script](https://gist.github.com/nickmoorman/0d883c2b5a7ede4f53c539d8fea7a76f#file-01-migrate_test-sh).  I tested with 1.19.1 (the version I'm currently using in a real project), 2.13.0-2.17.0, and the "latest", from this branch containing my fix.  The Gist linked previously also contains sample output from the script (the diffs between create/migrate, and the created state for the latest version of the plugin).  Everything came out as expected, with the exception that after I pulled in upstream changes the other day (specifically, #4598 / https://github.com/GoogleCloudPlatform/magic-modules/pull/2409), some of my "maintain ordering" logic no longer worked; https://github.com/GoogleCloudPlatform/magic-modules/pull/2489 addresses this, and the script output in my Gist was produced using those changes as well as the ones in this PR, but the `magic-modules` changes have their own problems that I'm hoping a maintainer will be able to help with.  (Without the changes in that PR, the diff output from my script isn't "empty" like I'd prefer, since the clusters get reordered.)

(Just a note, this is the first time I've ever tried my hand at Go, so feel free to call out anything that could be improved.  :slightly_smiling_face:)

cc @rileykarson @mackenziestarr


nickmoorman:
@rileykarson It [seems](https://github.com/terraform-providers/terraform-provider-google/issues/4701#issuecomment-546104496) that this is no longer required to fix #4701, but do you think it's still worthy of keeping around?  I'm kind of on the fence TBH.  The original bug is fixed, but these changes still do a few (maybe useful) things:
- For users like me who are still using pre-2.x versions of this provider, it removes the legacy attributes that have been marked as `Removed` from the state.
- It allows us to keep clusters in state in the order they're defined in config (with caveats; see https://github.com/GoogleCloudPlatform/magic-modules/pull/2489).
- Applies a state migration that, due to its placement in the resource lifecycle, would prevent crashes like this one from occurring if a future release contained code that tried to reference `cluster` as `[]interface{}` (like the above change) on state created pre-`2.17.0`.

I will say though that most of my testing has been with Terraform 0.11, and it seems like it's possible that Terraform 0.12 somewhat obviates the need for this due to the new state structure?  I defer to your knowledge here.  🙂

rileykarson:
I don't think we should perform a state migration we don't need to, and I'd rather close this out I think.

They're necessary in some cases, but finicky otherwise and can leave users' states in weird places. Additionally, TPG/TPGB `2.5.0+` are primarily targeted against `0.12` while this form of state migration isn't used by `0.12` at all. It's been superseded by "state upgraders", a slightly different implementation. I think the removed fields will get cleaned from state during either the `0.12` upgrade or the `3.0.0` upgrade when they're hard-removed.

For point 3, `3.0.0` will [remove `0.11` support](https://github.com/terraform-providers/terraform-provider-google/issues/4324), so not many changes will be made that could have this issue anymore since `0.12` isn't affected. I'm confident one won't happen by the time we drop support.

nickmoorman:
That all makes sense to me.  Thanks for taking a look at all this, and for all the additional info!  =]

PR #4795, opened 2019-11-01
website: Add subcategories to docs for display on the Terraform Registry

findkim:
We will soon be displaying documentation for this provider both on [terraform.io](https://www.terraform.io/docs/providers/index.html) and on [the Terraform Registry](https://registry.terraform.io/providers).

Documentation displayed on the Terraform Registry will not use the ERB layout containing the existing navigation, and will instead build the navigation dynamically based on the directory and YAML frontmatter of a file. For providers that group similar resources and data sources by service or use case (subcategories), we'll need to add this information to the Markdown source file.

This PR modifies resource and data source documentation source files by adding a subcategory to the metadata if those files are currently grouped on terraform.io.

For more information about how documentation is rendered on the Terraform Registry, please see this reference: [Terraform Registry - Provider Documentation](https://www.terraform.io/docs/registry/providers/docs.html).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ install:
 - make tools
 
 script:
+- make docscheck
 - make lint
 - make test
 - make website-test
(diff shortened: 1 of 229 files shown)


findkim:
Thanks @chrisst for moving these changes to the upstream docs! https://github.com/GoogleCloudPlatform/magic-modules/pull/2585

I'll close this one out and open up one with just the `docscheck` CI changes.

PR #4797, opened 2019-11-01
Add status argument (RUNNING/TERMINATED) to google_compute_instance

norbjd:
Hello,

I added a `status` argument on `google_compute_instance` to be able to
start/stop instances using Terraform, therefore answering :

- this issue : [google_compute_instance should allow to specify instance state #1719](https://github.com/terraform-providers/terraform-provider-google/issues/1719)
- and also this question on StackOverflow : [How to stop GCP vm instances using terraform](https://stackoverflow.com/questions/57986338/how-to-stop-gcp-vm-instances-using-terraform) (that made me discover this was not possible yet)

When searching I noticed that there is [an open PR](https://github.com/terraform-providers/terraform-provider-google/pull/2956) on that topic but sadly it looks like it has been abandoned (no update since 2019/03/13), so I've decided to implement it by following the valuable advices written by @rileykarson in this PR.

This is my first PR here so I gladly accept any comment :)

## Recap

I added a `status` field accepting only values `RUNNING` (default) or `TERMINATED`.
There are many other existing states : `PROVISIONING`, `STAGING`, `STOPPING`, `REPAIRING` (from the [docs]((https://cloud.google.com/compute/docs/instances/instance-life-cycle))), and I have also seen `STOPPED`, `SUSPENDED` or `SUSPENDING` on Compute Go client (https://godoc.org/google.golang.org/api/compute/v1#Instance, check `Status` field). Nevertheless, I think that only `RUNNING` and `TERMINATED` have sense here, because other are kind of "transition" states.

### Instance creation

On instance creation, I have chosen to tolerate only `RUNNING` status, because this is the default behavior in the current provider, and because I did not have an example that came to my mind where it is needed to create a `TERMINATED` instance. Also, modifications should have been more important (I think) if the possibiility to create `TERMINATED` instances were given.

On instance creation (or update), in order to wait for the instance to reach the expected `status`, I used `resource.StateChangeConf` and the `WaitForState()` function to avoid reinventing the wheel, but I don't know if parameters are correct here (`Delay`, `Timeout`, and even `Pending` states). I'm open to any suggestion.

### Instance update

On instance update, I have done two things :

- start the instance if `status` changes from `TERMINATED` to `RUNNING`, and stop the instance if `status` changes from `RUNNING` to `TERMINATED`
- if an instance have already `TERMINATED` status, and when updating fields requiring the instance to be stopped (like changing `machine_type`, `min_cpu_platform`, ...), do not try to stop (and after restart) the instance because it is already stopped (`TERMINATED` status)

### Acceptance tests

Finally, I have written acceptance tests covering and passing the following cases :

- a `RUNNING` instance is created
- an instance is `TERMINATED` (`RUNNING` => `TERMINATED`)
- an instance is started from a `TERMINATED` status (`TERMINATED` => `RUNNING`)
- a `TERMINATED` instance is updated :
    - on fields that do not require the instance to be stopped (for example `labels`, `metadata`, ...)
    - on fields that require the instance to be stopped (`machine_type`)
- a `TERMINATED` instance is deleted
- a `TERMINATED` instance cannot be created (expected to be `RUNNING` when creating)

I don't think adding this `status` field will impact the previous behavior of the provider (maybe wait a little more when creating an instance because it needs to wait until the instance is `RUNNING`, though when testing I've noticed that the instance was always `RUNNING` when it reached the part of the code waiting for the status). However, I did not run the whole acceptance tests suite (like in other PR I believe), so I'm wondering, is there a process that will run all tests before merging any PR to ensure that the modification did not break something unexpected?

I look forward to reading your comments.

```release-note:enhancement
compute: added the ability to manage the status of `google_compute_instance` resources with the `desired_status` field
```


google/resource_compute_instance.go
```diff
@@ -969,11 +981,61 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
...
+		instance, err := getInstance(config, d)
+		if err != nil || instance == nil {
+			log.Printf("Error on InstanceStateRefresh: %s", err)
+			return nil, "", err
+		}
+		return instance.Id, instance.Status, nil
+	}
+	stateChangeConf := resource.StateChangeConf{
+		Delay:      5 * time.Second,
+		Pending:    getAllStatusExcept(expectedStatus),
```
norbjd: I'm not sure that all the status (except the expected one) are needed here.

google/resource_compute_instance.go
```diff
@@ -1101,6 +1163,31 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
...
+			op, err = config.clientCompute.Instances.Start(project, zone, instance.Name).Do()
+			if err != nil {
+				return err
+			}
+		} else if expectedStatus == "TERMINATED" {
+			op, err = config.clientCompute.Instances.Stop(project, zone, instance.Name).Do()
+			if err != nil {
+				return err
+			}
+		} else {
```
norbjd: Since only `RUNNING` and `TERMINATED` status are allowed (validation when setting the argument), this code is not reachable but it is an additional precaution. Is it really needed?

google/resource_compute_instance_test.go
```diff
@@ -1744,6 +1855,16 @@ func testAccCheckComputeInstanceHasMinCpuPlatform(instance *compute.Instance, mi
 	}
 }
 
+func testAccCheckComputeInstanceHasMachineType(instance *compute.Instance, machineType string) resource.TestCheckFunc {
```
norbjd: There was no check about the `machine_type` so I've implemented one (used in `TestAccComputeInstance_statusTerminatedUpdateFields` acceptance tests above). However, I am not fully satisfied with this solution because `instance.MachineType` returns something like, IIRC, `https://googleapis.com/compute/.../projects/<my-project>/zones/<my-zone>/machineTypes/n1-standard-1` and since I cannot access neither project nor zone from the `instance`, I've decided to check only the last part of the machine type (`machineTypes/n1-standard-1`). Is this a suitable solution?

norbjd:
Bumping this up.

Is there anything else I can do for this PR?

tysen:
I'll review this soon.

google/resource_compute_instance.go
```diff
@@ -969,11 +981,64 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
...
+	}
+	for i, status := range allStatus {
+		if exceptedStatus == status {
+			return append(allStatus[:i], allStatus[i+1:]...)
+		}
+	}
+	return nil
+}
+
+func waitUntilInstanceHaveTheExpectedStatus(config *Config, d *schema.ResourceData) error {
```
tysen: This function breaks pre-existing user configs for an instance that already exists and is `TERMINATED`. `d.Get("status")` will return the schema default of `"RUNNING"` (because the user hasn't set anything for that field) and the function will wait for the `TERMINATED` instance to transition to `RUNNING` which will of course never happen (without outside action).

norbjd: You are right, I did not consider that case. I have updated the code to use the `waitUntilInstanceHaveTheExpectedStatus` method in `resourceComputeInstanceCreate` function instead of `resourceComputeInstanceRead`. Therefore, existing instances (with default `RUNNING` status if not specified) will not be affected. Integration tests I made are still passing.

google/resource_compute_instance_test.go
```diff
@@ -1948,6 +1984,32 @@ func testAccCheckComputeInstanceHasStatusTerminated(instance *compute.Instance)
...
+
+		op, err := config.clientCompute.Instances.Stop(config.Project, rs.Primary.Attributes["zone"], rs.Primary.ID).Do()
+		if err != nil {
+			return fmt.Errorf("Could not stop instance: %s", err)
+		}
+		err = computeOperationWait(config.clientCompute, op, config.Project, "Waiting on stop")
+		if err != nil {
+			return fmt.Errorf("Could not stop instance: %s", err)
+		}
+
```
norbjd: I used `testAccCheckComputeInstanceExistsInProject` as an example. I don't know if I should pass also here a `instance *compute.Instance` parameter and change its status (`instance.Status = "TERMINATED"`) before returning. It will simplify the acceptance test `TestAccComputeInstance_statusTerminatedManuallyAndRestart` (see my comment there).

google/resource_compute_instance.go
norbjd: @tysen I have also added an acceptance test to cover that case (`TestAccComputeInstance_statusTerminatedManuallyAndRestart`).

norbjd:
Hello @tysen, did you get a chance to look at the updates I made after your first review?

google/resource_compute_instance.go
```diff
@@ -969,6 +1034,7 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("name", instance.Name)
 	d.Set("description", instance.Description)
 	d.Set("hostname", instance.Hostname)
+	d.Set("status", instance.Status)
```
rileykarson: Can you check `desired_status` and only set this value if it's not `""`?

One of the problems that settable status fields have had historically is that we can't tell if the desired status is user-set or an old value from state. This should make a few interactions work predictably that could otherwise lock instances in undesirable states.

We expect the following behaviour, roughly:

* If a user never sets `desired_status`, we'll never wait for an expected status

* If a user has `desired_status` in config and imports the resource, it will not be set until an update.

* If a user unsets `desired_status` in config, Terraform will transition the value to unset (`""` represents unset) during a no-op update.

google/resource_compute_instance.go
```diff
@@ -563,6 +564,13 @@ func resourceComputeInstance() *schema.Resource {
 				},
 			},
 
+			"status": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "RUNNING",
```
```suggestion
```

Users who've stopped their instance out of band would see a diff with a default. Combined with the changes to `desired_state` that make state management optional below, I don't think we need this.

google/resource_compute_instance.go
```diff
@@ -718,6 +726,10 @@ func expandComputeInstance(project string, d *schema.ResourceData, config *Confi
 		return nil, fmt.Errorf("Error creating guest accelerators: %s", err)
 	}
 
+	if d.Get("status") != "RUNNING" {
```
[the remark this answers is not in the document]
norbjd: I'm sorry but I did not understand your comment after the first sentence :cry: Could you add further explanations please? Thanks :smile: 
To explain this line, the goal here is to check that only `RUNNING` value can be set at instance creation (there is no way to create a stopped instance). I put that check here because the `expandComputeInstance` function is only used in `resourceComputeInstanceCreate`.

google/resource_compute_instance.go
```diff
@@ -741,6 +753,54 @@ func expandComputeInstance(project string, d *schema.ResourceData, config *Confi
 	}, nil
 }
 
+func getAllStatusExcept(exceptedStatus string) []string {
```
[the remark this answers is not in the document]
norbjd: In fact, this is not an expected status passed as parameter but rather an excepted status :smile:. In the context of this method, there is no notion of "expected", but just a status that we want to except. I should have used the signature `getAllStatusBut(status string)` to avoid any confusion between `expected` and `excepted`, my mistake :smile:. I'm going to rename it and add a comment :+1:

google/resource_compute_instance.go
```diff
@@ -741,6 +753,54 @@ func expandComputeInstance(project string, d *schema.ResourceData, config *Confi
...
+	}
+	for i, status := range allStatus {
+		if exceptedStatus == status {
+			return append(allStatus[:i], allStatus[i+1:]...)
+		}
+	}
+	return nil
+}
+
+func waitUntilInstanceHaveTheExpectedStatus(config *Config, d *schema.ResourceData) error {
```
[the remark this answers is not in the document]
norbjd: I used `d` as parameter here because in the `StateRefreshFunc` function, there is a call to `getInstance(config *Config, d *schema.ResourceData)` to retrieve the current instance (and its status). I believe I can create my own function `getInstanceStatus(name, region, project string) (status string)` but I felt like reusing the existing would be a good idea. Am I wrong?

google/resource_compute_instance.go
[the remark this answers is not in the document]
rileykarson: https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_instance.go#L1407 calls it as well. You've got a few options:

1. Move this check out of expand
1. Stop using expand at 1407 (all of the data pulled from it is available in the `ResourceData` object `d`)
1. Implement a CustomizeDiff for this check, returning an error if it's create time and the value of the field is not `RUNNING`.

#3 is the hardest to implement but also the best user experience- it'll cause an error at plan time instead of apply time.

google/resource_compute_instance.go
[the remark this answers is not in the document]
rileykarson: Good point, this approach is :+1:

Since `d` is mutable I generally try not to pass it too far down into functions for the sake of clarity when reading the code, but this file has a precedent of passing it down.

google/resource_compute_instance.go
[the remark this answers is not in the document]
norbjd: Ok, got it! I did not understand because when I made this PR, `expandComputeInstance` was not used anywhere but in `resourceComputeInstanceCreate`, I did not notice that changed. I've implemented the 3rd solution (CustomizeDiff).

google/resource_compute_instance.go
```diff
@@ -1581,6 +1678,24 @@ func suppressEmptyGuestAcceleratorDiff(d *schema.ResourceDiff, meta interface{})
 	return nil
 }
 
+func desiredStatusDiff(diff *schema.ResourceDiff, meta interface{}) error {
+	// when creating an instance, name is not set
+	oldName, _ := diff.GetChange("name")
```
norbjd: I don't know if this is the best way to check if we are creating an instance, but this is a solution inspired by what is done in `google_bigtable_instance.go` (https://github.com/terraform-providers/terraform-provider-google/blob/07a8c1f9f7428cc490ed92071484797ab7dfe828/google/resource_bigtable_instance.go#L341).

google/resource_compute_instance_test.go
```diff
@@ -3881,3 +4104,178 @@ resource "google_compute_instance" "foobar" {
...
+	metadata = {
+		bar = "baz"
+	}
+
+	labels = {
+		only_me = "nothing_else"
+	}
+
+	allow_stopping_for_update = true
```
norbjd: I'm not sure about this. If the instance has status `TERMINATED`, there should be no issue updating some fields requiring stopping the instance (here, the machine_type). I could add the following check : if `instance.Status = "TERMINATED"`, then the `allow_stopping_for_update` is not required. What do you think?

google/resource_compute_instance.go
```diff
@@ -1077,6 +1147,33 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
 		d.SetPartial("scheduling")
 	}
 
+	if d.HasChange("desired_status") {
+		desiredStatus := d.Get("desired_status").(string)
```
norbjd: Should I check for `d.Get("desired_status") == nil`? If I understand correctly, the `if d.HasChange("desired_status")` part ensure that `d.Get("desired_status")` is not `nil`, but I would like to be sure :)

google/resource_compute_instance.go
```diff
@@ -715,6 +723,58 @@ func expandComputeInstance(project string, d *schema.ResourceData, config *Confi
...
+	for i, s := range allStatus {
+		if status == s {
+			return append(allStatus[:i], allStatus[i+1:]...)
+		}
+	}
+	return nil
+}
+
+func waitUntilInstanceHasDesiredStatus(config *Config, d *schema.ResourceData) error {
+	desiredStatus := d.Get("desired_status").(string)
```
norbjd: Is it possible here that `d.Get("desired_status")` returns `nil`? If so, I should add a check before doing the type assertion.

norbjd:
Hello @rileykarson, thanks for your review. I made some changes according to your comments. To sum up :

- rename `status` to `desired_status`
- add a `CustomizeDiff` to ensure that instances cannot be created when `desired_status = TERMINATED`
- update and fix the acceptance tests

google/resource_compute_instance.go
norbjd: I have updated acceptance tests to cover the 1st and 3rd cases. However, I'm not sure how to do the 2nd.

google/resource_compute_instance.go
```diff
@@ -715,6 +723,58 @@ func expandComputeInstance(project string, d *schema.ResourceData, config *Confi
 	}, nil
 }
 
+// return all possible Compute instances status except the one passed as parameter
+func getAllStatusBut(status string) []string {
```
norbjd: I'm not particulary proud of this function :
- I don't know if this is useful (it is used in `waitUntilInstanceHasDesiredStatus` `StateChangeConf` object)
- I can't find anywhere a list of possible Compute instance status so I must create the `allStatus` variable here (well, I could have created it in a `const` outside the function, even if it is the only usage here)
I am open to any suggestions to improve it.

google/resource_compute_instance.go
rileykarson: `d.Get` on a field that exists in the resource schema will return the set value or the zero value for the type (`""` in this case)

google/resource_compute_instance.go
rileykarson: Yep, this is the right way to check!

google/resource_compute_instance.go
rileykarson: See https://github.com/terraform-providers/terraform-provider-google/pull/4797#discussion_r377787758 (same rules apply)

google/resource_compute_instance.go
```diff
@@ -1581,6 +1678,24 @@ func suppressEmptyGuestAcceleratorDiff(d *schema.ResourceDiff, meta interface{})
 	return nil
 }
 
+func desiredStatusDiff(diff *schema.ResourceDiff, meta interface{}) error {
```
```suggestion
// Return an error if the desired_status field is set to a value other than RUNNING on Create.
func desiredStatusDiff(diff *schema.ResourceDiff, meta interface{}) error {
```

google/resource_compute_instance_test.go
rileykarson: I think we can skip checking `allow_stopping_for_update` if the instance is already terminated, or if the `desired_status` is `TERMINATED`. We'll treat it as "allow stopping+starting a RUNNING instance to perform an update".

Additionally, if the instance was RUNNING and we stopped it for an update, we should restore it to RUNNING if `desired_status` is `""` or `"RUNNING"`.

My expectations, assuming we run `terraform apply` changing the `machine_type`:

- a RUNNING instance with desired_status of `""` or `"RUNNING"` and `allow_stopping_for_update = true` will be updated and restored to RUNNING

- a RUNNING instance with desired_status of `""` or `"RUNNING"` and `allow_stopping_for_update = false` will fail to update

- a RUNNING instance with desired_status of `"TERMINATED"` with `allow_stopping_for_update = false` can be stopped early to allow updates

- a TERMINATED instance with desired_status of `"TERMINATED"` with `allow_stopping_for_update = false` is already stopped, and will remain stopped

- a TERMINATED instance with desired_status of `""` with `allow_stopping_for_update = false` is already stopped, and will remain stopped

- a TERMINATED instance with desired_status of `"RUNNING"` with `allow_stopping_for_update = false` is updated while stopped, and then started

Let me know if you'd expect different behaviour or if you feel there's an interesting case that isn't covered.

google/resource_compute_instance.go
rileykarson: Seems fine to me. There isn't anywhere else that lists all the states, so we need it. One implementation you could compare to is what I did in GKE cluster recently: https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_container_cluster.go#L1707-L1737 (+ [state utils](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/state_util.go))

The goal there is to wait for _any_ stable ("resting") state, and not just the user's desired state.

norbjd: I've kept the old implementation because it's simpler to understand. Using `RestingStates` could have been confusing because here, as you said, there is no need to wait for a "resting" state, but rather just the user's desired state.

google/resource_compute_instance_test.go
norbjd: Thanks for the details you provided, I made a table to summarize all possible cases. Each case is covered by an acceptance test :

```                                                                                                 
+----------------------------------------------------------------------+---------+-----------------------+-------------------------------------------------------------------------------------------+
| Input                                                                |         | Actions (ordered)     |                                                                                           |
+----------------------------------------------------------------------+         +-----------------------+                                                                                           |
| Current instance status | desired_status | allow_stopping_for_update | Success | Stop | Update | Start | Acceptance test                                                                           |
+-------------------------+----------------+---------------------------+---------+------+--------+-------+-------------------------------------------------------------------------------------------+
| RUNNING                 | ""             | TRUE                      | X       | X    | X      | X     | TestAccComputeInstance_updateRunning_desiredStatusNotSet_allowStoppingForUpdate           |
| RUNNING                 | RUNNING        | TRUE                      | X       | X    | X      | X     | TestAccComputeInstance_updateRunning_desiredStatusRunning_allowStoppingForUpdate          |
| RUNNING                 | ""             | FALSE                     |         |      |        |       | TestAccComputeInstance_updateRunning_desiredStatusNotSet_notAllowStoppingForUpdate        |
| RUNNING                 | RUNNING        | FALSE                     |         |      |        |       | TestAccComputeInstance_updateRunning_desiredStatusRunning_notAllowStoppingForUpdate       |
| RUNNING                 | TERMINATED     | TRUE                      | X       | X    | X      |       | TestAccComputeInstance_updateRunning_desiredStatusTerminated_allowStoppingForUpdate       |
| RUNNING                 | TERMINATED     | FALSE                     | X       | X    | X      |       | TestAccComputeInstance_updateRunning_desiredStatusTerminated_notAllowStoppingForUpdate    |
+-------------------------+----------------+---------------------------+---------+------+--------+-------+-------------------------------------------------------------------------------------------+
| TERMINATED              | ""             | TRUE                      | X       |      | X      |       | TestAccComputeInstance_updateTerminated_desiredStatusNotSet_allowStoppingForUpdate        |
| TERMINATED              | TERMINATED     | TRUE                      | X       |      | X      |       | TestAccComputeInstance_updateTerminated_desiredStatusTerminated_allowStoppingForUpdate    |
| TERMINATED              | ""             | FALSE                     | X       |      | X      |       | TestAccComputeInstance_updateTerminated_desiredStatusNotSet_notAllowStoppingForUpdate     |
| TERMINATED              | TERMINATED     | FALSE                     | X       |      | X      |       | TestAccComputeInstance_updateTerminated_desiredStatusTerminated_notAllowStoppingForUpdate |
| TERMINATED              | RUNNING        | TRUE                      | X       |      | X      | X     | TestAccComputeInstance_updateTerminated_desiredStatusRunning_allowStoppingForUpdate       |
| TERMINATED              | RUNNING        | FALSE                     | X       |      | X      | X     | TestAccComputeInstance_updateTerminated_desiredStatusRunning_notAllowStoppingForUpdate    |
+-------------------------+----------------+---------------------------+---------+------+--------+-------+-------------------------------------------------------------------------------------------+
```

All acceptance tests are passing :smile:

norbjd:
Hello @rileykarson, just updated the PR with your helpful comments :smile: 

To sum up, the biggest update was to handle all cases that requires stopping the instance before updating some of the attributes. Introducing `desired_status` created some conflicts, as for example there is no need to stop an instance already TERMINATED. I've made a table based on the different cases you provided and implement an acceptance test for each case :

```                                                                                                 
+----------------------------------------------------------------------+---------+-----------------------+-------------------------------------------------------------------------------------------+
| Input                                                                |         | Actions (ordered)     |                                                                                           |
+----------------------------------------------------------------------+         +-----------------------+                                                                                           |
| Current instance status | desired_status | allow_stopping_for_update | Success | Stop | Update | Start | Acceptance test                                                                           |
+-------------------------+----------------+---------------------------+---------+------+--------+-------+-------------------------------------------------------------------------------------------+
| RUNNING                 | ""             | TRUE                      | X       | X    | X      | X     | TestAccComputeInstance_updateRunning_desiredStatusNotSet_allowStoppingForUpdate           |
| RUNNING                 | RUNNING        | TRUE                      | X       | X    | X      | X     | TestAccComputeInstance_updateRunning_desiredStatusRunning_allowStoppingForUpdate          |
| RUNNING                 | ""             | FALSE                     |         |      |        |       | TestAccComputeInstance_updateRunning_desiredStatusNotSet_notAllowStoppingForUpdate        |
| RUNNING                 | RUNNING        | FALSE                     |         |      |        |       | TestAccComputeInstance_updateRunning_desiredStatusRunning_notAllowStoppingForUpdate       |
| RUNNING                 | TERMINATED     | TRUE                      | X       | X    | X      |       | TestAccComputeInstance_updateRunning_desiredStatusTerminated_allowStoppingForUpdate       |
| RUNNING                 | TERMINATED     | FALSE                     | X       | X    | X      |       | TestAccComputeInstance_updateRunning_desiredStatusTerminated_notAllowStoppingForUpdate    |
+-------------------------+----------------+---------------------------+---------+------+--------+-------+-------------------------------------------------------------------------------------------+
| TERMINATED              | ""             | TRUE                      | X       |      | X      |       | TestAccComputeInstance_updateTerminated_desiredStatusNotSet_allowStoppingForUpdate        |
| TERMINATED              | TERMINATED     | TRUE                      | X       |      | X      |       | TestAccComputeInstance_updateTerminated_desiredStatusTerminated_allowStoppingForUpdate    |
| TERMINATED              | ""             | FALSE                     | X       |      | X      |       | TestAccComputeInstance_updateTerminated_desiredStatusNotSet_notAllowStoppingForUpdate     |
| TERMINATED              | TERMINATED     | FALSE                     | X       |      | X      |       | TestAccComputeInstance_updateTerminated_desiredStatusTerminated_notAllowStoppingForUpdate |
| TERMINATED              | RUNNING        | TRUE                      | X       |      | X      | X     | TestAccComputeInstance_updateTerminated_desiredStatusRunning_allowStoppingForUpdate       |
| TERMINATED              | RUNNING        | FALSE                     | X       |      | X      | X     | TestAccComputeInstance_updateTerminated_desiredStatusRunning_notAllowStoppingForUpdate    |
+-------------------------+----------------+---------------------------+---------+------+--------+-------+-------------------------------------------------------------------------------------------+
```

google/resource_compute_instance_test.go
```diff
@@ -3881,3 +4524,337 @@ resource "google_compute_instance" "foobar" {
 }
 `, instance)
 }
+
+func testAccComputeInstance_desiredStatusNotSet(instance string) string {
```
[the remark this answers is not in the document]
norbjd: This config is "simpler" than t`estAccComputeInstance_basic`, but is the same as `testAccComputeInstance_basic2`, so I'll use `testAccComputeInstance_basic2` here.

google/resource_compute_instance.go
```diff
@@ -1403,35 +1510,40 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
...
 
-		// Retrieve instance from config to pull encryption keys if necessary
-		instanceFromConfig, err := expandComputeInstance(project, d, config)
-		if err != nil {
-			return err
-		}
+		if (statusBeforeUpdate == "RUNNING" && desiredStatus != "TERMINATED") ||
+			(statusBeforeUpdate == "TERMINATED" && desiredStatus == "RUNNING") {
+			// Retrieve instance from config to pull encryption keys if necessary
+			instanceFromConfig, err := expandComputeInstance(project, d, config)
```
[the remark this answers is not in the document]
norbjd: Oops, I missed it. I've extracted the common part in a method `startInstanceOperation` and reused it in both cases.

google/resource_compute_instance.go
```diff
@@ -1458,6 +1548,51 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
 	return resourceComputeInstanceRead(d, meta)
 }
 
+func startInstanceOperation(d *schema.ResourceData, config *Config) (*compute.Operation, error) {
+	project, err := getProject(d, config)
```
norbjd: When I started implementing this, I did not include lines 1552-1567 because retrieving the project, zone and instance was already done at the beginning of `resourceComputeInstanceUpdate`. So I decided to pass `project`, `zone` and `instance` as parameters of `startInstanceOperation`, but then I felt like having all those parameters was "too much", because we can retrieve them using `d` and `config` (what I did here). Tell me what you think and what you prefer :
- passing `project`, `zone` and `instance` as parameters of `startInstanceOperation`
- keep `startInstanceOperation` like this (it duplicates code)
- or something else :smile:

PR #4807, opened 2019-11-04
update examples from 0.11 syntax to 0.12 syntax

megan07:
Updating examples from 0.11 syntax to 0.12 syntax

### DIFF
--- a/examples/two-tier/main.tf
+++ b/examples/two-tier/main.tf
@@ -1,9 +1,9 @@
 # See https://cloud.google.com/compute/docs/load-balancing/network/example
 
 provider "google" {
-  region      = "${var.region}"
-  project     = "${var.project_name}"
-  credentials = "${file("${var.credentials_file_path}")}"
+  region      = var.region
+  project     = var.project_name
+  credentials = file(var.credentials_file_path)
 }
 
 resource "google_compute_http_health_check" "default" {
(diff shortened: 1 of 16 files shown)


examples/two-tier/main.tf
```diff
@@ -49,27 +49,57 @@ resource "google_compute_instance" "www" {
...
   }
 
   provisioner "file" {
-    source      = "${var.install_script_src_path}"
-    destination = "${var.install_script_dest_path}"
+    source      = var.install_script_src_path
+    destination = var.install_script_dest_path
 
     connection {
+      host = coalesce( # Simplify this to reference a specific desired IP address, if possible.
```
megan07: This was a "TO-DO" that `terraform 0.12upgrade` generated.  My assumption is that it will stay this way as log as the access_config nat_ip is ephemeral, but please correct me if I'm wrong.

ndmckinley: I don't quite understand why we can't just grab the first access config.  It seems like we know for sure we're going to have one in this example.  Does this get executed at a counter-intuitive time?

megan07: That could be - something like `self.network_interface.0.access_config.0.nat_ip`?

ndmckinley: That's my proposal, but I haven't tested it - I can't guarantee it'll work right.

megan07: I've updated it, ran the plan and it worked.  Thanks for the review!

PR #4819, opened 2019-11-05
Add docscheck make target/script for website docs

findkim:
Follow-up PR from https://github.com/terraform-providers/terraform-provider-google/pull/4795

This adds a script to verify expected fields for customized docs structure for the [Terraform Registry](https://www.terraform.io/docs/registry/providers/docs.html).
* Require custom `subcategories` to frontmatter for resources and data sources -- useful to organize large # of docs in navigation
* Require `page_title` for guides
* Error explicitly on any other files in the `website/docs` that will not be ingressed by the Registry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ install:
 - make tools
 
 script:
+- make docscheck
 - make lint
 - make test
 - make website-test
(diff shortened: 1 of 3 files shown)


chrisst:
depends on https://github.com/GoogleCloudPlatform/magic-modules/pull/2594 adding the last few categories.

PR #4864, opened 2019-11-11
Use vendored modules.

paddycarver:
Set the proper environment variables to use the modules in our vendor dir instead of downloading them again.


paddycarver:
Do those commits resolve your concerns?

PR #5021, opened 2019-11-29
Correct default GCP env var for credential path in Guide

draconteus:


website/docs/guides/getting_started.html.markdown
```diff
@@ -161,7 +161,7 @@ You supply the key to Terraform using the environment variable
 `GOOGLE_CLOUD_KEYFILE_JSON`, setting the value to the location of the file.
```
pradeepbhadani: Could you please replace `GOOGLE_CLOUD_KEYFILE_JSON` with  `GOOGLE_APPLICATION_CREDENTIALS` on this line as well?

draconteus: Oh, yeah sure.

pradeepbhadani:
@megan07  - Agree, but most of the places `GOOGLE_APPLICATION_CREDENTIALS` is being used. Maybe good to have this PR merged to make documentation consistent?

megan07:
Hi @pradeepbhadani, thank you for pointing that out!  I think the consistency that we want to keep would be with the recommended practice.  If `GOOGLE_APPLICATION_CREDENTIALS` is used elsewhere in the documentation, I think it would be great to change those cases to `GOOGLE_CLOUD_KEYFILE_JSON` instead.

draconteus:
> Sorry, I think I got ahead of myself here. I'm not sure this needs to be changed. To clarify how to set credentials you can read more here https://www.terraform.io/docs/providers/google/guides/provider_reference.html#credentials-1. `GOOGLE_CLOUD_KEYFILE_JSON` is one of the environment variables to use for the recommended Terraform-specific service accounts. `GOOGLE_APPLICATION_CREDENTIALS` is used as a fallback if a Terraform-specific service account key file is not provided. These credentials are your Google app default credentials.

The problem was for me, that I tried to use `GOOGLE_CLOUD_KEYFILE_JSON` as it was stated in the docs, but does not worked and I get credentials cannot be found error. 
I haven't tried the other two env var mentioned in the docs (I'll try it later), unfortunately I missed the mentioned documentation and found the `GOOGLE_APPLICATION_CREDENTIALS` in Google Cloud docs.

megan07:
Hi @draconteus !  Sorry about this, I accidentally merged this instead of another PR I was looking at (too many tabs open).  I will have to test this further, as I have `GOOGLE_CLOUD_KEYFILE_JSON` set in my environment variables and have never seen that error.  I will look at this ASAP and let you know what I think the reasoning might be behind your issue.  Meanwhile, could you please open a new PR (or issue) so we can track it there?
Sorry again, I appreciate your patience.  Thank you!

PR #5053, opened 2019-12-03
Fix documentation on Access Level

bcadiot:
This PR fixe a little mistake on the documentation.

Access Level members can not be Groups as expressed in the [API](https://cloud.google.com/access-context-manager/docs/reference/rest/v1/accessPolicies.accessLevels#condition).

A feature request is open in the Google public bug tracker : https://issuetracker.google.com/issues/119203606

Feel free to change the formulation if needed (the important fact is that you can't use groups as members for this resource).

### DIFF
--- a/website/docs/r/access_context_manager_access_level.html.markdown
+++ b/website/docs/r/access_context_manager_access_level.html.markdown
@@ -132,12 +132,13 @@ The `conditions` block supports:
 
 * `members` -
   (Optional)
-  An allowed list of members (users, groups, service accounts).
+  An allowed list of members (users, service accounts). 
+  Using groups is not supported yet.
   The signed-in user originating the request must be a part of one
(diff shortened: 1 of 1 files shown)


slevenick:
Thanks for the fix! I'll upstream this to our code generator so it gets preserved!

slevenick:
This resource is generated via a code generation tool, so the docs need to be changed in a separate file. This PR should fix this in a similar way: https://github.com/GoogleCloudPlatform/magic-modules/pull/2772

slevenick:
Closing in favor of the generated PRs that also change the description in the terraform resource schema

bcadiot:
> Closing in favor of the generated PRs that also change the description in the terraform resource schema

Thanks a lot @slevenick !
I didn't realized that the code is generated through a module. I will create my PR directly in it the next time

PR #5082, opened 2019-12-05
Update GCP env var for credential path in Getting started guide

draconteus:
Hi,

Opening a new PR requested by @megan07. This issue needs further investigation.
TL;DR: Using `GOOGLE_CLOUD_KEYFILE_JSON ` env var is not working for finding credential file

History can be found in: https://github.com/terraform-providers/terraform-provider-google/pull/5021

Cheers :)

### DIFF
--- a/website/docs/guides/getting_started.html.markdown
+++ b/website/docs/guides/getting_started.html.markdown
@@ -158,10 +158,10 @@ file. Name it something you can remember, and store it somewhere secure on your
 machine.
 
 You supply the key to Terraform using the environment variable
-`GOOGLE_CLOUD_KEYFILE_JSON`, setting the value to the location of the file.
+`GOOGLE_CREDENTIALS`, setting the value to the location of the file.
 
 ```bash
-export GOOGLE_CLOUD_KEYFILE_JSON={{path}}
+export GOOGLE_CREDENTIALS={{path}}
 ```
 
 -> Remember to add this line to a startup file such as `bash_profile` or


megan07:
Hi @draconteus!  Thanks again for opening this new PR, I appreciate it. Are you going through the "Getting Started with the Google Provider" guide? If so, I've gone through those steps after clearing my environment variables. After running `export GOOGLE_CLOUD_KEYFILE_JSON=path/to/creds.json`, I have not had a problem. I did get the error when `GOOGLE_CLOUD_KEYFILE_JSON` was not set, so after you run the export command, could you do an `echo $GOOGLE_CLOUD_KEYFILE_JSON` and make sure it's the same path that you're using when you used `GOOGLE_CREDENTIALS`?

draconteus:
Yeah, I did that's how I recognised this problem. Also, it was the first thing to check if this env var was set properly. One thing that comes to my mind that I don't gave `gsutil` installed. I don't any internal working of `gsutil` or `terraform`, thus I don't know which tool checks the `GOOGLE_CLOUD_KEYFILE_JSON` var. But setting neither `GOOGLE_CLOUD_KEYFILE_JSON` nor `GCLOUD_KEYFILE_JSON` like it stated in the Docs (https://www.terraform.io/docs/providers/google/guides/provider_reference.html#credentials-1).

pradeepbhadani:
@megan07  @draconteus  As per https://cloud.google.com/docs/authentication/getting-started, `GOOGLE_APPLICATION_CREDENTIALS` is used to provide the JSON Key for service account.

draconteus:
Ooo, okay, yeah I used service account's credentials. Thanks for finding this @pradeepbhadani!

megan07:
@draconteus , interesting.  I'm not sure why `GOOGLE_CLOUD_KEYFILE_JSON` would not work or is throwing an error, since it follows the same codepath as `GOOGLE_CREDENTIALS `.  If the file was in the wrong format it should still throw an error.
However, it looks like the issue is resolved, can we close?

draconteus:
Yeah, sure :)

megan07:
Thanks!

PR #5112, opened 2019-12-09
update doc related to use_ip_aliases in google_container_cluster

SezBulent:
Hello,
I noticed that the details were not consistent, and for the same resource block the information was different. I've updated the docs to have the same info for the related block.


rileykarson:
Sorry we didn't get to this, @SezBulent. We'd had some crossed wires, where two developers had touched the resource at the same time and made conflicting changes. We shipped the right change (removing the field) but these docs got left behind.

They've since been updated. As a result, I'm going to close this out.

PR #5156, opened 2019-12-12
remove unused pgp example

tonglil:


website/docs/r/google_service_account_key.html.markdown
```diff
@@ -48,20 +48,6 @@ resource "kubernetes_secret" "google-application-credentials" {
...
-
-```hcl
-resource "google_service_account" "myaccount" {
-  account_id   = "myaccount"
-  display_name = "My Service Account"
-}
-
-resource "google_service_account_key" "mykey" {
-  service_account_id = google_service_account.myaccount.name
-  public_key_type    = "TYPE_X509_PEM_FILE"
```
pradeepbhadani: `public_key_type` is a valid input. Ref: https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_google_service_account_key.go#L45

Could you please clarify why it's termed as `unused` ?

tonglil: Because this snippet is identical to the first snippet? pgp encryption is no longer supported by terraform.

pradeepbhadani: ahh .. thanks for the info.

PR #5292, opened 2020-01-02
updating big query table import syntax

szheng-nyc3:


emilymye:
This is a terraform-specific import format - we parse the given string for the necessary IDs here:

https://github.com/terraform-providers/terraform-provider-google/blob/183932fa28c9581690518a615dd9da25f316b924/google/resource_bigquery_table.go#L879-L897

PR #5340, opened 2020-01-08
Minor extra word typo

foxyblue:
I wasn't too sure on how to use the Magic Modules for a minor change.

I do think I've done this wrong. 😄


danawillow:
Hey @foxyblue, the line in MM that this comes from is at https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/monitoring/terraform.yaml#L84. Since it's a generated file, it would be easier on our end if you could open a PR against MM instead with the change. Would you mind doing that? Just the one line in the yaml file should be sufficient.

foxyblue:
Done: https://github.com/GoogleCloudPlatform/magic-modules/pull/2948

PR #5465, opened 2020-01-22
Ensure container node pools' instance_group_urls contains Instance Group URLs

### LINKED ISSUE #4628: data google_compute_instance_group broken with self_link
### Community Note

### Terraform Version

```
Terraform v0.12.10
+ provider.azurerm v1.35.0
+ provider.google v2.16.0
+ provider.kubernetes v1.9.0
```

### Affected Resource(s)

* google_compute_instance_group

```tf
data "google_compute_instance_group" "example" {
  self_link = "foo" // foo is just for illustration, but does trigger the error
}
```

### Expected Behavior

Data-source should be loaded when using a valid `self_link`.

### Actual Behavior

An error saying a zone is required is returned.

```
Error: A zone must be specified
```

Specifying zone, whilst defeating the purpose of `self_link`, indicates they shouldn't be used together:
```
Error: "self_link": conflicts with zone
Error: "zone": conflicts with self_link
```

### Steps to Reproduce

1. Define a `google_compute_instance_group` data-source using `self_link`
1. Run `terraform refresh`

weargoggles:
As noted in both resource_container_cluster.go and resource_container_node_pool.go, the instanceGroupUrls returned in a nodePool object are in fact Instance Group Manager URLs.

This pull request ensures that the URLs in the node pool resource/datasource `instance_group_urls` and cluster.node_pool[*].instance_group_urls are the URLs of Instance Groups, not Instance Group managers.

It uses an existing function whose comment explains it all: https://github.com/terraform-providers/terraform-provider-google/blob/5e9a1614e4950eb987c9e3e5bfc81d30116532d2/google/resource_container_cluster.go#L1719-L1737

Fixes #4628

### DIFF
--- a/google/resource_container_node_pool.go
+++ b/google/resource_container_node_pool.go
@@ -501,13 +501,17 @@ func flattenNodePool(d *schema.ResourceData, config *Config, np *containerBeta.N
 		}
 		size += int(igm.TargetSize)
 	}
+	instanceGroupUrls, err := getInstanceGroupUrlsFromManagerUrls(config, np.InstanceGroupUrls)
+	if err != nil {
+		return nil, err
+	}
 	nodePool := map[string]interface{}{
 		"name":                np.Name,
 		"name_prefix":         d.Get(prefix + "name_prefix"),
 		"initial_node_count":  np.InitialNodeCount,
 		"node_count":          size / len(np.InstanceGroupUrls),
 		"node_config":         flattenNodeConfig(np.Config),
-		"instance_group_urls": np.InstanceGroupUrls,
+		"instance_group_urls": instanceGroupUrls,
 		"version":             np.Version,
 	}
 


slevenick:
Thanks for the improvement! I'll check this out and get it into the beta provider as well

slevenick:
This is turning out to be trickier than I thought!

It turns out that the `instance_group_urls` attribute was undocumented, and could be currently being used by users for the existing functionality. This means that changing the function of the field (even if it would make more sense) would be a breaking change, which we won't be able to do in the near future.

Having the actual URLs for the instance groups as an attribute on container node pools definitely makes sense, but it may have to be a separate field within the node pool resource. If you can change this PR to add a new attribute that would be appreciated, otherwise you can file an issue to request a new attribute and we will put it in our queue of work

danawillow:
If we had an IGM data source, that would also be a way to get the IG info (and would mean we wouldn't need to worry about naming an `actual_instance_group_urls` field)

fxposter:
Any progress on this? It does actually complicate a lot of things that for google_container_cluster instance_group_urls return actual instance *group* urls, but for node pools instance *group manager* urls are returned. And there is no other way to get proper instance *group* urls for particular node pools.

fxposter:
@danawillow @slevenick https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_container_cluster.go#L2207-L2209 this comment (that is applied for container_cluster, but not container_node_pool) clearly says that the fact that API returns managers there is kinda wrong and should be fixed. Why can't the same fix be applied for container_node_pool side?

danawillow:
@fxposter, the reason is that changing the format that an existing field is returned in is a breaking change. If users are currently interpolating off of that field in other resources, then if we changed it, that could end up affecting their plan in destructive ways. If you'd like to have the format of an existing field changed, I'd recommend filing an issue against the provider asking for it so we can properly triage it for the next major release of the provider.

rileykarson:
Note that the documentation on the REST API actually states:

> The resource URLs of the managed instance groups associated with this node pool.

The cluster-level `instanceGroupUrls` field (which had ambiguous documentation at one point & we modified the value of) is also deprecated in the API now.

Echoing Dana's comment, mind filing an issue so that we can triage that into the `4.0.0` release? It's possible we'll want to do the opposite- ensure instance group managers are always returned- while also ensuring there's an IGM datasource to find the group value as well.

fxposter:
There is an issue for getting IG from IGMs already: https://github.com/hashicorp/terraform-provider-google/issues/7222.

What do you mean by "ensure instance group managers are always returned"? Aren't they returned currently? Problem is - we cannot do anything with those urls.

rileykarson:
GKE Cluster returns instance group URLs, but GKE Node Pool returns instance group manager URLs. We'd like them to be consistent, at least.

weargoggles:
dead

Issue #5502, opened 2020-01-27
Not managed `expr` to `google_compute_security_policy.rule.match`


diodfr:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Terraform Version

Terraform v0.12.13
google 3.5.0

### Affected Resource(s)

* google_compute_security_policy

### Terraform Configuration Files
```tf
# CLOUD ARMOR policy : WAF's default rules configuration
resource "google_compute_security_policy" "policy-WAF-default-stable" {
    name = "policy-WAF-default-stable"

    rule {
        action   = "deny(403)"
        priority = "1000"
        match {
            expr {
                expression = "evaluatePreconfiguredExpr('xss-stable') || evaluatePreconfiguredExpr('sqli-stable')"
            }
        }
        description = "Deny access to xss and sqli rules"
    }

    rule {
        action   = "allow"
        priority = "2147483647"
        description = "default rule"
    }
}
```

### Debug Output
https://gist.github.com/diodfr/f7b7d66090a9227f99578e10fb2043e1

### Expected Behavior
Accept "expr" block. See documentation https://www.terraform.io/docs/providers/google/r/compute_security_policy.html
The commit @megan07  https://github.com/terraform-providers/terraform-provider-google/pull/5070 made fields "config"  and "versioned_expr" optionaals but forgot to add the expr management even if it has been documented. 

Thank you for your work.

### Actual Behavior
A syntax error is reported.

### Steps to Reproduce

1. `terraform apply`

### References

* #0000

PR #5504, opened 2020-01-27
FIX ISSUE 5502 : unable to use "expr" in google_compute_security_policy.rule.match 

diodfr:
As described in the issue : https://github.com/terraform-providers/terraform-provider-google/issues/5502
I have tried to fix the use of "expr" in google_compute_security_policy.rule.match

**I was not able to launch the test to be sure it works (tests are skipped).**

### DIFF
--- a/google/resource_compute_security_policy.go
+++ b/google/resource_compute_security_policy.go
@@ -93,6 +93,21 @@ func resourceComputeSecurityPolicy() *schema.Resource {
 										Optional:     true,
 										ValidateFunc: validation.StringInSlice([]string{"SRC_IPS_V1"}, false),
 									},
+									
+									"expr": {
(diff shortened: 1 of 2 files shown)


slevenick:
It looks like the `expr` field is available only in the google-beta provider, but the field wasn't marked that way in the docs. You can use this field as expected by using the beta provider. See: https://www.terraform.io/docs/providers/google/guides/provider_versions.html

I'll check if this is ready come out of beta

slevenick:
https://github.com/GoogleCloudPlatform/magic-modules/pull/3028 should fix this

slevenick:
Closing in favor of https://github.com/GoogleCloudPlatform/magic-modules/pull/3028

PR #5541, opened 2020-01-29
update to Terraform 0.12 syntax and terraform fmt

pradeepbhadani:
Couldn't find the relevant file in `magic-modules` repo https://github.com/GoogleCloudPlatform/magic-modules/tree/master/third_party/terraform/website/docs/r , so created PR here.

### DIFF
--- a/website/docs/r/pubsub_topic_iam.html.markdown
+++ b/website/docs/r/pubsub_topic_iam.html.markdown
@@ -46,9 +46,9 @@ data "google_iam_policy" "admin" {
 }
 
 resource "google_pubsub_topic_iam_policy" "editor" {
-  project = "${google_pubsub_topic.example.project}"
-  topic = "${google_pubsub_topic.example.name}"
-  policy_data = "${data.google_iam_policy.admin.policy_data}"
+  project     = google_pubsub_topic.example.project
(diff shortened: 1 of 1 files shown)


danawillow:
Hey @pradeepbhadani, you couldn't find the files because they're autogenerated (you can tell by the comment at the top that says `AUTO GENERATED CODE`).

The overall IAM template markdown file is https://github.com/GoogleCloudPlatform/magic-modules/blob/master/templates/terraform/resource_iam.html.markdown.erb, and the file that generates the specific example is https://github.com/GoogleCloudPlatform/magic-modules/blob/master/templates/terraform/iam/iam_attributes.tf.erb. Getting the `=` to line up correctly won't be trivial, but you should at least be able to remove the `${`s.

We can't accept a PR into TPG that modifies an autogenerated file (since it'll just get overwritten the next time any other change gets made), so I'm going to close this, but that should give you the info needed to correctly make it against MM. Thanks!

pradeepbhadani:
I could figure out with the headers that this is Auto generated file but couldn't find the file in magic module which generates this file (as mentioned in my PR description).

Thanks for pointing out to the `erb` file which generate this specific documentation. I will raise separate PR in magic-module repo for this change.

Issue #5286, opened 2020-01-02
`availability_type` should be supported on MYSQL `sql_database_instance`


eyalzek:
https://www.terraform.io/docs/providers/google/r/sql_database_instance.html#availability_type

Regional HA is available for mysql instances (https://cloud.google.com/sql/docs/mysql/high-availability) but this attribute can only be set for postgres instance at the moment.

danawillow:
Hey @eyalzek, thanks for filing! Next time could you please fill out the issue template? That helps our bot label bugs / enhancements.

For this specific issue, does Terraform fail if you try to set availability_type on MySQL? I'm curious whether we need to update our code at all or just our docs.

eyalzek:
> Hey @eyalzek, thanks for filing! Next time could you please fill out the issue template? That helps our bot label bugs / enhancements.

Apologies! 

> For this specific issue, does Terraform fail if you try to set availability_type on MySQL? I'm curious whether we need to update our code at all or just our docs.

It fails, I can run it tomorrow and get you the output. The thing is, it looks like it's not supported in any Google api at the moment, nor gcloud. It's a manual process that has to be done via the cloud console.

eyalzek:
@danawillow the `waiting-response` label doesn't seem right to me :)

danawillow:
> It fails, I can run it tomorrow and get you the output. 
I was waiting for that :)

eyalzek:
@danawillow my bad! Here is trace level output for the failed request. Let me know if I can provide any additional info:

```
module.development_mysql_instance.google_sql_database_instance.database: Modifying... [id=development-mysql-master]
2020/02/03 10:52:48 [DEBUG] module.development_mysql_instance.google_sql_database_instance.database: applying the planned Update change
2020/02/03 10:52:48 [TRACE] GRPCProvider: ApplyResourceChange
2020-02-03T10:52:48.083+0100 [DEBUG] plugin.terraform-provider-google_v3.3.0_x5: 2020/02/03 10:52:48 [DEBUG] Waiting for state to become: [success]
2020-02-03T10:52:48.083+0100 [DEBUG] plugin.terraform-provider-google_v3.3.0_x5: 2020/02/03 10:52:48 [DEBUG] Google API Request Details:
2020-02-03T10:52:48.083+0100 [DEBUG] plugin.terraform-provider-google_v3.3.0_x5: ---[ REQUEST ]---------------------------------------
[...]

PR #5567, opened 2020-02-03
support MySQL 2nd gen to set up HA

ozashu:
#5286 
Confirmed that MySQL(Cloud SQL) 2nd generation 5.7 is also able to set up for high availability. 

The documentation needs to be fixed as:Availability type - (Optional) This specifies whether a PostgreSQL instance should be set up for high availability (`REGIONAL`) or single zone (`ZONAL`). It can be also applied for MySQL 2nd Gen. For MySQL, it should be set up for high availability (`REGIONAL`)without creating Failover replica instance using terraform apply command.

### DIFF
--- a/website/docs/r/sql_database_instance.html.markdown
+++ b/website/docs/r/sql_database_instance.html.markdown
@@ -232,6 +232,8 @@ The required `settings` block supports:
 
 * `availability_type` - (Optional) This specifies whether a PostgreSQL instance
     should be set up for high availability (`REGIONAL`) or single zone (`ZONAL`).
+    It can be also applied for MySQL 2nd Gen. For MySQL,
+    it should be set up for high availability (`REGIONAL`) without creating Failover replica instance using terraform apply command.
 
 * `crash_safe_replication` - (Optional, Deprecated) This property is only applicable to First Generation instances.
     First Generation instances are now deprecated, see [here](https://cloud.google.com/sql/docs/mysql/upgrade-2nd-gen)


ozashu:
I retry to set up for MySQL HA today.
Here is terraform plan result .

```
      ~ settings {
            activation_policy           = "ALWAYS"
            authorized_gae_applications = []
          + availability_type           = "REGIONAL"
```

And here is terraform apply result.

```
Apply complete! Resources: 1 added, 1 changed, 0 destroyed.
```

It failed when I created failover replica instance. 
I destroyed it and applied terraform setup MySQL HA.

rileykarson:
That's a failure for me as well, through both Terraform and the console. Can you share the actual config you're using? It's possible I'm missing some required setting on my instance.

ozashu:
Here is terraform plan's result of Excerpt.
```
      ~ settings {
            activation_policy           = "ALWAYS"
            authorized_gae_applications = []
          ~ availability_type           = "ZONAL" -> "REGIONAL"
            crash_safe_replication      = false
            disk_autoresize             = true
            disk_size                   = 10
            disk_type                   = "PD_SSD"
            pricing_plan                = "PER_USE"
            replication_type            = "SYNCHRONOUS"
            tier                        = "db-n1-standard-2"
            user_labels                 = {}
            version                     = 16

            backup_configuration {
                binary_log_enabled = true
                enabled            = true
                start_time         = "18:00"
            }
```
Couldn't you create MySQL HA instances by GCP console?

aharelick:
As @ozashu mentioned it looks like there is an API restriction and in order to use a `REGIONAL` instance you have to enable backups and binary logs (which share a setting with point in time backups). That restriction should probably be added to both the CloudSQL API docs and these terraform docs.

https://github.com/terraform-google-modules/terraform-google-sql-db/issues/71#issuecomment-604998264

ozashu:
#6254

PR #5578, opened 2020-02-04
In resource_google_project: Fix one bug with retries to Projects.Update(...) and add retries to GetBillingInfo(...) requests

spew:
This change affects two areas::
1. When the result of a Projects.Update(...) request is an error, the returned
    'project' value is 'nil'. Previously, this value was stored in 'p' which
    was then dereferenced when retrieving 'p.ProjectId' on the subsequent retry.
    This fix only allows 'p' to be overwritten if the Projects.Update(...)
    request is successful.
2. Requests to the BillingInfo service could fail due to transient HTTP timeouts
   (usually TLS handshake negotiation timeout) and for throttling (HTTP
   429). Adding retries addresses this.

```release-note:enhancement
resourcemanager: fixed retry behavior for updates in `google_project`, added retries for billing metadata requests
```

### DIFF
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -293,11 +293,8 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
 	if ok := d.HasChange("name"); ok {
 		p.Name = project_name
 		// Do update on project
-		if err = retryTimeDuration(func() (updateErr error) {
-			p, updateErr = config.clientResourceManager.Projects.Update(p.ProjectId, p).Do()
-			return updateErr
-		}, d.Timeout(schema.TimeoutUpdate)); err != nil {
-			return fmt.Errorf("Error updating project %q: %s", project_name, err)
+		if p, err = updateProject(config, d, project_name, p); err != nil {
+			return err
 		}
 
 		d.SetPartial("name")
@@ -310,11 +307,8 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
 		}
 
 		// Do update on project
-		if err = retryTimeDuration(func() (updateErr error) {
-			p, updateErr = config.clientResourceManager.Projects.Update(p.ProjectId, p).Do()
-			return updateErr
-		}, d.Timeout(schema.TimeoutUpdate)); err != nil {
-			return fmt.Errorf("Error updating project %q: %s", project_name, err)
+		if p, err = updateProject(config, d, project_name, p); err != nil {
+			return err
 		}
 		d.SetPartial("org_id")
 		d.SetPartial("folder_id")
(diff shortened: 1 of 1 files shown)


spew:
Failing check does not appear to be related to this change.

google/resource_google_project.go
```diff
@@ -346,6 +337,18 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
 	return resourceGoogleProjectRead(d, meta)
 }
 
+func updateProjectOverwriteOnSuccess(config *Config, d *schema.ResourceData, project_name string, projectPtrPtr **cloudresourcemanager.Project) error {
```
rileykarson: `**Project` feels a little awkward, we generally don't use double-pointers in TPG. I think that we can just pass in a `*Project`, and return `newProj` as a `*Project` on success, replacing `p` at the callsite.

```suggestion
func updateProjectOverwriteOnSuccess(config *Config, d *schema.ResourceData, projectName string, desiredProject *cloudresourcemanager.Project) (*cloudresourcemanager.Project, error) {
```

google/resource_google_project.go
```diff
@@ -429,7 +432,12 @@ func updateProjectBillingAccount(d *schema.ResourceData, config *Config) error {
 	if name != "" {
 		ba.BillingAccountName = "billingAccounts/" + name
 	}
-	_, err := config.clientBilling.Projects.UpdateBillingInfo(prefixedProject(pid), ba).Do()
+	updateBillingInfoFunc := func() error {
+		var err error
```
rileykarson: nit: this is equivalent to `_, err := config.clientBilling.Projects.UpdateBillingInfo(prefixedProject(pid), ba).Do()`, right?

google/resource_google_project.go
```diff
@@ -438,9 +446,13 @@ func updateProjectBillingAccount(d *schema.ResourceData, config *Config) error {
...
 	for retries := 0; retries < 3; retries++ {
-		ba, err = config.clientBilling.Projects.GetBillingInfo(prefixedProject(pid)).Do()
+		var ba *cloudbilling.ProjectBillingInfo
+		err = retryTimeDuration(func() (reqErr error) {
+			ba, reqErr = config.clientBilling.Projects.GetBillingInfo(prefixedProject(pid)).Do()
+			return reqErr
+		}, d.Timeout(schema.TimeoutRead))
 		if err != nil {
-			return err
+			return fmt.Errorf("Error getting billing info for project %q: %v", prefixedProject(pid), err)
```
rileykarson:
```suggestion
			return fmt.Errorf("error getting billing info for project %q: %v", prefixedProject(pid), err)
```

google/resource_google_project.go
spew: Yes, I think it is the same. I think that there is some worry about the value for 'err' persisting past the function closure (if err was actually declared before the function), but I am not even sure if that would happen if there was an 'err' defined before this function.

Will change.

google/resource_google_project.go
spew: That sounds fine, will change.

google/resource_google_project.go
spew: I used the capital 'Error' to match the other strings in this file even though it is 'bad' go style. I thought consistency would be preferred?

rileykarson: Yeah, either is fine.

spew:
Updated post comments.

PR #5644, opened 2020-02-11
google_dataflow_job - added job_id attribute

### LINKED ISSUE #5596: Support a "job_id" field for google_dataflow_job
### Community Note

### Description

Dataflow Jobs are uniquely identified by their job ID instead of their name. My request is to expose the job ID as a computed field, `job_id`.

It seems that the Terraform provider is already retrieving the job ID and setting it as the Terraform resource ID as seen [here](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_dataflow_job.go#L194-L198), so exposing it as a computed field should hopefully be reasonably doable.

Furthermore, there is precedence in using a computed identifier as the Terraform resource ID _and_ exposed as a computed field, as seen with the `name` field of `google_firestore_index` [here](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_firestore_index.go#L188-L196).

* google_dataflow_job

jeet-parekh:
Closes  #5596.

### DIFF
--- a/google/resource_dataflow_job.go
+++ b/google/resource_dataflow_job.go
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
-	"google.golang.org/api/dataflow/v1b3"
+	dataflow "google.golang.org/api/dataflow/v1b3"
 	"google.golang.org/api/googleapi"
 )
 
@@ -147,6 +147,11 @@ func resourceDataflowJob() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validation.StringInSlice([]string{"WORKER_IP_PUBLIC", "WORKER_IP_PRIVATE", ""}, false),
 			},
+
+			"job_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


google/resource_dataflow_job.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
-	"google.golang.org/api/dataflow/v1b3"
+	dataflow "google.golang.org/api/dataflow/v1b3"
```
jeet-parekh: `dataflow` was added automatically by the go tools... confused whether or not this is correct

google/resource_dataflow_job.go
```diff
@@ -147,6 +147,11 @@ func resourceDataflowJob() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validation.StringInSlice([]string{"WORKER_IP_PUBLIC", "WORKER_IP_PRIVATE", ""}, false),
 			},
+
+			"job_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
jeet-parekh: should I add `ForceNew: true` as well?

megan07: `ForceNew` is only applicable on input fields.  It signifies that the field is not updatable, and if the value changes a new resource needs to be recreated.  Because `job_id` is read only, it is not needed here.  Good question!

Issue #5222, opened 2019-12-18
Project is getting tainted in TF state file, even though it got created; when billing id is incorrect


pb0101:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment
* If an issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to "hashibot", a community member has claimed the issue already.

### Terraform Version

terraform -v
Terraform v0.12.17
+ provider.google v2.20.0
+ provider.random v2.2.1

### Affected Resource(s)

* google_project

### Terraform Configuration Files

```tf
resource "google_project" "my_project" {
 name       = var.gcp_project
 project_id = "xxxx-project-name-xxxxx"
 billing_account = var.billing_id
 folder_id = "77777777777777777"
}

```

### Expected Behavior

Project should update the correct billing id, instead of deleting and creating the project again.
 
### Actual Behavior

When running terraform for creating project (with incorrect billing id), project gets created. But when applying an update to the project with correct billing id; terraform is destroying the project and creating it again. This causes name conflict, because the project is pending deletion.

### Steps to Reproduce

1. Create project with incorrect billing id
2. Fix the billing id and apply changes.

1. `terraform apply`

### Important Factoids

terraform apply --auto-approve (with incorrect billing id)
google_project.my_project: Creating...

Error: Error setting billing account "yyyyyyyyyyyyy" for project "projects/xxxxxxxxxx": googleapi: Error 400: Request contains an invalid argument., badRequest

  on main.tf line 50, in resource "google_project" "my_project":
  50: resource "google_project" "my_project" {

Apply after fixing the billing id 

google_project.my_project: Destroying... [id=xxxxxxxxxxxxx]
google_project.my_project: Destruction complete after 5s
google_project.my_project: Creating...

[...]

PR #5671, opened 2020-02-13
project: check for billing account permissions as pre-requisite

umairidris:
Currently, the provider will create the project but then fail the billing account link step leaving the resource in a 'tainted' state. This pre-emptively checks whether the caller will have access to the billing account before the project is created.

We now get an error like:

```
Error: failed pre-requisites: missing permissions on "billingAccounts/abc": [billing.resourceAssociations.create]
```

https://github.com/terraform-google-modules/terraform-google-project-factory/issues/373

### DIFF
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -191,7 +191,7 @@ func resourceGoogleProjectCheckPreRequisites(config *Config, d *schema.ResourceD
 	if err != nil {
 		return fmt.Errorf("failed to check permissions on billing account %q: %v", ba, err)
 	}
-	if has := sliceContains(perm, resp.Permissions); !has {
+	if !stringInSlice(resp.Permissions); !has {
 		return fmt.Errorf("missing permissions on %q: %v", ba, perm)
 	}
 	return nil
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -191,7 +191,7 @@ func resourceGoogleProjectCheckPreRequisites(config *Config, d *schema.ResourceD
 		return fmt.Errorf("failed to check permissions on billing account %q: %v", ba, err)
 	}
 	if diff := diffStringSlices(resp.Permissions, req.Permissions); len(diff) > 0 {
-		return fmt.Errorf("missing permissions on org %q: %v", ba, diff)
+		return fmt.Errorf("missing permissions on billing account %q: %v", ba, diff)
 	}
 	return nil
 }
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -94,6 +94,10 @@ func resourceGoogleProject() *schema.Resource {
 func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
+	if err := resourceGoogleProjectCheckPreRequisites(config, d); err != nil {
+		return fmt.Errorf("failed pre-requisites: %v", err)
+	}
+
 	var pid string
 	var err error
 	pid = d.Get("project_id").(string)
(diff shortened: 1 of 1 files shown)


google/resource_google_project.go
```diff
@@ -173,6 +177,40 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
...
+	if !ok {
+		return nil
+	}
+	ba := "billingAccounts/" + ib.(string)
+	req := &cloudbilling.TestIamPermissionsRequest{
+		Permissions: []string{"billing.resourceAssociations.create"},
+	}
+	resp, err := config.clientBilling.BillingAccounts.TestIamPermissions(ba, req).Do()
+	if err != nil {
+		return fmt.Errorf("failed to check permissions on billing account %q: %v", ba, err)
```
chrisst: I think it's super edge case but I can see a possibility that a user will fail this call but still be able to successfully create the project and update the billing account. For example a transient network error could cause this to fail but since we wrap the call to create in retries it will survive these such errors.

You could wrap this in retries, but IMO we can just log this at WARN and return nil as this is good enough for the 99% case.

google/resource_google_project.go
```diff
@@ -173,6 +177,40 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
...
+	}
+	if diff := diffStringSlices(resp.Permissions, req.Permissions); len(diff) > 0 {
```
chrisst: This is a bit overly complicated. Can you just loop through `resp.Permissions` and make sure that `billing.resourceAssociations.create` is present instead of doing this slice comparison?

chrisst:
@rileykarson that's an interesting thought. It would have to be restricted to only call/fail if the resource was just about to be created. Also in theory a user could be generating a `plan` output with lower credentials to be delegated to an account with higher privileges and failing at plan time could break that work flow.
For now I think the approach in this PR is a good conservative option.

google/resource_google_project.go
umairidris: Done.

google/resource_google_project.go
```diff
@@ -173,6 +177,35 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
...
+	ba := "billingAccounts/" + ib.(string)
+	const perm = "billing.resourceAssociations.create"
+	req := &cloudbilling.TestIamPermissionsRequest{
+		Permissions: []string{perm},
+	}
+	resp, err := config.clientBilling.BillingAccounts.TestIamPermissions(ba, req).Do()
+	if err != nil {
+		return fmt.Errorf("failed to check permissions on billing account %q: %v", ba, err)
+	}
+	if has := sliceContains(perm, resp.Permissions); !has {
```
danawillow: nit: we already have a function in this package that does this, so you can use stringInSlice(resp.Permissions, perm). You also don't use the value of `has` anywhere so you could just do:
```
if !stringInSlice(resp.Permissions, perm) {
...
}
```

google/resource_google_project.go
danawillow: @chrisst, if this check failed as just a warning in the logs, how would a typical user notice?

chrisst: @danawillow  It will just revert to existing behavior:
a) if the permission isn't valid, it would  taint the resource
b) if the permission is valid, the resource would create correctly. 

Since this is just a precondition check it doesn't seem legitimate that a precondition should fail the resource creation for a transient error.

Relying on the failure to fetch permissions as a signal that api's aren't correctly enabled seems like it could just be another explicit check. We have a similar tainting problem when trying to delete the default network but `compute.googleapis.com` isn't enabled.

chrisst: @umairidris I'd be happy with either explicit checks for api enablement + silent failures or retries added. My primary concern is not adding a new category of failing to create a project

danawillow: Do you feel like having this check adds extra value then over the current behavior?

chrisst: It will fix https://github.com/terraform-providers/terraform-provider-google/issues/5222

danawillow: I misunderstood what @chrisst was saying at first, but I get it now. Basically, if the call to retrieve permissions fails, then we should ignore that because it was likely transient. but if the call succeeds and the permission isn't there, then fail before trying to create the project. Checking for API enablement is something we also can do first to make sure that the call to retrieve permissions fails for only transient or new reasons, not ones we already understand, and retries on the testIamPermissions call wouldn't fix that because they would just fail every time. I'm all caught up now :)

umairidris: Do you have a way to check if an API is enabled on the caller? I haven't been able to find a good way apart from trying to just use the API client to do something...

That also makes checking for compute api tricky as you need a project to be able to call the API on, but you don't have the project in the pre-requisites check...

The preconditions script tries to work around this by calling list APIs on the project set in `gcloud config` but this isn't always a valid check. For example if I am running from my local workstation with my own credentials then it doesn't matter what project I set in gcloud, I believe it uses the default gcloud project for my API calls.

google/resource_google_project.go
umairidris: Done.

google/resource_google_project.go
umairidris: friendly ping @chrisst

chrisst: Yeah I didn't think about the chicken/egg problem with enabling the compute API. I did some digging and I'm not sure if billing accounts are a parent type that can have services enabled/disabled on it. I think this change is definitely an improvement as is, so I'm happy accepting that there's a small chance somebody will fail due to a transitive error. If we see people filing issues about it we can reasses.

umairidris: @chrisst I was thinking, if we create a compute API client and make a call on a non-existent project, would the error be the project doesn't exist or would it error out that the API is not enabled on the caller? I am guessing it would be the latter. If so, would this be an appropriate "hack" to work around this? i.e. call the API and check if the error is API not enabled OR check for resource not found if it is a specific error code?

I can do some testing around this idea...

Issue #5580, opened 2020-02-04
Add support for `upgrade_settings` in `google_container_cluster` NAP


davidquarles:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

The upstream/v1 cluster api now exposes [`upgradeSettings`](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/UpgradeSettings) within [`NAP defaults`](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#Cluster.AutoprovisioningNodePoolDefaults).  These fields are already supported in the (beta) `google_container_node_pool` resource, and it would be great if exposed them here, too.

### New or Affected Resource(s)

* google_container_cluster

### Potential Terraform Configuration

```tf
resource "google_container_cluster" "foo" {
  name = "foo"
  # ...
  cluster_autoscaling {
    auto_provisioning_defaults {
      service_account = google_service_account.cluster_nodes_service_account.email

      # new field, mirroring google_container_node_pool.upgrade_settings structure
      upgrade_settings {
        max_surge = "25%"
        max_unavailable = "0"
      }
    }
  }
}
```

### References

* [terraform-providers/terraform-provider-google-beta#2669](https://github.com/terraform-providers/terraform-provider-google-beta/pull/2669)
* https://github.com/terraform-providers/terraform-provider-google-beta/pull/1400
* https://github.com/GoogleCloudPlatform/magic-modules/pull/2669

davidquarles:
I haven't contributed here in ~2 years as I've been in non-GCP environments (I can't adequately express how good it is to be back), so please forgive my ignorance around the new repo structure.  Should I be submitting issues over in magic-modules instead/too?  I'm happy to contribute, if that's helpful.  Thanks!

[...]

PR #5676, opened 2020-02-13
add support for `upgrade_settings` in both node config and NAP config

davidquarles:


davidquarles:
This fixes #5580.

davidquarles:
@rileykarson @danawillow I'm still finishing up writing tests here, but had two quick questions:
1. Is it actually preferable to factor the shared upgrade settings config out into its own file, as was done with e.g. `google/node_config.go`?
2. Should we do any sort of validation here of the upstream constraint that `max_surge` + `max_unavailable` <= 20, or does it make more sense to just let the failure occur and propagate down?  If we prefer the former, any guidance is appreciated, as I'm not sure how/where to do that sort of composite validation.

davidquarles:
> 2. Should we do any sort of validation here of the upstream constraint that `max_surge` + `max_unavailable` <= 20, or does it make more sense to just let the failure occur and propagate down?  If we prefer the former, any guidance is appreciated, as I'm not sure how/where to do that sort of composite validation.

Additional context:
After testing, it appears the google API will allow you to violate this constraint at both the node pool and NAP layers and silently fails in subtly different ways.  NAP config successfully applies (I used maxUnavailable=1, maxSurge=20), but NAP autoscale is subsequently broken.  I tried this same config at the node pool layer and terraform emitted no errors, but the node pool itself didn't update and nothing change in terraform state.

ndmckinley:
This looks great.

> 1. Is it actually preferable to factor the shared upgrade settings config out into its own file, as was done with e.g. `google/node_config.go`?

In my opinion, no - I'd prefer it if you put it in `container_cluster`.

> 2. Should we do any sort of validation here of the upstream constraint that `max_surge` + `max_unavailable` <= 20, or does it make more sense to just let the failure occur and propagate down?  If we prefer the former, any guidance is appreciated, as I'm not sure how/where to do that sort of composite validation.

I think it's probably best to add this validation - our general approach is that validation is desirable unless it would impose an undue maintenance burden.  In this case where the server-side can break if the validation isn't added, it seems extremely useful to add.

rileykarson:
Closing PR as stale

Issue #4141, opened 2019-07-31
google_logging_metric needs metric_descriptor.labels in a specific order or results in permadiff


dijitali:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment
* If an issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to "hashibot", a community member has claimed the issue already.

### Terraform Version

``` shell
terraform -v
Terraform v0.12.3
+ provider.google v2.11.0
+ provider.google-beta v2.11.0
+ provider.random v2.1.2
```

### Affected Resource(s)

* `google_logging_metric`

### Terraform Configuration Files

```tf
resource "google_logging_metric" "logging_metric" {
  name   = "mass-foo-bar"
  filter = "resource.type=gae_app AND severity>=ERROR"
  metric_descriptor {
    metric_kind = "DELTA"
    value_type  = "DISTRIBUTION"
    labels {
      key         = "mass"
      value_type  = "STRING"
      description = "amount of matter"
    }
    labels {
      key         = "foo"
      value_type  = "STRING"
      description = "amount of foo"
    }
    labels {
      key         = "bar"
      value_type  = "STRING"
      description = "amount of bar"
    }
  }
  value_extractor = "EXTRACT(jsonPayload.request)"
  label_extractors = {
    "mass" = "EXTRACT(jsonPayload.request)"
    "foo"  = "EXTRACT(jsonPayload.foo)"
    "bar"  = "EXTRACT(jsonPayload.bar)"
  }
  bucket_options {
    linear_buckets {
      num_finite_buckets = 3
      width              = 1
      offset             = 1
    }
  }
}
```

### Debug Output

N/A

### Panic Output

N/A

### Expected Behavior

After terraform apply creates the `google_logging_metric` resource, subsequent `plan` and `apply` commands do not result in a change to the resource.

### Actual Behavior

Resource results in a "permadiff" and continually tries to rearrange the resource metric_descriptor label order:

``` shell
An execution plan has been generated and is shown below.
[...]

PR #5732, opened 2020-02-21
Fix potential perma-diff with metric_descriptor.labels in logging_metric

xanonid:
Fixes: #4141

### DIFF
--- a/google/resource_logging_metric.go
+++ b/google/resource_logging_metric.go
@@ -87,7 +87,7 @@ example, the appengine.googleapis.com/http/server/response_latencies metric type
 for the HTTP response code, response_code, so you can look at latencies for successful responses
 or just for responses that failed.`,
 							Elem: loggingMetricMetricDescriptorLabelsSchema(),
-							// Default schema.HashSchema is used.
+							Set:  schema.HashResource(loggingMetricMetricDescriptorLabelsSchema()),
 						},
 						"unit": {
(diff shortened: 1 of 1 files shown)


xanonid:
It would be helpful, if anyone can shortly explain, why this change is needed at all. I would have expected that it have worked out-of-the-box (`Set: schema.HashResource(...)` auto-set) - but apparently it did not. Are there other places, where this needs to be explicitly set?

emilymye:
Hi @xanonid  - do you have an example of config that recreates this issue? I'm having trouble reproducing, and the default HashSchema calls out to HashResource(schema.Elem), so this shouldn't be necessary?

xanonid:
Hi @emilymye, I will prepare a minimal example, it should be reproducible by having more than one label and then simply reordering the labels after applying. The related bug #4141 has currently 11 thumps-up, so there seems to be really an issue.
I have it also in a slightly complicated setup where adjusting the order of the labels is not a solution as the metric resource is in a more than once used module (and depending on the instance, unfortunately different label orders would be currently needed to be free of perma-diffs).
I do also not understand yet why the fix worked as it should be already the default. Maybe somewhere a proper initialization of the Set is missing...

xanonid:
@emilymye 
I close this PR as the issue seems to be really already fixed - sorry. I had unfortunately a version constraints to an outdated 2.x provider version which had still the issue.

PR #5785, opened 2020-02-27
Update bigquery_table.html.markdown

minac:
Correct import structure


minac:
Ping?

rileykarson:
The slash format appears to be correct for current provider versions. Closing.

PR #5799, opened 2020-03-02
Style fix for cloudrun noauth example

kanolato:


website/docs/r/cloud_run_service.html.markdown
```diff
@@ -141,11 +141,11 @@ data "google_iam_policy" "noauth" {
...
 resource "google_cloud_run_service_iam_policy" "noauth" {
-  location    = google_cloud_run_service.default.location
-  project     = google_cloud_run_service.default.project
-  service     = google_cloud_run_service.default.name
+  location    = "${google_cloud_run_service.default.location}"
+  project     = "${google_cloud_run_service.default.project}"
+  service     = "${google_cloud_run_service.default.name}"
 
-  policy_data = data.google_iam_policy.noauth.policy_data
+  policy_data = "${data.google_iam_policy.security-scanner.policy_data}"
```
megan07: We don't want to change `noauth` to `security-scanner` here, because it looks like the datasource in line 134 has the name `noauth` and that is what we are referencing.

As for the other changes, the reason we dropped the `"${...}"` from the example is due to the fact that the google provider only supports Terraform 0.12.  You can see the change explained here https://www.terraform.io/upgrade-guides/0-12.html#first-class-expressions

megan07:
Hi @kanolato!  Since I haven't heard from you regarding this, I will close this PR.  If you have any issues running this example, please open an issue and we can help you there.  Thanks!

PR #5843, opened 2020-03-06
Fixed broken URL

svachmic:
As I was reading through the documentation while implementing automatic cloud build triggers I came across this broken link. Seems like a typo, so I fixed it. Just copy-pasted the link in the same description above (ignoredFiles/includedFiles).


svachmic:
Build seems to have failed on a broken link - interestingly enough, not the one I fixed.

PR #5859, opened 2020-03-09
correct import statement for uptime config

JaymoKang:


ndmckinley:
This documentation is correct as-is - you can see further up in the file that the format of `{{name}}` is `projects/[PROJECT_ID]/uptimeCheckConfigs/[UPTIME_CHECK_ID]`.

PR #5906, opened 2020-03-17
Update sql_database_instance.html.markdown

mehdicopter:
I noticed that the optional setting `settings.maintenance_window` starts from **Tuesday** and not from **Monday**.


megan07:
Hi @mehdicopter!  Thanks for this PR!  It appears in the API documentation that it should start on Monday https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1beta4/instances?hl=en#maintenancewindow. If you're seeing something different, please open an issue with a debug log and we can track it there. Thanks!

PR #5949, opened 2020-03-23
Update field requirements

AmandaHassoun:
Terraform apply was complaining that the project field was required, so updating the docs accordingly

### DIFF
--- a/website/docs/r/google_service_account.html.markdown
+++ b/website/docs/r/google_service_account.html.markdown
@@ -31,6 +31,9 @@ resource "google_service_account" "service_account" {
 
 The following arguments are supported:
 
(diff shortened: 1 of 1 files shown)


emilymye:
Hi @AmandaHassoun - while overall the project is required for creating the service account, it is optional on the resource because we allow for provider-default values. i.e. it's possible to do 

```
provider "google" {
   project = "my-project"
}

resource  "google_service_account" {
    name = "my-sa-name"
}
```

where project isn't provided and thus optional for the service account. 

I'm going to close this PR but feel free to re-open if you have other questions!

AmandaHassoun:
thanks @emilymye

PR #5953, opened 2020-03-23
add new project to tests

megan07:
TEMP TEST TO TEST AGAINST TC


megan07:
apparently I don't need to open a PR to test it, can just use the branch :)

PR #5983, opened 2020-03-26
Add missing quotes to example for saving GCP SA key as k8s secret

muru:
Without the quotes, terraform complains about missing quotes.


muru:
Oh, sorry, I just skimmed through the contribution guide and didn't realize that the auto-generated part applies to most of the repo. I noticed the missing CLA warning in the upstream PR, and have now signed it, things should be fine. Should this PR just be closed if it actually gets upstreamed?

danawillow:
I can merge both- what you did is totally fine. That specific file isn't autogenerated, so I was able to easily run a script to patch your change into the MM repo. For more complicated changes though, it's better to do in MM directly since the script can be finicky.

PR #6023, opened 2020-04-01
Standardize google_compute_instance import

jcanseco:
This patch changes the import ID of google_compute_instance to have a
format that is more consistent with other resources like
google_compute_subnetwork while still supporting the previously
supported import ID of {{project}}/{{zone}}/{{name}} in order to have
backwards compatibility.

### DIFF
--- a/google/resource_compute_instance.go
+++ b/google/resource_compute_instance.go
@@ -1773,16 +1773,21 @@ func resourceComputeInstanceDelete(d *schema.ResourceData, meta interface{}) err
 }
 
 func resourceComputeInstanceImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-	parts := strings.Split(d.Id(), "/")
-
-	if len(parts) != 3 {
-		return nil, fmt.Errorf("Invalid import id %q. Expecting {project}/{zone}/{instance_name}", d.Id())
+	config := meta.(*Config)
+	if err := parseImportId([]string{
+		"projects/(?P<project>[^/]+)/zones/(?P<zone>[^/]+)/instances/(?P<name>[^/]+)",
+		"(?P<project>[^/]+)/(?P<zone>[^/]+)/(?P<name>[^/]+)",
+		"(?P<name>[^/]+)",
+	}, d, config); err != nil {
+		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


emilymye:
Hi @jcanseco. Thanks for submitting! We'll go ahead and upstream this for you. In the future, if you want to do a similar change, it saves us a bit of work if you do so against [MM](https://github.com/GoogleCloudPlatform/magic-modules). In this case, since instance is still a handwritten resource, it'd be [these](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/resources/resource_compute_instance.go) [files](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/website/docs/r/compute_instance.html.markdown).

Otherwise for autogenerated resources, it'd be in our terraform.yaml (see import_format, [example](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/compute/terraform.yaml#L947)) or inferred from the id format. Not sure if you might have already known, but just in case!

google/resource_compute_instance.go
```diff
@@ -1773,16 +1773,20 @@ func resourceComputeInstanceDelete(d *schema.ResourceData, meta interface{}) err
...
+	}, d, config); err != nil {
+		return nil, err
 	}
-
-	d.Set("project", parts[0])
-	d.Set("zone", parts[1])
-	d.Set("name", parts[2])
-	d.SetId(fmt.Sprintf("projects/%s/zones/%s/instances/%s", parts[0], parts[1], parts[2]))
+	// Replace import id for the resource id
+	id, err := replaceVars(d, config, "{{project}}/{{zone}}/{{name}}")
```
slevenick, line 1785: The ID changed here, is that intentional? I think we actually want to keep the old long name version

jcanseco: Mhm, I don't think I remember anymore since I made this patch to our local copy of the Terraform provider awhile back. It doesn't seem like there's a reason not to do what you suggest though. Fixed.

jcanseco:
Hi @emilymye thanks, the links are helpful. I'll make sure to apply changes to Magic Modules next time. In the meantime, I hope it's ok if I keep my last remaining PR open in this repo (I'm in the process of addressing its comments currently).

PR #6028, opened 2020-04-01
Update container_node_pool.html.markdown

Kodyc1:
The node pool will not autoscale when the node pool is created with 0 nodes and node_count is set to < 1.

### DIFF
--- a/website/docs/r/container_node_pool.html.markdown
+++ b/website/docs/r/container_node_pool.html.markdown
@@ -142,7 +142,7 @@ cluster.
     [google_container_cluster](container_cluster.html) for schema.
 
 * `node_count` - (Optional) The number of nodes per instance group. This field can be used to
-    update the number of nodes per instance group but should not be used alongside `autoscaling`.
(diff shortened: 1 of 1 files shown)


rileykarson:
Hey @Kodyc1! I'm not familiar with this as a restriction at the node pool level. I believe that GKE won't allow a cluster to scale to less than 1 node altogether, but individual node pools can have a `node_count` of `0`. That's suggested [here](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler#minimum_and_maximum_node_pool_size).

How did you run into this restriction? Was it returned in an error message, or did you find it elsewhere in GKE docs?

danawillow:
Hey @Kodyc1, can you explain this change a bit more? The current docs say not to set that field if `autoscaling`  is getting set, so the change to now say **to** set the field is a bit surprising.

danawillow:
Oh whoops needed to refresh my page I guess

rileykarson:
Closing this as stale- feel free to reopen / send a new PR if you'd like to pick it back up.

PR #6036, opened 2020-04-03
google_source_repo_repository - allow topics to be specified as topic names

jcanseco:
This patches the google_source_repo_repository resource to allow the
`pubsub_configs.topic` field to be specified with topic names, whereas
previously, only values in the form of
`projects/{project}/topics/{topicName}` were allowed.

The implementation takes the same approach taken by
google_storage_notification [1].

[1] github.com/terraform-providers/terraform-provider-google/blob/f3c888bb41beade2dec51934185722d15b83195a/google/resource_storage_notification.go#L87-L95

### DIFF
--- a/google/resource_source_repo_repository.go
+++ b/google/resource_source_repo_repository.go
@@ -59,8 +59,9 @@ Keyed by the topic names.`,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"topic": {
-							Type:     schema.TypeString,
-							Required: true,
+							Type:             schema.TypeString,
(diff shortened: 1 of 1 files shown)


rileykarson:
This file is MM-generated, and the change doesn't immediately slot in to one of our existing custom code injection points. The required change isn't large, it might just take a couple days to do / get reviewed / merged. I won't merge this PR, since it's going to look a little bit different.

rileykarson:
Done in https://github.com/GoogleCloudPlatform/magic-modules/pull/3339

PR #6064, opened 2020-04-07
Fix docs bug in google_compute_instance_template

ArseniiPetrovich:
It seems that both `source` and `source_image` parameters are optional. This is essential, because otherwise there is no option to create disk during instance creation.

### DIFF
--- a/website/docs/r/compute_instance_template.html.markdown
+++ b/website/docs/r/compute_instance_template.html.markdown
@@ -267,6 +267,7 @@ The `disk` block supports:
     `projects/{project}/global/images/family/{family}`, `global/images/{image}`,
     `global/images/family/{family}`, `family/{family}`, `{project}/{family}`,
     `{project}/{image}`, `{family}`, or `{image}`.
+~> **Note:** Either `source` or `source_image` is **required** when creating a new instance except for when creating a local SSD. Check the API [docs](https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/insert) for details.
 
 * `interface` - (Optional) Specifies the disk interface to use for attaching
     this disk.
@@ -277,6 +278,7 @@ The `disk` block supports:
 
 * `source` - (Optional) The name (**not self_link**)
     of the disk (such as those managed by `google_compute_disk`) to attach.
+~> **Note:** Either `source` or `source_image` is **required** when creating a new instance except for when creating a local SSD. Check the API [docs](https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/insert) for details.
 
 * `disk_type` - (Optional) The GCE disk type. Can be either `"pd-ssd"`,
     `"local-ssd"`, or `"pd-standard"`.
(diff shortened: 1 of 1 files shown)


emilymye:
@ArseniiPetrovich this looks like our handwritten docs were written to show that either source or source_image are required. While this makes both `Optional`, we probably intentionally have them as `("Required if $otherField isn't specified")` in the docs since it confused someone prior. I'd be ok with your changes if we kept the note in the description of the field, i.e. 
`

* `source_image` - (Optional). ***Required if source is not set.*** The image from which to 
    initialize this disk. This can be one of: the image's `self_link`,	    initialize this disk. This can be one of: the image's `self_link`,
    `projects/{project}/global/images/{image}`,	    `projects/{project}/global/images/{image}`,
    `projects/{project}/global/images/family/{family}`, `global/images/{image}`,


> This is essential, because otherwise there is no option to create disk during instance creation.

I'm not sure I understand what "otherwise" means, could you clarify?  Like, giving both? Not giving either?

ArseniiPetrovich:
@emilymye, thank you for your answer! In my project I need to create a MIG that will automatically create a second disk for each of my instances inside of MIG. I was confused with the docs, because there were no option for me to do that. If I specify `source` - I would have to reuse the existing disk, if I specify `source_image` the second disk will be created, but it will not be empty. It worked only when I removed both `source` and `source_image` fields. I also checked the Go code you have for this resource. I could have missed something, but I didn't saw the `("Required if $otherField isn't specified")` check. I would be grateful if you double-check this so we can either continue with this PR or discard it. 
Many thanks in advance!

emilymye:
Ah, interesting. In the past, the docs and API were explicit about not allowing templates without some source. We defaulted to the API validation (i.e. returing the errors from the API) if you didn't specify one, so there isn't validation in Terraform regardless. 

Now it looks like:

- Local SSDs don't require a source
- Otherwise one of source_image, source OR source_snapshot must be set

In that case, would you be able to keep the `(Optional)` and move the Required note to the end of the description, edited to be more accurate? e.g. 

- source_image - (Optional) ...
${description}
When creating a new instance, one of `source_image`, `source_snapshot`, or `source` is required except for when creating a local SSD.
```

as mentioned in the API [docs](https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/insert)

ArseniiPetrovich:
Sure, great! :)
I have added new commit introducing docs changes. I've paraphrased things a bit, but keeping the same idea as in your post. However I have noticed that you've mentioned a `source_snapshot` parameter which is not allowed (as far as I can see from the Go code) and not mentioned at the docs. If it should be supported, should I probably create an issue for that?

Thanks, @emilymye

emilymye:
Ah, yes! we have issues for _image and region_disk but not instance template. Feel free to exclude snapshot for now

ArseniiPetrovich:
I didn't include snapshot in my commits for now, so they should be okay :)

Issue #5959, opened 2020-03-24
Support for one and two nodes production bigtable clusters


Lirt:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

Bigtable now  supports one- and two-node production clusters (https://groups.google.com/forum/#!topic/google-cloud-bigtable-discuss/7gKYvoLrj10). 

Can we support this in provider?

Currently we are still at error `expected cluster.0.num_nodes to be at least (3), got 1`. (Tested with stable and beta `3.14.0`)

### New or Affected Resource(s)

`google_bigtable_instance`

Lirt:
There was a mistake in release 3.18. The release mentioned that this change is in it, but somehow it doesn't contain the commits.

I think mistake happened here https://github.com/terraform-providers/terraform-provider-google/commit/486fd05631067f4cccec90d9b4803a844d5dccee

chrisst:
Apologies, I didn't upstream this fix correctly to our code generator and it did get overwritten. I should be able to able to get this in time to get into the 3.19 release.

ykodama:
Same here.
@chrisst [3.18 release note] should be fixed.

PR #6083, opened 2020-04-09
Allow one and two nodes production bigtable clusters

vitovitolo:
Fixes https://github.com/terraform-providers/terraform-provider-google/issues/5959

Relax number of nodes validation to one for PRODUCTION clusters.

```release-note:enhancement
bigtable: Reduced the minimum number of nodes for the `bigtable_instace` resource from 3 to 1.
```


erlendnils1:
Looks like this didn't make the 3.18 release after all? Overwritten after merge, perhaps by accident? Would really like to see this in the next release

Lirt:
Release notes mentions this, but it doesn't work for me either.

PR #6196, opened 2020-04-24
Add support for providing additional experiments to Dataflow job

wazim:
An example use case is as-per: https://cloud.google.com/dataflow/docs/guides/using-cloud-monitoring#receive_worker_vm_metrics_from_the_agent. It is needed that an additional experiment is configured: `enable_stackdriver_agent_metrics`.

An example looks like:
```
resource "google_dataflow_job" "with_additional_experiments" {
  name = "dataflow-job"

  additional_experiments = ["enable_stackdriver_agent_metrics"]

  template_gcs_path = "gs://testing/template"
  temp_gcs_location = "gs://testing/tmp"
  parameters = {
    inputFile = "gs://testing/input"
    output    = "gs://testing/output"
  }
  on_delete = "cancel"
}
```

### DIFF
--- a/google/resource_dataflow_job.go
+++ b/google/resource_dataflow_job.go
@@ -152,7 +152,7 @@ func resourceDataflowJob() *schema.Resource {
 			},
 
 			"additional_experiments": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Optional: true,
 				ForceNew: true,
 				Elem: &schema.Schema{
@@ -188,7 +188,7 @@ func resourceDataflowJobCreate(d *schema.ResourceData, meta interface{}) error {
 
 	params := expandStringMap(d, "parameters")
 	labels := expandStringMap(d, "labels")
-	additionalExperiments := convertStringArr(d.Get("additional_experiments").([]interface{}))
+	additionalExperiments := convertStringSet(d.Get("additional_experiments").(*schema.Set))
 
 	env := dataflow.RuntimeEnvironment{
 		MaxWorkers:            int64(d.Get("max_workers").(int)),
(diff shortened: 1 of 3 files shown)


google/resource_dataflow_job.go
```diff
@@ -151,6 +151,15 @@ func resourceDataflowJob() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{"WORKER_IP_PUBLIC", "WORKER_IP_PRIVATE", ""}, false),
 			},
 
+			"additional_experiments": {
+				Type:     schema.TypeList,
```
danawillow, line 155: Is the order here meaningful? If not, I'd recommend using a TypeSet instead.

google/resource_dataflow_job_test.go
```diff
@@ -182,6 +182,30 @@ func TestAccDataflowJob_withIpConfig(t *testing.T) {
 	})
 }
 
+func TestAccDataflowJobWithAdditionalExperiments(t *testing.T) {
+	t.Parallel()
+
+	randStr := randString(t, 10)
+	bucket := "tf-test-dataflow-gcs-" + randStr
+	job := "tf-test-dataflow-job-" + randStr
+	additionalExperiments := []string{"enable_stackdriver_agent_metrics"}
```
danawillow, line 191: Want to add a few more values to this list just for extra confidence?

google/resource_dataflow_job.go
wazim: Good idea! I've updated it to `TypeSet`. Thanks!

google/resource_dataflow_job_test.go
wazim: I've added `shuffle_mode=service` to the list as per https://cloud.google.com/blog/products/gcp/introducing-cloud-dataflow-shuffle-for-up-to-5x-performance-improvement-in-data-analytic-pipelines. I can add some more if you think there is value in it? Thanks!

Issue #5789, opened 2020-02-27
Move the provider from vendoring dependencies to gomodules


chrisst:
We are currently checking in the whole vendor folder into github and also the go.mod. The only thing keeping these in sync is a manual process. 

- [x] Ensure that magic modules can generate the correct imports once the vendor directory is removed
- [x] Change the makefiles in TPG and TPGB to remove mod=vendor references
- [x] Update the kokoro build files to remove any references to vendoring
- [x] Update the HashiCorp build process to work with go modules
- [x] Remove the vendor folder
- [x] Update the documentation that refers to the vendor process

ndmckinley:
@megan07, have you seen #6306 and are you planning on addressing it here?  If not we can discuss next week.

PR #6219, opened 2020-04-27
remove vendor directory, update makefile

megan07:
Remove vendor directory, update makefile.

Addresses https://github.com/terraform-providers/terraform-provider-google/issues/5789


GNUmakefile
```diff
@@ -1,22 +1,20 @@
...
 default: build
 
 build: fmtcheck generate
 	go install
 
 test: fmtcheck generate
-	go test -i $(TEST) || exit 1
-	echo $(TEST) | \
+	go test -i ./... || exit 1
+	echo ./... | \
```
y0ssar1an, line 13: Tiny mistake here: `echo ./...`

megan07: Hi @y0ssar1an !  Thank you for the review! I'm sorry, I'm learning as I go with this, could you explain what the mistake is here?  Thanks!

megan07:
Hi @y0ssar1an! Thank you for pointing that out! I've actually tested it with the `vendor` directory removed and it seemed to work fine, so I am leaving that part in for now. If we run into issues as time goes on, I'll come back and take a look at it.  Thanks!

GNUmakefile
y0ssar1an: No worries! Let me back way up and start with how `$(TEST)` was being used before.

```makefile
TEST?=$$(go list ./... |grep -v 'vendor')
```

`go list ./...` will print the import paths of all the packages in the repo.

```
github.com/terraform-providers/terraform-provider-google
github.com/terraform-providers/terraform-provider-google/google
github.com/terraform-providers/terraform-provider-google/scripts/affectedtests
github.com/terraform-providers/terraform-provider-google/scripts/sidebar
github.com/terraform-providers/terraform-provider-google/version
```

Notice that none of these packages has `vendor` in it, so there was no need to filter out vendored packages with `grep -v`. Very old versions of the Go compiler used to print vendored packages, but `go list ./...` hasn't printed vendored packages for a long time.

```makefile
TEST?=$$(go list ./... |grep -v 'vendor')
```

This line is saying "If `$TEST` is not set, set it to the list of all Go packages." When you pass that list to `go test`, it will run all the tests in those packages. Here's what that command will be if `$TEST` is not set:

```sh
go test github.com/terraform-providers/terraform-provider-google \
    github.com/terraform-providers/terraform-provider-google/google \
    github.com/terraform-providers/terraform-provider-google/scripts/affectedtests \
    github.com/terraform-providers/terraform-provider-google/scripts/sidebar \
    github.com/terraform-providers/terraform-provider-google/version
```

A more concise way of running the same tests is `go test ./...` ("run all tests"). So running `make test` without `$TEST` set just runs all the tests. Why not just replace `$(TEST)` with `./...` everywhere? Well, the author's intent for creating `$TEST` was so you could override the list of packages passed to `go test`.

```sh
# run all tests
make test

# run all tests (long version)
TEST=./... make tests

# run a single package's tests
TEST=github.com/terraform-providers/terraform-provider-google/version make test
```

If you `echo ./...` that will print the literal string `./...`, not the list of Go packages. Apparently, you _can_ pipe that to `go test`. Weirdly, it will only run tests for `github.com/terraform-providers/terraform-provider-google` and not any of the subpackages. Since there are no tests at the top level, it will always succeed and exit 0. So tests will always pass 😬 

I think the right thing to do here is to bring back `$(TEST)` everywhere, and simply change the top line to this.

```makefile
TEST?=$$(go list ./...)
```

We've deleted the reference to `vendor` (which wasn't doing anything), and we can still run a single package's tests if we want to. Does that make sense?

y0ssar1an: Actually, now that I think about it, I'm pretty sure `make test` can be simplified to just this.

```makefile
go test $(TESTARGS) -timeout=60s $(TEST)
```

I don't think there's any reason to limit test parallelism to 4, or to use `xargs`, or `go test -i`.  I'm not sure about the author's original intent. They may have been working around problems with the very old versions of `go`. The simplified version works just fine for me.

megan07: Thank you for this!!  This was a very helpful explanation, I really appreciate it!

megan07: I've updated this, but kept the parallelism in.  I ran this by another colleague who mentioned we limit test parallelism for our ci machines, so i'll keep that much in there.  Thank you so much!

y0ssar1an: Okay, it's almost there. There's one last change to make. `go test` is being run with `-i`, which doesn't actually run the test. It tries to install test dependencies and then quits. You can see the tests being run if you pass the `-v` (verbose) flag.

```makefile
test: fmtcheck generate
	go test -v $(TESTARGS) -timeout=30s -parallel=4 $(TEST)
```

With `-i`, no tests are run.

megan07: oh yes!  sorry, i read that and it registered - but not long enough for when i made the change.  thanks!  i'll update that now.

y0ssar1an: LGTM 🚀

PR #6254, opened 2020-04-30
Update High Availability flag to mention Cloud SQL instances

Kodyc1:
This line mentions only Postgres but the flag should apply to MySQL as well: * `availability_type` - (Optional) This specifies whether a PostgreSQL instance should be set up for high availability (`REGIONAL`) or single zone (`ZONAL`).

The REST API (https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1beta4/instances#sqlavailabilitytype) doesn't say it's Postgres specific.

### DIFF
--- a/website/docs/r/sql_database_instance.html.markdown
+++ b/website/docs/r/sql_database_instance.html.markdown
@@ -240,7 +240,7 @@ The required `settings` block supports:
     for information on how to upgrade to Second Generation instances.
     A list of Google App Engine (GAE) project names that are allowed to access this instance.
 
-* `availability_type` - (Optional) This specifies whether a PostgreSQL instance
+* `availability_type` - (Optional) This specifies whether a Cloud SQL instance
     should be set up for high availability (`REGIONAL`) or single zone (`ZONAL`).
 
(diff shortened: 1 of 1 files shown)


website/docs/r/sql_database_instance.html.markdown
```diff
@@ -240,7 +240,7 @@ The required `settings` block supports:
     for information on how to upgrade to Second Generation instances.
     A list of Google App Engine (GAE) project names that are allowed to access this instance.
 
-* `availability_type` - (Optional) This specifies whether a PostgreSQL instance
+* `availability_type` - (Optional) This specifies whether a Cloud SQL instance
```
rileykarson, line 243: Seems reasonable to me! While you're here, what do you think of changing it to the following:

```
The availability type of the Cloud SQL instance, high availability (`REGIONAL`) or single zone (`ZONAL`).'
```

That displays the actual reference docs' message ("The availability type of the given Cloud SQL instance.") while still listing both options concisely.

Kodyc1: Definitely a concise improvement - changed per your suggestion!

rileykarson:
Ah, never mind about the delay.

PR #6332, opened 2020-05-09
Clarify usage

mig4:
* add note about role required to be able to read organization info
* clarify format of `organization` argument to make it more obvious what values are expected

### DIFF
--- a/website/docs/d/google_organization.html.markdown
+++ b/website/docs/d/google_organization.html.markdown
@@ -22,13 +22,15 @@ resource "google_folder" "sales" {
 }
 ```
 
+Note that the identity whose credentials are used for authentication needs to have `roles/resourcemanager.organizationViewer` role to be able to read that info.
+
 ## Argument Reference
 
 The arguments of this data source act as filters for querying the available Organizations.
 The given filters must match exactly one Organizations whose data will be exported as attributes.
 The following arguments are supported:
 
-* `organization` (Optional) - The name of the Organization in the form `{organization_id}` or `organizations/{organization_id}`.
+* `organization` (Optional) - The numeric ID of the Organization in the form `{organization_id}` or `organizations/{organization_id}`.
 * `domain` (Optional) - The domain name of the Organization.
 
 ~> **NOTE:** One of `organization` or `domain` must be specified.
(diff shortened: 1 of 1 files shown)


website/docs/d/google_organization.html.markdown
```diff
@@ -22,13 +22,15 @@ resource "google_folder" "sales" {
...
+Note that the identity whose credentials are used for authentication needs to have `roles/resourcemanager.organizationViewer` role to be able to read that info.
+
 ## Argument Reference
 
 The arguments of this data source act as filters for querying the available Organizations.
 The given filters must match exactly one Organizations whose data will be exported as attributes.
 The following arguments are supported:
 
-* `organization` (Optional) - The name of the Organization in the form `{organization_id}` or `organizations/{organization_id}`.
+* `organization` (Optional) - The numeric ID of the Organization in the form `{organization_id}` or `organizations/{organization_id}`.
```
c2thorn, line 33:
```suggestion
* `organization` (Optional) - The name (numeric ID) of the Organization in the form `{organization_id}` or `organizations/{organization_id}`.
```
`name` is the actual term for the field here, as it is defined here: https://cloud.google.com/resource-manager/reference/rest/v1/organizations/get

So I'm hesitant to remove it completely. I understand the want for clarification, do you believe this suggestion will suffice?

c2thorn:
I'm very sorry, this PR slipped through the cracks for me. I have a small suggestion, but otherwise looks fine.

If you want to, you can make the change upstream in Magic Modules
https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/website/docs/d/google_organization.html.markdown
And your change will propagate to both providers. No worries otherwise!

website/docs/d/google_organization.html.markdown
mig4: Sure that makes sense. Initially I thought it means I can use either organization name or ID, that's why I thought it'd be good to clarify. Your suggestion works for me, thanks!

mig4:
> I'm very sorry, this PR slipped through the cracks for me. I have a small suggestion, but otherwise looks fine.

That's all right of course. Thanks!

> If you want to, you can make the change upstream in Magic Modules

OK, will do.

rileykarson:
Superseded with https://github.com/GoogleCloudPlatform/magic-modules/pull/4069

PR #6223, opened 2020-04-27
Upgrade from Go 1.13.4 -> 1.14.2

y0ssar1an:
This is a straightforward compiler version bump. All the tests passed with no issues. It's nice to finally delete this boilerplate.

```
GO111MODULE=on
GOFLAGS=-mod=vendor
```

Note that because it touches `vendor/modules.txt` there will be a merge conflict with #6219. It's easy to fix. When that merges, I'll simply rebase to delete `vendor/` entirely.

PR #6359, opened 2020-05-12
Upgrade go compiler 1.13.4 -> 1.14.3

ryboe:
* update `go` to version 1.14.3
* simplify `make test` command
* remove `GO111MODULE=on` and `GOFLAGS=-mod=vendor` env vars, which are no longer necessary in Go 1.14
  * `GO111MODULE=on` is the default behavior now
  * We don't have to pass a flag to tell `go build` and `go test` to use `vendor/`. It will use `vendor/` if the directory is present

This is a resubmit of #6223.


GNUmakefile
```diff
@@ -1,19 +1,14 @@
-TEST?=$$(go list ./... |grep -v 'vendor')
+TEST?=$$(go list ./...)
```
y0ssar1an, line 1: We don't need to filter out `vendor`. `go list ./...` doesn't print packages in `vendor/` anymore.

GNUmakefile
```diff
@@ -1,19 +1,14 @@
...
 default: build
 
 build: fmtcheck generate
 	go install
 
 test: fmtcheck generate
-	go test -i $(TEST) || exit 1
-	echo $(TEST) | \
-		xargs -t -n4 go test $(TESTARGS) -timeout=30s -parallel=4
+	go test $(TESTARGS) -timeout=30s $(TEST)
```
y0ssar1an, line 11: There's no need to pipe the list of packages to `xargs`. I was told parallelism was being limited to 4 for CI reasons, but I'm skeptical of that. Unless you have some very weird CI situation, limiting parallelism just slows down your tests a little. If a higher parallelism causes test failures, then you've got race conditions in your tests and they should be fixed or `parallelism` should be set to `1`.

vendor/modules.txt
```diff
@@ -9,6 +9,7 @@ cloud.google.com/go/internal/version
 cloud.google.com/go/longrunning
 cloud.google.com/go/longrunning/autogen
 # cloud.google.com/go/bigtable v1.1.0
+## explicit
```
y0ssar1an, line 12: `go mod vendor` added these. I believe this is new behavior in Go 1.14.

go.mod
```diff
@@ -24,4 +24,4 @@ require (
 	google.golang.org/api v0.22.0
 )
 
-go 1.13
+go 1.14
```
y0ssar1an, line 27: This line doesn't prevent recent Go 1.13 compilers from building the plugin. It just says "use the latest modules behavior." Keeping this up-to-date can prevent some dependency headaches.

ndmckinley:
Similarly handing off to @megan07 - related to the vendor removal work.  :)

ndmckinley:
It is very likely we will want to block this on the removal of `vendor` - We probably want one big change to the build environment at a time!  :)

y0ssar1an:
@ndmckinley Is that going to merge soon?

ndmckinley:
Great.  I'm not 100% sure whether this is going to require a change to magic-modules - I think probably not - but investigating that will take a while, and we'll also need to copy this change to the beta provider.  Is this urgent for you?  If not, we'll probably schedule that work through our normal sprint planning process.

y0ssar1an:
Nope. This is not urgent. It's just a stepping stone to some other PRs I'd like to submit.

ndmckinley:
Great, in that case I will let this get triaged and prioritized the normal way!

ryboe:
Thanks, can you please apply this to `terraform-provider-google-beta`? I tried to cherry-pick the commit, but couldn't get it to work.

GNUmakefile
```diff
@@ -2,15 +2,13 @@ TEST?=$$(go list ./...)
 WEBSITE_REPO=github.com/hashicorp/terraform-website
 PKG_NAME=google
 
-GO111MODULE=on
```
megan07, line 5: I'm so sorry. As I'm making these changes in beta, I saw this and I wonder if it does any harm leaving this in? I'm thinking in case the current environment already has it set, it might be nice to have this here to override it specifically for this makefile. What do you think?

ryboe: Hmm, I don't think there are too many people who have `GO111MODULE=off` in their shell configs. That would break too many things. The only time I've seen in set to `off` is when you're want the old `go get` behavior, e.g.

```sh
GO111MODULE=off go get github.com/foo/bar/cmd/baz
```

I could be wrong, though.

My preference is to embrace the new default behavior, and let people disable it inline (like the snippet above), but I'll leave it up to you.

megan07: Ok!  We'll try it and if we get any feedback we can always add it back in.  That sounds find to me! Thanks!

PR #6372, opened 2020-05-13
Fix make tools

ryboe:
NOTE: This includes all the changes from #6359 (the Go compiler update). #6359 should be reviewed and merged before this.

There was a slight regression from #6368, which removed `vendor/`. The `misspell` and `golangci-lint` tools are still being built out of `vendor/`, so `make tools` fails.

I updated `make tools` so that the compiled binaries are fetched from GitHub. Then I ran `go mod tidy` to remove unused packages from `go.mod`. It removed four packages. Removing dev dependencies from `go.mod` is a nice change for two reasons:
  1. The dependency solver will never hold back a package because of an incompatibility with the linter or `misspell`.
  2. Anybody importing packages from `terraform-provider-google` will not inherit the false dependency on `golangci-lint` or `misspell`.


GNUmakefile
```diff
@@ -30,9 +28,8 @@ lint:
 
 tools:
 	@echo "==> installing required tooling..."
-	go install github.com/client9/misspell/cmd/misspell
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint
-
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.18.0
+	curl -L https://git.io/misspell | BINDIR=$(shell go env GOPATH)/bin bash
```
y0ssar1an, line 32: Fetching the compiled binaries means we don't have to build them from source ourselves, which means we don't inherit their dependencies.

tools.go
```diff
@@ -1,8 +0,0 @@
-// +build tools
-
-package main
-
-import (
-	_ "github.com/client9/misspell/cmd/misspell"
-	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
-)
```
y0ssar1an, line 8: This underscore imports are how you vendor dev tools. Since we're not vendoring anymore, these should be deleted.

GNUmakefile
megan07: Hi @ryboe! Would you be able to provide me steps to reproduce `make tools` failing?  I've cloned a clean repo and run it and it still succeeds for me. Thanks!

ryboe: My mistake! `make tools` is working fine. I'm not sure why it was failing earlier for me. I'll close this PR. Thanks for catching this 🙂

ryboe:
This PR was misguided. `make tools` works fine 👍

PR #6387, opened 2020-05-15
Add "inbound_services" to google_app_engine_standard_app_version

pmarschik:


slevenick:
Hey @pmarschik thanks for the contribution!

Unfortunately merging this change would cause it to be overwritten by our code generator, as this entire file is generated by Magic Modules.

If you would like to contribute to that repository to get this generated that would be great!

Basically it would just require adding this field to the API description of the StandardAppVersion resource, and then it should add it to the resource and generate docs.

See: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/appengine/api.yaml#L215

pmarschik:
Hey @slevenick thanks for getting back so fast!

I've ported the change in https://github.com/GoogleCloudPlatform/magic-modules/pull/3537 (still need our legal to OK my CLA)

rileykarson:
Superseded by https://github.com/GoogleCloudPlatform/magic-modules/pull/3537

PR #6388, opened 2020-05-15
Fix fetching of app_engine_standard_app_version data

pmarschik:
When omitting the `view` query parameter some fields (such as `handlers`) are not returned and lead terraform to show changes in the resource when there are none.

### DIFF
--- a/google/resource_app_engine_standard_app_version.go
+++ b/google/resource_app_engine_standard_app_version.go
@@ -548,7 +548,7 @@ func resourceAppEngineStandardAppVersionCreate(d *schema.ResourceData, meta inte
 func resourceAppEngineStandardAppVersionRead(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	url, err := replaceVars(d, config, "{{AppEngineBasePath}}apps/{{project}}/services/{{service}}/versions/{{version_id}}")
+	url, err := replaceVars(d, config, "{{AppEngineBasePath}}apps/{{project}}/services/{{service}}/versions/{{version_id}}?view=FULL")
 	if err != nil {
 		return err
 	}


megan07:
Hi @pmarschik!  Thanks for pointing this out!  Unfortunately, this file is autogenerated and needs to be updated in the magic-modules repo. I started looking into it a bit, and it seems to be a little more involved that just adding `view=FULL`, but you can take a stab at it here: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/appengine/api.yaml#L228
From what I saw updating the `self_link` to include the `view=FULL` at the end would be the first step, but I think you may also have to add a bit in `appengine/terraform.yaml` under `StandardAppVersion` like this
```
      handlers: !ruby/object:Overrides::Terraform::PropertyOverride
        default_from_api: true
```
This will write the default handlers to state.
I didn't get to fully test it or see if anymore needed to be done, because I realized I was spending a little too much time on it, but if you'd rather one of us make the updates, please feel free to open a new bug and reference this PR and someone will take a look at it. Feel free to assign it to me if you'd like and I will get to it when I get the chance.
If you're feeling really ambitious, it looks like this also affects the `FlexibleAppVersion` resource as well, and what I noticed is that `handlers` isn't an option in there (but should be, since there is a default).  If you choose to fix this yourself, it would be awesome if you could make the change in both places, but again, if that's too much, we're happy to make the changes as well.
Thanks!

pmarschik:
@megan07 Thanks for getting back! I've opened an issue https://github.com/terraform-providers/terraform-provider-google/issues/6417 and started a PR https://github.com/GoogleCloudPlatform/magic-modules/pull/3536 which solves the concrete problem I faced when using the terraform-provider-google. I'm sorry but I couldn't invest too much time in catching the other edge cases you mentioned. Feel free to either use my PR as a base or start from scratch (I still need to get my employer to sign the CLA)

PR #6398, opened 2020-05-17
The tags argument specifics network tags to attach to the instance

dmyerscough:
The `tags` argument specifies `network tags` that are to be attached to the instance.

### DIFF
--- a/website/docs/r/compute_instance.html.markdown
+++ b/website/docs/r/compute_instance.html.markdown
@@ -156,7 +156,7 @@ The following arguments are supported:
     Structure is documented below.
     **Note**: [`allow_stopping_for_update`](#allow_stopping_for_update) must be set to true or your instance must have a `desired_status` of `TERMINATED` in order to update this field.
 
-* `tags` - (Optional) A list of tags to attach to the instance.
+* `tags` - (Optional) A list of network tags to attach to the instance.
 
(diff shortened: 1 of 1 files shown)


ndmckinley:
Hmm, thanks for the proposal, but I'm not sure that that's clearer - how would you distinguish a network tag from any other kind of tag?

dmyerscough:
When using the AWS module, they use `tags` for tagging instances.

```
tags          = {
    Name        = "Application Server"
    Environment = "production"
  }
```

However, in GCP the equivalent is metadata. I thought it would've been a little more clear since GCP "tags" in this scenario refer to network tags.

ndmckinley:
Got it - I agree, the docs do use the terms "network tags" and "tags" interchangeably, and that does sound clearer to an AWS user.  Thanks!

The docs I'm referring to are https://cloud.google.com/vpc/docs/add-remove-network-tags.  :)

Would you mind proposing this change to [this file](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/website/docs/r/compute_instance.html.markdown) instead?  This code is automatically generated from that code, so a change only here would be clobbered.

dmyerscough:
Thank you @ndarilek, PR opened: https://github.com/GoogleCloudPlatform/magic-modules/pull/3557

PR #6427, opened 2020-05-20
Add support to importing a project by projectnumber

spew:
Some google services, such as Cloud Asset Inventory, use the ProjectNumber as the ID of the project instead of the project id. This change adds support to importing by Number to the Project resource.

### DIFF
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -218,7 +218,10 @@ func resourceGoogleProjectRead(d *schema.ResourceData, meta interface{}) error {
 		return nil
 	}
 
-	d.Set("project_id", pid)
+	// after importing by project number, the id will be a project number, change to project-id so the rest of the logic
+	// in this file can be consistent
+	d.SetId(fmt.Sprintf("projects/%v", p.ProjectId))
+	d.Set("project_id", p.ProjectId)
 	d.Set("number", strconv.FormatInt(p.ProjectNumber, 10))
 	d.Set("name", p.Name)
 	d.Set("labels", p.Labels)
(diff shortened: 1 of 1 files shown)


emilymye:
Do you mind adding a test to resource_google_project_test.go?  You can add a test step (or 2) to  TestAccProject_create that uses the (Import step format)[https://github.com/GoogleCloudPlatform/magic-modules/blob/4540f9c4260c86d6d7361641097a18255da00228/third_party/terraform/tests/resource_google_project_test.go#L142-L147] - ideally we'd want one with `ImportStateId : "projects/" + projectNumber` and one with `ImportStateId : "projects/" + pid` to be explicit

google/resource_google_project.go
```diff
@@ -396,18 +399,19 @@ func resourceGoogleProjectDelete(d *schema.ResourceData, meta interface{}) error
...
 	if matched {
-		return nil, fmt.Errorf("Error importing project %q, please use project_id", pid)
+		// this is actually a project number not a project id
+		d.Set("number", pid)
+	} else {
+		d.Set("project_id", pid)
 	}
-
 	// Ensure the id format includes projects/
-	d.SetId(fmt.Sprintf("projects/%s", pid))
```
emilymye, line 410: Can you add a comment here and at the top of Read() saying that project ID might be either project ID or number?

spew:
Yes -- I will do so

c2thorn:
Hey @spew , we are cleaning up some lingering PR's. Is this something you are still intending to work on? Otherwise we can probably close after a few days.

spew:
Hi, yes, I would like to get this in. I am curious if unit tests are a standard requirement for all importers?

rileykarson:
Closing PR as stale

PR #6460, opened 2020-05-26
docs(cloudsql): complete POSTGRES version list

bdronneau:


slevenick:
Thanks for the update @bdronneau 

I've used this PR to create a PR against our upstream code generator so this will be preserved: https://github.com/GoogleCloudPlatform/magic-modules/pull/3568

PR #6571, opened 2020-06-11
Add instanceId validation to redis resource.

Mike-Dunton:
I got this error today while creating a redis-instance. 

```
Error: Error creating Instance: googleapi: Error 400: Instance ID must be 1 to 40 characters and use lowercase letters, numbers, or hyphens. It must start with a lowercase letter and end with a lowercase letter or number.
com.google.apps.framework.request.StatusException: <eye3 title='INVALID_ARGUMENT'/> generic::INVALID_ARGUMENT: Instance ID must be 1 to 40 characters and use lowercase letters, numbers, or hyphens. It must start with a lowercase letter and end with a lowercase letter or number.
```

### DIFF
--- a/google/resource_redis_instance.go
+++ b/google/resource_redis_instance.go
@@ -50,10 +50,11 @@ func resourceRedisInstance() *schema.Resource {
 				Description: `Redis memory size in GiB.`,
 			},
 			"name": {
-				Type:        schema.TypeString,
-				Required:    true,
-				ForceNew:    true,
-				Description: `The ID of the instance or a fully qualified identifier for the instance.`,
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateRFC1035Name(1, 40),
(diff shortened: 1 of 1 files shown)


megan07:
Hi @Mike-Dunton! Thanks for the contribution! This looks great, however, this file is generated by an upstream repository, https://github.com/GoogleCloudPlatform/magic-modules, so the change will have to be made there. I will close this, but you can create an issue and assign it to yourself (or to me) so we can track it there.

Specifically it will be adding a `validation` under `name` in `products/redis/terraform.yaml`, similar to what you see [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/compute/terraform.yaml#L1812)

Another thing to note is that the `validateRFC1035Name` function seems to need a min of at least 2, so you might want to look into making it `validateRegexp("<similar-regex>")` I think?

Let me know if you don't have the time to do this and I can jump in and do it, or if you need help, I'm happy to answer questions! Thanks again!

PR #6715, opened 2020-06-29
feat(sql_instance): add support for point_in_time_recovery

ykyr:
This allows enabling [Point in time recovery](https://cloud.google.com/sql/docs/mysql/backup-recovery/backups#tips-pitr) feature for Cloud SQL.

### DIFF
--- a/google/resource_sql_database_instance.go
+++ b/google/resource_sql_database_instance.go
@@ -40,6 +40,7 @@ var (
 		"settings.0.backup_configuration.0.enabled",
 		"settings.0.backup_configuration.0.start_time",
 		"settings.0.backup_configuration.0.location",
+		"settings.0.backup_configuration.0.point_in_time_recovery_enabled",
 	}
 
 	ipConfigurationKeys = []string{
@@ -182,6 +183,12 @@ settings.backup_configuration.binary_log_enabled are both set to true.`,
 										Optional:     true,
 										AtLeastOneOf: backupConfigurationKeys,
 										Description:  `Location of the backup configuration.`,
+									},
+									"point_in_time_recovery_enabled": {
+										Type:         schema.TypeBool,
+										Optional:     true,
+										AtLeastOneOf: backupConfigurationKeys,
+										Description:  `True if Point-in-time recovery is enabled.`,
 									},
 								},
 							},
@@ -841,11 +848,12 @@ func expandBackupConfiguration(configured []interface{}) *sqladmin.BackupConfigu
 
 	_backupConfiguration := configured[0].(map[string]interface{})
 	return &sqladmin.BackupConfiguration{
-		BinaryLogEnabled: _backupConfiguration["binary_log_enabled"].(bool),
-		Enabled:          _backupConfiguration["enabled"].(bool),
-		StartTime:        _backupConfiguration["start_time"].(string),
-		Location:         _backupConfiguration["location"].(string),
-		ForceSendFields:  []string{"BinaryLogEnabled", "Enabled"},
+		BinaryLogEnabled:           _backupConfiguration["binary_log_enabled"].(bool),
+		Enabled:                    _backupConfiguration["enabled"].(bool),
+		StartTime:                  _backupConfiguration["start_time"].(string),
+		Location:                   _backupConfiguration["location"].(string),
+		PointInTimeRecoveryEnabled: _backupConfiguration["point_in_time_recovery_enabled"].(bool),
+		ForceSendFields:            []string{"BinaryLogEnabled", "Enabled"},
 	}
 }
 
@@ -1050,10 +1058,11 @@ func flattenSettings(settings *sqladmin.Settings) []map[string]interface{} {
 
 func flattenBackupConfiguration(backupConfiguration *sqladmin.BackupConfiguration) []map[string]interface{} {
 	data := map[string]interface{}{
-		"binary_log_enabled": backupConfiguration.BinaryLogEnabled,
-		"enabled":            backupConfiguration.Enabled,
-		"start_time":         backupConfiguration.StartTime,
-		"location":           backupConfiguration.Location,
+		"binary_log_enabled":             backupConfiguration.BinaryLogEnabled,
+		"enabled":                        backupConfiguration.Enabled,
+		"start_time":                     backupConfiguration.StartTime,
+		"location":                       backupConfiguration.Location,
+		"point_in_time_recovery_enabled": backupConfiguration.PointInTimeRecoveryEnabled,
 	}
 
 	return []map[string]interface{}{data}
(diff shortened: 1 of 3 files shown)


ykyr:
Not sure why CI fails. I hope it's okay to have PR in this repo, and it will be upstreamed to MM 🤞 🙂 
```Community contributors can contribute directly to Magic Modules, or they can contribute directly to this repo```

File: google/resource_sql_database_instance.go [hunk 1]
```diff
@@ -841,11 +848,12 @@ func expandBackupConfiguration(configured []interface{}) *sqladmin.BackupConfigu
...
-		Enabled:          _backupConfiguration["enabled"].(bool),
-		StartTime:        _backupConfiguration["start_time"].(string),
-		Location:         _backupConfiguration["location"].(string),
-		ForceSendFields:  []string{"BinaryLogEnabled", "Enabled"},
+		BinaryLogEnabled:           _backupConfiguration["binary_log_enabled"].(bool),
+		Enabled:                    _backupConfiguration["enabled"].(bool),
+		StartTime:                  _backupConfiguration["start_time"].(string),
+		Location:                   _backupConfiguration["location"].(string),
+		PointInTimeRecoveryEnabled: _backupConfiguration["point_in_time_recovery_enabled"].(bool),
+		ForceSendFields:            []string{"BinaryLogEnabled", "Enabled"},
```
danawillow, line 856: You'll also want to add PointInTimeRecoveryEnabled to this list- it allows updating the field from true->false

ykyr: Totally make sense, wasn't aware of ForceSendFields.

google/resource_sql_database_instance_test.go
```diff
@@ -629,6 +629,29 @@ func testAccCheckGoogleSqlDatabaseRootUserDoesNotExist(t *testing.T, instance st
...
+	masterID := randInt(t)
+
+	vcrTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccSqlDatabaseInstanceDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(
+					testGoogleSqlDatabaseInstance_PointInTimeRecoveryEnabled, masterID),
```
danawillow, line 644: Would you mind also adding another test step (and then another import step) to update the value from true-> false? You can reuse the same config, just make the value of point_in_time_recovery modifiable.

gplasky:
I was going to pick up #6774 until I found this; thanks for making it happen!

google/resource_sql_database_instance_test.go
ykyr: added two test steps, with attribute checks during value update and also imports.

google/resource_sql_database_instance_test.go
```diff
@@ -959,3 +991,23 @@ resource "google_sql_database_instance" "instance" {
...
+  region           = "us-central1"
+  database_version = "POSTGRES_9_6"
+
+  settings {
+    tier = "db-f1-micro"
+
+    backup_configuration {
+      enabled                = true
+      start_time             = "00:00"
+      point_in_time_recovery = %t
```
danawillow, line 1008:
```suggestion
      point_in_time_recovery_enabled = %t
```

google/resource_sql_database_instance_test.go
```diff
@@ -959,3 +991,23 @@ resource "google_sql_database_instance" "instance" {
   }
 }
 `
+
+func testGoogleSqlDatabaseInstance_PointInTimeRecoveryEnabled(masterID int, PointInTimeRecoveryEnabled bool) string {
```
danawillow, line 995:
```suggestion
func testGoogleSqlDatabaseInstance_PointInTimeRecoveryEnabled(masterID int, pointInTimeRecoveryEnabled bool) string {
```

google/resource_sql_database_instance_test.go
```diff
@@ -959,3 +991,23 @@ resource "google_sql_database_instance" "instance" {
...
+    tier = "db-f1-micro"
+
+    backup_configuration {
+      enabled                = true
+      start_time             = "00:00"
+      point_in_time_recovery = %t
+    }
+  }
+}
+`, masterID, PointInTimeRecoveryEnabled)
```
danawillow, line 1012:
```suggestion
`, masterID, pointInTimeRecoveryEnabled)
```

google/resource_sql_database_instance_test.go
```diff
@@ -629,6 +629,38 @@ func testAccCheckGoogleSqlDatabaseRootUserDoesNotExist(t *testing.T, instance st
...
+	masterID := randInt(t)
+
+	vcrTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccSqlDatabaseInstanceDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testGoogleSqlDatabaseInstance_PointInTimeRecoveryEnabled(masterID, true),
+				Check:  resource.TestCheckResourceAttr("google_sql_database_instance.instance", "point_in_time_recovery", "true"),
```
danawillow, line 644: This check isn't necessary- if the value in state didn't match the value in config, this test step would fail with a non-empty plan. The import step below also checks that we properly read the value from the API, so we aren't just blindly putting the value from the config into state.

google/resource_sql_database_instance_test.go
```diff
@@ -629,6 +629,38 @@ func testAccCheckGoogleSqlDatabaseRootUserDoesNotExist(t *testing.T, instance st
...
+			},
+			{
+				ResourceName:      "google_sql_database_instance.instance",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			{
+				Config: testGoogleSqlDatabaseInstance_PointInTimeRecoveryEnabled(masterID, false),
+				Check:  resource.TestCheckResourceAttr("google_sql_database_instance.instance", "point_in_time_recovery", "false"),
```
danawillow, line 653: likewise

google/resource_sql_database_instance_test.go
ykyr: Appreciate your patience and inputs 🙏 . I would like to clarify to prevent such mistakes in the future. 
Isn't Check function executed after Config is applied? In that case, the Check ensures that value in the state is same as we set in the config. In the second test step I'm changing the value from true -> false, and again ensuring that it was changed properly in the state. 
If this is covered by the import step - I didn't know that.

danawillow: Eh it's not a mistake, just a bit redundant is all. The check does ensure that the value in the state is the same as the config. However, the overall test framework runs a `terraform plan` after every TestStep to ensure that the plan is empty. So if the value in the config didn't match the state, that would show a diff in the plan, and the TestStep would fail automatically.

ykyr: Got it. Thank you, Dana, for the explanation and steel nerves 🙂 Will remove it.

PR #6726, opened 2020-07-01
Update google_service_account markdown import example

louisvernon:
Terraform Version: 0.12.24
Google Provider Version: 3.27.0

The example for importing an existing google service account was not succesful. Updating the example to drop `projects` and `serviceAccounts` resulted in a succesful import.


slevenick:
Can you reproduce the issue with the longer form import ID not working? It seems like it is working fine to me

louisvernon:
Thanks for checking. I am not able to reproduce now. May have been transient or user error.

PR #6739, opened 2020-07-05
Update documentation with useful role to access secret

bodegus:
I spent an hour trying to figure out why my function did not have permission.  I suggest updating the docs so that the example allows you to read the secret


ndmckinley:
Thanks!  You're right, that's better.  I'll mirror this change into our upstream repository, github.com/googlecloudplatform/magic-modules.

PR #6776, opened 2020-07-10
Update Redis instance documentation to add v 5.0

rottenbytes:
Version 5.0 is now supported


danawillow:
See https://github.com/terraform-providers/terraform-provider-google-beta/pull/2305#issuecomment-664565229.

rottenbytes:
Sorry, I missed the point when checking if it was possible. Thanks for the explanation

PR #6781, opened 2020-07-11
Update google_project_iam.html.markdown

andy-caruso:
Remove the merge language as this conflicts with the guidance above that google_project_iam_policy is authoritative


c2thorn:
Hey @andy-caruso, thanks for the suggestion. Do you mind making this change upstream at https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/website/docs/r/google_project_iam.html.markdown ?

That way it will propagate to both google providers.

rileykarson:
Closing PR as stale

PR #6817, opened 2020-07-21
Fix line breaks

antondollmaier:
This change fixes missing/invalid line breaks to have a nice list without breaking the content.


slevenick:
These descriptions are generated via a code generation tool. I believe that https://github.com/GoogleCloudPlatform/magic-modules/pull/3767 should fix this. Closing for now

Issue #6889, opened 2020-07-29
Changing a node_pool's initial_node_count forces recreation of the whole container_cluster


mltsy:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Terraform Version

Terraform 0.12.24
Google provider version 3.22.0

### Affected Resource(s)

* google_container_node_pool

### Terraform Configuration Files

This issue should occur for any node_pool with `initial_node_count` set to a particular value.  HEre's an example

```hcl
resource "google_container_cluster" "cluster" {
  location = "us-east1"

  name               = "cluster-name"
  min_master_version = "latest"

  remove_default_node_pool = true
  initial_node_count = 1
}

resource "google_container_node_pool" "node_pool" {
  location           = google_container_cluster.cluster.location
  name               = "random-node-pool-name"
  cluster            = google_container_cluster.cluster.name
  initial_node_count = 1

  autoscaling {
    min_node_count = 1
    max_node_count = 3
  }

  node_config {
    image_type = "UBUNTU"

    metadata = {
      disable-legacy-endpoints = "true"
    }
  }
}
```

### Debug Output

I can't gather this without launching a new nodepool to recreate the problem. Let me know if it's really necessary.

### Expected Behavior

When I manually change the size of a node_pool in Google's Cloud Console (for whatever reason), it changes the `initial_node_count` setting for that node_pool based on my manual update of the current node_pool size.  Running TF (assuming the config is unchanged) after this node_pool update should have no effect.  It should not require re-creation of the entire cluster.

### Actual Behavior

Running TF after this node_pool update requires re-creation of the entire cluster, based on the fact that it sees `initial_node_count` has changed.

### Steps to Reproduce

1. `terraform apply` (This may require some finagling of google permissions, etc.)
2. Change the number of nodes to 2 per zone, using the Google Cloud Console web GUI
[...]

PR #6896, opened 2020-07-30
Add a warning about initial_node_count

mltsy:
This documents a known issue outlined in #6889


ndmckinley:
Great, it looks like we're going to pick up https://github.com/terraform-providers/terraform-provider-google/issues/6901, but for now we can use this.  It does need to be upstreamed to magic modules - this is a generated file.  The change needs to be made here: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/website/docs/r/container_node_pool.html.markdown

Would you like me to upstream that, or would you prefer to do it?  :)

mltsy:
I think I did this right 😄 ... https://github.com/GoogleCloudPlatform/magic-modules/pull/3851

Added a link to the workaround for now as well.

PR #6929, opened 2020-08-04
Fix validation of metric_sum_in_range.max

tmarszal:
Fixed validation of windows_based_sli.metric_sum_in_range.range.max. It was always required because of that.

### DIFF
--- a/google/resource_monitoring_slo.go
+++ b/google/resource_monitoring_slo.go
@@ -621,7 +621,7 @@ just one of min or max. Summed value 'X' should satisfy
 													Description: `max value for the range (inclusive). If not given,
 will be set to "infinity", defining an open range
 ">= range.min"`,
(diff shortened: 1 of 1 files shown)


c2thorn:
Hi @tmarszal, nice catch! Do you mind making this change upstream in our Magic Modules repo? This way the change can distribute to both google and google-beta. You can find the necessary code here:
https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/monitoring/api.yaml#L1565

tmarszal:
Created PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/3826
Thanks for pointing me to the right direction!

tmarszal:
Should I close this PR?

c2thorn:
I got it!

PR #6958, opened 2020-08-06
updating urls after migrating to hashicorp org

megan07:


megan07:
I _think_ the travis error will be fixed once i push the mm changes...
https://github.com/GoogleCloudPlatform/magic-modules/pull/3835

PR #6993, opened 2020-08-11
Use correct self_link in example for private IP example

timwsuqld:
`google_compute_network.private_network.id` isn't a valid network name for private_network, it needs to be a `self_link`. This updates the docs to reflect that.


rileykarson:
Closing PR as stale

timwsuqld:
@rileykarson this is a bit annoying that you have closed this documentation fix as stale, when the bug is still present in the docs. This is far from stale, the docs issue is still present. Please reopen and just merge the simple 1 line fix to the docs!

PR #6994, opened 2020-08-11
Fix documents GKE cluster_telemetry block Supported values

shiv3:
expected cluster_telemetry type to be one of [DISABLED ENABLED SYSTEM_ONLY]

https://godoc.org/google.golang.org/api/container/v1beta1#ClusterTelemetry


ndmckinley:
Okay, that's merged upstream.

PR #7004, opened 2020-08-12
Remove renamed SQL Instance Datasource

upodroid:
In https://github.com/GoogleCloudPlatform/magic-modules/pull/3854 I renamed it data_source_* instead of data_* and the files need to be deleted to pass go tests.


slevenick:
Looks like this was actually handled by the PR from magic modules. Only the data_source_ version exists on master now

upodroid:
I thought it required manual cleanup, `make test` failed locally until i deleted the old files

PR #7017, opened 2020-08-13
[Documentation] Type of member is required to import google_iap_web_backend_service_iam_member

AkarshSatija:
The type of member is needed to import the resource to terraform state


slevenick:
I have opened https://github.com/GoogleCloudPlatform/magic-modules/pull/3864 which should fix this for all generated IAM resources

PR #7037, opened 2020-08-16
Specify possible values for arg only once

ialidzhikov:
For https://www.terraform.io/docs/providers/google/r/compute_subnetwork.html#aggregation_interval, the possible values are specified 2 times.


ialidzhikov:
@ndmckinley , @c2thorn should I make this change also for https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/website/docs/d/compute_subnetwork.html.markdown? I am not pretty sure how the docs are generated. 😕  Any help will be appreciated!

ialidzhikov:
Ref https://github.com/GoogleCloudPlatform/magic-modules/pull/3874

ialidzhikov:
The change in the magic-modules is enough, right?

ndmckinley:
Yep!  Thank you!

PR #7095, opened 2020-08-25
Cloud Storage: docs improvements for automatic service account

mhuxtable:
I've improved the documentation with warnings, cross-references and
additional examples around appropriate use of the
google_storage_project_service_account resource to ensure the proper
existence of the Cloud Storage automatic internal service account prior
to utilising any features which depend on the same.

There are various subtleties around the lifecycle of this service
account which may be non-obvious when using these features.  Moreover,
the service account is always implicitly created as a side-effect of
operations expected to be read-only (clicking in the UI or making a GET
call to an API are sufficient). It is possible to write Terraform code
which assumes the existence of the service account through interpolating
the project ID into the well-known email address format and the same
Terraform will work in some or many cases when the service account
already exists. Such Terraform contains latent failure conditions and
cannot apply cleanly to new projects.

We must be defensive in these cases and ensure users are aware of the
relevant data source and the importance of its use.

The documentation was already good for the creation of
google_storage_notification objects and the associated permissions
required on pub/sub topics. I've added warnings relating to the use of
the data source when specifying Cloud KMS keys on storage buckets, which
did not have any hooks on the storage_bucket resource's documentation to
these subtleties.

### DIFF
--- a/website/docs/d/storage_project_service_account.html.markdown
+++ b/website/docs/d/storage_project_service_account.html.markdown
@@ -9,15 +9,40 @@ description: |-
 
 # google\_storage\_project\_service\_account
 
-Get the email address of a project's unique Google Cloud Storage service account.
+Get the email address of a project's unique [automatic Google Cloud Storage service account](https://cloud.google.com/storage/docs/projects#service-accounts).
 
-Each Google Cloud project has a unique service account for use with Google Cloud Storage. Only this
-special service account can be used to set up `google_storage_notification` resources.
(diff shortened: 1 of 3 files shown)


mhuxtable:
Hmm, I forgot all this was generated by modular-magician so I expect I need to contribute this upstream to GoogleCloudPlatform/magic-modules. Will do that.

mhuxtable:
Okay, I've raised this on GoogleCloudPlatform/magic-modules#3903 so I'll close this one.

PR #7157, opened 2020-08-30
Use cloud.json to find out cloud-netblocks

### LINKED ISSUE #6996: Switch to cloud.json for cloud-netblocks
From https://cloud.google.com/compute/docs/faq#find_ip_range

> Important: In the past, Google Cloud instructed users to inspect the _cloud-netblocks.googleusercontent.com DNS TXT record (and the records it referenced). Please update your scripts or software libraries so that they read from the cloud.json file instead. The JSON file includes additional information, such as the region to which a regional external IP address is attached.

while [the module is still querying the DNS record](https://github.com/hashicorp/terraform-provider-google/blob/584893fec6f9662ea7536afa72a4d64e05307d14/google/data_source_google_netblock_ip_ranges.go#L60). 

Maybe it would be better to switch to the JSON file as the documentation recommends.

cagataygurturk:
Fixes #6996

### DIFF
--- a/google/data_source_google_netblock_ip_ranges.go
+++ b/google/data_source_google_netblock_ip_ranges.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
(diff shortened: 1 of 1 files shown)


slevenick (approved):
Sorry I lost track of this! Thanks for the contribution

PR #7171, opened 2020-09-01
Update storage_bucket.html.markdown

vanessabodard-voi:
retention_policy block is optional, but retention_period is required if the block is present.

```release-note:none
```


rileykarson (approved):
LGTM- sorry, seems I lost track of this!

PR #7206, opened 2020-09-07
Added option 'all' for compute_firewall protocols as per #750

### LINKED ISSUE #750: Have firewall rule apply to all protocols, if specific protocol is omitted.
### Summary
There is no current way to specify all protocols in a single allow/deny block of a google_compute_firewall resource.  This is possible using the Google Cloud firewall API, where if no protocol is specified then the rule applies to all protocols.   If the equivalent is attempted  using TF, by omitting the mandatory protocol from the allow/deny block, then an error is generated.

### Terraform Version
Terraform v0.10.8

### Affected Resource(s)
- google_compute_firewall

```hcl
resource "google_compute_firewall" "default_deny_egress" {
  name    = "default-deny-egress"
  project = "${module.vpc_host_project.project_number}"
  network = "${google_compute_network.default.name}"
  description = "Default rule denies all egress traffic"

  deny {}

  direction = "EGRESS"
  destination_ranges = ["0.0.0.0/0"]
}
```

### Debug Output
```shell
terraform apply
                                                                                                                                                                                                                 
Error: google_compute_firewall.default_deny_egress: "deny.0.protocol": required field is not set
```

### Expected Behavior
Based on TF documentation, the error is expected.

### What Should Happen
Based on Google Cloud API, omitting the protocol from an allow/deny block should cause the rule to apply to all protocols.

### Important Factoids
N/A

### References
Google Firewall documentation (Protocols and Ports)
https://cloud.google.com/vpc/docs/firewalls#protocols_and_ports

> The protocol can be specified as a well-known protocol string (tcp, udp, icmp, esp, ah, sctp) or as the IP protocol number. **If no protocols are specified, the rule applies to all protocols.**

albertogeniola:
Fixes #750

Changes proposed in this pull request:

* Added 'all' protocol option for compute_firewall resource

### DIFF
--- a/website/docs/r/compute_firewall.html.markdown
+++ b/website/docs/r/compute_firewall.html.markdown
@@ -217,7 +217,7 @@ The `allow` block supports:
   The IP protocol to which this rule applies. The protocol type is
   required when creating a firewall rule. This value can either be
   one of the following well known protocol strings (tcp, udp,
-  icmp, esp, ah, sctp, ipip), or the IP protocol number.
+  icmp, esp, ah, sctp, ipip, all), or the IP protocol number.
 
 * `ports` -
   (Optional)


megan07 (changes requested):
Hi @albertogeniola! Great catch! I had to find out what happened here, since it looked like we made the change [here](https://github.com/hashicorp/terraform-provider-google/pull/769/files#diff-c02a4470f5406e6644e7bf6214f0d93eR92) a while ago. It seems when we changed to generating code with the magician, the description wasn't updated where we generated from.
With that being said, would you be able to make this change in the https://github.com/GoogleCloudPlatform/magic-modules repo?
Specifically would you please make it at these spots for [allow](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/compute/api.yaml#L3024) and [deny](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/compute/api.yaml#L3060)?
If you want you can assign it to me and I can merge it ASAP.
Thanks again for your contribution, we appreciate it!

albertogeniola:
Hi @megan07 ,

[done](https://github.com/GoogleCloudPlatform/magic-modules/pull/3962).

PR #7227, opened 2020-09-09
`point_in_time_recovery_enabled` for PostgreSQL

gabidavila:
This is only valid for PostgreSQL instances as shown by the error below:

```
google_sql_database_instance.primary: Creating...

Error: Error, failed to create instance store-8fc15b58: googleapi: Error 400: Invalid request: Point-in-time recovery can only be enabled for Postgres instances
```


ndmckinley:
Thanks for the help!  Would you mind making this change here: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/website/docs/r/sql_database_instance.html.markdown - this file is generated from that one in order to keep the GA and Beta provider in sync.  :)

PR #7282, opened 2020-09-16
Refer to the value of `labels` as a 'map'

jak-sdk:
Labels is described as a 'set of key/value label pairs'. This is correct from an English grammar perspective, however in our case a Terraform 'set' is a strictly defined collection type, described here https://www.terraform.io/docs/configuration/types.html#set-

To avoid misleading users, can we refer to it as a 'map' instead, as this is the actual Terraform collection type we want?

Thanks,
Jak


slevenick (approved):
Thanks! I forgot about this, sorry for the delay

PR #7291, opened 2020-09-17
upgrade to sdkv2

megan07:
Most of these files were mostly generated by https://github.com/GoogleCloudPlatform/magic-modules/pull/3971

`go.mod` and `go.sum` were updated here.


rileykarson (approved):
I think I'm around as confident as I can be, for a change of this size!

note: I reviewed the mm pr + the tpgb one in depth, this one was just a quick spot check.

PR #7308, opened 2020-09-21
[google_container_node_pool] Fix autoscaling.max_node_count validation rule

michael-bouvy:
The current validation rule for `autoscaling.max_node_count` checks that the value is >= 1, although the description of this field states that the value must be >= `autoscaling.min_node_count`:

https://github.com/hashicorp/terraform-provider-google/blob/23a4a7398a8d153781911d0b4f0e125814448c2f/google/resource_container_node_pool.go#L86-L87

I'm therefore not able to scale down to 0 nodes my pool, using 0 for both min and max.

As validation rules are not context-aware, I've added one further in the code.

### DIFF
--- a/google/resource_container_node_pool.go
+++ b/google/resource_container_node_pool.go
@@ -83,7 +83,7 @@ var schemaNodePool = map[string]*schema.Schema{
 				"max_node_count": {
 					Type:         schema.TypeInt,
 					Required:     true,
-					ValidateFunc: validation.IntAtLeast(1),
+					ValidateFunc: validation.IntAtLeast(0),
 					Description:  `Maximum number of nodes in the NodePool. Must be >= min_node_count.`,
 				},
 			},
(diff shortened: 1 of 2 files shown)


ndmckinley:
Thanks for the contribution!  It seems you're right, so we'd like to accept this.  Would you mind adding a test for this case, updating a node pool to 0 nodes?  You can see the `TestAccContainerNodePool_update` series of tests for examples of how to do this.

michael-bouvy:
Sure, will do this.

michael-bouvy:
Done @ndmckinley

michael-bouvy:
Actually, according to https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters.nodePools#NodePool.NodePoolAutoscaling, it is not possible to have `minNodeCount = 0`. So I guess this is more an issue in the description / documentation rather than in the code of the provider, what do you think @ndmckinley?

michael-bouvy:
Note that I successfully set `autoscaling.min_node_count = 0` using Terraform and this provider. Odd to see such different behaviour.

ndmckinley:
I think that the confusion is an issue with serialization - when you set a value to "0" (or any 'empty' value, like "", false, [], {}, etc), the value doesn't get sent.  Since Terraform doesn't send the value, the server doesn't provide an error, so terraform thinks it was set correctly.  So I think that what happened when you were "able" to set the size to 0 was that Terraform mistakenly believed it had succeeded.

Since it seems the value can't be zero, it seems like the validation rule is correct, and I'd recommend you close this PR.  Does that seem right to you?

michael-bouvy:
In this case, the documentation shall be updated to state that `autoscaling.max_node_count` should be `> 0` AND `>=` `autoscaling.max_node_count`, what do you think?

ndmckinley:
Hm, I think we'd be fine to have that documentation, but it seems like the sort of validation that's best left to the server side.  it seems clear to most readers that a minimum node count would need to be smaller than a maximum node count.  If you would like, feel free to open a PR in Magic Modules, which generates those files - github.com/GoogleCloudPlatform/magic-modules - against the docs file, which is under third_party/website/docs/r, I believe.  :)

I'm going to close this one, though, since it doesn't seem like we should merge it.

PR #7331, opened 2020-09-22
temp change to show to riley

ScottSuarez:


google/resource_compute_instance.go
```diff
@@ -1291,108 +1286,57 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
...
-			// the only way to do it.  In future this should be revised to only change what is
-			// necessary, and also add before removing.
-
-			// Delete any accessConfig that currently exists in instNetworkInterface
-			for _, ac := range instNetworkInterface.AccessConfigs {
-				op, err := config.clientCompute.Instances.DeleteAccessConfig(
-					project, zone, instance.Name, ac.Name, networkName).Do()
+		if d.HasChange(prefix + ".subnetwork") {
+			if !d.HasChange(prefix + ".network") {
+				subnetProjectField := d.Get(prefix + ".subnetwork_project").(string)
```
rileykarson, line 1317: I think we just want to send in the `prefix + .subnetwork_project` string? This sends the value in where I thought the path to the field was expected.

ScottSuarez: nice catch! that's correct

PR #7335, opened 2020-09-23
Add exclusions option for logging sink resources

pmoncadaisla:
Added as defined in https://cloud.google.com/logging/docs/reference/v2/rest/v2/projects.sinks#LogExclusion

Log entries that match any of the exclusion filters will not be exported.
If a log entry is matched by both filters and one of exclusion_filters it will not be exported.

Affects to the following resources:

- google_logging_billing_sink
- google_logging_project_sink
- google_logging_organization_sink

### DIFF
--- a/google/resource_logging_project_sink_test.go
+++ b/google/resource_logging_project_sink_test.go
@@ -117,6 +117,29 @@ func TestAccLoggingProjectSink_heredoc(t *testing.T) {
 	})
 }
 
+func TestAccLoggingProjectSink_loggingbucket(t *testing.T) {
+	t.Parallel()
+
+	sinkName := "tf-test-sink-" + randString(t, 10)
+	logBucketID := "tf-test-logbucket-" + randString(t, 10)
+
+	vcrTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLoggingProjectSinkDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccLoggingProjectSink_loggingbucket(sinkName, getTestProjectFromEnv(), logBucketID),
+			},
+			{
+				ResourceName:      "google_logging_project_sink.loggingbucket",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
 func testAccCheckLoggingProjectSinkDestroyProducer(t *testing.T) func(s *terraform.State) error {
 	return func(s *terraform.State) error {
 		config := googleProviderConfig(t)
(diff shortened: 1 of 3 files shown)


google/resource_logging_sink.go
```diff
@@ -30,6 +31,38 @@ func resourceLoggingSinkSchema() map[string]*schema.Schema {
...
+						Optional:    true,
+						Description: `A description of this exclusion.`,
+					},
+					"filter": {
+						Type:        schema.TypeString,
+						Required:    true,
+						Description: `An advanced logs filter that matches the log entries to be excluded. By using the sample function, you can exclude less than 100% of the matching log entries`,
+					},
+					"disabled": {
+						Type:        schema.TypeString,
```
slevenick, line 57: Why is this a string but it seems to be a boolean in the API client? We could just use a TypeBool which should remove some of the conversions

slevenick (commented):
Thanks for adding this! I'm running the tests, but is there a reason to have `disabled` be a string rather than a bool?

google/resource_logging_sink.go
pmoncadaisla: Absolutely, there was no reason why `disabled` had to be `TypeString`. I have changed it to `TypeBool`. 
Thanks for the catch!

pmoncadaisla:
Having `disabled` as `TypeString` didn't make sense. I have changed it to `TypeBool`.
Thanks for the catch @slevenick !

slevenick (changes requested):

google/resource_logging_project_sink_test.go
```diff
@@ -248,3 +271,32 @@ resource "google_bigquery_dataset" "logging_sink" {
...
+	exclusions {
+		name = "ex2"
+		description = "test-2"
+		filter = "resource.type = k8s_container"
+	}
+
+  unique_writer_identity = false
+}
+
+resource "google_logging_project_bucket_config" "logbucket" {
```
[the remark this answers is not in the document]
pmoncadaisla: Sorry for this! I was running tests incorrectly locally and I thought they ran in CI. I have fixed this issue and a couple issues more.

```
stAccLoggingProjectSink_loggingbucket'
go generate  ./...
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google -v -run=TestAccLoggingProjectSink_loggingbucket -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccLoggingProjectSink_loggingbucket
=== PAUSE TestAccLoggingProjectSink_loggingbucket
=== CONT  TestAccLoggingProjectSink_loggingbucket
--- PASS: TestAccLoggingProjectSink_loggingbucket (9.01s)
PASS
ok  	github.com/hashicorp/terraform-provider-google/google	9.955s
```

pmoncadaisla:
I have fixed an issue with tests, where unnecessary `google_logging_project_bucket_config` was created wrongly.
Also fixed some old bool conversions and removed `Computed` field from `exclusions` as it does not need to be computed.

slevenick (commented):
This looks great! Can you add documentation for these fields to the relevant resources?

For example here: https://github.com/hashicorp/terraform-provider-google/blob/master/website/docs/r/logging_project_sink.html.markdown

pmoncadaisla:
Awesome, thanks for the catch again.

I have added documentation with an example.

slevenick (approved):
Thanks for the contribution!

PR #7338, opened 2020-09-23
Add NotificationConfig to google_container_cluster

### LINKED ISSUE #7337: Add cluster upgrade notifications (gke)
### Community Note

* google_container_cluster

```tf
resource "google_container_cluster" "primary" {
  name     = "my-gke-cluster"
  location = "us-central1"

  # We can't create a cluster with no node pool defined, but we want to only use
  # separately managed node pools. So we create the smallest possible default
  # node pool and immediately delete it.
  initial_node_count       = 1

  notification_config {
    enabled = true
    topic = projects/project/topics/topic-name
  }

  master_auth {
    username = ""
    password = ""

    client_certificate_config {
      issue_client_certificate = false
    }
  }
}
```

### References
* https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-upgrade-notifications?hl=el
* https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.NotificationConfig

freddd:
This PR aims to introduce a new attribute called `notification_config` on `google_container_cluster` which enables a feature to send pubsub notifications on cluster upgrades. It's currently in beta. As this is my first contributing to a provider, I'm a bit unsure how to test it past writing tests. 

References:
* https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-upgrade-notifications?hl=el
* https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.NotificationConfig

Closes #7337

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -167,6 +167,38 @@ func resourceContainerCluster() *schema.Resource {
 				Description: `The list of zones in which the cluster's nodes are located. Nodes must be in the region of their regional cluster or in the same region as their cluster's zone for zonal clusters. If this is specified for a zonal cluster, omit the cluster's zone.`,
 			},
 
+			"notification_config": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Computed:    true,
+				Description: `The notification config for sending cluster upgrade notifications`,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"pubsub": {
+							Type:        schema.TypeList,
+							Optional:    true,
+							Computed:    true,
+							MaxItems:    1,
+							Description: `Notification config for pubsub`,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:        schema.TypeBool,
+										Required:    true,
+										Description: `Whether or not the notification config is enabled`,
+									},
+									"topic": {
+										Type:        schema.TypeString,
+										Required:    true,
+										Description: `The pubsub topic to send the notification to, must be in the format: projects/{project}/topics/{topic}`,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+
 			"addons_config": {
 				Type:        schema.TypeList,
 				Optional:    true,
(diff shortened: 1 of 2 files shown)


google/resource_container_cluster.go
```diff
@@ -167,6 +167,38 @@ func resourceContainerCluster() *schema.Resource {
 				Description: `The list of zones in which the cluster's nodes are located. Nodes must be in the region of their regional cluster or in the same region as their cluster's zone for zonal clusters. If this is specified for a zonal cluster, omit the cluster's zone.`,
 			},
 
+			"notification_config": {
```
megan07, line 170: Because this is technically an object and not a list, we will want to add `MaxItems: 1` to this block.

google/resource_container_cluster.go
```diff
@@ -167,6 +167,38 @@ func resourceContainerCluster() *schema.Resource {
...
+				Type:        schema.TypeList,
+				Optional:    true,
+				Computed:    true,
+				Description: `The notification config for sending cluster upgrade notifications`,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"pubsub": {
```
megan07, line 177: This can be `Required: true` since it's the only attribute on the `notification_config` block.

google/resource_container_cluster.go
```diff
@@ -167,6 +167,38 @@ func resourceContainerCluster() *schema.Resource {
...
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:        schema.TypeBool,
+										Required:    true,
+										Description: `Whether or not the notification config is enabled`,
+									},
+									"topic": {
+										Type:        schema.TypeString,
+										Required:    true,
```
megan07, line 192: If `enabled` is `false`, is `topic` still required? Do we maybe want to make this `Optional: true`?

megan07 (changes requested):
Hi @freddd!  Thank you so much for your contribution! This looks great!
I've added a few comments in the code, but have a few more general thoughts.
First, I see this is beta-only. Our `google-beta` provider is actually in a different repository, however, both repositories are generated from an upstream repository called [magic-modules](https://github.com/GoogleCloudPlatform/magic-modules). So I'm going to direct you there, you can make your changes in `third_party/terraform/resources/resource_container_cluster.go.erb`.  You'll see in that file that there are some blocks surrounded by 
```
<% unless version == 'ga' -%>
...
<% end -%>
```
The code in those blocks is only generated for `google-beta`, so you can put your code in those blocks.
I think the last things missing from this are setting it in state and allowing it to be updated.
To set it in the state file, you'll want to add a `d.Set("notification_config", flattenNotificationConfig(cluster.NotificationConfig));` in the Read function, where the flatten function will take the object returned from the API and translate it back into the Terraform Schema.
You'll then also want to check for/make updates in the Update function.
Please feel free to add me as the reviewer there, and don't hesitate to ask any questions along the way! I can run the tests for you once the new PR is created in magic-modules.
Thanks again!

freddd:
Thanks for the feedback @megan07 - really appreciate it. I will open a PR in Magic Modules shortly. Closing this meanwhile.

PR #7349, opened 2020-09-24
Update google appengine iap documentation

Tony-Proum:
This part of the documentation were missing and we were in trouble to find out what were going wrong without reading the code here : https://github.com/GoogleCloudPlatform/magic-modules/blob/af45a3148d9ccfc9d5f43f7489bda3f60de34ce3/third_party/terraform/resources/resource_app_engine_application.go#L120
the flag documentation used is the same as the google API documentation source : 
https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1beta/apps#identityawareproxy


rileykarson:
Closing PR as stale

PR #7380, opened 2020-09-29
Remove redundant issue template file.

ScottSuarez:
File is no longer used. We have the Issue_Template folder in its place.


danawillow:
Actually, this is intentional. There's a sentence on the issue chooser page that says "Don’t see your issue here? Open a blank issue." that links to a blank issue. We want to make sure that if people end up there, they still get a template.

ScottSuarez:
> Actually, this is intentional. There's a sentence on the issue chooser page that says "Don’t see your issue here? Open a blank issue." that links to a blank issue. We want to make sure that if people end up there, they still get a template.

Ah I see! Thanks for the info :)

PR #7391, opened 2020-09-30
Include devstorage.read_only scope in example

Eranikid:
OAuth scope `https://www.googleapis.com/auth/devstorage.read_only` is required for nodes to be able to pull images from GCR

Unfortunately, it is not included by default when `oauth_scopes` aren't specified

I was struggling with "Why my pod goes to ImagePullBackoff", before actually reading through the docs.

I think it would be much more obvious if example in docs incuded that scope

### DIFF
--- a/website/docs/r/container_cluster.html.markdown
+++ b/website/docs/r/container_cluster.html.markdown
@@ -61,6 +61,7 @@ resource "google_container_node_pool" "primary_preemptible_nodes" {
     }
 
     oauth_scopes = [
+      "https://www.googleapis.com/auth/devstorage.read_only",
       "https://www.googleapis.com/auth/logging.write",
       "https://www.googleapis.com/auth/monitoring",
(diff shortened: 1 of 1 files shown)


tshak:
We should discourage use of oauth_scopes as they are considered a [legacy feature](https://cloud.google.com/kubernetes-engine/docs/how-to/access-scopes) of GKE. The [recommended approach](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#best_practices) is to only include the scope `https://www.googleapis.com/auth/cloud-platform` and use a service account to manage access to resources. In the case of pulling images, the service account must be a member of the `roles/storage.objectViewer` role in the project that the container registry resides.

c2thorn:
As @tshak pointed out, we want to avoid recommending more `oauth_scopes`. Closing this PR.

PR #7439, opened 2020-10-06
Clarification on the use of the 'roles/owner' role

jamesmealing:
Removed the note that references 'roles/owner' as this role is not possible to grant with this resource and is, therefore, misleading to have as an example. Added an additional note in the initial introduction to advise around this.

"A user cannot be granted the owner role using setIamPolicy(). The user must be granted the owner role using the Cloud Platform Console and must explicitly accept the invitation."

https://cloud.google.com/resource-manager/reference/rest/v1/projects/setIamPolicy

### DIFF
--- a/website/docs/r/google_project_iam.html.markdown
+++ b/website/docs/r/google_project_iam.html.markdown
@@ -21,6 +21,8 @@ Four different resources help you manage your IAM policy for a project. Each of 
 
 ~> **Note:** `google_project_iam_binding` resources **can be** used in conjunction with `google_project_iam_member` resources **only if** they do not grant privilege to the same role.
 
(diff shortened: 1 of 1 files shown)


megan07 (approved):

jamesmealing:
Hey @megan07 - Looks like the Travis CI build failed - Not sure if this is something I did wrong or an issue with the build itself - Are you able to take a look? 🙂

megan07:
@jamesmealing thanks for taking a notice at that! it looks like it's something from a previous commit that was fixed now in a later commit, so we should be good to go.  thanks for bringing it to my attention so i could have a look!

jamesmealing:
No problem, however, it doesn't look like the merge has completed due to the build failure as the documentation page is still out-of-date (unless I am just being impatient! 🙈 ) Is there a way to kick the build off again to see if it completes this time round?

megan07:
Oh, sorry, the change is on [master](https://github.com/hashicorp/terraform-provider-google/blame/master/website/docs/r/google_project_iam.html.markdown#L24) and should go out in the next release! After that you should be able to see that change on the website. Does that answer your question?

jamesmealing:
Aha, okay, sorry, just me being impatient then! 🙈  Thanks again for your help and for getting this merged. 🙂

megan07:
No worries!  Thanks for your help!

PR #7441, opened 2020-10-06
GKE documentation recommends default oauth scope

tshak:
The `oauth_scopes` section of `google_container_cluster` has generated a lot of confusion since GCP [no longer uses access scopes](https://cloud.google.com/kubernetes-engine/docs/how-to/access-scopes). The [best practice](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#best_practices) is to use the `https://www.googleapis.com/auth/cloud-platform` scope and constrain permissions at the service account level. As currently documented, the examples guide the developer down the path of using legacy access scopes. This PR updates the documentation with the recommended configuration. 

Related: https://github.com/hashicorp/terraform-provider-google/issues/1962, https://github.com/hashicorp/terraform-provider-google/issues/1817, https://github.com/hashicorp/terraform-provider-google/pull/7391

### DIFF
--- a/website/docs/r/container_cluster.html.markdown
+++ b/website/docs/r/container_cluster.html.markdown
@@ -61,8 +61,7 @@ resource "google_container_node_pool" "primary_preemptible_nodes" {
     }
 
     oauth_scopes = [
-      "https://www.googleapis.com/auth/logging.write",
-      "https://www.googleapis.com/auth/monitoring",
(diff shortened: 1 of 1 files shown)


slevenick (approved):
Thanks for the contribution! I've also added similar documentation to the `node_config.oauth_scopes` field in the upstream PR.

Issue #6222, opened 2020-04-27
Disabled serviceusage.googleapis.com API leads to tainted project


morgante:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Terraform Version
```
Terraform v0.12.24
+ provider.google v3.19.0
+ provider.google-beta v3.19.0
+ provider.null v2.1.2
+ provider.random v2.2.1
```

### Affected Resource(s)

- `google_project`

### Terraform Configuration Files
From [Project Factory](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/core_project_factory/main.tf#L96):
```tf
resource "google_project" "main" {
  name                = var.name
  project_id          = local.temp_project_id
  org_id              = local.project_org_id
  folder_id           = local.project_folder_id
  billing_account     = var.billing_account
  auto_create_network = false

  labels = var.labels

  depends_on = [null_resource.preconditions]
}
```

### Debug Output
https://gist.github.com/morgante/cbb394bcadd6e81d90acd2dc2dab86d5

### Expected Behavior
When creating a project with `auto_create_network = false`, Terraform needs to activate the compute API on the project. This requires the **seed project** have the serviceusage API active.

I would expect this failure to be recoverable so Terraform bails out early (ie. before creating the project) if the correct APIs are not active similar to #5671.

### Actual Behavior
Terraform creates the project, then attempts to delete the network - leading to a tainted project it attempts to recreate (which is impossible because projects are unique).

### Steps to Reproduce
1. Disable the Service Usage APi on your credential project
1. `terraform apply` and observe failure
1. `terraform plan` and observe tainted project

### References
* #5671

morgante:
[...]

PR #7447, opened 2020-10-07
Adds serviceusage api in precheck function

thiagonache:
This PR address the issue https://github.com/hashicorp/terraform-provider-google/issues/6222.

### DIFF
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -213,6 +213,8 @@ func resourceGoogleProjectCheckPreRequisites(config *Config, d *schema.ResourceD
 	if !d.Get("auto_create_network").(bool) {
 		_, err := config.NewServiceUsageClient(userAgent).Services.Get("projects/00000000000/services/serviceusage.googleapis.com").Do()
 		switch {
+		// We are querying a dummy project since the call is already coming from the quota project.
+		// If the API is enabled we get a not found message or accessNotConfigured if API is not enabled.
 		case err.Error() == "googleapi: Error 403: Project '00000000000' not found or permission denied., forbidden":
 			return nil
 		case strings.Contains(err.Error(), "accessNotConfigured"):
--- a/google/resource_google_project.go
+++ b/google/resource_google_project.go
@@ -210,6 +210,15 @@ func resourceGoogleProjectCheckPreRequisites(config *Config, d *schema.ResourceD
 	if !stringInSlice(resp.Permissions, perm) {
 		return fmt.Errorf("missing permission on %q: %v", ba, perm)
 	}
+	if !d.Get("auto_create_network").(bool) {
+		_, err := config.NewServiceUsageClient(userAgent).Services.Get("projects/00000000000/services/serviceusage.googleapis.com").Do()
+		switch {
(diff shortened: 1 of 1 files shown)


thiagonache:
Hello @rileykarson and @slevenick .
It is a simple check that we need to perform in the provider in order to disable local execs we have in the project-factory module.
It is a simple change. Please, let me know any considerations.

google/resource_google_project.go
```diff
@@ -210,6 +210,17 @@ func resourceGoogleProjectCheckPreRequisites(config *Config, d *schema.ResourceD
 	if !stringInSlice(resp.Permissions, perm) {
 		return fmt.Errorf("missing permission on %q: %v", ba, perm)
 	}
+	errExpected := "googleapi: Error 403: Project '00000000000' not found or permission denied., forbidden"
+	errAPINotEnabled := "accessNotConfigured"
```
rileykarson, line 214: nit: inlining this seems fine

google/resource_google_project.go
```diff
@@ -210,6 +210,17 @@ func resourceGoogleProjectCheckPreRequisites(config *Config, d *schema.ResourceD
 	if !stringInSlice(resp.Permissions, perm) {
 		return fmt.Errorf("missing permission on %q: %v", ba, perm)
 	}
+	errExpected := "googleapi: Error 403: Project '00000000000' not found or permission denied., forbidden"
```
rileykarson, line 213: Does using a dummy project work here? I would expect us to need the quota project- the seed project in most cases- instead.

thiagonache: Actually, this is just to make the string easier to match. The idea was discussed in issue #6222 and the suggestion was that we should use 0000000 because we always get not found, meaning the API is enabled. The API I'm using does not matter actually. I just want to force an not found.

google/resource_google_project.go
thiagonache: Okay. I thought it would make it easier to read, but I was wrong. Going to revert the commit.

google/resource_google_project.go
thiagonache: just friendly FYI, I've compiled and tested. It does work as expected.

rileykarson: Ah, right, that makes sense because we're sending a request that's based on the quota project anyways. Can you comment inline with that context?

thiagonache: ```We are querying a dummy project since the call is already coming from the quota project. So, we always need to match the two types of message.``` how does it look like?

rileykarson: sgtm!

thiagonache:
Hi guys,

Please, let me know if you want any additional changes.

rileykarson (approved):
LGTM!

I've got to upstream this to [Magic Modules](https://github.com/GoogleCloudPlatform/magic-modules) so it applies to both providers. I'll do that, then merge here.

umairidris:
@thiagonache @rileykarson do the http error codes differ? Perhaps checking that would be a more reliable method than checking error strings, which could change without notice.

thiagonache:
> @thiagonache @rileykarson do the http error codes differ between the two switch cases? Perhaps checking that would be a more reliable method than checking error strings, which could change without notice.

no. it's both 403

PR #7490, opened 2020-10-10
Fixing the gcloud service list command example

emmekappa:
From the man page:

```
    This command lists the services that are enabled or available to be enabled
    by a project. You can choose the mode in which the command will list
    services by using exactly one of the --enabled or --available flags.
    --enabled is the default.
```

So running `gcloud service list` without any args will result in list of services already enabled

```release-note:none
```


rileykarson (approved):
LGTM!

PR #7509, opened 2020-10-13
google_active_folder data lookup: make parent optional

ngodec:
Issue: when using a nested folder structure, it is proving challenging to use the existing data resource due to the need to know the parent ID.
Google folder lookup API finds folders perfectly fine without providing the parent:

```
"query": "lifecycleState=ACTIVE AND displayName=\"Engineering\""

{
  "folders": [
    {
      "name": "folders/<Redacted>",
      "parent": "organizations/<Redacted>",
      "displayName": "Engineering",
      "lifecycleState": "ACTIVE",
      "createTime": "2019-06-14T09:20:35.233Z"
    }
  ]
}

"query": "lifecycleState=ACTIVE AND displayName=\"Global Dev\""

Results returned just fine:
{
  "folders": [
    {
      "name": "folders/<Redacted>",
      "parent": "folders/<Redacted>",
      "displayName": "Global Dev",
      "lifecycleState": "ACTIVE",
      "createTime": "2020-06-16T11:41:45.998Z"
    }
  ]
}
```


google/data_source_google_active_folder.go
```diff
@@ -14,7 +14,8 @@ func dataSourceGoogleActiveFolder() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"parent": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+				Default:  "",
```
rileykarson, line 18:
```suggestion
```

This default is implicit, we can omit it.

google/data_source_google_active_folder.go
```diff
@@ -14,7 +14,8 @@ func dataSourceGoogleActiveFolder() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"parent": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
```
rileykarson, line 17: Can you update https://github.com/hashicorp/terraform-provider-google/blob/master/website/docs/d/active_folder.html.markdown as well?

google/data_source_google_active_folder.go
```diff
@@ -35,10 +36,15 @@ func dataSourceGoogleActiveFolderRead(d *schema.ResourceData, meta interface{})
 		return err
 	}
 
-	parent := d.Get("parent").(string)
 	displayName := d.Get("display_name").(string)
 
-	queryString := fmt.Sprintf("lifecycleState=ACTIVE AND parent=%s AND displayName=\"%s\"", parent, displayName)
+	// parent is optional
+	if parent, ok := d.GetOk("parent"); ok {
+		queryString := fmt.Sprintf("lifecycleState=ACTIVE AND parent=%s AND displayName=\"%s\"", parent.(string), displayName)
```
rileykarson, line 43: You'll need to initialize the variable outside of this if block, and then set a value inside. As-is it isn't in scope when it gets used later.

rileykarson (commented):
Generally looks good- we'll just want to update the website as well.

google/data_source_google_active_folder.go
ngodec: Thank you! This is very helpful, I was wondering whether it was the case.

google/data_source_google_active_folder.go
ngodec: Done~

google/data_source_google_active_folder.go
rileykarson: Ah- now, you'll want to change `:=` to `=` on lines 43 and 45. As-is, you're shadowing the `queryString` variable above instead of assigning a value to it.

ngodec: Hey, thank you - fixed that now. This should be good to go I think?

google/data_source_google_active_folder_test.go
```diff
@@ -111,3 +131,16 @@ data "google_active_folder" "my_folder" {
 }
 `, parent, displayName)
 }
+
+func testAccDataSourceGoogleActiveFolderConfig_no_parent(parent string, displayName string) string {
```
rileykarson, line 135: One last small thing from the linter! This is unused, so we can delete it.

ngodec:
Hey @rileykarson apologies, I just came to fixing this. Tests pass now

rileykarson (approved):
I'll need to make a pull request against https://github.com/GoogleCloudPlatform/magic-modules cloning this PR and merge this at the same time as the one there.

rileykarson:
Closing PR as stale

PR #7527, opened 2020-10-14
Update CHANGELOG.md

megan07:


CHANGELOG.md
```diff
@@ -1,4 +1,32 @@
 ## 3.44.0 (Unreleased)
+NOTES:
```
rileykarson, line 2: We should mark this breaking change even though we're not bumping versions. https://github.com/hashicorp/terraform-provider-google/blob/master/CHANGELOG.md#3220-may-18-2020 for precedent

rileykarson (approved):

PR #7541, opened 2020-10-15
Adds new resource to manage default service accounts

thiagonache:
Hi guys,

Since this is my first terraform resource I'm asking for an early review to avoid re-work. This is the skeleton of the new resource that has been discussed on the issue https://github.com/hashicorp/terraform-provider-google/issues/4135.
It is a new resource in order to remove the gcloud dependency and also to avoid the huge extra cost when a project is tainted and needs to be recreated.

Please, let me know your thoughts.

```release-note:new-resource
`google_project_default_service_accounts`
```


thiagonache:
```
resource "google_project_default_service_accounts" "name" {
  project_id = "my-seed-project"
  action = "delete"
  restore_policy = "REACTIVATE"
}
```
Resource usage example

google/resource_google_project_default_service_accounts_test.go
```diff
@@ -0,0 +1,4 @@
+package google
+
+// IDK if I should do TDD for terraform providers development
+// Do you guys do?
```
danawillow, line 4: It's up to you whether you want to write the tests first or write the tests after you've written the code, but we'll definitely need to see working acceptance tests in order to merge the resource.

danawillow (changes requested):
I know this is still WIP, but can you also add an outline for the documentation of the resource on the website so that we don't forget to do it later?

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,151 @@
...
+// to manage all the default serviceAccounts.
+// It does mean that terraform tried to perform the action in the SA at some point but does not ensure that
+// all defaults serviceAccounts where managed. Eg.: API was activated after project creation.
+func resourceGoogleProjectDefaultServiceAccounts() *schema.Resource {
+	return &schema.Resource{
+		SchemaVersion: 0,
+
+		Create: resourceGoogleProjectDefaultServiceAccountsCreate,
+		Read:   resourceGoogleProjectDefaultServiceAccountsReadAndUpdate,
+		Update: resourceGoogleProjectDefaultServiceAccountsReadAndUpdate,
```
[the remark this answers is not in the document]
thiagonache: The Update is only for one value that's not stored on the API. Meaning, it's just a state update. Maybe when I create all logics I'll have to split them, but so far I know it may sound odd, but it's behaving correctly.
```
carvalho-c02ccan9md6m:pdsa thiagocarvalho$ terraform destroy
google_project_default_service_accounts.name: Refreshing state... [id=projects/arc-seed-project]

  - destroy

  # google_project_default_service_accounts.name will be destroyed
  - resource "google_project_default_service_accounts" "name" {
      - action         = "delete" -> null
      - id             = "projects/arc-seed-project" -> null
      - project_id     = "arc-seed-project" -> null
      - restore_policy = "REACTIVATE" -> null
    }

Plan: 0 to add, 0 to change, 1 to destroy.

Do you really want to destroy all resources?
  Terraform will destroy all your managed infrastructure, as shown above.
  There is no undo. Only 'yes' will be accepted to confirm.

  Enter a value: yes

google_project_default_service_accounts.name: Destroying... [id=projects/arc-seed-project]
google_project_default_service_accounts.name: Destruction complete after 0s

Destroy complete! Resources: 1 destroyed.
carvalho-c02ccan9md6m:pdsa thiagocarvalho$ terraform apply

  + create

  # google_project_default_service_accounts.name will be created
  + resource "google_project_default_service_accounts" "name" {
      + action         = "delete"
      + id             = (known after apply)
      + project_id     = "arc-seed-project"
      + restore_policy = "REACTIVATE"
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

google_project_default_service_accounts.name: Creating...
google_project_default_service_accounts.name: Creation complete after 2s [id=projects/arc-seed-project]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
carvalho-c02ccan9md6m:pdsa thiagocarvalho$ vim terraform.tf
carvalho-c02ccan9md6m:pdsa thiagocarvalho$ terraform apply
google_project_default_service_accounts.name: Refreshing state... [id=projects/arc-seed-project]

  ~ update in-place

  # google_project_default_service_accounts.name will be updated in-place
  ~ resource "google_project_default_service_accounts" "name" {
        action         = "delete"
        id             = "projects/arc-seed-project"
        project_id     = "arc-seed-project"
      ~ restore_policy = "REACTIVATE" -> "NONE"
    }

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

google_project_default_service_accounts.name: Modifying... [id=projects/arc-seed-project]
google_project_default_service_accounts.name: Modifications complete after 1s [id=projects/arc-seed-project]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
carvalho-c02ccan9md6m:pdsa thiagocarvalho$ 
```

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,151 @@
...
+		// },
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(10 * time.Minute),
+			Read:   schema.DefaultTimeout(10 * time.Minute),
+			Delete: schema.DefaultTimeout(10 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"project_id": {
```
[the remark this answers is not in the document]
thiagonache: I've copied from google_project, but no problem at all.

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,151 @@
...
+				ForceNew:     true,
+				ValidateFunc: validateProjectID(),
+				Description:  `The project ID where service accounts are created.`,
+			},
+			"action": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				Default:      "deprivilege",
+				ValidateFunc: validateServiceAccountAction(),
```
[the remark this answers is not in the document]
thiagonache: even having to validate three different options?

google/resource_google_project_default_service_accounts_test.go
thiagonache: sure. I was more interested in knowing if you guys follow TDD or not when coding terraform providers.

google/resource_google_project_default_service_accounts.go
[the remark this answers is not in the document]
thiagonache: we have deprivileged, delete and disable as valid values.

thiagonache:
> I know this is still WIP, but can you also add an outline for the documentation of the resource on the website so that we don't forget to do it later?

done!

google/resource_google_project_default_service_accounts.go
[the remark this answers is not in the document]
thiagonache: I could create another function for the update which just set the new state... I'll keep that in mind for the next updates

google/resource_google_project_default_service_accounts.go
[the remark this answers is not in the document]
thiagonache: > we have deprivilege, delete and disable as valid values.

That's the only doubt I have before I can continue to work on this.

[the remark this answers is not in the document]
thiagonache: NVM! I assumed that I knew what the function does. You're right!

thiagonache:
So, this is how I've implemented the main function:
```
	for _, sa := range serviceAccounts {
		// As per documentation https://cloud.google.com/iam/docs/service-accounts#default
		// we have just two default SAs and the e-mail may change. So, it is been filtered
		// by the Display Name
		switch sa.DisplayName {
		case "Compute Engine default service account":
			err := resourceGoogleProjectDefaultServiceAccountsDoAction(d, meta, action, sa.Email, pid)
			if err != nil {
				return fmt.Errorf("Error doing action %s on Service Account %s", action, sa.Email)
			}
		case "App Engine default service account":
			err := resourceGoogleProjectDefaultServiceAccountsDoAction(d, meta, action, sa.Email, pid)
			if err != nil {
				return fmt.Errorf("Error doing action %s on Service Account %s", action, sa.Email)
			}
		default:
			continue
		}
	}
```

How does it look like?

thiagonache:
Functions implementation:
```
func resourceGoogleProjectDefaultServiceAccountsDeleteAction(d *schema.ResourceData, meta interface{}, selflink string) error {
	config := meta.(*Config)
	userAgent, err := generateUserAgentString(d, config.userAgent)
	if err != nil {
		return err
	}

	_, err = config.NewIamClient(userAgent).Projects.ServiceAccounts.Delete(selflink).Do()
	if err != nil {
		return fmt.Errorf("Cannot delete service account: %v", err)
	}
	return nil
}

func resourceGoogleProjectDefaultServiceAccountsDoAction(d *schema.ResourceData, meta interface{}, action, email, project string) error {
	switch action {
	case "delete":
		var serviceAccountSelfLink = fmt.Sprintf("projects/%s/serviceAccounts/%s", project, email)
		err := resourceGoogleProjectDefaultServiceAccountsDeleteAction(d, meta, serviceAccountSelfLink)
		if err != nil {
			return fmt.Errorf("Cannot delete %s: %v", serviceAccountSelfLink, err)
		}
		return nil
	case "disable":
		return fmt.Errorf("not implemented yet")
	case "deprivilege":
		return fmt.Errorf("not implemented yet")
	default:
		return fmt.Errorf("Action %s is not a valid action", action)
	}
}
```

thiagonache:
Current execution:
```
carvalho-c02ccan9md6m:pdsa thiagocarvalho$ gcloud iam service-accounts list --project $TF_VAR_project_id
NAME                                    EMAIL                                                             DISABLED
my-project Project Service Account     project-service-account@my-project-ff05.iam.gserviceaccount.com  False
Compute Engine default service account  260650732864-compute@developer.gserviceaccount.com                True


carvalho-c02ccan9md6m:pdsa thiagocarvalho$ TF_LOG=trace terraform apply -auto-approve >/tmp/tf-run.log 2>&1


carvalho-c02ccan9md6m:pdsa thiagocarvalho$ gcloud iam service-accounts list --project $TF_VAR_project_id
NAME                                 EMAIL                                                             DISABLED
my-project Project Service Account  project-service-account@my-project-ff05.iam.gserviceaccount.com  False


carvalho-c02ccan9md6m:pdsa thiagocarvalho$ env | grep TF_VAR
TF_VAR_restore_policy=NONE
TF_VAR_action=delete
TF_VAR_project_id=my-project-ff05
```

thiagonache:
One more question:
```
	p, err := readGoogleProject(d, config, userAgent)
	if err != nil {
		if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 403 && strings.Contains(gerr.Message, "caller does not have permission") {
			return fmt.Errorf("the user does not have permission to access Project %q or it may not exist", p.ProjectId)
		}
		return handleNotFoundError(err, d, fmt.Sprintf("Project %q", p.ProjectId))
	}
```
I added it but giving a second thought I changed my mind. If the project does not exist, the call list service accounts will return that. Agree?

Sorry for too many questions.

thiagonache:
hi @danawillow ,

I've been thinking about the drift logic and I think it's not required since we are not storing the service accounts in the state. Do you agree?

thiagonache:
@danawillow first of all, sorry for spam you. At the end of the day, I could answer my own questions.
I think it's finally ready for review. I've created just a test for when the `policy_restore` is `NONE`. Please, let me know if that's enough.
Thank you for your patience and help.

website/docs/r/google_project_default_service_accounts.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+## Example Usage
+
+```hcl
+resource "google_project_default_service_accounts" "my_project" {
+  project = "my-project-id"
+  action = "delete"
+}
+```
+
+To try to reactivate the default service account on the resource destroy
```
danawillow, line 27:
```suggestion
To reactivate the default service accounts on the resource destroy:
```

website/docs/r/google_project_default_service_accounts.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+  restore_policy = "REACTIVATE"
+}
+
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `project` - (Required) The project ID. Changing this forces a new project to be created.
```
danawillow, line 42: Changing this deletes and recreates this resource, but it doesn't actually create a new project.

danawillow (changes requested):
Thanks @thiagonache! Here's another round of comments.

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,253 @@
...
+				ForceNew:     true,
+				ValidateFunc: validateProjectID(),
+				Description:  `The project ID where service accounts are created.`,
+			},
+			"action": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				Default:      "deprivilege",
+				ValidateFunc: validation.StringInSlice([]string{"deprivilege", "delete", "disable"}, false),
```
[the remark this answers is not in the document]
thiagonache: makes sense

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,253 @@
...
+	serviceAccounts, err := resourceGoogleProjectDefaultServiceAccountsList(config, d, userAgent)
+	if err != nil {
+		return fmt.Errorf("Error listing service accounts on project %s: %v", pid, err)
+	}
+	changedServiceAccounts := make(map[string]interface{})
+	for _, sa := range serviceAccounts {
+		// As per documentation https://cloud.google.com/iam/docs/service-accounts#default
+		// we have just two default SAs and the e-mail may change. So, it is been filtered
+		// by the Display Name
+		switch sa.DisplayName {
```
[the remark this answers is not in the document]
thiagonache: Honestly, I think the API should have a field to say if it's the default service account as it does for the disabled status.
The documentation says that the e-mail may change. So, I could not think of any other way to filter it. I'm going to convert to lowercase unless you have any other idea how to filter it.

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,253 @@
...
+			"restore_policy": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "NONE",
+				ValidateFunc: validation.StringInSlice([]string{"NONE", "REACTIVATE"}, false),
+				Description: `The action to be performed in the default service accounts on the resource destroy.
+				Valid values are NONE and REACTIVATE. If set to REACTIVATE it will attempt to restore all default SAs.`,
+			},
+			"service_accounts": {
+				Type:        schema.TypeMap,
```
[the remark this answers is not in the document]
thiagonache: I'm storing the service account changed and switch action I performed so I can revert it on the destroy. I'm storing uniqueId as key and service_account:action as value to know what to do on the destroy. The problem with a computed field is if the action is deleted I've no information about the SA on the API anymore.

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	response, err := config.NewIamClient(userAgent).Projects.ServiceAccounts.List(prefixedProject(pid)).Do()
+	if err != nil {
+		return nil, fmt.Errorf("failed to list service accounts on project %q: %v", pid, err)
+	}
+	return response.Accounts, nil
+}
+
+func resourceGoogleProjectDefaultServiceAccountsRead(d *schema.ResourceData, meta interface{}) error {
+	if err := d.Set("project", d.Get("project").(string)); err != nil {
```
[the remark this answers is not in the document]
thiagonache: thank you!

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,253 @@
...
+			}
+		}
+	}
+
+	d.SetId("")
+
+	return nil
+}
+
+func resourceGoogleProjectDefaultServiceAccountsUpdate(d *schema.ResourceData, meta interface{}) error {
```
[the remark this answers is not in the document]
thiagonache: thank you!

google/resource_google_project_default_service_accounts.go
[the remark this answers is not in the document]
thiagonache: Doh! I'm st*p*d. Thank you!!!

thiagonache:
hey @danawillow !
friendly message in case you missed the PR review request. if you are just busy, please, ignore me.

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,222 @@
...
+			ForceSendFields: []string{},
+			NullFields:      []string{},
+		}).Do()
+		if err != nil {
+			return fmt.Errorf("cannot get IAM policy on project %s: %v", project, err)
+		}
+
+		for _, bind := range iamPolicy.Bindings {
+			newMembers := []string{}
+			if bind.Role == "roles/editor" {
```
danawillow, line 105: Can you add a comment explaining why this is specifically for the editor role?

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,222 @@
...
+					if member != fmt.Sprintf("serviceAccount:%s", email) {
+						newMembers = append(newMembers, member)
+					}
+				}
+			}
+			bind.Members = newMembers
+		}
+		_, err = config.NewResourceManagerClient(userAgent).Projects.SetIamPolicy(project, &cloudresourcemanager.SetIamPolicyRequest{
+			Policy:          iamPolicy,
+			UpdateMask:      "",
```
danawillow, line 116: Since these three lines are just their default values, they can be removed

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,222 @@
...
+		// we have just two default SAs and the e-mail may change. So, it is been filtered
+		// by the Display Name
+		switch strings.ToLower(sa.DisplayName) {
+		case "compute engine default service account":
+			changedServiceAccounts[sa.UniqueId] = sa.Email
+			err := resourceGoogleProjectDefaultServiceAccountsDoAction(d, meta, action, sa.UniqueId, sa.Email, pid)
+			if err != nil {
+				return fmt.Errorf("error doing action %s on Service Account %s: %v", action, sa.Email, err)
+			}
+		case "app engine default service account":
```
danawillow, line 161: Since this logic is exactly the same as the compute engine one (and I imagine if more default SAs came up they would have the same logic as well), this feels more like it should be an if statement (with compute || app engine) instead of a case/switch

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,222 @@
...
+	pid, ok := d.Get("project").(string)
+	if !ok {
+		return fmt.Errorf("cannot get project")
+	}
+	for saUniqueID, saEmail := range d.Get("service_accounts").(map[string]interface{}) {
+		action := d.Get("action").(string)
+		switch action {
+		case "DISABLE":
+			action := "ENABLE"
+			err := resourceGoogleProjectDefaultServiceAccountsDoAction(d, meta, action, saUniqueID, saEmail.(string), pid)
```
danawillow, line 206: This is also a situation where the code inside the case statements are nearly identical. A simpler version would be:
```
origAction := d.Get("action").(string)
newAction := ""
if origAction == "DISABLE" {
  newAction = "ENABLE"
} else if origAction == "DELETE" {
  newAction = "UNDELETE"
} else {
  // error
}
err := resourceGoogleProjectDefaultServiceAccountsDoAction(d, meta, newAction, saUniqueID, saEmail.(string), pid)
```

google/resource_google_project_default_service_accounts_test.go
```diff
@@ -0,0 +1,287 @@
...
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckGoogleProjectDefaultServiceAccountsBasic(org, project, billingAccount),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "id", "projects/"+project),
+					resource.TestCheckResourceAttrSet(resourceName, "project"),
+					resource.TestCheckResourceAttrSet(resourceName, "action"),
+					resource.TestCheckResourceAttrSet(resourceName, "project"),
```
danawillow, line 32: This line is identical to line 30

google/resource_google_project_default_service_accounts_test.go
```diff
@@ -0,0 +1,287 @@
...
+func testAccCheckGoogleProjectDefaultServiceAccountsBasic(org, project, billingAccount string) string {
+	return fmt.Sprintf(`
+resource "google_project" "acceptance" {
+	project_id = "%s"
+	name       = "%s"
+	org_id  = "%s"
+	billing_account = "%s"
+}
+
+resource "google_project_default_service_accounts" "acceptance" {
```
danawillow, line 48: Looking at this resource, I as a user don't really know what it's intending to accomplish because the behavior is all in default values. I wonder whether it actually makes sense to make some fields required instead of default so that the user's intent is more explicit. What do you think?

google/resource_google_project_default_service_accounts_test.go
```diff
@@ -0,0 +1,287 @@
...
+					resource.TestCheckResourceAttrSet("google_project_default_service_accounts.acceptance", "project"),
+					preventRaceCondition(), // Looks like the API is eventually consistent by couple seconds
+					testAccCheckGoogleProjectDefaultServiceAccountsChanges(t, project, action),
+				),
+			},
+		},
+	})
+}
+
+func preventRaceCondition() func(s *terraform.State) error {
```
danawillow, line 141: What if instead this function were moved to provider_test.go and given a more generic name like `sleepForTest` and the ability to choose how long you sleep for? I know there's at least one other test in the provider (in provider_test.go) that has a sleep step.

google/resource_google_project_default_service_accounts_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+func testAccCheckGoogleProjectDefaultServiceAccountsChanges(t *testing.T, project, action string) func(s *terraform.State) error {
+	return func(s *terraform.State) error {
+		config := googleProviderConfig(t)
+		response, err := config.NewIamClient(config.userAgent).Projects.ServiceAccounts.List(prefixedProject(project)).Do()
+		if err != nil {
+			return fmt.Errorf("failed to list service accounts on project %q: %v", project, err)
+		}
+		for _, sa := range response.Accounts {
+			switch strings.ToLower(sa.DisplayName) {
```
danawillow, line 181: Same comment here as in the resource, where I think this switch statement should actually just be an if statement with an || clause

google/resource_google_project_default_service_accounts_test.go
```diff
@@ -0,0 +1,287 @@
...
+				case "DISABLE":
+					if !sa.Disabled {
+						return fmt.Errorf("compute engine default service account is not disabled, disable field is %t", sa.Disabled)
+					}
+				case "DELETE":
+					return fmt.Errorf("compute engine default service account is not deleted")
+				case "DEPRIVILEGE":
+					iamPolicy, err := config.NewResourceManagerClient(config.userAgent).Projects.GetIamPolicy(project, &cloudresourcemanager.GetIamPolicyRequest{
+						Options:         &cloudresourcemanager.GetPolicyOptions{},
+						ForceSendFields: []string{},
```
danawillow, line 193: same comment wrt not needing to include empty fields

google/resource_google_project_default_service_accounts_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+func testAccCheckGoogleProjectDefaultServiceAccountsRevert(t *testing.T, project, action string) func(s *terraform.State) error {
+	return func(s *terraform.State) error {
+		config := googleProviderConfig(t)
+		response, err := config.NewIamClient(config.userAgent).Projects.ServiceAccounts.List(prefixedProject(project)).Do()
+		if err != nil {
+			return fmt.Errorf("failed to list service accounts on project %q: %v", project, err)
+		}
+		for _, sa := range response.Accounts {
+			switch strings.ToLower(sa.DisplayName) {
```
danawillow, line 253: same here. Also, now that this has shown up a few times, it actually probably makes sense to have a separate `func isDefaultServiceAccount(displayName string) bool` or something like that

danawillow (changes requested):
This is coming together well- thanks for adding the tests! Most of my comments now are around clarity and finding places to consolidate code. In addition to the places I specifically pointed out, I'd also recommend doing a proofread before sending it back to see if there are any others that I missed where common logic can be pulled out.

google/resource_google_project_default_service_accounts_test.go
thiagonache: agreed

google/resource_google_project_default_service_accounts_test.go
thiagonache: I'm okay with creating a function for the main code, but my concern about using the same function for tests is that I broke the function I may break get false results from the tests. Thoughts?

danawillow: Wouldn't that be the exact same scenario as if you had two identical functions?

thiagonache: yes, but I would not have two identical functions. We would have code function and test function with will eventually being the same, but allow me to test the logic in two steps. Update the code function, make sure the test fails, update the function code, make sure the tests pass... just thought I had.....

danawillow: That's fine! This is a pretty simple situation (just comparing strings) so I don't have a strong preference.

google/resource_google_project_default_service_accounts.go
thiagonache: thank you! updated. we should not error in this case, that's the only difference.

google/resource_google_project_default_service_accounts_test.go
thiagonache: Sorry, I could not find this function. I honestly don't think we need a function for it. time.Sleep is enough.
what if we use just an inline function? 
`func(s *terraform.State) error { time.Sleep(5 * time.Second); return nil },`

thiagonache: I did it... it's simple enough :D

danawillow: That's fine!

thiagonache: should I update the two functions that use the sleep builtin?

danawillow: I'm not totally sure what you mean by that, but do what you think makes the most sense and I'll let you know if there's something else I'd prefer!

thiagonache: I mean that I end up creating the function in the provider_test.go. Should I update the other two functions that use time.Sleep inside the provider_test.go to the function I just created? Sorry for the confusing message.
I think I shouldn't but just double-checking.

thiagonache: > I'm not totally sure what you mean by that, but do what you think makes the most sense and I'll let you know if there's something else I'd prefer!

k... i will do that

thiagonache:
Alright... I went over it multiple times. Implemented changes you mentioned and renamed the REACTIVATE to REVERT. I'll give it one more review tomorrow with a fresh and rested brain before requesting the review.

danawillow:
You too! Today is actually my last day on the team, so I'll hand off the review to @rileykarson for whenever you're ready.

thiagonache:
If you want to finish this work before leaving I’m comfortable to ask a review. I’d review once more just in case

thiagonache:
@rileykarson Hello sir, this is ready for review!

google/resource_google_project_default_service_accounts.go
```diff
@@ -0,0 +1,202 @@
...
+	if d.Get("restore_policy").(string) == "NONE" {
+		d.SetId("")
+		return nil
+	}
+
+	pid := d.Get("project").(string)
+	for saUniqueID, saEmail := range d.Get("service_accounts").(map[string]interface{}) {
+		origAction := d.Get("action").(string)
+		newAction := ""
+		// We agreed to not revert the DEPRIVILEGE because Morgante said it is not required.
```
rileykarson, line 173: Can you link the comment here?

website/docs/r/google_project_default_service_accounts.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+---
+
+# google_project_default_service_accounts
+
+Allows management of Google Cloud Platform project default service accounts.
+
+When certain service APIs are enabled, Google Cloud Platform automatically creates service accounts to help get started, but
+this is not recommended for production environments as per [Google's documentation](https://cloud.google.com/iam/docs/service-accounts#default).
+See the [Organization documentation](https://cloud.google.com/resource-manager/docs/quickstarts) for more details.
+
```
rileykarson, line 17:
```suggestion
~> This resource works on a best-effort basis, as no API formally describes the default service accounts. If the default service accounts change their name/email or additional service accounts are added, this resource will need to be updated.

```

Let's add a disclaimer so that it's clear what our contract is.

website/docs/r/google_project_default_service_accounts.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `project` - (Required) The project ID. Changing this forces the actions to be applied in the new project.
+
+- `action` - (Optional) The action to be performed in the default service accounts. Valid values are: deprivilege, delete, disable.
```
rileykarson, line 44: required?

rileykarson, line 44: Make sure the valid values match the user-specifiable ones exactly. I'd also recommend backticks to `code font` the values.

website/docs/r/google_project_default_service_accounts.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+
+- `restore_policy` - (Optional) The action to be performed in the default service accounts on the resource destroy.
+  Valid values are NONE and REACTIVATE. If set to REACTIVATE it will attempt to restore all default SAs.
```
rileykarson, line 47: REVERT?

website/docs/r/google_project_default_service_accounts.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+```hcl
+resource "google_project_default_service_accounts" "my_project" {
+  project = "my-project-id"
+  action = "DISABLE"
+  restore_policy = "REACTIVATE"
+}
+
+```
+
+## Argument Reference
```
rileykarson, line 38: Can you sync the descriptions here + on the fields in their `Description` field?

website/docs/r/google_project_default_service_accounts.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+The following arguments are supported:
+
+- `project` - (Required) The project ID. Changing this forces the actions to be applied in the new project.
+
+- `action` - (Optional) The action to be performed in the default service accounts. Valid values are: deprivilege, delete, disable.
+
+- `restore_policy` - (Optional) The action to be performed in the default service accounts on the resource destroy.
+  Valid values are NONE and REACTIVATE. If set to REACTIVATE it will attempt to restore all default SAs.
+
+## Attributes Reference
```
rileykarson, line 49: service_accounts is an exported attribute, right?

thiagonache:
Thank you for the review! I'm sorry, I decided to change from REACTIVATE to REVERT at the last minute and forgot to update the documentation.

thiagonache:
@rileykarson just a friendly message saying it's ready for review.

rileykarson (approved):
LGTM- upstreaming to Magic Modules so this gets applied to the `google-beta` provider too. That's https://github.com/GoogleCloudPlatform/magic-modules/pull/4167. Once that's approved, I'll merge this one too.

thiagonache:
> LGTM- upstreaming to Magic Modules so this gets applied to the `google-beta` provider too. That's [GoogleCloudPlatform/magic-modules#4167](https://github.com/GoogleCloudPlatform/magic-modules/pull/4167). Once that's approved, I'll merge this one too.

sorry for bothering you with friendly reminders... I'm very excited about this work and I got anxious. I'll do better next time.

thiagonache:
> LGTM- upstreaming to Magic Modules so this gets applied to the `google-beta` provider too. That's [GoogleCloudPlatform/magic-modules#4167](https://github.com/GoogleCloudPlatform/magic-modules/pull/4167). Once that's approved, I'll merge this one too.

Please, do not merge it yet. There's a bug in the revert implementation of deleted service accounts... they can only be undeleted after 30 days

thiagonache:
thank you @byte_count

thiagonache:
Riley,

How do we want to handle this? Store deletion date in the state and use for comparison or just ignore errors on the delete operation?

rileykarson:
We'd return an error if someone tried to undelete a service account that had been deleted for 30 days (they can only be undeleted for the first 30 days at which point they're deleted, I believe), right? Then they could just change their `restore_policy` to `NONE`.

That's my understanding of what would happen, at least.

thiagonache:
> We'd return an error if someone tried to undelete a service account that had been deleted for 30 days (they can only be undeleted for the first 30 days at which point they're deleted, I believe), right? Then they could just change their `restore_policy` to `NONE`.

right.

> 
> That's my understanding of what would happen, at least.

my idea is to add a struct (resource) to store more info about the changed SA
```
                         ...

                        "service_accounts": {
				Type:        schema.TypeList,
				Elem:        changedServiceAccountResource(),
				Computed:    true,
				Description: `The Service Accounts changed by this resource. It is used for revert the action on the destroy.`,
			},
		},
	}
}

func changedServiceAccountResource() *schema.Resource {
	return &schema.Resource{
			"unique_id": {
				Type:     schema.TypeString,
				Computed: true,
			},
			"email": {
				Type:     schema.TypeString,
				Computed: true,
			},
			"action_date": {
				Type:     schema.TypeInt,
				Computed: true,
			},
			"roles": {
				Type:     schema.TypeList,
				Computed: true,
				Elem: &schema.Resource{
						"id": {
							Type:     schema.TypeString,
							Optional: true,
						},
					},
				},
			},
		},
	}
}
```

It would look like that in the state
```
            ...

            "service_accounts": [
              {
                "action_date": 1604031401,
                "email": "1008146422730-compute@developer.gserviceaccount.com",
                "roles": [],
                "unique_id": "109212719101262793432"
              }
            ],
            "timeouts": null
```

I think it is better to store when the resource changed the service account instead of parsing the error returned since the API does not implement/expose the lifecycle as it does for a project.
We can choose between fail and return the message to the user to change the `restore_policy` or just ignore the error since the undelete may fail for another reason as noted in the API docs.
Also, I've added roles to the struct just to avoid break changes when implementing the revert for deprivilege action.

I've coded this change, just need to update how I read the state on the destroy. What I mean is that I should be able to solve it quickly.

Thank you!

thiagonache:
so, there's no bug... hehe
that's the current behavior for all APIs calls

thiagonache:
you can go ahead and merge whenever you want.

PR #7550, opened 2020-10-16
Adding missing new line for dnssec_config block

jtyr:
This PR is adding a missing new line to make the arguments for the `dnssec_config` block as a separate paragraph.


ScottSuarez:
Great catch on this. Looks like this was a bug in our generator logic.
https://github.com/GoogleCloudPlatform/magic-modules/pull/4131/files

upstream change here

PR #7566, opened 2020-10-19
Bumped google api dependency to 0.33.0

melinath:
This is a prerequisite for #7480.


melinath:
Test run & discussion are currently on https://github.com/hashicorp/terraform-provider-google-beta/pull/2616 since that should include both things that are GA and beta. LMK if I should run tests here as well.

slevenick (approved):
Should be fine with just the tests run against beta. We will get additional test runs nightly before this gets released anyways, so we should just be aware of this as a possible cause of any failures that come up

PR #7577, opened 2020-10-20
Update container_node_pool.html.markdown

alexdicianu:
It seems there are a few discrepancies between the official Google GKE docs and the terraform provider docs. Please see correction below.

### DIFF
--- a/website/docs/r/container_node_pool.html.markdown
+++ b/website/docs/r/container_node_pool.html.markdown
@@ -149,7 +149,7 @@ cluster.
     [google_container_cluster](container_cluster.html) for schema.
 
 * `node_count` - (Optional) The number of nodes per instance group. This field can be used to
-    update the number of nodes per instance group but should not be used alongside `autoscaling`.
+    update the number of nodes per instance group.
 
 * `project` - (Optional) The ID of the project in which to create the node pool. If blank,
     the provider-configured project will be used.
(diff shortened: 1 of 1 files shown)


website/docs/r/container_node_pool.html.markdown
```diff
@@ -149,7 +149,7 @@ cluster.
     [google_container_cluster](container_cluster.html) for schema.
 
 * `node_count` - (Optional) The number of nodes per instance group. This field can be used to
-    update the number of nodes per instance group but should not be used alongside `autoscaling`.
+    update the number of nodes per instance group.
```
alexdicianu, line 152: As per the documentation below, `node_count` is used alongside `autoscaling` and represents the starting point of the pool's size.

https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler#minimum_and_maximum_node_pool_size

website/docs/r/container_node_pool.html.markdown
```diff
@@ -167,10 +167,10 @@ cluster.
 
 The `autoscaling` block supports:
 
-* `min_node_count` - (Required) Minimum number of nodes in the NodePool. Must be >=0 and
+* `min_node_count` - (Required) Minimum number of nodes per instance group in the NodePool. Must be >=0 and
     <= `max_node_count`.
 
-* `max_node_count` - (Required) Maximum number of nodes in the NodePool. Must be >= min_node_count.
+* `max_node_count` - (Required) Maximum number of nodes per instance group in the NodePool. Must be >= min_node_count.
```
alexdicianu, line 173: The min and max node count seem to be per instace group (i.e. per zone for multi-zonal clusters).

https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler#minimum_and_maximum_node_pool_size

website/docs/r/container_node_pool.html.markdown
ndmckinley: Yes, but in a declarative config system like Terraform, this will produce a diff if the autoscaler changes the node count.  We'll take the other change, though!  This repository is generated, though, so if you don't mind adding it to GoogleCloudPlatform/magic-modules, under third_party/terraform/website, that would be great.

ndmckinley (changes requested):

PR #7582, opened 2020-10-20
Update bigquery_dataset.html.markdown

burnzy:
Correcting bigquery dataset id attribute output format. 
Output returned is `{{project_id}}:{{dataset_id}}`

### DIFF
--- a/website/docs/r/bigquery_dataset.html.markdown
+++ b/website/docs/r/bigquery_dataset.html.markdown
@@ -259,7 +259,7 @@ The `default_encryption_configuration` block supports:
 
 In addition to the arguments listed above, the following computed attributes are exported:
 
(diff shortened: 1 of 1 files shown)


ScottSuarez:
Upstreaming this change 
https://github.com/GoogleCloudPlatform/magic-modules/pull/4132

ScottSuarez:
Won't change after looking into this. We don't surface id in the same way as the official rest api we actually manually craft it to follow the pattern documented. Any reason for this change?

burnzy:
Sorry for the confusion @ScottSuarez — my dataset resources are returning the project_id:dataset_id format still, even using the latest provider. So that through me off. I'm guessing the id has since changed to the documented format, but my Terraform state file hasn't updated with those changes. I'm not quite sure why the state hasn't updated since using the new provider format, but I will need to investigate on my side.

PR #7634, opened 2020-10-26
Updated docs for switching to/from google-beta

Justin-W:
The "Resources can safely be converted from one version to the other without needing to rebuild infrastructure." statement and associated migration steps in the previous docs appears to be inaccurate as of Terraform v0.13.

Following the previously documented instructions for switching existing resources to use the google-beta provider resulted in `terraform plan` that attempted to destroy (using the google provider) all existing resources and to (re)create new instances of each resource (using the google-beta provider), instead of a no-op plan.

Also, if the existing provider's type was changed (rather than adding a new, separate provider, while retaining the old provider) in the Terraform configuration, then `terraform refresh` would also result in errors similar to the following:

```
Error: Provider configuration not present

To work with
google_compute_RESOURCE_TYPE.RESOURCE-NAME
its original provider configuration at
provider["registry.terraform.io/hashicorp/google"].PROVIDER-ALIAS
is required, but it has been removed. This occurs when a provider
configuration is removed while objects created by that provider still exist in
the state. Re-add the provider configuration to destroy
google_compute_RESOURCE_TYPE.RESOURCE-NAME,
after which you can remove the provider configuration again.
```

This new/undocumented behavior when switching between the google and google-beta providers appears to be related to changes in Terraform v0.12 and v0.13 that affected the way in which providers are referenced within the Terraform state data.

### DIFF
--- a/website/docs/guides/provider_versions.html.markdown
+++ b/website/docs/guides/provider_versions.html.markdown
@@ -119,11 +119,26 @@ terraform import google_compute_instance.beta-instance my-instance
 
 Resources can safely be converted from one version to the other without needing to rebuild infrastructure.
 
-To go from GA to beta, change the `provider` field from `"google"` to `"google-beta"`.
+To go from GA to beta:
+
+1. Change the `provider` field from `"google"` to `"google-beta"`.
+1. Migrate the Terraform state of your existing infrastructure, to migrate the ownership of existing resources from the `"google"` provider to the `"google-beta"` provider. For example, the following command will migrate all existing resources in the Terraform state that used the `"google"` provider to use the `"google-beta"` provider instead: 
+    
+    ```bash
+    terraform state replace-provider registry.terraform.io/hashicorp/google registry.terraform.io/hashicorp/google-beta
+    ```
+
+1. (Optional) Run  `terraform plan` or `terraform refresh`+`terraform show` to verify that the migrated Terraform configuration and state result in a correct plan (e.g. typically "Plan: 0 to add, 0 to change, 0 to destroy."), and that beta-only fields are added to the state.
 
 To go from beta to GA, do the reverse. If you were previously using beta fields that you no longer wish to use:
 
 1. (Optional) Explicitly set the fields back to their default values in your Terraform config file, and run `terraform apply`.
 1. Change the `provider` field to `"google"`.
 1. Remove any beta fields from your Terraform config.
+1. Migrate the Terraform state of your existing infrastructure, to migrate the ownership of existing resources from the `"google-beta"` provider to the `"google"` provider. For example, the following command will migrate all existing resources in the Terraform state that used the `"google-beta"` provider to use the `"google"` provider instead: 
+    
+    ```bash
+    terraform state replace-provider registry.terraform.io/hashicorp/google-beta registry.terraform.io/hashicorp/google
+    ```
+
 1. Run  `terraform plan` or `terraform refresh`+`terraform show` to see that the beta fields are no longer in state.


Justin-W:
Can I get some confirmation from the google provider team of whether this doc update is correct or not? I want to make sure that the new step I documented is now expected to be a necessary step (vs if I am unaware of some other, better way to handle switching providers without impacting already-provisioned infrastructure). Thanks.

rileykarson:
Sorry! I hadn't experienced this myself, so I wanted to dig into it more before reviewing. I'm still able to switch providers at will- can you post a before+after config?

my versions (`5.0.0` is a made up version that refers to locally built copies:

```
$ terraform -v
Terraform v0.13.4
+ provider registry.terraform.io/hashicorp/google v3.44.0
+ provider registry.terraform.io/hashicorp/google-beta v3.44.0
```

Justin-W:
> I'm still able to switch providers at will- can you post a before+after config?

Terraform v0.13.3

Original Terraform configuration:

```terraform
provider "google" {
  alias = "xyz-gcp-us-east1"

  version = "~> 3.41.0"

  credentials = "..."
  project     = "..."
  region      = "us-east1"
}

//resources...
```

New Terraform configuration:

```terraform
provider "google-beta" {
  alias = "xyz-gcp-us-east1"

  version = "~> 3.41.0"

  credentials = "..."
  project     = "..."
  region      = "us-east1"
}

//resources...
//NOTE: The resources' original configurations were unchanged, except as described below in #2 and #3.
```

During the migration to the google-beta provider, the only HCL configuration modifications made were:

1. Modified all TF providers' declarations by replacing `provider "google" {` with `provider "google-beta" {`.
2. For each existing resource declaration of a resource type that has a google-beta-only `labels` attribute, we added a `labels = local.common_labels` attribute.
3. Modified each resources' `provider` attribute, by replacing `provider = google.xyz-gcp-us-east1` with `provider = google-beta.xyz-gcp-us-east1`.

What we expected was for `terraform plan` to generate a plan (based on the existing tfstate from the previous `terraform apply` which used the original GA google provider) that would update only the resources with the newly-added `labels` attribute (due to #2 above), and perform those updates using the modified google-beta providers (due to #1 above).

We expected that the plan would require zero resources to be added or removed, and for all of the "unchanged" resources (those which hadn't received a new `labels` attribute) to be completely unaffected.

Instead, after changing the TF HCL as described above, `terraform plan` failed (during the `terraform refresh` that runs automatically during `terraform plan`), emitting many errors with output like the following:

```
Error: Provider configuration not present

To work with
google_compute_RESOURCE_TYPE.RESOURCE-NAME
its original provider configuration at
provider["registry.terraform.io/hashicorp/google"].PROVIDER-ALIAS
is required, but it has been removed. This occurs when a provider
configuration is removed while objects created by that provider still exist in
the state. Re-add the provider configuration to destroy
google_compute_RESOURCE_TYPE.RESOURCE-NAME,
after which you can remove the provider configuration again.
```

where the `google_compute_RESOURCE_TYPE` in the error above included the following resources types:

- google_compute_backend_service
- google_compute_global_forwarding_rule
- google_compute_health_check
- google_compute_ssl_certificate
- google_compute_target_https_proxy
- google_compute_url_map

I suspect that the reason that those particular resources generated errors (while many other resources did not) is because the resources which generated the errors were among the ones which have a beta-only `labels` attribute, and which we therefore expected would be "updated" in the plan which we expected would be generated by `terraform plan`.

It seems that none of the errors we saw were for resources whose configuration was unchanged (i.e. those without a beta-only labels attribute).

However, only some of the resources with modified HCL generated refresh errors. Other resources (e.g. google_compute_address resources) with a newly-added label did not result in such errors.

But for each resource which generated such an error, it seems like the refresh and/or plan incorrectly was trying to do a full delete+add of the resource, instead of the simple "update resource" which we expected.

And that such delete+add operations were attempting to delete using the (changed/removed) "google.xyz-gcp-us-east1" provider (based on the tfstate data's association with the GA provider), and then planning to re-add each deleted resource using the changed/new "google-beta.xyz-gcp-us-east1" provider (based on the migrated HCL). However, since the original provider declaration was modified/replaced (by changing the provider type, but not the `alias`), the (undesirably-)planned resource deletes all failed, since the "google.xyz-gcp-us-east1" provider no longer existed in the migrated HCL.

Justin-W:
@rileykarson Any update on this?

Justin-W:
Can I get an update on this issue, please?

rileykarson:
Sorry, I tried to reproduce this and couldn't. I earmarked it to dig into more, but didn't find the time.

What version are you using? I can't reproduce your issue when using `0.13.4`.

PR #7688, opened 2020-10-30
Bump the SDK version to 2.1.0

rileykarson:
This picks up changes in `2.0.4` that fix `CheckDestroy` functions, which were broken in earlier versions of SDKv2. See https://github.com/hashicorp/terraform-plugin-sdk/blob/master/CHANGELOG.md#204-october-06-2020. We caught the issue in https://github.com/GoogleCloudPlatform/magic-modules/pull/4129.

Once I'm upgrading anyways, may as well go to latest (`2.1.0`).


melinath (approved):
LGTM

PR #7794, opened 2020-11-12
Add google_composer_environment data source

FromZeus:
Hello! I propose to add google composer environment data source because it's safer way to reference existing environments without a chance to destroy it.

### DIFF
--- a/google/data_source_google_composer_environment.go
+++ b/google/data_source_google_composer_environment.go
@@ -8,41 +8,21 @@ import (
 
 func dataSourceGoogleComposerEnvironment() *schema.Resource {
 	return &schema.Resource{
-		Read: dataSourceGoogleComposerEnvironmentRead,
-		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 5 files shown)


ndmckinley:
Hi there!  Absolutely, this is great.  I do have one request - this repository is generated code, it comes from github.com/GoogleCloudPlatform/magic-modules.  That generator does have the ability to copy handwritten code - which helps us maintain parity between the GA provider (this one) and the beta provider.

Are you able to add these files there, instead?  The best place for them would be `third_party/terraform/` and then the subdirectories `data_sources`, `tests`, and you'd need to modify `provider.go` in the `util` subdirectory.  Once you create a PR, the automation will create the diffs and we can review those.

Does that make sense?

FromZeus:
@ndmckinley Yep, done GoogleCloudPlatform/magic-modules#4238

ndmckinley:
Great, will close this in favor of that one.

PR #7795, opened 2020-11-12
Add description and disabled fields to logging sinks

dabbertorres:
While doing some work with analyzing logs, higher priority tasks came up, and it was useful to be able to temporarily disable the log sink when not working on the analysis task(s). I noticed `disabled` (and then `description` as well) were missing from the Terraform resource schemas compared to the GCP API.

In short: all this PR does is add `description` and `disabled` fields to the logging sink (project, folder, organization, and billing account) schemas. Tests for these new fields are included.

### DIFF
--- a/google/resource_logging_billing_account_sink_test.go
+++ b/google/resource_logging_billing_account_sink_test.go
@@ -83,6 +83,64 @@ func TestAccLoggingBillingAccountSink_update(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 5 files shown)


ScottSuarez:
hi @dabbertorres ! I've got a change to upstream this [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/4233)

dabbertorres:
@ScottSuarez, thanks! I added the fields I missed that @slevenick noted on the upstream PR.

PR #7813, opened 2020-11-15
docs(kms-crypto-key): fix wrong import format

elsbrock:
Fixes wrong import format missing location, project name:

```
Error: Invalid CryptoKey id format, expecting `{projectId}/{locationId}/{KeyringName}/{cryptoKeyName}` or `{locationId}/{keyRingName}/{cryptoKeyName}, got id: mykeyring/mytoken
```

### DIFF
--- a/website/docs/r/kms_crypto_key.html.markdown
+++ b/website/docs/r/kms_crypto_key.html.markdown
@@ -168,8 +168,8 @@ This resource provides the following
 CryptoKey can be imported using any of these accepted formats:
 
 ```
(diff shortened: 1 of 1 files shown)


slevenick (commented):
Unfortunately this code is generated from this template: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/templates/terraform/resource.html.markdown.erb

So this change would be overwritten immediately if this change were accepted. We should look into better support for these short forms that expect specific formats as it can be difficult formats like this

PR #7835, opened 2020-11-17
Clarify name format for importing uptime checks

joshk0:
The {{name}} is not obvious, and needs to be a longer string involving the project.

### DIFF
--- a/website/docs/r/monitoring_uptime_check_config.html.markdown
+++ b/website/docs/r/monitoring_uptime_check_config.html.markdown
@@ -306,10 +306,10 @@ This resource provides the following
 ## Import
 
 
-UptimeCheckConfig can be imported using any of these accepted formats:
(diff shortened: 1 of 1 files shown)


slevenick (commented):
Unfortunately this code is generated from this template: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/templates/terraform/resource.html.markdown.erb

So this change would be overwritten immediately if this change were accepted. We should look into better support for `{{name}}` as it can be difficult formats like this

joshk0:
Hello, just checking for an update here. Is there a better place to submit this doc change now?

slevenick:
Not at the moment. I have filed a bug for better support for this, but it would likely need to happen at the code generation level. https://github.com/hashicorp/terraform-provider-google/issues/8499

Closing this for the time being

Issue #7778, opened 2020-11-11
Incorrect docs for google_service_account_iam


zbjornson:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Issue

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account_iam says

> This resource is to add iam policy bindings to a service account resource **to configure permissions for who can edit the service account.** To configure permissions for a service account to act as an identity that can manage other GCP resources, use the google_project_iam set of resources.

I think that's incorrect: I think that resource can be used for any of the purposes that correspond to the Service Account roles, https://cloud.google.com/iam/docs/understanding-roles#service-accounts-roles, which notably includes `iam.serviceAccountUser`.

I'm not 100% certain here, otherwise I'd open a PR.

edwardmedia:
@zbjornson that statement is correct. Closing this issue then. Feel free to reopen it if you have code backing your thought.

zbjornson:
@edwardmedia the examples on that page include:

```
resource "google_service_account_iam_binding" "admin-account-iam" {
  service_account_id = google_service_account.sa.name
  role               = "roles/iam.serviceAccountUser"

  members = [
    "user:jane@example.com",
  ]
}
```

The `"roles/iam.serviceAccountUser"` role "Allows members to indirectly access all the resources that the service account can access" ([ref](https://cloud.google.com/iam/docs/impersonating-service-accounts#allow-impersonation)).

I created a `google_service_account_iam_binding` with that role like the one I pasted above, and the GCP console shows the `members` have the "Service Account User" role for that `service_account_id`.

PR #7874, opened 2020-11-23
Correct documentation for google_service_account_iam

zbjornson:
I think #7778 was closed incorrectly, but I didn't receive a follow-up response there. In that issue I provided an example supporting this change.


rileykarson (approved):
LGTM. I've created https://github.com/GoogleCloudPlatform/magic-modules/pull/4289 to upstream this so that it gets applied to the `google-beta` provider as well, I'll merge here once that's merged.

PR #7879, opened 2020-11-24
bigtable_gc_policy: Adding support for duration lower than day

Eraac:
Allow GC policy to have a duration lower than a day, bigtable supports GC policies down to the hour. In case one day the support go to minutes/seconds the code will be already good

### DIFF
--- a/google/resource_bigtable_gc_policy.go
+++ b/google/resource_bigtable_gc_policy.go
@@ -58,13 +58,13 @@ func resourceBigtableGCPolicy() *schema.Resource {
 				Optional:     true,
 				ForceNew:     true,
 				Description:  `GC policy that applies to all cells older than the given age.`,
-				ExactlyOneOf: []string{"max_age.0.days", "max_age.0.seconds"},
+				ExactlyOneOf: []string{"max_age.0.days", "max_age.0.duration"},
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"days": {
 							Type:        schema.TypeInt,
 							Optional:    true,
-							Deprecated:  "Deprecated in favor of seconds",
+							Deprecated:  "Deprecated in favor of duration",
 							Description: `Number of days before applying GC policy.`,
 						},
 						"duration": {
@@ -271,10 +271,12 @@ func generateBigtableGCPolicy(d *schema.ResourceData) (bigtable.GCPolicy, error)
 }
 
 func getMaxAgeDuration(values map[string]interface{}) (time.Duration, error) {
-	d := values["seconds"].(string)
-	if d == "" {
-		d = values["days"].(string)
+	d := values["duration"].(string)
+	if d != "" {
+		return time.ParseDuration(d)
 	}
 
-	return time.ParseDuration(d)
+	days := values["days"].(int)
+
+	return time.Hour * 24 * time.Duration(days), nil
 }
--- a/google/resource_bigtable_gc_policy.go
+++ b/google/resource_bigtable_gc_policy.go
@@ -54,24 +54,25 @@ func resourceBigtableGCPolicy() *schema.Resource {
 			},
 
 			"max_age": {
-				Type:         schema.TypeList,
-				Optional:     true,
-				ForceNew:     true,
-				Description:  `GC policy that applies to all cells older than the given age.`,
-				ExactlyOneOf: []string{"max_age.0.days", "max_age.0.duration"},
+				Type:        schema.TypeList,
+				Optional:    true,
+				ForceNew:    true,
+				Description: `GC policy that applies to all cells older than the given age.`,
 				Elem: &schema.Resource{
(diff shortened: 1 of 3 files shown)


google/resource_bigtable_gc_policy.go
```diff
@@ -62,9 +62,15 @@ func resourceBigtableGCPolicy() *schema.Resource {
...
 						"days": {
 							Type:        schema.TypeInt,
-							Required:    true,
+							Optional:    true,
+							Deprecated:  "Deprecated in favor of seconds",
 							Description: `Number of days before applying GC policy.`,
 						},
+						"seconds": {
+							Type:        schema.TypeInt,
+							Required:    true,
```
slevenick, line 71: We cannot add a `Required: true` in a minor version release of the Terraform provider, as it will break existing configurations.

slevenick (changes requested):
Is there another way we could handle this? Potentially adding new fields and calculating the max age policy depending on which is set? We need to handle this without adding a new required field at least until 4.0.0 is ready to release

google/resource_bigtable_gc_policy.go
Eraac: Indeed, I've update the code to consider this optional and throw an error if both fields aren't set

slevenick:
So, the existing tests pass using the seconds field, but using the `days` field seems to be causing errors for me. It seems to always be producing a duration of 0 even when days is set to real values.

google/resource_bigtable_gc_policy.go
```diff
@@ -62,9 +63,15 @@ func resourceBigtableGCPolicy() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"days": {
 							Type:        schema.TypeInt,
-							Required:    true,
+							Optional:    true,
+							Deprecated:  "Deprecated in favor of seconds",
 							Description: `Number of days before applying GC policy.`,
 						},
+						"seconds": {
```
slevenick, line 70: We will want to add `ExactlyOneOf` to the schema to ensure one of these are set. `ExactlyOneOf` must use hard coded resource indexes to work with lists, so we will want to limit this max_age list to a length of 1 (this is effectively how it works anyways)

Example: https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_compute_resource_policy.go#L142

slevenick: After this is added you can remove the log around "Either seconds or days must be set". Not setting either will cause a panic before that error is hit anyways

google/resource_bigtable_gc_policy.go
```diff
@@ -257,3 +267,17 @@ func generateBigtableGCPolicy(d *schema.ResourceData) (bigtable.GCPolicy, error)
 
 	return policies[0], nil
 }
+
+func getMaxAgeDuration(values map[string]interface{}) (time.Duration, error) {
+	d, ok := values["seconds"]
```
slevenick, line 272: Terraform will default the value of `seconds` to 0 if it is unset (because int is a primitive value I believe). This prevents us from telling if `seconds` is set to the value 0, or unset. `ok` will always be true because of this, causing this code to return a duration of 0 seconds if days is specified.

Because 0 is an invalid duration, we should probably assume that a value of 0 is equivalent to `seconds` not being set.

On that note, we should enforce validation on the `seconds` field to make sure users set a positive int value. This could look like `ValidateFunc: validation.IntAtLeast(1000)` or some similar value

slevenick (changes requested):

google/resource_bigtable_gc_policy.go
Eraac: I've change the type to string and use `time.ParseDuration`, and set a validation function

slevenick: Ah, that won't quite work either. Changing the type of the `days` field will cause existing user configs to stop working, which would be a breaking change.

I think what we want is to set them back to ints, set the ValidateFunc on the `seconds` field, to make sure anyone who specifies it enters a number greater than 0. Then we can assume that a 0 value is the same as unset, and use the value for `days` instead.

Eraac: Sorry, I'm dumb ... furthermore I've update only the half of the reference to second ... 

Anyway, days stay a int and the new field `duration` accept a valid duration (in string). If duration is set use it, otherwise fallback on `days`

slevenick (changes requested):
Looks like validation fails:
```
1 error occurred:
  * resource google_bigtable_gc_policy: ExactlyOneOf: max_age configuration
block reference (max_age.0.days) can only be used with TypeList and MaxItems:
```

I think that the `ExactlyOneOf` needs to be on the `days` block rather than the `max_age` block

slevenick (approved):
Thanks! I've added a couple fixes on the upstream build, they should be included once that PR is merged

PR #7901, opened 2020-11-30
Add DaysSinceNoncurrentTime to the bucket life cycle rules

TimotheBarbe:
It's now possible to use DaysSinceNoncurrentTime in the bucket life cycle rules https://cloud.google.com/storage/docs/lifecycle#dayssincenoncurrenttime

> The DaysSinceNoncurrentTime condition is typically only used in conjunction with Object Versioning. The condition is satisfied when the specified number of days have passed since the object became noncurrent, either because the live version was deleted or replaced. For example, if an object became noncurrent at 2020/07/08 15:00 UTC and the DaysSinceNoncurrentTime condition is 10 days, then the condition is satisfied for the object on and after 2020/07/18 15:00 UTC.

### DIFF
--- a/google/resource_storage_bucket.go
+++ b/google/resource_storage_bucket.go
@@ -182,6 +182,11 @@ func resourceStorageBucket() *schema.Resource {
 										Optional:    true,
 										Description: `Relevant only for versioned objects. The number of newer versions of an object to satisfy this condition.`,
 									},
+                                    "days_since_noncurrent_time": {
+                                        Type:        schema.TypeInt,
+                                        Optional:    true,
+                                        Description: `Relevant only for versioned objects. The number of days since the object became noncurrent, either because the live version was deleted or replaced.`,
+                                    },
 								},
 							},
 							Description: `The Lifecycle Rule's condition configuration.`,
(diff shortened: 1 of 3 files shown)


TimotheBarbe:
Hey @c2thorn, it's my 1st PR here. I think the build is failing but I don't get why my PR broke it, could you give me pointers please ?

waiter-melon:
Up 👍  This would be very useful indeed!

c2thorn:
Hey @TimotheBarbe, it looks like our Travis build was failing before your PR, so this certainly did not introduce the failure. After a quick glance, do you mind adding the field to one of the storage bucket tests? It looks like https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_storage_bucket_test.go#L129 would be a good candidate.

TimotheBarbe:
Hey @c2thorn, thank you for the quick answer. I update the PR to add the new field inside the test, let me know if that's ok

TimotheBarbe:
No longer needed since https://github.com/hashicorp/terraform-provider-google/pull/7937

c2thorn:
@TimotheBarbe That's true, I'll close this.

PR #7963, opened 2020-12-07
Making location required in CloudRunServiceIamSchema

gpkc:
According to [the docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_service_iam#location), the `location` argument is supposed to be required. Indeed, if you don't provide one, the API calls fail like:

```
Error: Error retrieving IAM policy for cloudrun service "v1/.../locations//services/worker": Get "https://-run.googleapis.com/v1/projects/.../locations//services/worker:getIamPolicy?alt=json": dial tcp: lookup -run.googleapis.com: no such host
```

### DIFF
--- a/google/iam_cloud_run_service.go
+++ b/google/iam_cloud_run_service.go
@@ -30,8 +30,7 @@ var CloudRunServiceIamSchema = map[string]*schema.Schema{
 	},
 	"location": {
 		Type:     schema.TypeString,
-		Computed: true,
-		Optional: true,
+		Required: true,
 		ForceNew: true,
 	},
 	"service": {


ScottSuarez:
Its designed like this because we inherit this parameter from multiple locations if not supplied. I agree we can get a better error message/sooner propagated to the user. Tagging @rileykarson  to comment

gpkc:
@ScottSuarez Sorry if this is somehow obvious but can you provide some examples of where it inherits the parameter from?

rileykarson:
We draw the project from the provider configuration (either the `provider "google" block or one of several env vars) when it's not specified. Normally we'd return an error, but it looks like taht got changed because of an issue in imports: https://github.com/GoogleCloudPlatform/magic-modules/pull/3078/files

@slevenick: Do you happen to remember why we can't use `parseImportId`? It seems like we're using a function that's pretty close, [`getImportIdQualifiers`](https://github.com/hashicorp/terraform-provider-google/blame/master/google/iam_cloud_run_service.go#L118)

Assuming we can't use `parseImportId`, I wonder if we can move `getProject` and `getLocation` after the `getImportIdQualifiers` call, and only call them is `values["project"]` or `values["location`] (respectively) is the empty string.

slevenick:
> We draw the project from the provider configuration (either the `provider "google" block or one of several env vars) when it's not specified. Normally we'd return an error, but it looks like taht got changed because of an issue in imports: https://github.com/GoogleCloudPlatform/magic-modules/pull/3078/files
> 
> @slevenick: Do you happen to remember why we can't use `parseImportId`? It seems like we're using a function that's pretty close, [`getImportIdQualifiers`](https://github.com/hashicorp/terraform-provider-google/blame/master/google/iam_cloud_run_service.go#L118)
> 
> Assuming we can't use `parseImportId`, I wonder if we can move `getProject` and `getLocation` after the `getImportIdQualifiers` call, and only call them is `values["project"]` or `values["location`] (respectively) is the empty string.

Yeah, `parseImportId` sets values in state. That's ok normally for most resources, but in IAM we expect that we may get either a long or short name for a `{{name}}` field or similar.

So if a resource is identified by: `projects/{{project}}/resources/{{name}}` we want to handle either of the following configs:
```
{
  project = "my-project"
  name = "a-name"
}

{
  name = "projects/my-project/resources/a-name"
}
```

Calling `parseImportId` on the second of those would overwrite the `name` value in state to only be `a-name` rather than the full name

ScottSuarez:
@gpkc for reference on the default values being inherited from the provider block. I suggest checking out https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference

In this case for location I think we fall back to zone. I'll create an issue to potentially hand back a better error message to you.

ScottSuarez:
Created [issue](https://github.com/hashicorp/terraform-provider-google/issues/7993) to track this.. Closing this PR

PR #7982, opened 2020-12-10
Added v3 to module path

melinath:
This satisfies semantic versioning requirements of golang. See https://github.com/golang/go/wiki/Modules#semantic-import-versioning

This is a first step to being able to upgrade the version of the provider used by terraform-validator.

This is related to https://github.com/GoogleCloudPlatform/terraform-validator/issues/158


melinath:
The failure seems to be related to the website tests rather than this change.

c2thorn (approved):

Issue #7512, opened 2020-10-13
Add support for IAM authentication to google_sql_user


shanesavoie:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

This is a preview feature, but would love support for this in the future.

https://cloud.google.com/sql/docs/postgres/create-manage-iam-users?authuser=1#gcloud

It looks like there's a new flag --type=cloud_iam_user|cloud_iam_service_account used to specify between a normal user and IAM authenticated user.

### New or Affected Resource(s)

* google_sql_user

### Potential Terraform Configuration

```tf
# Propose what you think the configuration to take advantage of this feature should look like.
# We may not use it verbatim, but it's helpful in understanding your intent.
resource "google_sql_user" "user" {
  ...
  type="cloud_iam_user"
}

```

### References

* #0000

dabbertorres:
I'd be happy to take a crack at adding this feature (over in [magic-modules](https://github.com/GoogleCloudPlatform/magic-modules)?), unless of course someone is already working on it.

rileykarson:
@dabbertorres: Nobody is that I'm aware! You could edit in Magic Modules (there's a handwritten file @ https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/resources/resource_sql_user.go) but sending a PR in this repo is fine. One of us will upstream it to MM.

rileykarson:
Looks like the change downstream was missing a `Fixes:`. Closing.

PR #8017, opened 2020-12-15
add support for IAM authentication to google_sql_user for #7512

dabbertorres:
Support for the `type` field requested in #7512. Adds `type` as an optional field, with accepted values being unset/empty, or the values as defined in the [REST API reference](https://cloud.google.com/sql/docs/postgres/admin-api/rest/v1beta4/SqlUserType).

### DIFF
--- a/google/resource_sql_user.go
+++ b/google/resource_sql_user.go
@@ -53,10 +53,20 @@ func resourceSqlUser() *schema.Resource {
 			},
 
 			"password": {
-				Type:        schema.TypeString,
-				Optional:    true,
-				Sensitive:   true,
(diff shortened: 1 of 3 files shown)


slevenick (changes requested):
This looks very reasonable! Thanks for the addition

We will also need to add documentation for this field to the markdown file: https://github.com/hashicorp/terraform-provider-google/blob/master/website/docs/r/sql_user.html.markdown

Can you add the documentation over there and then I can get this merged in?

dabbertorres:
@slevenick, documentation added!

slevenick (approved):
Great, thanks!

PR #8030, opened 2020-12-16
Fix documentation link

joe-a-t:
The existing link only mentioned `SQL Server` version but the new link is a superset that lists the supported versions for all types of SQL databases. Also updated the list of supported values.


joe-a-t:
@ndmckinley I'm not sure what's going on with the test, it looks completely unrelated to my PR since I only touched 1 link which definitely exists

ndmckinley:
Awesome, thanks for the update - don't worry about the test, it's failing at `master`, frankly we don't know where the broken link is supposed to be either!

However, this is actually generated code.  Would you mind making this same change in github.com/GoogleCloudPlatform/magic-modules?  I think this one is in `third_party/terraform/website` there, but if not, you should find/replace the url in `products/sql/api.yaml`.

joe-a-t:
Changed via https://github.com/GoogleCloudPlatform/magic-modules/pull/4327, thanks!

PR #8036, opened 2020-12-16
Correct documentation for importing google_service_account_iam

joshk0:
iam.serviceAccountUser needs to be the full role expression, roles/iam.serviceAccountUser.


slevenick (approved):
Thanks!

PR #8082, opened 2020-12-24
Allow dependabot to check go dependencies

jlosito:
This change allows dependabot to check go dependencies which this project uses and automatically submit PRs with version bumps.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"


ndmckinley:
I don't have much of an opinion on that, but Cameron may?

c2thorn:
I'm not quite sure we want automated version bumps of our go dependencies. I'll check with the team

c2thorn:
Looks like this is something we want to try!

c2thorn (changes requested):
@jlosito Do you know if there is a way to modify the pull request message? I'd want to add a message to reviewers reminding them to keep https://github.com/hashicorp/terraform-provider-google-beta in sync as well.

jlosito:
I'm not sure too about that one. 

I just double-scanned the docs again. It looks like there's a way to change commit message, not what we're looking for. It doesn't look like there's a way to change the comment that the bot would leave when it creates the pull request.

I'd be willing to open up another PR against the beta provider. I also use that one on a daily basis. The idea being if both projects have the same config, I would imagine the bot would open pull requests against both of the projects at the same time.

jlosito:
https://github.com/hashicorp/terraform-provider-google-beta/pull/2832

jlosito:
Looks like someone is going to be busy will all of the PRs the bot just submitted. 😆 Have fun!

PR #8108, opened 2021-01-03
Correct import directions

AdrienneCohea:
I believe that the documentation might be incorrect, or the behavior of the API changed. It appears that `location` must be provided. I tried to import with `terraform import google_kms_crypto_key.worker_auth asgard-boundary/worker-auth` and encountered the error:

```
Error: Invalid CryptoKey id format, expecting `{projectId}/{locationId}/{KeyringName}/{cryptoKeyName}` or `{locationId}/{keyRingName}/{cryptoKeyName}, got id: asgard-boundary/worker-auth`
```

By prepending `location` according to the directions from the error, I achieved the desired import.

### DIFF
--- a/website/docs/r/kms_crypto_key.html.markdown
+++ b/website/docs/r/kms_crypto_key.html.markdown
@@ -167,9 +167,11 @@ This resource provides the following
 
 CryptoKey can be imported using any of these accepted formats:
 
+{projectId}/{locationId}/{KeyringName}/{cryptoKeyName}`
+
 ```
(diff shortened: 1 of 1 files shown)


AdrienneCohea:
I believe that the failing checks are not due to my change.

ndmckinley:
Thanks!  This is generated code - would you be interested in making this change in the generator - github.com/GoogleCloudPlatform/magic-modules - instead?  It's handled here: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/products/kms/terraform.yaml.  If not, can you open an issue in this repository and we'll handle it?

PR #8123, opened 2021-01-06
Use connection_name in Cloud Run Sql example

hectorj:
It gives the same result as the previous string, but is more robust because the project ID and the instance region are not hardcoded.

```release-note:none

```


rileykarson (approved):
LGTM. I've got to merge this to https://github.com/GoogleCloudPlatform/magic-modules so it applies to the beta provider as well, so I'll merge this once my upstreaming PR there is approved.

PR #8131, opened 2021-01-07
Add `temp_bucket` parameter for resource google_dataproc_cluster (fixes #7927)

### LINKED ISSUE #7927: Temp bucket parameter for google_dataproc_cluster
### Community Note

### Description

Dataproc cluster create api does have an option to provide --temp-bucket, would be nice to have this option via terraform
https://cloud.google.com/sdk/gcloud/reference/dataproc/clusters/create

User Case: Be able to provide a pre provisioned temp bucket to create a Dataproc cluster

* google_dataproc_cluster

bartcode:
Allow using `temp_bucket` when creating a Dataproc cluster as addressed in #7927 .

Closes #7927 .

If this PR is for Terraform, I acknowledge that I have:
* [x]  Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
* [x]  [Generated Terraform](https://github.com/GoogleCloudPlatform/magic-modules#generating-downstream-tools), and ran `make test` and `make lint` to ensure it passes unit and linter tests.
* [x]  Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/third_party/terraform/tests) (for handwritten resources or update tests).
* [x]  [Ran](https://github.com/hashicorp/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#tests) relevant acceptance tests (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
* [x]  Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
dataproc: Added field `temp_bucket` to `google_dataproc_cluster` cluster config.
```

### DIFF
--- a/google/resource_dataproc_cluster.go
+++ b/google/resource_dataproc_cluster.go
@@ -165,7 +165,7 @@ func resourceDataprocCluster() *schema.Resource {
 							Computed:     true,
 							AtLeastOneOf: clusterConfigKeys,
 							ForceNew:     true,
-							Description:  `The Cloud Storage temp bucket used to store ephemeral cluster and jobs data, such as Spark and MapReduce history files.. Note: If you don't explicitly specify a temp_bucket then GCP will auto create / assign one for you.`,
+							Description:  `The Cloud Storage temp bucket used to store ephemeral cluster and jobs data, such as Spark and MapReduce history files. Note: If you don't explicitly specify a temp_bucket then GCP will auto create / assign one for you.`,
 						},
 
 						"gce_cluster_config": {
--- a/google/resource_dataproc_cluster.go
+++ b/google/resource_dataproc_cluster.go
@@ -43,6 +43,7 @@ var (
 
 	clusterConfigKeys = []string{
 		"cluster_config.0.staging_bucket",
+		"cluster_config.0.temp_bucket",
 		"cluster_config.0.gce_cluster_config",
 		"cluster_config.0.master_config",
 		"cluster_config.0.worker_config",
@@ -156,6 +157,14 @@ func resourceDataprocCluster() *schema.Resource {
 							Type:        schema.TypeString,
 							Computed:    true,
 							Description: ` The name of the cloud storage bucket ultimately used to house the staging data for the cluster. If staging_bucket is specified, it will contain this value, otherwise it will be the auto generated name.`,
+						},
+
+						"temp_bucket": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							AtLeastOneOf: clusterConfigKeys,
+							ForceNew:     true,
+							Description:  `The Cloud Storage temp bucket used to store ephemeral cluster and jobs data, such as Spark and MapReduce history files.. Note: If you don't explicitly specify a temp_bucket then GCP will auto create / assign one for you.`,
 						},
 
 						"gce_cluster_config": {
@@ -779,6 +788,10 @@ func expandClusterConfig(d *schema.ResourceData, config *Config) (*dataproc.Clus
 		conf.ConfigBucket = v.(string)
 	}
 
+	if v, ok := d.GetOk("cluster_config.0.temp_bucket"); ok {
+		conf.TempBucket = v.(string)
+	}
+
 	c, err := expandGceClusterConfig(d, config)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


google/resource_dataproc_cluster_test.go
```diff
@@ -419,6 +419,37 @@ func TestAccDataprocCluster_withStagingBucket(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataprocCluster_withTempBucketAndCluster(clusterName, bucketName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDataprocClusterExists(t, "google_dataproc_cluster.with_bucket", &cluster),
+					resource.TestCheckResourceAttr("google_dataproc_cluster.with_bucket", "cluster_config.0.temp_bucket", bucketName)),
+			},
+			{
+				// Simulate destroy of cluster by removing it from definition,
+				// but leaving the storage bucket (should not be auto deleted)
```
melinath, line 443: Could you update this docstring to say "temp bucket"?

melinath (approved):
LGTM - seems pretty straightforward! I'll get started on upstreaming this.

google/resource_dataproc_cluster_test.go
bartcode: Yes, done!

google/resource_dataproc_cluster.go
```diff
@@ -158,6 +159,14 @@ func resourceDataprocCluster() *schema.Resource {
 							Description: ` The name of the cloud storage bucket ultimately used to house the staging data for the cluster. If staging_bucket is specified, it will contain this value, otherwise it will be the auto generated name.`,
 						},
 
+						"temp_bucket": {
+							Type:         schema.TypeString,
+							Optional:     true,
```
melinath, line 164:
```suggestion
							Optional:     true,
							Computed:     true,
```

melinath (changes requested):
It looks like this ended up breaking all the DataprocCluster tests that _don't_ define an explicit temp bucket, since the server then assigns one for you.

The fix for this should be to mark the field as Computed as well as optional. That should still allow setting an explicit value while also accepting whatever the server sends back if there's not a value defined.

bartcode:
Good find! I've updated it and also executed `TestAccDataprocCluster_singleNodeCluster`, `TestAccDataprocCluster_withStagingBucket`, and `TestAccDataprocCluster_withTempBucket` which all passed.

google/resource_dataproc_cluster.go
bartcode: Good find! I've updated it and also executed TestAccDataprocCluster_singleNodeCluster, TestAccDataprocCluster_withStagingBucket, and TestAccDataprocCluster_withTempBucket which all passed.

melinath: great, glad to hear they pass locally now! I've upstreamed this change & the CI tests are running.

melinath (approved):
LGTM and tests are passing

bartcode:
I noticed the comment in `magic-modules`, but it's been quiet for a while. Should I just wait or shall I remove the test that's suggested there? Your interpretation was correct, by the way, I copied some tests to ensure they also pass with this temporary bucket.

melinath:
@bartcode thanks for checking back in. I lost track of this, sorry. Let me ping Sam.

melinath (changes requested):
Okay, so there's just one thing to change - since this is a handwritten resource, the docs also need to be updated manually. They're in website/docs/r/. Thanks!

bartcode:
Thanks! I wasn't aware of those docs, but they should be in order now!

melinath (approved):

PR #8237, opened 2021-01-17
Add support for configuring disk resource policies for google_compute_instance_template.

### LINKED ISSUE #6719: Add support for configuring disk resource policies for google_compute_instance_template
### Community Note

### Description

Today, if managing disks with the `google_compute_disk` resource, you're able to set and manage resource policies (for things like scheduled snapshots, etc) on disks. When using instance templates (`google_compute_instance_template`) for example with a MIG, there is no ability to set the resource policies that should be attached to a disk when creating the template (consider stateful MIGs, where this would be incredibly helpful)

In the same way that you can attach resource policies to the disk resource, it would be great to be able to do the same with instance templates.

[instanceTemplates.insert](https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/insert) does seem to have support for this (`properties.disks[].initializeParams.resourcePolicies[]`), so it "should" just be a matter of surfacing the functionality in Terraform.

* google_compute_instance_template

```tf
resource "google_compute_instance_template" "template" {
  ..
  disk {
    source_image = ..
    resource_policies = [
      # similar to google_compute_disk, resource policy uris
    ]
  }
}
```

### References

* #5863 - discusses adding support for instance level resource policies (such as a placement policy) to the instance template, but does not specifically raise support for them on disks.

rayward:
Fixes #6719

Per documentation (and checked), https://pkg.go.dev/google.golang.org/api/compute/v0.beta#AttachedDiskInitializeParams only the resource policy _name_ is supported for instance templates, a uri can't be used. A link can be used in user code, but it will be converted to the resource name.

### DIFF
--- a/google/resource_compute_instance_template.go
+++ b/google/resource_compute_instance_template.go
@@ -208,6 +208,18 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 								},
 							},
 						},
+
+						"resource_policies": {
+							Type:        schema.TypeList,
+							Optional:    true,
+							ForceNew:    true,
+							MaxItems:    1,
+							Description: `A list of short names or self_links of resource policies to attach to this disk. Currently a max of 1 resource policy is supported.`,
+							Elem: &schema.Schema{
+								Type:             schema.TypeString,
+								DiffSuppressFunc: compareResourceNames,
+							},
+						},
 					},
 				},
 			},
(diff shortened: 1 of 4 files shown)


ScottSuarez:
can you add this field to the documentation? Possibly an example there-in too?

rayward:
@ScottSuarez I'm second guessing the decision to support self links on this field when the underlying API doesn't. Do you think I should keep this behaviour (of converting a link) or make it short names only, it's easy enough for a user to reference the name:

`resource_policies = [google_compute_resource_policy.my_snapshot_schedule.name]`

The `source` field for disks also doesn't support self-links for example - https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance_template#source

rayward:
I've (tentatively) documented the field, but will adjust it pending the above question.

ScottSuarez:
Hi, sorry for the delay on this. I was out of town. I spoke with my team on this and we prefer resources to reference each other by the `.id` property if possible. Its something we've consistently maintained within the provider.

rayward:
> I spoke with my team on this and we prefer resources to reference each other by the `.id` property if possible.

I think `.id` is effectively interchangeable for `.self_link` from what I can see. In any case the resource itself only supports the short name so I can keep the conversion code in and update the documentation and example to use `.id`.

ScottSuarez:
I'll focus on trying to merge this for you today! Sorry for the delay my friend :>

PR #8258, opened 2021-01-20
3.54.0 - release notes and cherry pick

ScottSuarez:
contains the release notes and cherry-pick of [c216b91568068600a59b99c375d42f9400d2e943](https://github.com/hashicorp/terraform-provider-google/commit/c216b91568068600a59b99c375d42f9400d2e943)

### DIFF
--- a/.changelog/4412.txt
+++ b/.changelog/4412.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+healthcare: add retry logic on healthcare dataset not initialized error
+```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,28 @@
 ## 3.54.0 (Unreleased)
+
+FEATURES:
+* **New Data Source:** `google_cloud_run_locations` ([#8192](https://github.com/hashicorp/terraform-provider-google/pull/8192))
+* **New Resource:** `google_privateca_certificate_authority` ([#8233](https://github.com/hashicorp/terraform-provider-google/pull/8233))
+* **New Resource:** google_privateca_certificate_authority_iam_binding ([#8249](https://github.com/hashicorp/terraform-provider-google/pull/8249))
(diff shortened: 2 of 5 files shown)


ScottSuarez:
hi @rileykarson, I'm going to be out of town for the remainder of the week but just wanted to make sure I got this out there. Could you shepherd this cherrypick and audit these notes before the release Monday?

There was a failing test on a newly added resource [TestAccPrivatecaCertificateAuthority_privatecaCertificateAuthorityCmekExample](https://ci-oss.hashicorp.engineering/project.html?tab=testDetails&testNameId=7482435357168603072&projectId=GoogleCloudBeta&buildTypeId=GoogleCloudBeta_ProviderGoogleCloudBetaGoogleProject) in tpgb. Could you take a look at it? I think we should be good to go. Since it's a new resource any flaws won't break existing users.

c2thorn (approved):

rileykarson (approved):

m1n0:
hi, ```bigquery: made incompatible changes to the `google_bigquery_table.schema` field to cause the resource to be recreated ``` seems like a quite disruptive change for a small version bump. I see that 54 was unreleased but I could not find any info on that, what happened here please?

rileykarson:
We treated this as a compatible change, eg one that we can release in a minor release. The change should cause Terraform to attempt to recreate the resource in cases where the API would have returned an error indicating that changing from one schema to another was impossible. However the interaction with some other behaviour(s) on the field caused a larger impact than expected, and we'll revert for the next release. See https://github.com/hashicorp/terraform-provider-google/issues/8282

We plan to reintroduce the change during the `3.X` release series at this time, although feedback is appreciated. https://github.com/hashicorp/terraform-provider-google/issues/1144 is a good place for it, I think.

maroshmka:
hello @rileykarson, I've wrote feedback in https://github.com/hashicorp/terraform-provider-google/issues/1144#issuecomment-768930080, but I see issue is closed, so pinging here :) thanks!

PR #8267, opened 2021-01-21
docs: fix incorrect argument name in time_rotating example

joemiller:
https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/rotating#argument-reference

fixes an error when using the documented example:

```
Error: Unsupported argument

  on main.tf line 17, in resource "time_rotating" "rotation":
  17:   rotate_days = 30

An argument named "rotate_days" is not expected here.
```


ScottSuarez:
upstreaming this pr https://github.com/GoogleCloudPlatform/magic-modules/pull/4437

PR #8301, opened 2021-01-26
Add 'ignore_deleted' to service account key resource

### LINKED ISSUE #8300: Add argument to ignore deleted service account keys
### Community Note

### Description

I have a `google_service_account_key` resource that I create as a part of bootstrapping a Terraform Cloud workspace and Google Cloud Platform project. An automated system later rotates that service account key on some cadence, which involves deleting the service account key and replacing the contents of the Terraform Cloud workspace variable.

Fortunately, once this happens the next time `terraform plan` is run in the workspace that contains the `google_service_account_resource`, [this piece of code](https://github.com/hashicorp/terraform-provider-google/blob/4373d10c86d06498cb2297bbe4c1ddf0a1309140/google/resource_google_service_account_key.go#L160-L167) catches that the key is missing and doesn't raise an error.

Unfortunately for my use case, this means that it tries to create the service account key again, which, while I can workaround in the rest of the Terraform configuration (with an `ignore_changes` block), is not ideal because I'd rather not have valid service account key material stay in my Terraform state. It also causes unneeded visual noise in the plan.

**tl;dr:** would it be possible to add an argument to the `google_service_account_resource` to have it ignore a missing key? I would expect the resource to behave similarly to how it does with the `private_key` field, which is no longer returned in the `Projects.ServiceAccounts.Keys.Get` call - it could continue returning the historical values of the key, but with the understanding that when deleted, the private key will no longer be valid.

* google_service_account_key

```tf
resource "google_service_account_key" "key" {
  service_account_id = google_service_account.account.id

  # not tied to this argument name
  ignore_deleted = true
}
```

### Note

If the maintainers are okay with this approach, I would be more than happy to attempt a PR.

ericnorris:
Potentially closes #8300, was easy enough to try myself. Tested locally and works as expected. Open to feedback!

### DIFF
--- a/google/resource_google_service_account_key.go
+++ b/google/resource_google_service_account_key.go
@@ -59,6 +59,13 @@ func resourceGoogleServiceAccountKey() *schema.Resource {
 				Optional:    true,
 				ForceNew:    true,
 			},
+			"ignore_deleted": {
+				Type:        schema.TypeBool,
+				Default:     false,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "If true, then this resource will not be recreated upon deletion by an external entity. Attributes (such as private_key) will be invalid once deleted.",
+			},
 			// Computed
 			"name": {
 				Type:        schema.TypeString,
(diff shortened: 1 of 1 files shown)


SarahFrench:
Closing this PR as stale, and [the linked issue](https://github.com/hashicorp/terraform-provider-google/issues/8300) is closed

PR #8305, opened 2021-01-26
Update required role bindings on KMS key

pmansour:
Per the [official docs](https://cloud.google.com/certificate-authority-service/docs/reference/permissions-and-roles#service_agent), the CAS service agent needs two role bindings on a CMEK. The current doc only includes one of these (`roles/cloudkms.signerVerifier`), so this change just adds the second one (`roles/viewer`).

### DIFF
--- a/website/docs/r/privateca_certificate_authority.html.markdown
+++ b/website/docs/r/privateca_certificate_authority.html.markdown
@@ -134,10 +134,20 @@ resource "google_project_service_identity" "privateca_sa" {
   service  = "privateca.googleapis.com"
 }
 
-resource "google_kms_crypto_key_iam_binding" "privateca_sa_keyuser" {
+resource "google_kms_crypto_key_iam_binding" "privateca_sa_keyuser_signerverifier" {
   provider      = google-beta
   crypto_key_id = "projects/keys-project/locations/us-central1/keyRings/key-ring/cryptoKeys/crypto-key"
   role          = "roles/cloudkms.signerVerifier"
+
+  members = [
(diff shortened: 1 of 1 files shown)


website/docs/r/privateca_certificate_authority.html.markdown
```diff
@@ -134,7 +134,7 @@ resource "google_project_service_identity" "privateca_sa" {
   service  = "privateca.googleapis.com"
 }
 
-resource "google_kms_crypto_key_iam_binding" "privateca_sa_keyuser" {
+resource "google_kms_crypto_key_iam_binding" "privateca_sa_keyuser_signerverifier" {
```
drebes, line 137: If you change the TF resource name, you also need to change the depends_on below.

website/docs/r/privateca_certificate_authority.html.markdown
```diff
@@ -144,6 +144,16 @@ resource "google_kms_crypto_key_iam_binding" "privateca_sa_keyuser" {
   ]
 }
 
+resource "google_kms_crypto_key_iam_binding" "privateca_sa_keyuser_viewer" {
```
drebes, line 147: Please also add this to the depends_on below.

website/docs/r/privateca_certificate_authority.html.markdown
drebes: I see its fixed on the upstream PR, so please ignore this. :)

website/docs/r/privateca_certificate_authority.html.markdown
drebes: Likewise.

PR #8309, opened 2021-01-27
Fix time_rotating argument name

nicosingh:
One of the example definitions shows `rotate_days` as argument of `time_rotating` resource.

This argument should be called `rotation_days`, as defined at the current [resource documentation](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/rotating#rotation_days) (as of today using version `0.6.0`).


slevenick (approved):
Thanks!

PR #8349, opened 2021-01-30
data.google_compute_network: subnetworks_self_links to set

stephengroat:
since all `subnetworks_self_links` should be unique, shouldn't this be a `set` instead of a `list`?

would allow looping through these outputs without conversion

### DIFF
--- a/google/data_source_google_compute_network.go
+++ b/google/data_source_google_compute_network.go
@@ -37,7 +37,7 @@ func dataSourceGoogleComputeNetwork() *schema.Resource {
 			},
 
 			"subnetworks_self_links": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Computed: true,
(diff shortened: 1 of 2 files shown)


melinath (approved):
This seems plausible to me.

melinath (changes requested):
So, it looks like changing to a TypeSet would probably break the attribute for some users. For example, someone probably references `subnetworks_self_links.0` and expects it to be the first value in the list. If this were a typeset instead, it could change which link is considered "first", if it's possible to reference by index at all.

Is there a particular use case that you have that requires changing this to a set? And/or is there a case to be made that referencing subnetworks by index is definitely not a useful operation?

stephengroat:
I was getting the list back in different orders, causing me an issue when looping in an order-based list (so i converted i to a set)

I was using an older version of the provider, assuming that the behavior hadn't changed. seems to have been fixed in a newer version. thanks!

Issue #8000, opened 2020-12-11
github checks failing during `make website`


melinath:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Debug Output

Github checks for terraform-provider-google and terraform-provider-google-beta are currently failing for the `make website` step. See https://travis-ci.org/github/hashicorp/terraform-provider-google/builds/748916234

rileykarson:
Likely handled by doing https://github.com/hashicorp/terraform-provider-google/issues/8001

melinath:
It looks like this is no longer happening - specifically, it looks like the github actions tests are not running `make website`.

PR #8353, opened 2021-02-01
Fix broken directory in make docscheck

upodroid:
Fixes: #8000

GH Actions is already enabled and running but there is directory being checked in `make docscheck` which doesn't exist anymore so it has been removed.


melinath (changes requested):
Hi upodroid - this LGTM; it doesn't seem like this causes `make docscheck` to fail but it's always nice to remove extraneous error messages. Thanks!

However, it looks like a change to the Github Issue Template snuck in there; could you remove that change? Thanks!

upodroid:
The shell script isn't running with -e flag.

The issue template change is on purpose. The line I removed is something that isn't done anymore such as hashibot assigning issues.

melinath (approved):
Confirmed that the language in the issue template is good.

PR #8378, opened 2021-02-03
Add changelog for 3.56.0 release

melinath:


CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 ## 3.56.0 (Unreleased)
 
+NOTES:
+* billingBudget: doc updated for `google_billing_budget` ([#8316](https://github.com/hashicorp/terraform-provider-google/pull/8316))
```
c2thorn, line 4: I don't think this is worth adding to the changelog, unless it was going to call out something important that was added to the documentation. My bad for not catching it
```suggestion
```

c2thorn (changes requested):
I think at least one PR that shows up in the google-beta logs is missing here:
https://github.com/hashicorp/terraform-provider-google/pull/8377

c2thorn:
Or rather: it should be taken out of the google-beta release notes? I'll address it in the other PR

c2thorn (approved):

melinath:
yeah I accidentally generated the beta changelog on master at first & missed that one in the cleanup - thanks for catching!

melinath:
Fixed in the other PR

PR #8498, opened 2021-02-18
Update dns_record_set.html.markdown

joshk0:
If you don't add this space, your plans will always see a space between the records on the remote side of things, and will always have a non-empty plan for such records.

```release-note:none

```


c2thorn (approved):
Thanks!

PR #8540, opened 2021-02-24
Bigtable GC policy: force new

### LINKED ISSUE #8539: deprecation warning for bigtable gc policies creates impossible situation
### Community Note

* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

Hello
We started getting deprecation warnings like this:
```
for bt gc policies, we got the following error:
Warning: Deprecated Attribute
  on bigtable.tf line 34, in resource "google_bigtable_gc_policy" "foo":
  34:     days = 3
Deprecated in favor of duration
(and 19 more similar warnings elsewhere)
```

That's fine, so Ok, we change 3d to, say, 72h. At that point, though, terraform apply returns the error
`Error: doesn't support update` if we try to a change like this
```diff
  ~ resource "google_bigtable_gc_policy" "foo" {
        id            = "(age() > 3d || versions() > 1)"
        # (5 unchanged attributes hidden)

      ~ max_age {
          - days     = 3 -> null
          + duration = "72h"
        }

        # (1 unchanged block hidden)
    }
```

We can remove the state and try to reimport it, however, then we find out that this is not an importable resource
```
Error: resource google_bigtable_gc_policy doesn't support import
```

It may be safe to simply remove the state, reapply everything, and not worry about dupe resources being left behind, but I don't know enough about bt garbage-collection policies to know if this is safe / sane. In either event, the current experience is frustrating.

### Terraform Version
```
Terraform v0.14.7
+ provider registry.terraform.io/hashicorp/google v3.56.0
+ provider registry.terraform.io/hashicorp/google-beta v3.56.0
```
(tried with 3.57.0 as welll)

### Affected Resource(s)
* google_bigtable_gc_policy

```tf
resource "google_bigtable_gc_policy" "foo" {
  instance_name = module....
  table         = module....
  column_family = "data"

  max_age {
    duration = "72h"
  }
}
```
### Debug Output
can provide if needed

### Expected Behavior
When changing from, say, 30d to 720h for `duration`, we'd expect that the change will be a noop or state only change

### Actual Behavior
Terraform gives the error
```
Error: doesn't support update
```

### Steps to Reproduce

1. `terraform apply`

Eraac:
If you update `max_version.number`, `max_age.duration` or `max_age.days` terraform try to update, but it's not supported by this resource

Fixes #8539

### DIFF
--- a/google/resource_bigtable_gc_policy.go
+++ b/google/resource_bigtable_gc_policy.go
@@ -64,6 +64,7 @@ func resourceBigtableGCPolicy() *schema.Resource {
 						"days": {
 							Type:         schema.TypeInt,
 							Optional:     true,
+							ForceNew:     true,
 							Deprecated:   "Deprecated in favor of duration",
 							Description:  `Number of days before applying GC policy.`,
 							ExactlyOneOf: []string{"max_age.0.days", "max_age.0.duration"},
@@ -71,6 +72,7 @@ func resourceBigtableGCPolicy() *schema.Resource {
 						"duration": {
 							Type:         schema.TypeString,
 							Optional:     true,
+							ForceNew:     true,
 							Description:  `Duration before applying GC policy`,
 							ValidateFunc: validateDuration(),
 							ExactlyOneOf: []string{"max_age.0.days", "max_age.0.duration"},
@@ -89,6 +91,7 @@ func resourceBigtableGCPolicy() *schema.Resource {
 						"number": {
 							Type:        schema.TypeInt,
 							Required:    true,
+							ForceNew:    true,
 							Description: `Number of version before applying the GC policy.`,
 						},
 					},


slevenick:
I'm not sure this is the correct way to handle this. It may be more appropriate as a CustomizeDiff function that forces recreation if the user moves from specifying the deprecated format to the new format. As I understand it we should be able to update the new format in place

slevenick:
I think this will be fixed by https://github.com/GoogleCloudPlatform/magic-modules/pull/4533

And I was wrong, this looks like the right way to handle this

PR #8632, opened 2021-03-09
Add support for GKE Autopilot in `google_container_cluster` resource

paulwilljones:
Adds support for GKE Autopilot to `google_container_cluster` resource.

Due to the [defaults](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview#comparison) set in GKE Autopilot, some features of GKE aren't available when in Autopilot mode.


google/resource_container_cluster.go
```diff
@@ -1079,43 +1086,52 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
...
 			DatapathProvider:          d.Get("datapath_provider").(string),
-		},
-		MasterAuth:     expandMasterAuth(d.Get("master_auth")),
-		ResourceLabels: expandStringMap(d, "resource_labels"),
+		}
+		cluster.ReleaseChannel = expandReleaseChannel(d.Get("release_channel"))
+		cluster.EnableTpu = d.Get("enable_tpu").(bool)
+		cluster.MasterAuth = expandMasterAuth(d.Get("master_auth"))
+		cluster.ResourceLabels = expandStringMap(d, "resource_labels")
 	}
```
ScottSuarez, line 1135: Its seems a little off to ignore half the parameters if gke auto pilot is enabled. This would result in a continuous diff if the user does define those parameters in the terraform configuration. Instead of refactoring this piece can you add look into utilizing the `ConflictsWith` keyword on the resource schema ?

ScottSuarez: adding @c2thorn who is more familiar with auto_pilot

google/resource_container_cluster.go
```diff
@@ -1079,43 +1086,52 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
...
+
+	if d.Get("enable_autopilot").(bool) {
+		cluster.Autopilot = &containerBeta.Autopilot{
+			Enabled:         d.Get("enable_autopilot").(bool),
+			ForceSendFields: []string{"Enabled"},
+		}
+	} else {
+		cluster.InitialNodeCount = int64(d.Get("initial_node_count").(int))
+		cluster.MaintenancePolicy = expandMaintenancePolicy(d, meta)
+		cluster.InitialClusterVersion = d.Get("min_master_version").(string)
```
c2thorn, line 1108: I believe this field is actually settable for autopilot clusters

google/resource_container_cluster.go
```diff
@@ -1079,43 +1086,52 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
...
+		}
+		cluster.NetworkConfig = &containerBeta.NetworkConfig{
 			EnableIntraNodeVisibility: d.Get("enable_intranode_visibility").(bool),
 			DefaultSnatStatus:         expandDefaultSnatStatus(d.Get("default_snat_status")),
 			DatapathProvider:          d.Get("datapath_provider").(string),
-		},
-		MasterAuth:     expandMasterAuth(d.Get("master_auth")),
-		ResourceLabels: expandStringMap(d, "resource_labels"),
+		}
+		cluster.ReleaseChannel = expandReleaseChannel(d.Get("release_channel"))
```
c2thorn, line 1131: Same with `release_channel`, it should be settable with autopilot

google/resource_container_cluster.go
```diff
@@ -370,6 +370,13 @@ func resourceContainerCluster() *schema.Resource {
 				Description: `Enable Shielded Nodes features on all nodes in this cluster. Defaults to false.`,
 			},
 
+			"enable_autopilot": {
+				Type:        schema.TypeBool,
+				Optional:    true,
```
c2thorn, line 375: It does not appear that the autopilot flag is updatable, so we should add `ForceNew: true` here to force recreate the cluster on the field update.

c2thorn (changes requested):
@paulwilljones Thanks for starting this process. With the amount of differences between autopilot and standard GKE, this is not a straightforward change. We want to be cautious and ensure that standard GKE UX is not disrupted.

In addition to the comments already suggested, I would request more extensive testing:
- A test config with all available fields set
- A test checking the error for autopilot on a zonal cluster (similar to https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_container_cluster_test.go#L1468)

google/resource_container_cluster.go
c2thorn: +1 to `ConflictsWith` on incompatible fields. Example here: https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_container_cluster.go#L326

paulwilljones:
> @paulwilljones Thanks for starting this process. With the amount of differences between autopilot and standard GKE, this is not a straightforward change. We want to be cautious and ensure that standard GKE UX is not disrupted.
> 
> In addition to the comments already suggested, I would request more extensive testing:
> 
> * A test config with all available fields set
> * A test checking the error for autopilot on a zonal cluster (similar to https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_container_cluster_test.go#L1468)

Thanks both. I'll take a look at those changes.
One thing I've noticed is certain features will need enabling when `enable_autopilot: true`, such as shielded nodes in order for the API to accept the request:
```
Error: googleapi: Error 400: Autopilot clusters requires shielded nodes to be enabled., badRequest                                            
```

c2thorn:
> One thing I've noticed is certain features will need enabling when `enable_autopilot: true`, such as shielded nodes in order for the API to accept the request:
> 
> ```
> Error: googleapi: Error 400: Autopilot clusters requires shielded nodes to be enabled., badRequest                                            
> ```
Omitting shielded nodes from the request altogether works correct?

glerchundi:
This is absolutely fantastic! Cannot wait to see this merged and ready to be consumed :D

c2thorn (approved):
LGTM, thanks for the changes!

I'll work on getting this upstreamed to https://github.com/GoogleCloudPlatform/magic-modules where there'll be more testing and another review.

cmcga1125:
not trying to be pesky - any idea when this will hit the provider?

c2thorn:
@cmcga1125 My estimate is version `v3.62` scheduled for March 29th. Running the full test suite for `google_container_cluster` showed that this implementation introduces new test failures that need to be addressed before releasing. The cutoff for the `v3.61` release was last night, so this looks like it will be released in `v3.62`.

c2thorn:
@paulwilljones If you are interested in taking a stab at the failing tests, I have found there are a few types of failures:

- `enable_shielded_nodes` shows a diff after applying - seen in `TestAccContainerCluster_withMasterAuthConfig`
- `enable_autopilot` shows a diff after importing - seen in `TestAccContainerCluster_withReleaseChannelEnabled`
- A container cluster datasource tests are also failing: see `TestAccContainerClusterDatasource_zonal`

Otherwise, I plan to address them in https://github.com/GoogleCloudPlatform/magic-modules/pull/4591

paulwilljones:
> @paulwilljones If you are interested in taking a stab at the failing tests, I have found there are a few types of failures:
> 
> * `enable_shielded_nodes` shows a diff after applying - seen in `TestAccContainerCluster_withMasterAuthConfig`
> * `enable_autopilot` shows a diff after importing - seen in `TestAccContainerCluster_withReleaseChannelEnabled`
> * A container cluster datasource tests are also failing: see `TestAccContainerClusterDatasource_zonal`
> 
> Otherwise, I plan to address them in [GoogleCloudPlatform/magic-modules#4591](https://github.com/GoogleCloudPlatform/magic-modules/pull/4591)

Address in https://github.com/hashicorp/terraform-provider-google/pull/8632/commits/8ea195f0af66ecb71632c923c5a3597e1f6dbaa8

google/resource_container_cluster.go
```diff
@@ -364,10 +367,19 @@ func resourceContainerCluster() *schema.Resource {
 			},
 
 			"enable_shielded_nodes": {
+				Type:          schema.TypeBool,
+				Optional:      true,
+				Default:       true,
```
paulwilljones, line 372: @c2thorn Enabling Shielded Nodes by default is in alignment with GKE 1.18.

I'm seeing test failures (for example `TestAccContainerCluster_withMasterAuthConfig`) when shielded nodes is `false`:

```
# google_container_cluster.with_master_auth_no_cert will be updated in-place
          ~ resource "google_container_cluster" "with_master_auth_no_cert" {
              ~ enable_shielded_nodes       = true -> false
```

c2thorn:
@paulwilljones I've made a few changes to the upstream PR in https://github.com/GoogleCloudPlatform/magic-modules/pull/4591 and believe I have a working iteration. Syncing this PR isn't required, since the upstream PR will cover the differences.

PR #8689, opened 2021-03-15
Allow MYSQL_5_5 for database_version

EvyBongers:
Per the [GCP documentation](https://cloud.google.com/sql/docs/mysql/replication/replication-from-external#setup-source-instance), MYSQL_5_5 is a valid value for source representation instance. As such, terraform-provider-google should support this as well. 

This is particularly useful when migrating from a legacy instance (hosted elsewhere) to a Cloud SQL instance of a newer version.

### DIFF
--- a/website/docs/r/sql_source_representation_instance.html.markdown
+++ b/website/docs/r/sql_source_representation_instance.html.markdown
@@ -61,7 +61,7 @@ The following arguments are supported:
 * `database_version` -
   (Required)
   The MySQL version running on your source database server.
-  Possible values are `MYSQL_5_6`, `MYSQL_5_7`, and `MYSQL_8_0`.
+  Possible values are `MYSQL_5_5`, `MYSQL_5_6`, `MYSQL_5_7`, and `MYSQL_8_0`.
 
 * `host` -
   (Required)
--- a/website/docs/r/sql_source_representation_instance.html.markdown
+++ b/website/docs/r/sql_source_representation_instance.html.markdown
@@ -61,7 +61,7 @@ The following arguments are supported:
 * `database_version` -
   (Required)
   The MySQL version running on your source database server.
-  Possible values are `MYSQL_5_5`, `MYSQL_5_6`, `MYSQL_5_7`, and `MYSQL_8_0`.
+  Possible values include the following `MYSQL_5_5`, `MYSQL_5_6`, `MYSQL_5_7`, and `MYSQL_8_0`.
 
 * `host` -
   (Required)
--- a/google/resource_sql_source_representation_instance.go
+++ b/google/resource_sql_source_representation_instance.go
@@ -47,8 +47,8 @@ func resourceSQLSourceRepresentationInstance() *schema.Resource {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     true,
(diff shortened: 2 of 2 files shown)


EvyBongers:
@rileykarson Can you give an estimate when you will be able to look at this? This is currently a blocking issue for us.

rileykarson (commented):
Sorry for the delay in getting to this! Can you make the change here: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/resources/resource_sql_database_instance.go.erb

Additionally, want to just remove the `ValidateFunc` altogether? These values change frequently enough, let's just stop checking.

website/docs/r/sql_source_representation_instance.html.markdown
```diff
@@ -61,7 +61,7 @@ The following arguments are supported:
 * `database_version` -
   (Required)
   The MySQL version running on your source database server.
-  Possible values are `MYSQL_5_6`, `MYSQL_5_7`, and `MYSQL_8_0`.
+  Possible values are `MYSQL_5_5`, `MYSQL_5_6`, `MYSQL_5_7`, and `MYSQL_8_0`.
```
rileykarson, line 64: w/ the other suggested change, we could consider this:
```suggestion
  Possible values include the following `MYSQL_5_5`, `MYSQL_5_6`, `MYSQL_5_7`, and `MYSQL_8_0`.
```

EvyBongers:
> Sorry for the delay in getting to this! Can you make the change here: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/resources/resource_sql_database_instance.go.erb
> 
> Additionally, want to just remove the `ValidateFunc` altogether? These values change frequently enough, let's just stop checking.

Either would be fine for me, though I feel a descriptive error message is important. Should this validation be removed, what kind of error message would be returned to the user when the API doesn't support the value entered?

EvyBongers:
> > Sorry for the delay in getting to this! Can you make the change here: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/resources/resource_sql_database_instance.go.erb
> 
> Sure! Should the list be updated here like in this PR, or should the validation be removed altogether, should that be the decision here?

On second thought, this looks like it's applicable to the Cloud SQL instance, rather than the SQL source representation. Unless I'm mistaken, this should remain as is, as MySQL 5.5 is not supported by Cloud SQL.

rileykarson:
Sorry! Wrong place on my part.

The change you'll want to make is here: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/products/sql/api.yaml#L702-L710

I think you can change it to the following to make an equivalent change:

```
      - !ruby/object:Api::Type::String
        name: 'databaseVersion'
        description: |
          The MySQL version running on your source database server. Possible values include the following `MYSQL_5_5`, `MYSQL_5_6`, `MYSQL_5_7`, and `MYSQL_8_0`.
        required: true
```

EvyBongers:
Hi @rileykarson,

Considering #8806, it seems this PR can be closed now?

rileykarson:
Thanks! Yep

PR #8698, opened 2021-03-16
Fix broken kalm_config link

newtorob:
Going through the documentation for the container cluster resource, there is a broken link under the `kalm_config` section of the documentation. There seems to be an extra space in the link which breaks the url.


rileykarson (approved):

rileykarson:
Merging w/ https://github.com/GoogleCloudPlatform/magic-modules/pull/4610, so that it applies to the google-beta repo as well.

PR #8734, opened 2021-03-19
Addressing issue 8732

ThomasAntoniusCrowley:
[Issue](https://github.com/hashicorp/terraform-provider-google/issues/8732)  was caused by a mistake in the documentation which this PR will fix


rileykarson:
Unusual! The API actually the exact configuration shown in the example for us. Can you make the change against https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/templates/terraform/examples/region_target_http_proxy_basic.tf.erb? That'll update a test as well.

PR #8794, opened 2021-03-30
fix: Adding a triggers block for alert policy to force recreation - u…

abdrehma:
…seful for external dependency e.g. notification channels

### DIFF
--- a/google/resource_monitoring_alert_policy_test.go
+++ b/google/resource_monitoring_alert_policy_test.go
@@ -43,7 +43,7 @@ func testAccMonitoringAlertPolicy_basic(t *testing.T) {
 		CheckDestroy: testAccCheckAlertPolicyDestroyProducer(t),
 		Steps: []resource.TestStep{
 			{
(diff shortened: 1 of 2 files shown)


abdrehma:
Hi, this PR is raised as an inspiration from https://github.com/adamdecaf/terraform-provider-aws/commit/cb8ec6564e98e3039fca96d1450df53f3674e198 

Currently a terraform failure occurs with alertpolicies when a notification channel requires destruction - the notification channel destruction is blocked by the alertpolicy. After some quick googling I found there isn't a Terraform provided solution as of yet, so a triggers block is best to suffice for now.

abdrehma:
Just noticed magic modules generation, will fix this PR and raise PR there shortly

abdrehma:
https://github.com/GoogleCloudPlatform/magic-modules/pull/4633

added triggers schema entry to magic modules

abdrehma:
ah just thinking logically and testing locally, this PR wont work, whilst force recreation will work, it will still error out on the other resource that is removing it, this is a terraform limitation, but, i think i can experiment with local-exec when=destroy to modify the notification channels as Destroy provisioners are run before the resource is destroyed

PR #8825, opened 2021-04-01
Add E2_HIGHCPU_8 and E2_HIGHCPU_32 MachineType validations

thedmeyer:
New machine types supported by Cloud Build. Ref: https://cloud.google.com/build/docs/api/reference/rest/v1/projects.builds#machinetype


slevenick (approved):
Thanks! I've included this in our code generator upstream so it will be preserved

PR #8848, opened 2021-04-06
fix(docs): broken markdown links in compute_instance

aslafy-z:
Fix broken markdown links in `compute_instance` documentation.


rileykarson (changes requested):
Do you mind making these changes against https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/website/docs/r/compute_instance.html.markdown instead? That'll apply the changes to `google-beta` as well as update this file. Thanks!

PR #8911, opened 2021-04-14
Added labels to issue templates

melinath:
GitHub now supports labels being declared in the issue templates.


rileykarson (commented):
You can remove line 7 / 8 (original / revised) in bug+enhancement's templates now, I think

melinath:
ah, that's the only automation they were helping? 🔥

rileykarson (approved):

PR #9004, opened 2021-04-27
add Android and iOS to supported os_type values

keeleysam:
these are supported per https://cloud.google.com/access-context-manager/docs/reference/rest/Shared.Types/OsType


keeleysam:
Closing in favor of https://github.com/GoogleCloudPlatform/magic-modules/pull/4736

Issue #9051, opened 2021-05-02
google_compute_instance_template missing disk_labels at documentation


gcaracuel:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

[An old PR](https://github.com/hashicorp/terraform-provider-google/pull/3766) to google_compute_instance_template added labels to disk section but it is not yet added to documentation, found it by checking old PRs

### New or Affected Resource(s)

* google_compute_instance_template disk section

### Potential Terraform Configuration

```tf
  disk {
    // Instance Templates reference disks by name, not self link
    source      = google_compute_disk.foobar.name
    auto_delete = false
    boot        = false
    labels.     = ['foo']
  }
```

### References

* #2026

gcaracuel:
Fixed by ( #9052 )

upodroid:
@rileykarson This has been fixed. OP forgot to use fixed by in the pull request.

Issue #2026, opened 2018-09-11
Add disk labels to google_compute_instance_template


andyhky:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Terraform Version

Terraform v0.11.7
    provider.google v1.17.1

### Description

Requesting that google_compute_instance_template gets supports for disk labels. Support is present in the Compute API  already.

### New or Affected Resource(s)

* google_compute_instance_template

### Potential Terraform Configuration
https://www.terraform.io/docs/providers/google/r/compute_instance_template.html

```tf
# Propose what you think the configuration to take advantage of this feature should look like.
# We may not use it verbatim, but it's helpful in understanding your intent.
resource "google_compute_instance_template" "foo" {
  name_prefix        = "foo"
  ###....
  disk { 
  	boot = true
  	auto_delete = true
  	labels = {
	    cost-center = "marketing",
	    env = "production"
	}
  }
###...
}

```

### References

* https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/insert
(properties.disks[].initializeParams.labels)

marcboudreau:
I've created a PR (#3766) to address this issue.

chrisst:
fixed by https://github.com/terraform-providers/terraform-provider-google/pull/3766

PR #9052, opened 2021-05-02
Adds labels to disk on instance templates

gcaracuel:
Created to address #9051 

In an old issue #2026 labels was added to disk on an instance template. 

```release-note:none

```


melinath (approved):
LGTM

PR #9061, opened 2021-05-03
google_storage_bucket_object_content fix visible spelling

m1:
As title says, fixes spelling for visible!


rileykarson (commented):
Thanks! Mind changing this in https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/website/docs/d/storage_bucket_object_content.html.markdown? That'll apply it to the `google-beta` provider repo as well (which is the "real" source of our documentation)

PR #9083, opened 2021-05-05
add details to google_service_account_key documentation

Sayrus:
I do think this is actually important because if you use `account_id` to reference the account and the provider has no specified project, `terraform plan` will return no error while `terraform apply` will complain about a missing `project` field.

The issue is that, this pass the **plan** but not the **apply**:
```
provider "google" {
  version = "3.64.0"
}
resource "google_service_account" "account" {
  account_id   = "account"
  project      = local.project_id
}
resource "google_service_account_key" "account_key" {
  service_account_id = google_service_account.account.account_id
}
```
```
Error: project: required field is not set
```

While this works:
```
provider "google" {
  version = "3.64.0"
}
resource "google_service_account" "account" {
  account_id   = "account"
  project      = local.project_id
}
resource "google_service_account_key" "account_key" {
  service_account_id = google_service_account.account.id
}
```

As the `project` field is not available on this resource, this can be very confusing. To match the behavior discussed here (https://github.com/hashicorp/terraform-provider-google/issues/7170), it might make sense to always get the `project` from the `google_service_account` instead, even if `account_id` is used.

### DIFF
--- a/google/resource_google_service_account_key.go
+++ b/google/resource_google_service_account_key.go
@@ -21,7 +21,7 @@ func resourceGoogleServiceAccountKey() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 				ForceNew:    true,
-				Description: `The ID of the parent service account of the key. This can be a string in the format {ACCOUNT} or projects/{PROJECT_ID}/serviceAccounts/{ACCOUNT}, where {ACCOUNT} is the email address or unique id of the service account. If the {ACCOUNT} syntax is used, the project will be inferred from the account.`,
+				Description: `The ID of the parent service account of the key. This can be a string in the format {ACCOUNT} or projects/{PROJECT_ID}/serviceAccounts/{ACCOUNT}, where {ACCOUNT} is the email address or unique id of the service account. If the {ACCOUNT} syntax is used, the project will be inferred from the provider's configuration.`,
 			},
 			// Optional
 			"key_algorithm": {


melinath (approved):
@Sayrus I'm not sure I understand the example HCL you posted. Both samples should fail because `google_service_account.account_key` doesn't exist. According to the [docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account_key), the expected way to use that field would be:

```
resource "google_service_account" "myaccount" {
  account_id   = "myaccount"
  display_name = "My Service Account"
}

resource "google_service_account_key" "mykey" {
  service_account_id = google_service_account.myaccount.name
  public_key_type    = "TYPE_X509_PEM_FILE"
}
```

That being said, looking at the code, you're right that the project is inferred from the provider configuration, not from "the account".

Regarding the error you're seeing (`Error: project: required field is not set`), it sounds like you might not [have a project set at the provider level](https://registry.terraform.io/providers/hashicorp/google/latest/docs), so the fallback doesn't work.

Sayrus:
> @Sayrus I'm not sure I understand the example HCL you posted. Both samples should fail because `google_service_account.account_key` doesn't exist.

You're absolutely right, I made an error while writing the example in the description. I edited my PR.

> Regarding the error you're seeing (`Error: project: required field is not set`), it sounds like you might not [have a project set at the provider level](https://registry.terraform.io/providers/hashicorp/google/latest/docs), so the fallback doesn't work.

This is exactly the reason. I have a provider which handles several environments so I have no `project` specified in the provider block.

melinath:
This was upstreamed & merged in the linked PRs.

Issue #9096, opened 2021-05-06
Can't create or update a Big Query View with a schema


avercoutere:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Terraform Version

Terraform v0.15.0
on linux_amd64
+ provider registry.terraform.io/hashicorp/google v3.64.0

### Affected Resource(s)

google_bigquery_table

### Terraform Configuration Files

```tf
provider "google" {
  project = "{{your_gcp_project}}
  region  = "europe-west1"
  zone    = "europe-west1-b"
}

resource "google_bigquery_dataset" "TestDS" {

  dataset_id                      = "Test"
  default_partition_expiration_ms = "0"
  delete_contents_on_destroy      = "false"
  location                        = "EU"

  access {
    role          = "OWNER"
    special_group = "projectOwners"
  }

  access {
    role          = "READER"
    special_group = "projectReaders"
  }

  access {
    role          = "WRITER"
    special_group = "projectWriters"
  }
}

resource "google_bigquery_table" "testViewSchema" {
  dataset_id = google_bigquery_dataset.TestDS.dataset_id
  expiration_time = "0"
  schema = jsonencode(
  [

    {
      "description":"Column 1 Decription",
      "mode":"NULLABLE",
      "name":"Column1",
      "type":"STRING"
    },
    {
      "description":"Column 2 Decription",
      "mode":"NULLABLE",
      "name":"Column2",
      "type":"STRING"
    }
  ]
  )
  table_id = "testViewSchema"

  view {
    query = <<SQL
select "value1" as col1, "value2" as col2
SQL
    use_legacy_sql = "false"
  }
}
```

### Debug Output

https://gist.github.com/avercoutere/8facad5e6b03b0dc42d6eeaf995037d1

### Expected Behavior

Create the Big Query view (table with a view) with the schema.

[...]

PR #9101, opened 2021-05-06
Add the ability to create a big query view with schema

avercoutere:
Make a 2 steps creation/update when the table is a view and it contains a schema. Google API doesn't support creating/modifying a view query with setting a schema at the same time.


avercoutere:
Fix #9096

rileykarson:
Note: we likely want to wait on a resolution for issuetracker.google.com/171967405 before moving forward with this change

tchiotludo:
@rileykarson : The google ticket 171967405 is not the same goal than this PR.

Google ticket 171967405 want to copy description from based tabled to the view. 
This PR only want to allow terraform to add description on the view (that can be a custom one and different from based table). 
The scope is larger with this PR since we can provide a description for any columns, example : I have an aggregation from many columns, the google ticket will not help in this case since no description will be forwarded on this standard case. 

This PR will allow to have the same behaviour and usage than the gcp console here : 
image
a simple manual description on columns of the view. 

Can this PR can be reconsidered please ? We have a lot of usage here that is completely blocked and this PR do exactly what we need. 

Thanks

rileykarson:
@ScottSuarez: Is this a reasonable change to make w/o the issuetracker issue linked above? Not sure I understand the product well enough to make a call.

ScottSuarez:
I think this is reasonable.

originally I tracked this ask to the api team and they refused this altercation on their side. I think this would be good QoL change

ScottSuarez:
@rileykarson are we to consider allowing this then?

@c2thorn shall you assist in following this through or I?

rileykarson:
Seems fine to me- y'all can choose between yourselves.

tchiotludo:
@ScottSuarez @rileykarson @c2thorn any progress please ? seems the just need a reviewer here ?

ScottSuarez:
I'll take over and review this change. I'll try to get to it this week.
Releases are paused for now for the holidays but we should have it in the release we do the second week of January.

Thank you for your patience.

c2thorn:
Hey folks, I had already started work on this before @ScottSuarez realized.

> any progress please ? seems the just need a reviewer here ?

It is not as simple as just a review. The code change needed to be upstreamed to Magic Modules and also a test added to show the feature working correctly. I've made the upstream PR here: https://github.com/GoogleCloudPlatform/magic-modules/pull/5583

I'm having trouble getting a proper test going, but once that is complete we should be good to go.

c2thorn:
upstream PR is merged, I will close this one.

PR #9270, opened 2021-06-01
replace hashibot with github actions

megan07:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/9136


.github/workflows/issue-comment-created.yml
```diff
@@ -0,0 +1,16 @@
...
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  issue_comment_triage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
bflad, line 11: This was my bad in the migration plan, I don't think this is necessary for this workflow 👍 

```suggestion
```

.github/workflows/pull-request-size.yml
```diff
@@ -0,0 +1,17 @@
+name: "Pull Request Size Labeler"
+
+on: [pull_request]
+
+jobs:
+  labeler:
+    runs-on: ubuntu-latest
+    name: Label the PR size
+    steps:
+      - uses: codelytv/pr-size-labeler@v1
```
bflad, line 10: Upstream enhancement to allow specifying the size labels is still pending: https://github.com/CodelyTV/pr-size-labeler/pull/26

Up to you folks whether this is worth skipping this migration for now (I will ensure there is a solution before decommissioning) or if it is worth renaming the repository `size/*` labels with lowercase to match this action after merging this.

.github/workflows/issue-comment-created.yml
```diff
@@ -0,0 +1,16 @@
+name: Issue Comment Created Triage
```
rileykarson, line 1: This will just remove `stale` and `waiting-reply` if a maintainer adds them at creation time, right? Is it possible a maintainer would do that on first file intentionally / do we expect it to happen unintentionally?

bflad (approved):

.github/workflows/issue-comment-created.yml
megan07: i think it's when a comment is created on the issue that we'll remove `stale` and `waiting-reply`, does that answer your question?

rileykarson: Oh, yup, that makes more sense!

rileykarson (approved):

.github/workflows/pull-request-size.yml
megan07: we have no preference, so i'll rename the labels to lowercase. thanks!

PR #9326, opened 2021-06-08
Expose filtered CIDRs ranges for default domains

bpineau:
Per https://cloud.google.com/vpc/docs/configure-private-google-access#ip-addr-defaults :

> The IP addresses used by the default domains for Google APIs and services fit within
> the list of ranges computed by taking away all ranges in cloud.json from those in
> goog.json. The following example shows you how to get this range using Python.

`terraform-provider-google` was offering access to both `goog.json`
(`google-netblocks`) and `cloud.json` (`cloud-netblocks`) but not to
that useful subset of addresses in goog.json but not in cloud.json,
which is hard to process otherwise (using only terraform native transforms).

This mirrors the above doc's python script's behavior.

The "default domains" name comes straight from the documentation linked above,
but I'm not sure if that the best choice here?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,7 @@ require (
 	google.golang.org/api v0.46.0
 	google.golang.org/genproto v0.0.0-20210503173045-b96a97608f20 // indirect
 	google.golang.org/grpc v1.38.0
+	gopkg.in/netaddr.v1 v1.5.1
 )
 
 go 1.16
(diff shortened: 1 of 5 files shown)


slevenick:
This looks pretty good! The only thing I'm hesitant about is adding the dependency on another repo.

Is there a way to replicate that behavior without the external dependency? If not I'll likely approve this, but it will take me a little bit to upstream to preserve it between this repo and the beta provider

joshuagibeonm:
Hi, Is there any follow up for this PR? We stumbled across a problem that requires set of filtered CIDRs ranges for GCP default domains.

If needed, I would like to provide support that needed to make this PR accepted.

Looking at the dependency that needed for this PR, it looks like https://github.com/go-netaddr/netaddr is not maintained anymore?

slevenick:
No follow up, and I'm going to close this PR for now as it's been so long. My only concern for this PR in general was the dependency, and we probably shouldn't use it if it's not maintained.

PR #9328, opened 2021-06-08
Update dcl

slevenick:
Required for https://github.com/GoogleCloudPlatform/magic-modules/pull/4852

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/hashicorp/terraform-provider-google
 
 require (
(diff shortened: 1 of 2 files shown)


melinath (changes requested):
It looks like there are build failures.

slevenick:
@melinath yeah it requires the linked PR to be merged to work. There will be a couple minutes of brokenness here, but merging both at the same time should get us to a good place

melinath (approved):
fun!

PR #9356, opened 2021-06-11
Updated metadata

fsadykov:
Hello HashiCorp team,

Please approve this PR we don't need to re-create the resource if metadata was updated

### DIFF
--- a/google/node_config.go
+++ b/google/node_config.go
@@ -111,9 +111,9 @@ func schemaNodeConfig() *schema.Schema {
 					Type:        schema.TypeMap,
 					Optional:    true,
 					Computed:    true,
-					ForceNew:    true,
+					ForceNew:    false,
 					Elem:        &schema.Schema{Type: schema.TypeString},
-					Description: `The metadata key/value pairs assigned to instances in the cluster.`,
+					Description: `If someone will modify metada the resource will not be re-created`,
 				},
 
 				"min_cpu_platform": {


google/node_config.go
```diff
@@ -111,9 +111,9 @@ func schemaNodeConfig() *schema.Schema {
 					Type:        schema.TypeMap,
 					Optional:    true,
 					Computed:    true,
-					ForceNew:    true,
+					ForceNew:    false,
 					Elem:        &schema.Schema{Type: schema.TypeString},
-					Description: `The metadata key/value pairs assigned to instances in the cluster.`,
+					Description: `If someone will modify metada the resource will not be re-created`,
```
megan07, line 116: `Description` shouldn't have to change here, as the current description describes what the field is.

google/node_config.go
```diff
@@ -111,9 +111,9 @@ func schemaNodeConfig() *schema.Schema {
 					Type:        schema.TypeMap,
 					Optional:    true,
 					Computed:    true,
-					ForceNew:    true,
+					ForceNew:    false,
```
megan07, line 114: We don't need to explicitly set this to `false` here, removing `ForceNew` should suffice.

megan07 (changes requested):
By making `metadata` updatable, we'll need to check if there's a change in the update function so that if there is a change, it will send the updates. See the `if d.HasChange` blocks [here](https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_container_cluster.go#L1587), and similarly in the `nodePoolUpdate` function.

Other things to consider would be if this is the same across the `NodeConfigs` in NodePools and the Cluster. I'm assuming it's the same either way, just want to call out that this is shared by both resources. If, for some reason, one is updatable and the other is not, take a look at this [PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/3512/files?diff=unified&w=1#).

Thanks for the contribution!

PR #9386, opened 2021-06-17
Added support for pubsubConfig trigger in the cloudbuild resource.

iamsumit:
Google have [API documentation](https://cloud.google.com/build/docs/api/reference/rest/v1/projects.triggers) to create a pub/sub trigger in the cloud build resource, but, [the terraform google provider](https://github.com/hashicorp/terraform-provider-google/issues/8692) doesn't. I have updated the terraform resource to add that support.

There is an issue in creating the pub/sub trigger using rest API and that seems a bug in Google Cloud API. When we create the pub/sub trigger using rest API and edit the trigger, it looks like this:

configedit

If I update the cloud build manually and check the get response for the cloud build, there is no difference in the response and UI shows the configuration checked properly. I hope there is nothing I am missing at the moment which makes this a bug in Google Cloud APIs.


iamsumit:
The failing log says: `2021-06-18T19:15:29.2195766Z Can't find any online and idle self-hosted runner in current repository that matches the required labels: 'ubuntu-latest'`

I am not sure if this is anything I am missing or there was some system issue. Can anyone please confirm?

iamsumit:
@megan07 @rileykarson We discussed about the missing webhookConfig support with the product manager in Google. It turns out that only documentation was missing which has been updated. I have added the support for webhookConfig in this PR only.

rileykarson:
Hey @iamsumit! This file is generated by [Magic Modules](https://github.com/GoogleCloudPlatform/magic-modules) which means we're unable to accept contributions made directly against this repository. Instead, can you edit it under [this directory](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/products/cloudbuild)?

MM's repo has a README with basic contribution instructions included. You can refer to similar fields in `api.yaml` and `terraform.yaml` to configure the field, and add your test to https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/tests/resource_cloudbuild_trigger_test.go so that this repo + `google-beta` are updated.

iamsumit:
@rileykarson That is so cool. Thank you. I have raised a PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/4931

PR #9389, opened 2021-06-17
remove exception about minscale key for cloudrun

czedyousfi:
I use fully managed Cloud Run and setting the minscale with Terraform works perfectly.


slevenick (approved):
I'll need to upstream this so it sticks around as these files are auto-generated

PR #9493, opened 2021-07-02
Added h3 to Argument Reference sub blocks to help easier navigation on bigquery_table

jonathanlawhh:
# Purpose
To allow users to immediately jump onto that sub section they need for easier navigation. 
For example when browsing and they want to see what structure the `view` block has, they could click on it and the page will scroll to that section immediately.

_If this pull request is helpful, I will definitely help adding similar commits to other docs for Google provider!_


SarahFrench:
Closing as stale, though if you want to propose changes to the documentation in future please open a PR against https://github.com/GoogleCloudPlatform/magic-modules, which controls the files in this repo.

PR #9506, opened 2021-07-05
Fix role definition in docs: bigquery_dataset_iam.html.markdown

timvermaercke:
The docs defined a role that throws errors, because it doesn't exist in GCP.


megan07 (approved):
Thanks!

PR #9555, opened 2021-07-13
cloudfunctions: validate allowed regions

bboughton:
Add validation for regions that support cloudfunctions to improve the user experience of using this provider.

### DIFF
--- a/google/resource_cloudfunctions_function.go
+++ b/google/resource_cloudfunctions_function.go
@@ -26,6 +26,31 @@ var allowedVpcConnectorEgressSettings = []string{
 	"PRIVATE_RANGES_ONLY",
 }
 
+var allowedRegionsSettings = []string{
+	"asia-east1",
+	"asia-east2",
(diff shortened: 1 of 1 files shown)


melinath:
Thanks for your contribution! Unfortunately, validations like this end up causing problems down the line because the list of supported regions gets updated frequently. This can end up being a UX issue because a person with an older version of the provider might not able to use a newly-supported region. It also ends up being a significant maintenance burden for the team when you look at the number of resources that could use this kind of validation.

If you think there's a particular reason that this use case is different in a way that warrants validation, please open an issue with additional details about why.

bboughton:
Your reasons for not allowing validation here make sense. The problem I ran into was that I was able to successfully create a plan for a new CloudFunction but the plan failed to apply because I had chosen an invalid region. Ideally the provider would provide feedback to the user that input is invalid at plan creation however it make sense why this may not be an easy problem to solve.

Issue #8465, opened 2021-02-12
Add security_level field to google_cloudfunctions_function


upodroid:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Affected Resource(s)

* google_cloudfunctions_function

### Potential Terraform Configuration

```tf
security_level = "secure-always"
```

### References

* https://cloud.google.com/functions/docs/reference/rest/v1/projects.locations.functions#SecurityLevel
* https://cloud.google.com/functions/docs/writing/http#security_levels

millouzgc:
Hello @gomezjdaniel, were you able to find a compromise on the code? You seem to be very close to the solution to implement.

melinath:
resolved by https://github.com/GoogleCloudPlatform/magic-modules/pull/6007

melinath:
duplicate of https://github.com/hashicorp/terraform-provider-google/issues/11086

PR #9583, opened 2021-07-19
cloudfunctions: allow to configure `trigger_http` security level

gomezjdaniel:
- This is not a backwards compatible change
- Possible values are `SECURE_OPTIONAL` (http/s) and `SECURE_ALWAYS` (https)

### DIFF
--- a/google/data_source_google_cloudfunctions_function_test.go
+++ b/google/data_source_google_cloudfunctions_function_test.go
@@ -53,7 +53,7 @@ resource "google_cloudfunctions_function" "function_http" {
   available_memory_mb   = 128
   source_archive_bucket = google_storage_bucket.bucket.name
   source_archive_object = google_storage_bucket_object.archive.name
-  trigger_http          = true
+  trigger_http          = "SECURE_ALWAYS"
   timeout               = 61
   entry_point           = "helloGET"
 }
--- a/google/iam_cloudfunctions_function_generated_test.go
+++ b/google/iam_cloudfunctions_function_generated_test.go
@@ -139,7 +139,7 @@ resource "google_cloudfunctions_function" "function" {
   available_memory_mb   = 128
   source_archive_bucket = google_storage_bucket.bucket.name
   source_archive_object = google_storage_bucket_object.archive.name
-  trigger_http          = true
+  trigger_http          = "SECURE_ALWAYS"
   timeout               = 60
   entry_point           = "helloGET"
 }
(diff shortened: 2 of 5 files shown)


slevenick (changes requested):
Unfortunately we won't be able to accept this change.

This would be a breaking change because changing the type of this field from Bool -> String will break existing user's configs if they define this field as a boolean. This is *possible* to do in a major version of the provider, but unlikely to be something we would do.

I think the better approach would be to add a new optional field that controls this setting

gomezjdaniel:
@slevenick got it.

I then added a new `security_field` as https://github.com/hashicorp/terraform-provider-google/issues/8465 suggests :)

In case everything is good, I will squash commits afterwards.

slevenick:
> @slevenick got it.
> 
> I then added a new `security_level` field as #8465 suggests :)
> 
> In case everything is good, I will squash commits afterwards.
> 
> Also note [cloudfunctions.HttpsTrigger.SecurityLevel](https://pkg.go.dev/google.golang.org/api/cloudfunctions/v1#HttpsTrigger) possible values are `["SECURITY_LEVEL_UNSPECIFIED", "SECURE_ALWAYS", "SECURE_OPTIONAL"]` but I omit the unspecified one.

I don't think we can include `RequiredWith: []string{"trigger_http"},` as that would also make this change backwards incompatible. If the change would cause an existing config to fail it is considered breaking and would not be allowed until a major version of the provider.

Removing that should make this feasible at the cost of some validation

gomezjdaniel:
My wrong @slevenick, I was trying to make `trigger_http` a required field in case `security_level` was set. Otherwise the http trigger would be enabled with either only of the previous fields being set which would make it confusing in my opinion.

Therefore I remove `RequiredWith: []string{"trigger_http"},` from `security_level` and add `RequiredWith: []string{"security_level"},` to `trigger_http`.

gomezjdaniel:
Ping @slevenick

slevenick (commented):
I've upstreamed this to our code generator so it can be replicated across the `google` and `google-beta` providers. 

We will need docs describing this new field before we can merge this, can you add that? It would go into this file: https://github.com/hashicorp/terraform-provider-google/blob/master/website/docs/r/cloudfunctions_function.html.markdown

gomezjdaniel:
Sure, there you go ;). I also fixed a typo in the code.

slevenick:
So it looks like RequiredWith causes the dependency both ways:

```
Error: RequiredWith
on terraform_plugin_test.tf line 20, in resource "google_cloudfunctions_function" "function":
20:   trigger_http          = true
"trigger_http": all of `security_level,trigger_http` must be specified
```

which may make this a breaking change. We will likely have to remove that RequiredWith validation

PR #9649, opened 2021-07-28
Docs update for `google_compute_global_address`

kskumgk63:
Hi, 

I got this error when `terraform apply`. 

```
│ Error: Error creating GlobalAddress: googleapi: Error 400: Invalid value for field 'resource.purpose': 'VPC_PEERING'. The field cannot be specified with external address., invalid
│ 
│   with google_compute_global_address.static_ip["foo"],
│   on main.tf line 37, in resource "google_compute_global_address" "static_ip":
│   37: resource "google_compute_global_address" "static_ip" {
│ 
```

I think `The field cannot be specified with external address.` should be on the document to prevent this error before `terraform apply`.

### DIFF
--- a/google/resource_compute_global_address.go
+++ b/google/resource_compute_global_address.go
@@ -113,7 +113,7 @@ or addressType=INTERNAL when purpose=PRIVATE_SERVICE_CONNECT`,
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
-				Description: `The purpose of the resource. Possible values include:
+				Description: `The purpose of the resource. This field cannot be specified with `EXTERNAL` address_type. Possible values include:
 
 * VPC_PEERING - for peer networks
 
--- a/google/resource_compute_global_address.go
+++ b/google/resource_compute_global_address.go
@@ -113,11 +113,14 @@ or addressType=INTERNAL when purpose=PRIVATE_SERVICE_CONNECT`,
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
-				Description: `The purpose of the resource. This field cannot be specified with `EXTERNAL` address_type. Possible values include:
-
-* VPC_PEERING - for peer networks
-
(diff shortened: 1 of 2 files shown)


kskumgk63:
@ScottSuarez Could you review it,please?

ScottSuarez:
Looks like this documentation is outdated in general since we fork these values at release.

This allows you to match any allowed enumeration in the official api.. `NAT_AUTO` should allow external addresses. Sorry for the confusion here.

```
The purpose of this resource, which can be one of the following values:

GCE_ENDPOINT for addresses that are used by VM instances, alias IP ranges, load balancers, and similar resources.
DNS_RESOLVER for a DNS resolver address in a subnetwork for a Cloud DNS inbound forwarder IP addresses (regional internal IP address in a subnet of a VPC network)
VPC_PEERING for global internal IP addresses used for private services access allocated ranges.
NAT_AUTO for the regional external IP addresses used by Cloud NAT when allocating addresses using <a href="https://cloud.google.com/nat/docs/ports-and-addresses#addresses" automatic NAT IP address allocation.
IPSEC_INTERCONNECT for addresses created from a private IP range that are reserved for a VLAN attachment in an IPsec-encrypted Cloud Interconnect configuration. These addresses are regional resources. Not currently available publicly.
SHARED_LOADBALANCER_VIP for an internal IP address that is assigned to multiple internal forwarding rules.
PRIVATE_SERVICE_CONNECT for a private network address that is used to configure Private Service Connect. Only global internal addresses can use this purpose.
```

https://cloud.google.com/compute/docs/reference/rest/v1/globalAddresses

kskumgk63:
@ScottSuarez 

Thanks for the review.
I edited it following the official document. e00d1be

ScottSuarez:
I've upstreamed this to our code generator, so that pr will supersede this one. Thanks for the contribution! :) 

https://github.com/hashicorp/terraform-provider-google/pull/9649

PR #9655, opened 2021-07-28
correct import syntax for alert policy

philwhln:
Resolves https://github.com/hashicorp/terraform-provider-google/issues/7693


slevenick:
Unfortunately because this file is generated we cannot accept this change currently.

I have a ticket open for updating these {{name}}s in documentation for imports with their expanded full name: https://github.com/hashicorp/terraform-provider-google/issues/8499

PR #9681, opened 2021-07-30
Fix format for import in google_service_networking_connection

Jasstkn:
Signed-off-by: Maria Kotlyarevskaya <jasssstkn@yahoo.com>


megan07 (approved):
Thanks!

PR #9741, opened 2021-08-07
Docs update for google_compute_network: default values

gajdaw:
Hi,
I find missing info about default values for `auto_create_subnetworks` and `routing_mode` a bit confusing.
Maybe we should add them?

### DIFF
--- a/website/docs/r/compute_network.html.markdown
+++ b/website/docs/r/compute_network.html.markdown
@@ -86,7 +86,7 @@ The following arguments are supported:
   it will create a subnet for each region automatically across the
   `10.128.0.0/9` address range.
   When set to `false`, the network is created in "custom subnet mode" so
-  the user can explicitly connect subnetwork resources.
+  the user can explicitly connect subnetwork resources. Defaults to `true`.
 
 * `routing_mode` -
   (Optional)
@@ -95,7 +95,7 @@ The following arguments are supported:
   of this network in the same region as the router. If set to `GLOBAL`,
   this network's cloud routers will advertise routes with all
   subnetworks of this network, across regions.
-  Possible values are `REGIONAL` and `GLOBAL`.
+  Possible values are `REGIONAL` and `GLOBAL`. Defaults to `REGIONAL`.
 
 * `mtu` -
   (Optional)


gajdaw:
@rileykarson Can you take a look at this PR?

gajdaw:
@megan07 What do you think about this PR?

website/docs/r/compute_network.html.markdown
```diff
@@ -95,7 +95,7 @@ The following arguments are supported:
   of this network in the same region as the router. If set to `GLOBAL`,
   this network's cloud routers will advertise routes with all
   subnetworks of this network, across regions.
-  Possible values are `REGIONAL` and `GLOBAL`.
+  Possible values are `REGIONAL` and `GLOBAL`. Defaults to `REGIONAL`.
```
megan07, line 98: This one seems a bit tricky as it looks like the default comes from the API, which typically means the default will change depending on different settings. I'd be hesitant to add the default here.

website/docs/r/compute_network.html.markdown
```diff
@@ -86,7 +86,7 @@ The following arguments are supported:
   it will create a subnet for each region automatically across the
   `10.128.0.0/9` address range.
   When set to `false`, the network is created in "custom subnet mode" so
-  the user can explicitly connect subnetwork resources.
+  the user can explicitly connect subnetwork resources. Defaults to `true`.
```
megan07, line 89: This appears to be [correct](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/products/compute/terraform.yaml#L1407), however this description is auto-generated and needs to be changed upstream. Specifically here: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/products/compute/api.yaml#L7924.

megan07 (commented):
I've added comments, but wanted to mention that this file is auto-generated. I've included the link to the upstream file where `auto_create_subnetworks` description should be updated. Thanks!

website/docs/r/compute_network.html.markdown
gajdaw: Thank you for info. I moved PR with default value for `auto_create_subnetworks` to `magic-modules` repo.

PR #9848, opened 2021-08-18
Fix:9723 google_sql_user - "Root resource was present, but now absent"

jangolano:


google/resource_sql_user.go
```diff
@@ -186,9 +186,12 @@ func resourceSqlUserRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	var user *sqladmin.User
-	for _, currentUser := range users.Items {
 
-		name = strings.Split(name, "@")[0]
+	for _, currentUser := range users.Items {
+		var databaseInstance *sqladmin.DatabaseInstance
+		if !strings.Contains(databaseInstance.DatabaseVersion, "POSTGRES") {
```
megan07, line 192: `databaseInstance` is `nil` here. We'll need to find a way to GET the associated instance if we want to check the version.

megan07 (changes requested):
Thanks for this fix! Per the comment, we're going to need to GET the associated databaseInstance in order to check what version it is. Thanks!

google/resource_sql_user.go
megan07: I think something like this would do:
```
		databaseInstance, err := config.NewSqlAdminClient(userAgent).Instances.Get(project, currentUser.Instance).Do()
		if err != nil {
			return err
		}
```

I haven't dug too deep into this, my only suggestion would be to look quick if we can do the GET call outside the loop or not. I'm not sure if that's possible or not.

google/resource_sql_user.go
```diff
@@ -186,9 +186,15 @@ func resourceSqlUserRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	var user *sqladmin.User
-	for _, currentUser := range users.Items {
+	databaseInstance, err := config.NewSqlAdminClient(userAgent).Instances.Get(project, instance).Do()
```
megan07, line 189: Oh, great! Testing now and if it passes I'll approve and get this in. Thanks!

megan07 (approved):
Looks great! Thanks so much!

PR #9896, opened 2021-08-24
Updated migrate_traffic description in google_app_engine_service_split_traffic resource

thedmeyer:
Updating description of the `migrate_traffic` variable to be more clear on it's purpose and be in line with GCP documentation.

See: https://cloud.google.com/appengine/docs/admin-api/migrating-splitting-traffic

### DIFF
--- a/google/resource_app_engine_service_split_traffic.go
+++ b/google/resource_app_engine_service_split_traffic.go
@@ -73,7 +73,7 @@ func resourceAppEngineServiceSplitTraffic() *schema.Resource {
 			"migrate_traffic": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Description: `If set to true traffic will be migrated to this version.`,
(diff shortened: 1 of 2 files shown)


ndmckinley (approved):

pjeannet:
Hello, found this PR while looking at existing issues with `migrate_traffic`. You could also add that it won't work with flex environment (I tried to use it and got `Traffic migration does not support App Engine Flexible Environment.`).
I also noticed that it's not really optionnal as if we don't set it to false then we get `Invalid value at 'migrate_traffic' (TYPE_BOOL), \"\"`.

ndmckinley (approved):
Hi there - this is correct, but as you can see at the top of this file, this code is autogenerated.  Would you mind making the change at the source?  That's https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/products/appengine/api.yaml#L1437.

Issue #7811, opened 2020-11-14
google_logging_project_sink does not work for automatically created sinks


matthewblain:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Terraform Version

> Terraform v0.13.5
> + provider registry.terraform.io/hashicorp/google v3.46.0
> + provider registry.terraform.io/hashicorp/google-beta v3.37.0

### Affected Resource(s)

* google_logging_project_sink

### Terraform Configuration Files

Elided config file:

```tf
resource "google_logging_project_sink" "log_sink" {
  name = "_Default"

  destination = vars.logging_bucket

  # While this is the default filter, it appears that it must be set explicitly.
  filter = <<-endfilter
    NOT LOG_ID("cloudaudit.googleapis.com/activity") AND 
    NOT LOG_ID("externalaudit.googleapis.com/activity") AND
    NOT LOG_ID("cloudaudit.googleapis.com/system_event") AND
    NOT LOG_ID("externalaudit.googleapis.com/system_event") AND
    NOT LOG_ID("cloudaudit.googleapis.com/access_transparency") AND
    NOT LOG_ID("externalaudit.googleapis.com/access_transparency")
  endfilter
}

```

### Debug Output

The error is this:

`googleapi: Error 409: Sink _Default already exists, alreadyExists`

Attached is a non-debug log: 
[elidedlog.txt](https://github.com/hashicorp/terraform-provider-google/files/5540241/elidedlog.txt)
 It should be enough info to understand this issue but I can provide the full version if necessary.

### Expected Behavior

The always-present logging sink should have been updated to point to the non-default bucket.

### Actual Behavior

Terraform attempted to create a new sink, which is not possible for _Default.

Note: I also experimented with importing the existing _Default sink, but that failed at the 'delete' stage.

[...]

PR #9903, opened 2021-08-24
Make Logging Sink resources work with auto-created sinks

estliberitas:
Hey, for quite a long time I wanted to re-route logs to regional logging buckets using automatically created sinks (UPDATE: which turns out to be not really possible). I saw that acquire'n'update feature already exists for logging bucket configs, but not for sinks. Hence, this PR. Another reason for this update is that it closes a gap of configuring automatically created sinks, which was mentioned in #7811.

This is my first contribution to a Terraform provider, so pls do a thorough review, because it's my first Go experience as well.

### DIFF
--- a/google/logging_utils.go
+++ b/google/logging_utils.go
@@ -33,11 +33,11 @@ func (l LoggingSinkId) parent() string {
 	return fmt.Sprintf("%s/%s", l.resourceType, l.resourceId)
 }
 
-// parseLoggingSinkId parses a canonical id into a LoggingSinkId, or returns an error on failure.
-func parseLoggingSinkId(id string) (*LoggingSinkId, error) {
+// parseLoggingSinkParentId parses a canonical id to get sink parent resource id
+func parseLoggingSinkParentId(id string) (string, error) {
 	parts := loggingSinkIdRegex.FindStringSubmatch(id)
 	if parts == nil {
-		return nil, fmt.Errorf("unable to parse logging sink id %#v", id)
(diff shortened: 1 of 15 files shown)


estliberitas:
Now that I've created a PR I realized that there are no tests for automatically created sinks. I should add them, right?

estliberitas:
Hola, just giving a small update.

Blocked by this ATM: https://groups.google.com/g/google-stackdriver-discussion/c/BgfBHD7C5bU.

Lack of the `writerIdentity` in the API output breaks import state verification for some project logging sink tests. In those tests we usually set `uniqueWriterIdentity=true` for `POST` / `PATCH` requests. And when we get no `writerIdentity` in the `GET` request, we are unable to determine if `uniqueWriterIdentity` is actually true or not.

The Cloud Logging API description doesn't describe how to interpret lack of the `writerIdentity` field. And something tells me that we can't use `res["writerIdentity"] != nonUniqueWriterIdentity` condition (the one we have in in [here](https://github.com/hashicorp/terraform-provider-google/blob/f88a05a3d292a1ccaacb4755540a253cd758eb4f/google/resource_logging_project_sink.go#L114-L122) for `google_logging_project_sink`) to determine if we can set `unique_writer_identity` to `true` or `not`.

Correct me if I'm wrong, please.

estliberitas:
@rileykarson set as ready for review, so that we can start a conversation.

rileykarson:
Closing this as stale, as it's several years old.

PR #9958, opened 2021-09-01
Redis Instance Resource no longer supports auth_string, updating docs to reflect

NeckBeardPrince:
Support for `auth_string` for `google_redis_instance` was removed in https://github.com/GoogleCloudPlatform/magic-modules/pull/4492. But the docs were never updated to reflect this.


megan07:
Hi @NeckBeardPrince! Sorry for the delay on reviewing this! It looks like this file is generated. If you'd like, you can remove these lines [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/products/redis/terraform.yaml#L17)! Otherwise I'd be happy to make this change as well if you don't have the time.
Thanks for your patience!

PR #10018, opened 2021-09-07
✨ Add support for Log Views

estliberitas:
`google_logging_view` resource was added


website/docs/r/logging_project_view.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+  Manages a log bucket view.
+---
+
+# google\_logging\_view
+
+Manages a log bucket view. For more information see
+[the official logging documentation](https://cloud.google.com/logging/docs/) and
+[Managing log views on your log buckets](https://cloud.google.com/logging/docs/logs-views).
+
+~> **Note:** Log views are automatically created for a log bucket. Creating a resource of this type will fail in Terraform. The log views that are currently automatically created are `_AllLogs` and `_Default`.
```
estliberitas, line 16: I think later we can add a datasource for that if will be needed ever. The problem with `_AllLogs` and `_Default` views is that they are not modifiable, which kinda breaks the whole idea of a Terraform-managed resource.

c2thorn:
Hi @estliberitas, thanks for getting started on this resource. 

This repository is maintained by the upstream repository https://github.com/GoogleCloudPlatform/magic-modules. In order for your changes to appear in both GA and beta providers, we'll have to move the PR there. 

Checkout https://github.com/GoogleCloudPlatform/magic-modules#getting-started-with-magic-modules, and I'll point you to the directories you'll be interested in:

You'll want to put the resource file here:
https://github.com/GoogleCloudPlatform/magic-modules/tree/master/mmv1/third_party/terraform/resources

Test file: https://github.com/GoogleCloudPlatform/magic-modules/tree/master/mmv1/third_party/terraform/tests

Documentation: https://github.com/GoogleCloudPlatform/magic-modules/tree/master/mmv1/third_party/terraform/website/docs/r

Provider file update: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/utils/provider.go.erb

The `website/google.erb` file will update automatically.

Feel free to @ me in the PR for review/questions!

SarahFrench:
Closing as the provider now contains a [ google_logging_log_view  resource](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/logging_log_view)

PR #10210, opened 2021-10-01
Don't test the scripts/ directory since it contains code that can't be tested.

ndmckinley:
Specifically, can't be tested because it includes phony packages that don't exist.  See https://github.com/GoogleCloudPlatform/magic-modules/pull/5268 for context.  :)

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,4 +1,4 @@
-TEST?=$$(go list ./...)
+TEST?=$$(go list ./... | grep -v scripts)
 WEBSITE_REPO=github.com/hashicorp/terraform-website
 PKG_NAME=google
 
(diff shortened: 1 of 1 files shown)


ndmckinley:
Sibling: https://github.com/hashicorp/terraform-provider-google-beta/pull/3687

GNUmakefile
```diff
@@ -1,4 +1,4 @@
-TEST?=$$(go list ./...)
+TEST?=$$(go list ./... | grep -v scripts)
```
rileykarson, line 1: Can we filter on the repo-specific prefix like `github.com/hashicorp/terraform-provider-google/scripts/`? That'll minimise the chance of false positives in the future.

rileykarson (approved):

GNUmakefile
ndmckinley: Yep!

PR #10251, opened 2021-10-05
Remove misleading word from the node pool - node_config docs. 

d4v3y0rk:
The use of the word `default` here leads one to believe this relates to a "default" node pool as is referenced in the examples and in the cluster config settings. This block can be used for any node pool we want to setup and has nothing to do with the "default" node pool that is created when creating a cluster unless it is actually used in the cluster creation config.


d4v3y0rk:
> Hey @d4v3y0rk! Do you mind making this change against https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/website/docs/r/container_node_pool.html.markdown? That'll apply it to the `google-beta` provider as well.

Will do!

PR #10402, opened 2021-10-25
fix crash when object owner is missing

marcb1:
For context: https://github.com/hashicorp/terraform-provider-google/issues/10342

### DIFF
--- a/google/resource_storage_object_acl.go
+++ b/google/resource_storage_object_acl.go
@@ -80,7 +80,10 @@ func resourceStorageObjectAclDiff(_ context.Context, diff *schema.ResourceDiff, 
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


melinath (changes requested):
Please add tests that demonstrate this issue and show that it is fixed. Here are a couple examples that you could base it off of:

- https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_compute_instance_template_test.go#L164
- https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_logging_project_sink_test.go#L188 

(The core change will be extracting the logic you want to test into a separate function that is more testable.)

melinath:
@marcb1 are you still working on this PR?

melinath:
Closing for now.

PR #10576, opened 2021-11-17
add files for gha release workflow

megan07:
Adding the appropriate files for automated releases with GHA


scripts/changelog-links.sh
```diff
@@ -1,31 +0,0 @@
-#!/bin/bash
```
anGie44, line 1: so this is an interesting one..this script is usually run by the TeamCity configuration? perhaps to re-use this behavior, an additional job in the `release.yml` can run this  script and commit the updated changelog file🤔 in the `awscc` provider i guess we haven't run into many situations where the changelog entries are referencing specific PRs so that part has been pretty manual

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+        with:
+          aws-access-key-id: ${{ secrets.TF_PROVIDER_RELEASE_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.TF_PROVIDER_RELEASE_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+          role-to-assume: ${{ secrets.TF_PROVIDER_RELEASE_AWS_ROLE_ARN }}
+          role-duration-seconds: 3600
+          # TODO: allow session tagging once IAM permission changes addressed
+          # Reference: https://github.com/hashicorp/hc-releases/issues/124
+          role-skip-session-tagging: true
+      - name: goreleaser release
```
anGie44, line 62: so one thing we're not doing in the `awscc` provider is uploading the manifest.json to the release artifacts for each new release...which from my understanding is _ok_ since the protocol version and versions are sent through the `hc-release upload-file` header flags for each provider binary pushed to S3. nevertheless, a good example to look at  for manifest file copying is seen here https://github.com/bflad/ghaction-terraform-provider-release/blob/55a0972946e4619847280e1057cee14b947247cd/.github/workflows/community.yml#L41-L42. so before this `gorelease release` step, the manifest could be copied so that the new file has the new release version in the filename. To make sure `goreleaser` doesn't complain about this new file, you'll want to add `*_manifest.json` to the repo's .gitignore as mentioned in https://github.com/hashicorp/terraform-plugin-framework/issues/87#issuecomment-944713973

.goreleaser.yml
```diff
@@ -0,0 +1,60 @@
...
+    env:
+      - AWS_DEFAULT_REGION={{ .Env.AWS_DEFAULT_REGION }}
+      - AWS_REGION={{ .Env.AWS_REGION }}
+      - AWS_ACCESS_KEY_ID={{ .Env.AWS_ACCESS_KEY_ID }}
+      - AWS_SECRET_ACCESS_KEY={{ .Env.AWS_SECRET_ACCESS_KEY }}
+      - AWS_SESSION_TOKEN={{ .Env.AWS_SESSION_TOKEN }}
+release:
+  ids:
+    - none
+signs:
```
anGie44, line 46: one thing that came up recently is that the artifacts found in https://releases.hashicorp.com for each provider directory should technically contain both this signature file with the public key in the filename and the `plain` filename as seen here for the google provider:

Screen Shot 2021-11-18 at 12 03 10 AM

so to do so with this new goreleaser configuration, you'll need to add an additional block that creates the "plain" signature file e.g.

```suggestion
signs:
  # Default Signature file (i.e. terraform-provider-awscc_VERSION_SHA256SUMS.sig)
  - cmd: sh
    args:
    - -c
    - >-
      signore
      sign
      --dearmor
      --file ${artifact}
      --signer {{ .Env.SIGNER }}
      --out ${signature}
    artifacts: checksum
  # Signature file with GPG Public Key ID in filename (i.e. terraform-provider-awscc_VERSION_SHA256SUMS.7685B676.sig)
  - id: sig-with-gpg-public-key-id
    signature: ${artifact}.72D7468F.sig
    cmd: sh
    args:
    - -c
    - >-
      signore
      sign
      --dearmor
      --file ${artifact}
      --signer {{ .Env.SIGNER }}
      --out ${signature}
    artifacts: checksum
```

anGie44 (commented):
Only major addition needed would be to make sure `goreleaser` creates the 2 signature files (one with the public key in the filename and one without) as described in the comments, otherwise this LGTM 👍 I dropped in suggestions related to the manifest file copying and changelong-links scripts as I haven't dealt with those yet but lmk your thoughts 😄

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+*'
+
+env:
+  GOPROXY: https://proxy.golang.org/
+  PROTOCOL_VERSION: "6.0"
+  PROTOCOL_VERSIONS: "6.0"
```
anGie44, line 11: ooo i almost overlooked this! ah so use of these protocol versions (6.0) assume that the provider is using `terraform-plugin-framework`  but i imagine the provider is still compatible w/protocol 5..so looking at the last TC build for the google provider, these params should probably be inline with

Screen Shot 2021-11-18 at 12 23 28 AM

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+        run: |
+          manifestFile=${{ github.event.repository.name }}_manifest.json
+          if [[ -f  $manifestFile ]] && [[ $(jq '.metadata.protocol_versions' $manifestFile | jq length) -gt 0 ]]; then
+            versions=`jq '.metadata.protocol_versions' $manifestFile | jq 'sort'`
+            protoV=`echo $versions | jq '.[0]'`
+            protoVs=`echo $versions | jq 'join(", ")'`
+
+            echo "PROTOCOL_VERSION=${protoV}" >> $GITHUB_ENV
+            echo "PROTOCOL_VERSIONS=${protoVs}" >> $GITHUB_ENV
+          fi
```
anGie44, line 51: another thing i overlooked 😅 since this provider is built on SDK v2 (not plugin-framework..for now), then this step can be removed as it's specific to providers that use protocol 6 per https://www.terraform.io/docs/plugin/framework/publishing.html

.github/workflows/release.yml
megan07: yes, yes! will change, thanks!

scripts/changelog-links.sh
megan07: ah! gotcha, i just read in the RFC that we could remove this. for tpg/tpgb we generate our changelog entries anyway. i'll have to look at that changelog generation to see if our changelog generates to the `[GH-nnnn]` format or just includes the links? I'd assume the latter, but will double-check! Thanks!

anGie44 (approved):

melinath (approved):
I don't know what all of the parts of this do precisely but this seems reasonable.

Issue #9180, opened 2021-05-19
Add support for the rest of the AutoprovisioningNodePoolDefaults settings in `google_container_cluster`


pib:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

[AutoprovisioningNodePoolDefaults](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#Cluster.AutoprovisioningNodePoolDefaults) in the v1 API has support for several more fields than those currently supported by the [auto_provisioning_defaults block](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#auto_provisioning_defaults). 

Currently only `min_cpu_platform`, `oauth_scopes`, and `service_account` are supported.

Options currently missing are
- `upgradeSettings`
- `management`
- `diskSizeGb`
- `diskType`
- `shieldedInstanceConfig`
- `bootDiskKmsKey`

It might makes sense for these to be implemented in separate PRs. There are already a couple of existing PRs that add `upgrade_settings` and `management` (linked in the references section below).

### New or Affected Resource(s)

* google_container_cluster

### Potential Terraform Configuration

```tf
resource "google_container_cluster" "example" {
  name = "example"

  cluster_autoscaling {
    enabled = true

    auto_provisioning_defaults {
      service_account  = google_service_account.cluster_nodes_service_account.email
      min_cpu_platform = "Intel Haswell"
      oauth_scopes     = []

      # New options
      upgrade_settings {
        max_surge       = "25%"
        max_unavailable = "0"
      }

      management {
        auto_repair  = true
        auto_upgrade = true
      }

      disk_size_gb      = 12
      disk_type         = "pd-ssd"
      boot_disk_kms_key = google_kms_crypto_key.example.id

      shielded_instance_config {
        enable_secure_boot          = true
[...]

Issue #5579, opened 2020-02-04
Add support for management settings in `google_container_cluster` NAP


davidquarles:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

It would be amazing if the provider supported the v1 upstream NAP [NodeManagement](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/NodeManagement) fields.  For our use case, I'm not super comfy turning on node autoprovisioning in production without node-autorepair.

### New or Affected Resource(s)

* google_container_cluster

### Potential Terraform Configuration

```tf
resource "google_container_cluster" "foo" {
  name = "foo"
  # ...
  cluster_autoscaling {
    auto_provisioning_defaults {
      service_account = google_service_account.cluster_nodes_service_account.email

      # new field, mirroring google_container_node_pool.management structure
      management {
        auto_repair = true
        auto_upgrade = true
      }
    }
  }
}
```

### References

* #655
* #669

davidquarles:
I contributed the linked PR for this functionality at the `google_container_node_pool` layer back when (and very much appreciated the community vibe over here!) and I'm happy to contribute here as well, assuming it's not a wild amount of additional complexity to do so.

danawillow:
That would be great! I'll assign this issue to you. I imagine it'll be quite similar to the other one.

mitj04:
This issue is to add support for `management` in the `google_container_cluster` NAP. But, this has already been added as part of this [PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/6794) which was part of this [issue](https://github.com/hashicorp/terraform-provider-google/issues/9180). Closing this issue please reach out to us in case of any queries.

We can close this [PR](https://github.com/hashicorp/terraform-provider-google/pull/5672) as well which was resolving this issue.

mitj04:
b/263545635

PR #10610, opened 2021-11-20
`google_container_cluster` implement missing auto provisioning defaults fields

dysosmus:
This implements multiple GA fields currently missing in auto-provisioning defaults for `google_container_cluster`. 
It resolves multiple issues: #5580, #5579, #9180. 

All the settings already exist either at cluster level or in the node pool configure I used the same convention. 

### Affected Resource
`google_container_cluster`

### Terraform Configuration
```terraform
resource "google_container_cluster" "example" {
  name = "example"

  cluster_autoscaling {
    enabled = true

    auto_provisioning_defaults {
      service_account = google_service_account.cluster_nodes_service_account.email

      oauth_scopes = []

      # New options
      disk_size_gb      = 12
      disk_type         = "pd-ssd"
      boot_disk_kms_key = google_kms_crypto_key.example.id
      image_type        = "UBUNTU_CONTAINERD"
      min_cpu_platform  = "Intel Haswell"

      upgrade_settings {
        max_surge       = 1
        max_unavailable = 0
      }

      management {
        auto_repair  = true
        auto_upgrade = true
      }

      shielded_instance_config {
        enable_secure_boot          = true
        enable_integrity_monitoring = false
      }
    }
  }
}
```

### DIFF
--- a/google/resource_container_cluster.go
+++ b/google/resource_container_cluster.go
@@ -315,6 +315,106 @@ func resourceContainerCluster() *schema.Resource {
 										Default:     "default",
 										Description: `The Google Cloud Platform Service Account to be used by the node VMs.`,
 									},
+									"shielded_instance_config": {
+										Type:        schema.TypeList,
+										Optional:    true,
+										Description: `Shielded Instance options.`,
+										MaxItems:    1,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"enable_secure_boot": {
+													Type:        schema.TypeBool,
+													Optional:    true,
+													Default:     false,
+													Description: `Defines whether the instance has Secure Boot enabled.`,
+												},
+												"enable_integrity_monitoring": {
+													Type:        schema.TypeBool,
+													Optional:    true,
+													Default:     true,
+													Description: `Defines whether the instance has integrity monitoring enabled.`,
+												},
+											},
+										},
+									},
+									"upgrade_settings": {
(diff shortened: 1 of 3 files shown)


google/resource_container_cluster.go
```diff
@@ -315,6 +315,106 @@ func resourceContainerCluster() *schema.Resource {
...
+												"auto_upgrade": {
+													Type:        schema.TypeBool,
+													Optional:    true,
+													Default:     false,
+													Description: `Whether the nodes will be automatically upgraded.`,
+												},
+											},
+										},
+									},
+									"min_cpu_platform": {
```
slevenick, line 386: min_cpu_platform already exists in the beta provider. We can add it by removing the version guard tags in the upstream repo here and in the flatten/expand methods: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/resources/resource_container_cluster.go.erb#L442

Similarly, the other fields can be added to that file to add to both the google and google-beta providers.

We will also need these fields to be documented in the container_cluster markdown file, and test the update behavior for these fields by changing their values in the test.

slevenick (changes requested):

google/resource_container_cluster.go
dysosmus: Hey, thanks for the indications, I gave it a try in https://github.com/GoogleCloudPlatform/magic-modules/pull/5490 .

PR #10623, opened 2021-11-23
Change link to moved article about Knative Serving

sorenhansendk:
Update link to article about Knative Serving in `google_cloud_run_service`


slevenick (approved):
Thanks!

PR #10633, opened 2021-11-24
Clarification in spanner_instance doc

davidcueva:
Clarify that exactly one of num_nodes or processing_units is required.
If none of them is specified, you get an `Error: ExactlyOne`


c2thorn:
Thanks for bringing this to our attention @davidcueva. This file is managed upstream by Magic Modules. I've replicated the change in https://github.com/GoogleCloudPlatform/magic-modules/pull/5494, so we can close this here.

PR #10638, opened 2021-11-25
Add support GCE_VM_IP type for network endpoint group

irwanshofwan:
Please review this pull request, the pull request to support GCE_VM_IP type for network endpoint group.
So we can setup internal lb using NEG as backend using terraform.


google/resource_compute_network_endpoint.go
```diff
@@ -65,7 +65,7 @@ range).`,
 			},
 			"port": {
 				Type:        schema.TypeInt,
-				Required:    true,
+				Optional:    true,
```
melinath, line 68: From the docs it looks like "If not specified, the defaultPort for the network endpoint group will be used." Do you know if this means that this will be nil (but fall back to the network endpoint group's value) or if it will be set on creation to whatever the default is? In the latter case this would need to be computed as well as optional.

website/docs/r/compute_network_endpoint.html.markdown
```diff
@@ -103,7 +103,7 @@ The following arguments are supported:
   The instance must be in the same zone of network endpoint group.
 
 * `port` -
-  (Required)
+  (Optional)
   Port number of network endpoint.
```
melinath, line 107: This should also include ([from the API docs](https://cloud.google.com/compute/docs/reference/rest/beta/networkEndpointGroups/attachNetworkEndpoints)): `If not specified, the defaultPort for the network endpoint group will be used.`

google/resource_compute_network_endpoint_group.go
```diff
@@ -80,8 +80,8 @@ you create the resource.`,
 				Type:         schema.TypeString,
 				Optional:     true,
 				ForceNew:     true,
-				ValidateFunc: validation.StringInSlice([]string{"GCE_VM_IP_PORT", ""}, false),
-				Description:  `Type of network endpoints in this network endpoint group. Default value: "GCE_VM_IP_PORT" Possible values: ["GCE_VM_IP_PORT"]`,
+				ValidateFunc: validation.StringInSlice([]string{"GCE_VM_IP", "GCE_VM_IP_PORT", ""}, false),
```
melinath, line 83: From the [API docs](https://cloud.google.com/compute/docs/reference/rest/beta/networkEndpointGroups) it looks like this has a number of potential values: GCE_VM_IP_PORT, NON_GCP_PRIVATE_IP_PORT, INTERNET_FQDN_PORT, INTERNET_IP_PORT, SERVERLESS, PRIVATE_SERVICE_CONNECT (but `GCE_VM_IP` doesn't seem to be an option)

But regardless - usually in cases like this (where the set of potential values changes frequently) we like to remove the validation on the Terraform side, so that we're not preventing users from using all available values that the API provides.

google/resource_compute_network_endpoint_group.go
```diff
@@ -80,8 +80,8 @@ you create the resource.`,
...
+				Description:  `Type of network endpoints in this network endpoint group. Default value: "GCE_VM_IP_PORT" Possible values: ["GCE_VM_IP", "GCE_VM_IP_PORT"]`,
```
melinath, line 84: Updating the docs here & in the related markdown file is great, but the list should match what's in the [API docs](https://cloud.google.com/compute/docs/reference/rest/beta/networkEndpointGroups)

melinath (changes requested):
@irwanshofwan thanks for your contribution! It looks like there are a few small things it would be great to address.

google/resource_compute_network_endpoint.go
irwanshofwan: I've try this changes to and I think any API bug in google, so as documentation if we set the Network Endpoint type as "GCE_VM_IP" we don't need port and will use default_port variable which set in Network Endpoint Group.

But In console, port input is required (Different with the docs). 

The another thing when we implement the changes (modified provider) to infrastructure, the provider will set port to 0 and the network endpoint still created (it's diff with the spec of port which must greater than zero in console). 
So we can conclude that GCP not filter the API when creation by API and have different behaviour with docs. The proven is when we want to destroy the resource via terraform, it will turns to error regarding zero port.


Here's the docs I refer to https://cloud.google.com/load-balancing/docs/negs/setting-up-zonal-negs

website/docs/r/compute_network_endpoint.html.markdown
irwanshofwan: Here's the docs I refer to https://cloud.google.com/load-balancing/docs/negs/setting-up-zonal-negs

google/resource_compute_network_endpoint_group.go
irwanshofwan: it's little bit different with https://cloud.google.com/load-balancing/docs/negs

melinath: Actually, looking into this some more, I think this validation is important - load balancers are weird.

google/resource_compute_network_endpoint_group.go
melinath: This list should not match what's in the API docs; load balancers are weird, and this resource only corresponds to one type of load balancer.

google/resource_compute_network_endpoint.go
melinath: Okay... I _think_ this probably needs to be computed as well as optional.

website/docs/r/compute_network_endpoint.html.markdown
melinath: In addition to the change specified above, this docs file also need to be updated to specify the new possible value of GCE_VM_IP.

melinath:
It looks like this will need a test. It might be easier to make this change and add a test by contributing directly to https://github.com/GoogleCloudPlatform/magic-modules/

google/resource_compute_network_endpoint_group.go
irwanshofwan: the another thing maybe the API currently doesn't support `GCE_VM_IP` type for network endpoint resource, but I agree to validate the type if the API doesn't support the API it should be return the API error. Just to avoid the network endpoint resource created with zero port. 

Based on experience if the network endpoint created with zero port it can't remove via IAC and we should remove manually from console and continue with remove network endpoint from state.

irwanshofwan:
> It would be great to add this field to a test. It might be easier to make this change and add a test by contributing directly to https://github.com/GoogleCloudPlatform/magic-modules/

will look into this

google/resource_compute_network_endpoint_group.go
melinath: Sorry, I was wrong about removing the validation. You should keep the change validation and add GCE_VM_IP to the list. Load balancers are unusual so we can't remove this the way we would with other resources.

google/resource_compute_network_endpoint.go
melinath: If you're upstreaming this to magic-modules, you would want to use `default_from_api: true` to indicate that it should be optional & computed. I believe that marking it as computed will probably / hopefully resolve the issue you're seeing with a 0 port? It could also be a problem with Terraform not being able to tell the difference between nil and zero.

melinath (changes requested):
removing from my review queue pending the changes requested above

melinath:
@irwanshofwan are you still working on this PR?

melinath:
Closing this PR for now.

PR #10640, opened 2021-11-26
Correct google_bigquery_data_transfer_config docs

rdeknijf:
`service_account_name` may be called "name" but you need to supply the *email*.

Supplying a name results in a this error: 
```Error: Error creating Config: googleapi: Error 400: Request contains an invalid argument```

### DIFF
--- a/website/docs/r/bigquery_data_transfer_config.html.markdown
+++ b/website/docs/r/bigquery_data_transfer_config.html.markdown
@@ -154,7 +154,7 @@ The following arguments are supported:
 
 * `service_account_name` -
   (Optional)
-  Optional service account *email*. If this field is set, transfer config will
+  Optional service account email. If this field is set, transfer config will
   be created with this service account credentials. It requires that
   requesting user calling this API has permissions to act as this service account.
 
--- a/website/docs/r/bigquery_data_transfer_config.html.markdown
+++ b/website/docs/r/bigquery_data_transfer_config.html.markdown
@@ -154,7 +154,7 @@ The following arguments are supported:
 
 * `service_account_name` -
   (Optional)
-  Optional service account name. If this field is set, transfer config will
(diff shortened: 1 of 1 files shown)


melinath (commented):
It looks like the current documentation matches the [API docs](https://cloud.google.com/bigquery-transfer/docs/reference/datatransfer/rest/v1/projects.locations.transferConfigs/create). While it may be somewhat confusing, we tend to adhere to the terminology used in the API docs.

Is there something that "service account name" could refer to apart from the service account's email address?

rdeknijf:
There's actually three candidates for "name". See the GSA [API docs](https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts).
Google service accounts have a `name` value, but that can be either the `email` or the `accountId` (which is the username part of that same email). And there's `displayName` which is referred to as "name" everywhere in the console.

Adding to the confusion is that the Terraform `google_service_account` [resource](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account#attributes-reference) has four outputs:

- `name` (which is the "fully qualified name", which is another format again) 
- `display_name`
- `email`
- `id`

Of these, *only* `email` is a valid input for `google_bigquery_data_transfer_config.service_account_name`.

Basically it's all very confusing and I suggest we make the docs clearer to save people some time. (I spent a long time debugging the `location` while the problem turned out to be this.) Perhaps not in the way I've now proposed. Maybe just adding a warning at the end of the text is more appropriate.
Let me know if you want me to change this PR to reflect that.

melinath:
Note for self: b/208822038

melinath (changes requested):
I can confirm this behavior & agree it's pretty confusing. The one change I'd request is removing the emphasis on email.

rdeknijf:
There you go. :)

melinath (approved):
LGTM

PR #10647, opened 2021-11-29
Container_cluster: complete the feature of enabling GKE workload metrics

ZachGoldberg:
I believe PR #10321 only updated the docs and perhaps accidentally neglected to update the associated code/tests.  When using the provider version 4.1.0 (which is post PR-10321) I get errors that `WORKLOAD` is not a valid value for `enable_components`.


rileykarson (commented):
Ah- in this repo that's correct that only docs got changed, but in `terraform-provider-google-beta` the new value is supported. See the change at https://github.com/GoogleCloudPlatform/magic-modules/pull/5292/files which applied version guards on the value.

If that's incorrect and the value is at GA, we can remove those (which we'll need to do in the `magic-modules` repo rather than here.

PR #10652, opened 2021-11-30
pubsub_subscription: switch topic.name to topic.id

j0hnsmith:
Topic names are not fully qualified, they're only valid when the topic is in the same project as the subscription. Google Cloud supports subscriptions to topics in other accounts, you just have to use the fully qualified name... which `google_pubsub_topic.example.id` is.

This PR switches all usages of `google_pubsub_topic.example.name` to `google_pubsub_topic.example.id` since the former is only valid in some circumstances whereas the latter is always valid (even for same account).

### DIFF
--- a/website/docs/r/pubsub_subscription.html.markdown
+++ b/website/docs/r/pubsub_subscription.html.markdown
@@ -51,7 +51,7 @@ resource "google_pubsub_topic" "example" {
 
 resource "google_pubsub_subscription" "example" {
   name  = "example-subscription"
-  topic = google_pubsub_topic.example.name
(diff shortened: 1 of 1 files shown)


rileykarson:
Do you mind making these changes against https://github.com/GoogleCloudPlatform/magic-modules? That's a generated files based on sources there. They're sourced from https://github.com/GoogleCloudPlatform/magic-modules/tree/master/mmv1/templates/terraform/examples with one file per example, such as https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/templates/terraform/examples/pubsub_subscription_pull.tf.erb

PR #10668, opened 2021-12-02
[WIP] Allow module output when configuring firewall

jackwhelpton:
Floated as a possible fix for https://github.com/hashicorp/terraform-provider-google/issues/10494, but very much a work in progress.

### DIFF
--- a/google/resource_compute_firewall.go
+++ b/google/resource_compute_firewall.go
@@ -88,10 +88,10 @@ func resourceComputeFirewallSourceFieldsCustomizeDiff(_ context.Context, diff *s
 		_, sasOk := diff.GetOk("source_service_accounts")
 
 		_, tagsExist := diff.GetOkExists("source_tags")
-		// ranges is computed, but this is what we're trying to avoid, so we're not going to check this
+		_, rangesExist := diff.GetOkExists("source_ranges")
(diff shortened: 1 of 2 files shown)


rileykarson:
@slevenick, if you don't mind reviewing this! One other thing to look into is `NewValueKnown` (https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/resource_diff.go#L416-L423), which can help distinguish between values that are known / unknown. Unknown values will get returned as their zero value in a CustomizeDiff, and I forget what GetOkExists will report.

jackwhelpton:
Any updates on this?

slevenick (commented):
So, this seems to not quite work. If a user specifies source_ranges during creation and then removes it from their config, Terraform does not throw an error because it finds that source_ranges exists (due to the computed value being in state)

But, I think this is a better behavior than what currently exists, and I'm not really sure that we can handle this in a better way

jackwhelpton:
Ah, I see the problem. I tested the scenario where no `source_ranges` exists to begin with, but not what happens if it's initially there and then removed. I think I agree that this proposal may be better than the current behavior, as at the moment perfectly valid setups are failing, something that can be seen in the examples for other modules.

I am curious if we can do better, though... I'm new to the provider codebase so I'll take your steer on that.

PR #10670, opened 2021-12-02
Adds support for Public Access Prevention

jackwhelpton:


jackwhelpton:
If this looks good I'll probably need some guidance getting it applied correctly. From a preliminary reading it sounds like I should be modifying this file instead:

https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/resources/resource_storage_bucket.go

and renaming that to `resource_storage_bucket.go.erb` to ensure it applies only to the beta provider. I'm not totally clear yet why the corresponding test file doesn't seem to exist in the `magic-modules` repo.

rileykarson:
That file's at https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/tests/resource_storage_bucket_test.go!

jackwhelpton:
Great, thanks. Do you want me to start on a PR in that `magic-modules` repo based on what's here? Does it look correct?

ScottSuarez:
taking a look now ~ thanks for your patience

google/resource_storage_bucket.go
```diff
@@ -1119,13 +1129,19 @@ func expandBucketWebsite(v interface{}) *storage.BucketWebsite {
...
 
 func expandIamConfiguration(d *schema.ResourceData) *storage.BucketIamConfiguration {
-	return &storage.BucketIamConfiguration{
+	cfg := &storage.BucketIamConfiguration{
 		ForceSendFields: []string{"UniformBucketLevelAccess"},
 		UniformBucketLevelAccess: &storage.BucketIamConfigurationUniformBucketLevelAccess{
 			Enabled:         d.Get("uniform_bucket_level_access").(bool),
 			ForceSendFields: []string{"Enabled"},
 		},
 	}
```
ScottSuarez, line 1138: This is changing existing behavior... we should craft the object manually so that we only configure the object with these properties  if `UniformBucketLevelAccess` is present on the config. I recommend checking `uniform_bucket_level_access` and configuring it appropriately.

ScottSuarez (changes requested):
Only concerns that I have currently as long as tests pass. If you could make a pull request at the upstream repository as asked and tag me that would be most appreciated.

google/resource_storage_bucket.go
jackwhelpton: Isn't that what the code currently does at L592? `expandIamConfiguration` is only called if:

```golang
	if d.HasChange("uniform_bucket_level_access") || d.HasChange("public_access_prevention") {
		sb.IamConfiguration = expandIamConfiguration(d)
	}
```

This check was already there, I added the test for `public_access_prevention`.

The tests look promising locally, although the CI/CD workflow looks like it still needs approval. I'll prepare the upstream changes now.

jackwhelpton: Raised https://github.com/GoogleCloudPlatform/magic-modules/pull/5519 and tagged you there. Thanks!

ScottSuarez:
I've requested the relevant changes and expanded on my requests there... closing this pr as it is superseded by the new one.

PR #10674, opened 2021-12-02
Increase to warning in markdown in google_folder_iam.html.markdown

brandonbirdj:
Match the warning on https://github.com/hashicorp/terraform-provider-google/blob/master/website/docs/r/google_organization_iam.html.markdown?plain=1#L26 since it is warning of the same loss of permission.


slevenick (approved):
Looks fine to me, I'll just upstream this to our code generation repo so it gets preserved before merging

Issue #10559, opened 2021-11-15
dataflow: DRAINING should be treated as a terminal state if lifecycle.create_before_destroy is true


n-oden:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

Consider a dataflow job with `lifecycle.create_before_destroy` set to `true`.  If a plan is applied that forces the replacement of the resource, terraform will first attempt to create the replacement job (presumably/hopefully there is something interpolated into the job name in order to allow the job to exist in parallel with its replacement) and then issues the drain command to the old job... and waits.  And waits.  Plan execution will not finish until the job reaches the DRAINED state, and there are multiple scenarios (mostly involving windowing) in which this could take hours, plural.  In the case of a terraform cloud or atlantis workspace, this potentially blocks any other use of the workspace for a substantial amount of time.

But for the purposes of the user, there is no reason to wait: the replacement job has been created.  The old job is in state DRAINING, and it will either reach DRAINED or FAILED depending on the local weather, but there is no way for it to revert to RUNNING.

I think it would be reasonable to treat `DRAINING` as a terminal state, delete the old job from tf state and move on at this point.

### New or Affected Resource(s)

* google_dataflow_job
* google_dataflow_flex_template_job

### Potential Terraform Configuration

As noted above, I think this should simply be the default behavior, but potentially this could also be a configurable boolean, e.g. something like:

```tf
resource "random_id" "job_suffix" {
  byte_length = 4
  keepers = {
    region = var.region
    tmpl = var.myjob_gcs_tmpl
  }
}

resource "google_dataflow_job" "myjob" {
  name = "myjob"
  template_gcs_path = var.myjob_gcs_tmpl
  region = var.region
  on_delete = "drain"
  wait_for_drain = false
  lifecycle {
    create_before_destroy = true
  }
}
```

n-oden:
[...]

PR #10728, opened 2021-12-10
Add `skip_wait_on_job_termination` option for dataflow job resources

n-oden:
This addresses https://github.com/hashicorp/terraform-provider-google/issues/10559

See also https://github.com/hashicorp/terraform-provider-google-beta/pull/3868

### DIFF
--- a/google/resource_dataflow_job.go
+++ b/google/resource_dataflow_job.go
@@ -204,6 +204,13 @@ func resourceDataflowJob() *schema.Resource {
 				Optional:    true,
 				Description: `Indicates if the job should use the streaming engine feature.`,
 			},
+
+			"skip_wait_on_job_termination": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: `If true, treat DRAINING and CANCELLING as terminal job states and do not wait for further changes before removing from terraform state and moving on. WARNING: this will lead to job name conflicts if you do not make sure that the job names are different, e.g. by using a random_id.`,
+			},
 		},
 		UseJSONNumber: true,
 	}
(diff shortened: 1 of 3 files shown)


n-oden:
@ScottSuarez Happy new year!  Did you have any thoughts on this one?

ScottSuarez:
I think I understand the use case but could you elaborate or go into more detail on your change?

n-oden:
@ScottSuarez https://github.com/hashicorp/terraform-provider-google/issues/10559 goes into a lot more detail; the 5s summary is "sometimes draining streaming dataflow jobs takes a ridiculously long time" -- minutes to hours depending on what sort of internal windowing/watermark strategy your job is using.  This means that potentially your terraform workspace is locked for that entire time while waiting for the job to drain.

With this in place, operators will have the option of waiting only until the job reaches the DRAINING state before moving on.  Obviously this has the caveat that if you try to create a replacement job with the same name before the job actually exits it will fail, hence the digression about random_id in the doc examples.

n-oden:
@rileykarson happy new year!  Thoughts on this?

n-oden:
@ScottSuarez @rileykarson nudge?  If this isn't an approach to #10559 that you like, please let me know.

ScottSuarez:
taking over review here. I'll give this a look over soon

google/resource_dataflow_job.go
```diff
@@ -471,6 +478,11 @@ func resourceDataflowJobDelete(d *schema.ResourceData, meta interface{}) error {
 
 	// Wait for state to reach terminal state (canceled/drained/done)
 	_, ok := dataflowTerminalStatesMap[d.Get("state").(string)]
+	if !ok {
+		if d.Get("skip_wait_on_job_termination").(bool) {
+			_, ok = dataflowTerminatingStatesMap[d.Get("state").(string)]
+		}
+	}
```
ScottSuarez, line 485: This logic is repeated several times. Can you extrapolate this to a function

```golang
func isDataflowJobResourceDeleted(state string, skipWait bool) bool {
	_, stopQuery := dataflowTerminalStatesMap[state]
	if !stopQuery && skipWait {
		_, stopQuery = dataflowTerminatingStatesMap[d.Get("state").(string)]
	}
	return stopQuery
}
```

ScottSuarez (changes requested):
Could you also add a test case to the testsuite for this resource and ensure it passes

https://github.com/hashicorp/terraform-provider-google/blob/main/google/resource_dataflow_job_test.go

n-oden:
All reasonable requests!  I'll try to get it sorted out both here and in the beta-provider PR on monday or tuesday.

ScottSuarez:
Can you make the changes against the upstream repository https://github.com/GoogleCloudPlatform/magic-modules

We use this to generate the two version of the provider. Tag me on the PR when you open it

n-oden:
continued in https://github.com/GoogleCloudPlatform/magic-modules/pull/5844

PR #10735, opened 2021-12-13
docs: composer: format list of warnings

twilly:
It appears this block of warnings is a list of warnings that got formatted into
a single block. Adding some newlines breaks it into the bullet points which
helps highlight the GCS bucket limitation and link to known issues.


twilly:
Closing this PR since it's been a year without comment.

PR #10764, opened 2021-12-15
feat: add type to security policy

jasondamour:
https://github.com/hashicorp/terraform-provider-google/issues/10761

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,5 @@
 module github.com/hashicorp/terraform-provider-google
+
 require (
 	cloud.google.com/go/bigtable v1.10.1
 	github.com/GoogleCloudPlatform/declarative-resource-client-library v0.0.0-20211215020606-79616333f950
@@ -9,7 +10,7 @@ require (
 	github.com/gammazero/deque v0.0.0-20180920172122-f6adf94963e4 // indirect
 	github.com/gammazero/workerpool v0.0.0-20181230203049-86a96b5d5d92
 	github.com/golangci/golangci-lint v1.40.1
-	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0 // indirect
+	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
 	github.com/hashicorp/errwrap v1.0.0
 	github.com/hashicorp/go-cleanhttp v0.5.2
 	github.com/hashicorp/go-multierror v1.1.1
@@ -19,7 +20,7 @@ require (
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.5.0
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/hashstructure v1.1.0
-	github.com/sirupsen/logrus v1.8.1 // indirect
+	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/afero v1.2.2 // indirect
 	github.com/zclconf/go-cty v1.5.1 // indirect
 	golang.org/x/mod v0.5.0 // indirect
--- a/google/resource_compute_security_policy.go
+++ b/google/resource_compute_security_policy.go
@@ -53,6 +53,14 @@ func resourceComputeSecurityPolicy() *schema.Resource {
 				Description: `The project in which the resource belongs. If it is not provided, the provider project is used.`,
 			},
 
+			"type": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Default:     "CLOUD_ARMOR",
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -9,7 +10,7 @@ require (
 	github.com/gammazero/deque v0.0.0-20180920172122-f6adf94963e4 // indirect
 	github.com/gammazero/workerpool v0.0.0-20181230203049-86a96b5d5d92
 	github.com/golangci/golangci-lint v1.40.1
-	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0 // indirect
+	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
```
aucampia, line 13: Are the changes to this file needed for your change? If not it may be best to leave it as was.

google/resource_compute_security_policy.go
```diff
@@ -53,6 +53,14 @@ func resourceComputeSecurityPolicy() *schema.Resource {
 				Description: `The project in which the resource belongs. If it is not provided, the provider project is used.`,
 			},
 
+			"type": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
```
aucampia, line 59: Should this be true?

From this it seems it is possible to update it: https://cloud.google.com/compute/docs/reference/rest/beta/securityPolicies/patch

Maybe a comment explaining why ForceNew is true is warranted.

google/resource_compute_security_policy.go
```diff
@@ -53,6 +53,14 @@ func resourceComputeSecurityPolicy() *schema.Resource {
...
+				Default:     "CLOUD_ARMOR",
```
aucampia, line 60: It may be best to not specify a default value for this but instead just not send it the value is not set.

google/resource_compute_security_policy.go
```diff
@@ -53,6 +53,14 @@ func resourceComputeSecurityPolicy() *schema.Resource {
 				Description: `The project in which the resource belongs. If it is not provided, the provider project is used.`,
 			},
 
+			"type": {
```
aucampia, line 56: Given this is an enum, a validateFunc for enum may be the best option.

aucampia: See this for an example: https://github.com/hashicorp/terraform-provider-google/blob/9eef5a485ec52fa52fcfd11e2902dd062c7a0b37/google/resource_cloudbuild_trigger.go#L417-L422

ScottSuarez:
Waiting for response to aucampia's questions before reviewing myself

google/resource_compute_security_policy.go
jasondamour: I've tried updating policies from edge to backend, or vice versa. Neither has worked. I've opened a case with GCP support to see if the docs are wrong, or if I've used it incorrectly

google/resource_compute_security_policy.go
jasondamour: Added validation in latest commit

go.mod
jasondamour: Not intended, reverted in latest commit

google/resource_compute_security_policy.go
jasondamour: Removed default

google/resource_compute_security_policy.go
jasondamour: @aucampia FYI, gcp support confirmed that it cannot be updated in place.  

> This is Shrikant from the GCP Support Networking team based in Sydney. I will be assisting you further with this case during APAC hours.
> 
> Thank you for providing a detailed description of the issue. I appreciate you taking the time to fill this out.
> 
> Looking at the equivalent gcloud command (documented at [1]), we can see that the type of the policy cannot be updated.
> Similarly, the Cloud Console UI also does not enable editing the type of the policy once created.
> 
> Therefore, I believe that this wouldn't be supported through the APIs either.
> 
> I do agree with you, that our documentation should reflect this and additionally, I believe the API response should reflect this as well.
> 
> I will reach out to the product team and keep you updated with their response.
> I hope to have an update for you within the next 48 hours.
> 
> Please let me know if you have any queries or concerns regarding the above.
> 
> Thanks and regards,
> Shrikant
> 
> [1] https://cloud.google.com/sdk/gcloud/reference/compute/security-policies/update

jasondamour:
@ScottSuarez @aucampia Addressed all comments, can you please review again? Specifically, is there somewhere I didn't add code that needs it? Thanks

go.mod
```diff
@@ -1,4 +1,5 @@
 module github.com/hashicorp/terraform-provider-google
```
ScottSuarez, line 1: go mod changes still persist?

jasondamour: My pesky IDE kept making those changes after I saved, no idea why. Definitely resolved now

ScottSuarez (changes requested):
you'll need to send this field on create... see line 215 for how we send `rule`.. We will need to do something similar for `type`.

Did you test this locally?

Can you add a test at https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_compute_security_policy_test.go

ManojPachigolla:
@aucampia  Even, I was stuck with this option "type" to make use of the security policy for GCS backend.
Any timeline for this feature addition to include the option type in "google_compute_security_policy"

Thanks

PR #10766, opened 2021-12-16
Added TF_LOG=TRACE to test run examples

melinath:
I feel like I always have to explain this separately; it seems potentially useful to suggest it as the default that can be excluded instead.


slevenick (approved):
You may want to include information on directing the logs to a file if you include this. Setting TRACE without setting a log file will produce a ridiculous amount of output, especially for a long test like ContainerNodePool

melinath:
+1 that makes sense! I've added redirecting the output to a file.

PR #10862, opened 2022-01-07
Fix some documentation formatting issues

multani:
Wrap Mardown-sensitive content into fixed-size text.


multani:
This should be done in the [magic-modules repository](https://github.com/GoogleCloudPlatform/magic-modules) instead.

PR #10878, opened 2022-01-10
Update f1-micro to e2-micro to be part of the "always free" tier

roopeshvs:
Updating the guide examples to reflect the upgrade to the free-tier VM from f1-micro to e2-micro.


c2thorn:
Hi @roopeshvs,

The upstream source that manages this file is located here: https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/website/docs/guides/getting_started.html.markdown

Would you mind moving your change there?

PR #10893, opened 2022-01-12
Update monitoring_alert_policy.html.markdown

jdegger:
- Add hint that alignment_period must be given in 60s increments.
- Language and wording copied from the `terraform apply` error from Google Cloud:

```
googleapi: Error 400: Field alert_policy.conditions[0].condition_threshold.aggregations[0].alignment_period had an invalid value of "{"seconds":450}": alignment period must be in multiples of one minute
```

### DIFF
--- a/website/docs/r/monitoring_alert_policy.html.markdown
+++ b/website/docs/r/monitoring_alert_policy.html.markdown
@@ -235,7 +235,9 @@ The following arguments are supported:
   The alignment period for per-time
   series alignment. If present,
   alignmentPeriod must be at least
-  60 seconds. After per-time series
+  60 seconds. The alignment period
+  must be given in multiples of one
+  minute. After per-time series
   alignment, each time series will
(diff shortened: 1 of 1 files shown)


shuyama1 (commented):
@jdegger  Thank you for your contribution! Looks like this is a generated file. Would you mind making the change against https://github.com/GoogleCloudPlatform/magic-modules following the instruction [here](https://github.com/GoogleCloudPlatform/magic-modules#making-changes-to-resources). I believe modifying the `description` of the field in the `api.yaml` should update the website documentation for the field in the downstreams https://github.com/hashicorp/terraform-provider-google and https://github.com/hashicorp/terraform-provider-google-beta. Thanks a lot!

shuyama1:
Also you can assign the PR to me for reviewing after you make the change in https://github.com/GoogleCloudPlatform/magic-modules. Let me know if you have any questions!

PR #10935, opened 2022-01-20
docs: Fix a URL in provider's index webpage

olliefr:
Purpose: this PR fixes the link to _GCP Community Slack_ and adds guidance for the new users on how to sign up for it (as it's not obvious).

The old link is 404. Since there is no way for the new users from outside Google to sign up to that Slack Workspace from the login screen, I've also added a link to the Google Developer Center page where an up-to-date public sign-up link can be found. Slack is confusing! :/


slevenick (approved):
Thanks!

PR #10951, opened 2022-01-21
Update google_sql_database_instance documentation

MatthewSBarnes:
`settings.backup_configuration.binary_log_enabled` may be true even if `settings.backup_configuration.enabled` is false.

This behaviour as changed in https://github.com/GoogleCloudPlatform/magic-modules/pull/4907

## Related Material
https://cloud.google.com/sql/docs/mysql/replication#bin-log-replica
https://github.com/GoogleCloudPlatform/magic-modules/pull/4907


megan07 (approved):
LGTM! Thanks!

PR #10955, opened 2022-01-22
docs(cloudsql): add `POSTGRES_14` to version list

utamori:
It is possible to specify v14, but it is missing from the documentation

https://cloud.google.com/sql/docs/postgres/db-versions


wyardley (approved):

c2thorn:
Hi @utamori,
these files are managed in a separate repository. You can find the files at https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown and https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/resources/resource_sql_database_instance.go.erb

PR #11126, opened 2022-02-17
IAM member(s) plan time validation

DrFaust92:
Add plan time validation to all IAM member(s) permutations based on docs:

* https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket_iam#member/members
* https://cloud.google.com/iam/docs/overview#cloud-iam-policy

ACCTests output for example resource:

```
make testacc TEST=./google TESTARGS='-run=TestAccStorageBucketIam' > output.log 

--- PASS: TestAccStorageBucketIamMemberGenerated (20.57s)
--- PASS: TestAccStorageBucketIamBindingGenerated (32.93s)
```

### DIFF
--- a/google/resource_iam_binding.go
+++ b/google/resource_iam_binding.go
@@ -25,7 +25,13 @@ var iamBindingSchema = map[string]*schema.Schema{
 		Elem: &schema.Schema{
 			Type:             schema.TypeString,
 			DiffSuppressFunc: caseDiffSuppress,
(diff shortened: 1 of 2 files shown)


rileykarson:
Those docs are fairly out of date, unfortunately, and there's at least a dozen types. I'm not aware of public docs listing them I can share.

rileykarson:
Probably not, given there's no public documentation on what's supported and I'm not sure how accurate the internal list is either (since some of them may be obselete)

PR #11141, opened 2022-02-22
Handle scheduler job state

lmartelli:
Adds new attribute `state` that can be `PAUSED` or `ENABLED`
Also enables scheduler prior to any update, because a paused job cannot be updated
We should probably validate the `state` value, but I'm not sure what the standard approach for this.

### DIFF
--- a/google/resource_cloud_scheduler_job.go
+++ b/google/resource_cloud_scheduler_job.go
@@ -217,6 +217,11 @@ A duration in seconds with up to nine fractional digits, terminated by 's'. Exam
 				Description: `A human-readable description for the job. 
 This string must not contain more than 500 characters.`,
 			},
+			"state": {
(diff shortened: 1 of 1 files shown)


slevenick (changes requested):
Unfortunately this resource is auto generated, so we won't be able to accept this patch.

Contributions for this resource can be added at https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/products/cloudscheduler/api.yaml but this is an unusual use case where custom code will be needed via a custom update or similar.

I'm a little worried about the implementation enabling jobs that users haven't specified that they want to enable. If a job was created via TF and paused manually we would enable via TF without telling the user what was going to happen

SarahFrench:
Closing as superseded by https://github.com/GoogleCloudPlatform/magic-modules/pull/6304

PR #11142, opened 2022-02-22
Clarify project ID requirement in google_storage_bucket docs

gtirloni:
I had a situation where the provider was not configured with a project ID and I also didn't specify the project in the resource configuration. Terraform failed to apply the update and the API error said that the project ID was a required field. It's optional in Terraform but only because it falls back to using the provider's configured project ID instead. When that does not exist, the field is omitted from the API call and it fails. From Terraform's perspective, a developer could omit the `project` parameter in the resource configuration but not in the API call, which is why I'm proposing to clarify it in this PR.

### DIFF
--- a/website/docs/r/storage_bucket.html.markdown
+++ b/website/docs/r/storage_bucket.html.markdown
@@ -76,8 +76,7 @@ The following arguments are supported:
 
 * `location` - (Required) The [GCS location](https://cloud.google.com/storage/docs/bucket-locations)
 
-* `project` - (Optional) The ID of the project in which the resource belongs. If it
-    is not provided, the provider project is used.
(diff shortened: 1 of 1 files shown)


melinath:
I get what you're saying and I agree there's room for improvement; however, this is the standard language that we use for this field across all resources. Changing it in just one place would probably be more confusing than leaving it as-is. If you feel strongly about this change, you could make a PR to https://github.com/GoogleCloudPlatform/magic-modules/, which is used to generate this repository. You would need to change:

- All the handwritten docs in https://github.com/GoogleCloudPlatform/magic-modules/tree/master/mmv1/third_party/terraform/website
- The docs template at https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/templates/terraform/resource.html.markdown.erb
- Probably also the docs generation template for tpgtools but I'm not sure how it works; you might need to ask your reviewer.

PR #11143, opened 2022-02-22
cloud function: keep source archive object fragment

lmartelli:
This allows to use "source.zip#<md5sum>" and not have the function
updated even when source has not changed


ScottSuarez:
Could you make a pull request against our upstream repository ?
https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/resources/resource_cloudfunctions_function.go

PR #11214, opened 2022-03-03
Nat router dynamic port allocation

rabun788:
https://cloud.google.com/nat/docs/set-up-network-address-translation#set-up-dynamic-port

```
{
    "creationTimestamp": "2021-07-30T08:13:25.581-07:00",
    "id": ".....",
    "kind": "compute#router",
    "name": "nat-name",
    "nats": [
      {
        "enableDynamicPortAllocation": true,
        "enableEndpointIndependentMapping": false,
        "icmpIdleTimeoutSec": 30,
        "logConfig": {
          "enable": true,
          "filter": "ERRORS_ONLY"
        },
...
```
[fixes](https://github.com/hashicorp/terraform-provider-google/commit/6c677a016fb9ced05a8e92e408414977bc96bbe9) https://github.com/hashicorp/terraform-provider-google/issues/11052

### DIFF
--- a/google/resource_compute_router_nat.go
+++ b/google/resource_compute_router_nat.go
@@ -188,6 +188,12 @@ valid static external IPs that have been assigned to the NAT.`,
 see the [official documentation](https://cloud.google.com/nat/docs/overview#specs-rfcs).`,
 				Default: true,
 			},
+			"enable_dynamic_port_allocation": {
+				Type:        schema.TypeBool,
+				Optional:    true,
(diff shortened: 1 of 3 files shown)


c2thorn:
Hi @rabun788,
This resource in managed in Magic Modules from this file: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/compute/api.yaml#L13204

Please make open a PR making the change upstream, as any changes applied here will be overwritten.

PR #11235, opened 2022-03-08
feat(spanner): added support for setting database_dialect on `google_spanner_database`

rahul2393:
Add support for setting database_dialect on google_spanner_database

### DIFF
--- a/google/resource_spanner_database.go
+++ b/google/resource_spanner_database.go
@@ -136,6 +136,13 @@ in the same location as the Spanner Database.`,
 				Computed: true,
 				ForceNew: true,
 			},
+			"database_dialect": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "DATABASE_DIALECT_UNSPECIFIED",
+				ValidateFunc: validateRegexp(`^(DATABASE_DIALECT_UNSPECIFIED|GOOGLE_STANDARD_SQL|POSTGRESQL)\s`),
+				Description:  `The dialect of the Cloud Spanner Database.`,
+			},
 		},
 		UseJSONNumber: true,
 	}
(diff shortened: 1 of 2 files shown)


google/resource_spanner_database.go
```diff
@@ -485,6 +500,13 @@ func flattenSpannerDatabaseInstance(v interface{}, d *schema.ResourceData, confi
 	return ConvertSelfLinkToV1(v.(string))
 }
 
+func flattenSpannerDatabaseDialect(v interface{}, d *schema.ResourceData, config *Config) interface{} {
+	if v == nil {
+		return v
+	}
+	return ConvertSelfLinkToV1(v.(string))
```
aseering, line 507: `ConvertSelfLinkToV1()` cleans URLs.  But database dialects are enums, I didn't think they can contain URLs?  So that function should be a no-op for any valid dialect.  Should this be a no-op like (for example) `flattenSpannerDatabaseState()` above?

website/docs/r/spanner_database.html.markdown
```diff
@@ -100,6 +100,10 @@ The following arguments are supported:
 * `deletion_protection` - (Optional) Whether or not to allow Terraform to destroy the instance. Unless this field is set to false
 in Terraform state, a `terraform destroy` or `terraform apply` that would delete the instance will fail.
 
+* `database_dialect` - (Optional) The dialect of the Cloud Spanner Database.
+  If it is not provided, "DATABASE_DIALECT_UNSPECIFIED" will be used.
```
aseering, line 104: Could I request adding an extra bit of information to this comment?

```
If it is not provided, "DATABASE_DIALECT_UNSPECIFIED" will be used, which will create a Google Standard SQL database.
```

aseering (commented):
Thanks for putting this together!

I left a couple comments.  One general comment -- can/should there be tests?

rahul2393:
Patch be auto-generated by https://github.com/GoogleCloudPlatform/magic-modules/pull/5799

PR #11343, opened 2022-03-24
docs: `name` in `secret_manager_secret` uses project number

jakebiesinger-storyhealth:
The `secret_manager_secret` attribute `name` format does not use the provided `{{project}}` value, but rather always resolves to the project *number*, which often causes plans that appear to be out of date yet re-application changes nothing. E.g., this is the underlying problem in https://github.com/hashicorp/terraform-provider-google/issues/5963.

This PR calls out this behavior explicitly in the docs, so that at least people can see the difference vs `id` and choose accordingly.

### DIFF
--- a/website/docs/r/secret_manager_secret.html.markdown
+++ b/website/docs/r/secret_manager_secret.html.markdown
@@ -177,7 +177,7 @@ In addition to the arguments listed above, the following computed attributes are
 
 * `name` -
   The resource name of the Secret. Format:
-  `projects/{{project}}/secrets/{{secret_id}}`
+  `projects/{{project_number}}/secrets/{{secret_id}}`
 
 * `create_time` -
   The time at which the Secret was created.


jakebiesinger-storyhealth:
Friendly ping?

rileykarson:
Hey! Real sorry about the extremely delayed response here, I completely lost the email(s). 

The import format we expect here _is_ the project id, represented by the `project` field, and the project number mostly works by coincidence. Based on that, I'm going to close this specific PR.

However, I recognise that that fixes issues with the API- it chose to return project numbers when users send project ids based on a misreading of some old AIP guidance that's since been fixed. If you're encountering issues with the secret manager resource(s), I'd recommend filing a bug detailing your expectations and what you think should happen- that'll let us re-evaluate a fix, and/or bring up the feedback to the API team.

PR #11351, opened 2022-03-24
Changed 'displayName' to 'display_name'

jsushetski:
The example for the google_apigee_environment used the argument 'displayName' when it should have been 'display_name', as per the argument reference.


ScottSuarez:
super-seeded by https://github.com/GoogleCloudPlatform/magic-modules/pull/5892

PR #11398, opened 2022-04-04
4.16.0 Patch Notes

ScottSuarez:


CHANGELOG.md
```diff
@@ -1,4 +1,33 @@
 ## 4.16.0 (Unreleased)
+BREAKING CHANGES:
+* composer: made the `google_composer_environment.config.software_config.image_version` field immutable as this functionality is only available in beta. ([#11309](https://github.com/hashicorp/terraform-provider-google/pull/11309))
```
rileykarson, line 3: @megan07 @ScottSuarez: The code change made suggests that _updating_ is only available in beta, is that right? The changelog here suggests the feature as a whole is beta-only.

CHANGELOG.md
```diff
@@ -1,4 +1,33 @@
 ## 4.16.0 (Unreleased)
+BREAKING CHANGES:
```
rileykarson, line 2:
```suggestion
NOTE: We're marked a change in this release as a `BREAKING CHANGE` to indicate that the change may cause undesirable behavior for users in some circumstances. This is done to increase visibility on the change, which otherwise would have been marked under the `BUG FIXES` category, and it is not believed to be a change that breaks the backwards compatibility of the provider requiring a major version change.

BREAKING CHANGES:
```

rileykarson: @megan07 @ScottSuarez: We got some feedback at one point where users were confused why we were marking changes as BCs in minors. I'm a fan of the practice- it increases visibility on risky or technically breaking (but not breaking in practice) changes. WDYT of sending this message to clarify what a BREAKING CHANGE in a minor means?

CHANGELOG.md
ScottSuarez: its available in the GA api. updated the comment correspondingly 

https://cloud.google.com/composer/docs/reference/rest/v1/projects.locations.environments

CHANGELOG.md
ScottSuarez: This is okay with me IMO

rileykarson (approved):

Issue #11537, opened 2022-04-19
Error creating google_monitoring_uptime_check_config with private Service Directory service


mikepietruszka:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Terraform Version
```
$ terraform -v
Terraform v1.1.7
on darwin_amd64
+ provider registry.terraform.io/hashicorp/google v4.17.0
+ provider registry.terraform.io/hashicorp/google-beta v4.17.0
```

### Affected Resource(s)

* google_monitoring_uptime_check_config

### Terraform Configuration Files

```tf
resource "google_monitoring_uptime_check_config" "http" {
  display_name = "http-uptime-check"
  timeout = "60s"
  project  = "test_project_name"

  http_check {
    path = "/web"
    port = "80"
    request_method = "POST"
    content_type = "URL_ENCODED"
    body = "ZmXXXXXXXXXI="
  }

  monitored_resource {
    type = "servicedirectory_service"
    labels = {
      project_id = "test_project_name"
      location = "us-central1"
      namespace_name = "custom_namespace"
      service_name = "web"
    }
  }
}
```

### Debug Output

```
google_monitoring_uptime_check_config.http: Creating...
╷
│ Error: Error creating UptimeCheckConfig: googleapi: Error 400: Configs with Service Directry targets must use checker_type VPC_CHECKERS
│ 
│   with google_monitoring_uptime_check_config.http,
│   on main.tf line 41, in resource "google_monitoring_uptime_check_config" "http":
│   41: resource "google_monitoring_uptime_check_config" "http" {
│ 
```

### Expected Behavior

Private Uptime Check should have been created with Service Directory target. This typically works using `gcloud` command.

### Actual Behavior

[...]

PR #11545, opened 2022-04-19
Added support for checker_type when using Service Directory for Uptime Checks

mikepietruszka:
Added support for checker_type when using Service Directory as a target for Uptime Checks. This fixes issue #11537


melinath:
Duplicate PR - closing

PR #11376, opened 2022-03-31
[DO NOT MERGE] Onboard project to Releases API

jeanneryan:
Configure `terraform-provider-google` to use Releases API

***Note: RelEng  will merge this PR when ready***

PR #11621, opened 2022-05-02
update release to new github actions process

megan07:
I simply copied and pasted from Google Workspace, which has been working appropriately for some time now.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,5 +1,8 @@
 archives:
-  - format: zip
+  - files:
+      # Only include built binary in archive
+      - 'none*'
+    format: zip
     name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}'
 builds:
   - # Special binary naming is only necessary for Terraform CLI 0.12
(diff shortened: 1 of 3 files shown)


bflad:
This may require some coordination with Release Engineering and https://github.com/hashicorp/terraform-provider-google/pull/11376.

Other than that, it looks like the `terraform-registry-manifest.json` file is not in place so this will require some additional changes if we do move forward with it:

- Ensure the `terraform-registry-manifest.json` file is in the root directory with protocol versions set to `["5.0"]`
- Ensure the `.goreleaser.yml` includes the manifest file in `checksums` via `extra_files` configuration
- Ensure the `.goreleaser.yml` includes the manifest file in `publishers` via `extra_files` configuration

c2thorn (approved):

.goreleaser.yml
```diff
@@ -1,8 +1,5 @@
 archives:
-  - files:
-      # Only include built binary in archive
-      - 'none*'
```
bflad, line 4: I'd recommend keeping this configuration, unless there is a specific reason to ship additional files inside the archives (which Terraform CLI will unpack into plugin directories)

bflad (approved):
Looks good to me with one thing to consider, and ensuring folks in #11376 know to rebase afterwards. 🚀

PR #11646, opened 2022-05-05
website/google.erb: Fix the sidebar names

xescugc:


slevenick:
Thanks for bringing this up! I upstreamed this to our code generator so this will stick around, as this specific file is generated

PR #11720, opened 2022-05-18
authenticator_groups_config should not require new resource

derek-burdick:


google/resource_container_cluster.go
```diff
@@ -413,7 +413,7 @@ func resourceContainerCluster() *schema.Resource {
 				Type:        schema.TypeList,
 				Optional:    true,
 				Computed:    true,
-				ForceNew:    true,
+				ForceNew:    false,
```
slevenick, line 416: I believe we would also need to add update support to the `resourceContainerClusterUpdate` method to make this work.

We manage this file in another repo to keep it in sync across both the google and google-beta providers, can you make the changes there instead? https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/resources/resource_container_cluster.go.erb

derek-burdick: Yes, just realized these were generated. I'm working on a PR now. Will also look to cover resourceContainerClusterUpdate if applicable.

slevenick:
Thanks!

google/resource_container_cluster.go
derek-burdick: PR opened. Still testing.
https://github.com/GoogleCloudPlatform/magic-modules/pull/6045

PR #11807, opened 2022-06-02
Add MAC to possible values of google_kms_crypto_key purpose attribute

ori-shalom:
https://github.com/hashicorp/terraform-provider-google/issues/11806


melinath:
Closing this PR as the change will need to be made upstream - see above comment.

PR #11870, opened 2022-06-13
Add new dataSourceDatastoreIndex

AndreasBergmeier6176:
This allows for readonly Index usage.
Usage as a resource may not be sufficient since Index creation can take
multiple hours and block Terraform in that time.
Also supports ensuring kind, ancestor and properties have (custom)
valid values.

### DIFF
--- a/google/data_source_datastore_index.go
+++ b/google/data_source_datastore_index.go
@@ -1,0 +1,130 @@
+package google
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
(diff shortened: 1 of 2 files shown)


melinath (changes requested):
@AndreasBergmeier6176 Thanks for your contribution! Since you're adding a new data source, it would be great if you could contribute this code to [Magic Modules](https://github.com/GoogleCloudPlatform/magic-modules/), which will copy the code into both the beta and GA providers.

You will also need to add tests and documentation for the new data source.

Example PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/6042

PR #11889, opened 2022-06-16
The regular expression in doc should be presented as it is

gecko655:
The regular expression in doc `[a-z]([a-z0-9-]{0,61}[a-z0-9])?` should be presented as it is, but it is parsed as a link of markdown format.

example:
スクリーンショット 2022-06-16 15 01 36

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/apikeys_key#name

### DIFF
--- a/google/resource_apikeys_key.go
+++ b/google/resource_apikeys_key.go
@@ -49,7 +49,7 @@ func resourceApikeysKey() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 				ForceNew:    true,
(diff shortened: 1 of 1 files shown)


gecko655:
I read https://github.com/hashicorp/terraform-provider-google/blob/main/.github/CONTRIBUTING.md and I understand that update to the website will be applied by the maintainers.
If I don't understand correctly, let me know and I will generate the update of https://github.com/hashicorp/terraform-provider-google/blob/main/website/docs/r/apikeys_key.html.markdown file.

PR #11931, opened 2022-06-21
read-only fields updateTime or version

nikhil-goenka:


melinath (changes requested):
Thank you for your contribution! This repository & the beta provider are both generated from https://github.com/GoogleCloudPlatform/magic-modules/; could you please make this change in that repository so that it gets generated into both repositories? Thanks!

PR #11966, opened 2022-06-27
Split changelog on major versions

rileykarson:


melinath (changes requested):
LGTM overall

- v3 changelogs includes v4 notes
- CHANGELOG.md still includes pre-v4 notes; unsure if that's intentional or not.

PR #11989, opened 2022-06-30
docs: fix broken example hcl in google_logging_project_sink

kurochan:
fix broken example hcl in google_logging_project_sink


melinath:
Hello! Thanks for your contribution. This codebase (including docs) is generated from https://github.com/GoogleCloudPlatform/magic-modules/, so this change will need to be made there; otherwise it'll get overwritten. This also has the benefit of propagating the change to the beta provider as well.

kurochan:
@melinath Thank you for your guidance. I will try to make the changes there.

PR #12044, opened 2022-07-07
docs: org_policy_policy wrong attribute for project

Eraac:
Small error on `google_org_policy_policy`


shuyama1 (changes requested):
Thanks for making the changes! The code in this repo are generated using https://github.com/GoogleCloudPlatform/magic-modules. Would you mind creating a PR against the upstream repo https://github.com/GoogleCloudPlatform/magic-modules. Also, I left some small comments for the change. Many thanks!

shuyama1:
Okay. Looks like this is a DCL-based resource. The source code need to be modified internally. I will file a bug internally then. Thank you so much for your contribution.

PR #12048, opened 2022-07-07
add support for MatchesPrefix and MatchesSuffix in condition of lifecycle rule (google_storage_bucket)

Rob4t:
Adds Support for [MatchesPrefix and MatchesSuffix](https://cloud.google.com/storage/docs/lifecycle#matchesprefix-suffix) like requested here: https://github.com/hashicorp/terraform-provider-google/issues/11990


slevenick (changes requested):
Can you target the upstream repo at  https://github.com/GoogleCloudPlatform/magic-modules rather than this repository? Making the changes there will allow us to keep the `google` and `google-beta` repos and providers in sync

PR #12187, opened 2022-07-27
chore: fix header of gke_hub_feature resource

DanielMSchmidt:


slevenick:
Can you target this against https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/gke_hub_feature.html.markdown instead?

PR #12192, opened 2022-07-27
docs: vpc-access-connector refd via id not name

hewholived:
The metadata annotations for a cloud run service should refer to the vpc-access-connector via the id not by the name.


ScottSuarez:
not applicable anymore

PR #12203, opened 2022-07-29
Renamed deadline -> timeout for golangci

melinath:
See https://github.com/golangci/golangci-lint/pull/822 for background


megan07 (approved):
LGTM! Thanks!

PR #12258, opened 2022-08-04
Run `go mod download` instead of `go get`

rileykarson:
Running `go get` during builds means that we're able to pull new dependencies not in the `go.sum` file during builds, allowing the two to drift. Other tooling such as TeamCity doesn't do so, causing failures in later build stages.

In theory, this will pull only the versions represented in `go.sum` and move failures to the appropriate place.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -23,7 +23,7 @@ jobs:
 
     - name: Get dependencies
       run: |
-        go get
+        go mod download
         make tools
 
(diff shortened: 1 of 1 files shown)


rileykarson:
Unusual! It seems to still pass, and I'm not sure why. Works (well, fails) on my machine.

rileykarson:
Superseded with https://github.com/hashicorp/terraform-provider-google/pull/12259

PR #12288, opened 2022-08-09
See what fmt errors we're getting

rileykarson:


rileykarson:
Fix in https://github.com/GoogleCloudPlatform/magic-modules/pull/6398

PR #12301, opened 2022-08-11
google_dns_managed_zone: add zone ID attribute

### LINKED ISSUE #7128: Add DNS Managed Zone ID
### Community Note

### Description

Each Managed DNS zone in GCP has unique ID ([see the id field here](https://cloud.google.com/dns/docs/reference/v1/managedZones)). The `google_dns_managed_zone` resource does not have this field (it has [`id` field](https://www.terraform.io/docs/providers/google/r/dns_managed_zone.html#attributes-reference) but it is something different).

I would need to get the `ID` so I can pass it to ExternalDNS controller configuration which can identify DNS zones either by domain or by ID. In cases where I have both private and public zone with the same domain, I need to use the ID. Because the ID is not available in Terraform resource, it is not possible to automate the ExternalDNS deployment.

* google_dns_managed_zone

There should be attribute with managed zone ID, e.g.
```tf
zone_id=1930303173991938365
```

seh:
In [the data source for DNS managed zones](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/dns_managed_zone), include a new "managed_zone_id" attribute reporting the zone's ID, matching the corresponding [`google_dns_managed_zone` resource](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dns_managed_zone)'s [output attribute with the same name](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dns_managed_zone#managed_zone_id).

While we're here, revise the order in which we transcribe the fields from the fetched object to the data source's attributes to match the declaration order in the schema.

Fixes #7128.

### DIFF
--- a/google/data_source_dns_managed_zone.go
+++ b/google/data_source_dns_managed_zone.go
@@ -26,6 +26,12 @@ func dataSourceDnsManagedZone() *schema.Resource {
 				Computed: true,
 			},
 
+			"managed_zone_id": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: `Unique identifier for the resource; defined by the server.`,
+			},
+
 			"name_servers": {
 				Type:     schema.TypeList,
 				Computed: true,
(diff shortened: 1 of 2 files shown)


seh:
Well, now I feel silly having posted this patch here, but it's not clear to me how one tells _MagicModules_ that [this "managedZoneId" field](https://github.com/prateek2408/magic-modules/blob/d62527e575a7caeac45bfed3da17bbdacfdcf290/mmv1/products/dns/api.yaml#L139-L142) should be included in the generated Terraform data source too.

ScottSuarez:
hi @seh 

You can make the change [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/data_sources/data_source_dns_managed_zone.go) in the other repository. Just make sure to enhance whatever tests to ensure your change works !

seh:
> You can make the change [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/data_sources/data_source_dns_managed_zone.go) in the other repository.

Oh, that looks much more like what I expected to have to change. I didn't think to look around in that _mmv1/third_party_ directory.

Do you have any objection to the intention of the change? I assume that this proposed attribute's omission was unintentional.

ScottSuarez:
The change makes sense to me~ ! 
I think it's a good idea to sync up the data source with the resource.

For context this was likely added by a community contributor a long time ago. The datasource could have either drifted away from the resource as it evolved or the not be completely mirrored due to the contributor fulfilling their own requirements.

seh:
Please see GoogleCloudPlatform/magic-modules#6406.

PR #12362, opened 2022-08-23
Add a developer productivity issue type, for issues about our develop…

rileykarson:
…ment process

This'll give non-maintainers like self-service folks a way to file `technical-debt` issues directly


c2thorn (approved):
Small note:
This will insert itself between Bug and Enhancement during https://github.com/hashicorp/terraform-provider-google/issues/new/choose
The ordering seems nice, even though it has been by chance up to now assumedly. https://github.com/orgs/community/discussions/21601 suggests we can order by the filename, I think it would be slightly better to have this one towards the bottom?

rileykarson:
Sounds good! Prefixed 'em all, `0X` for user-facing and `1X` for developer-facing ones.

ScottSuarez:
Taking myself off as its already been reviewed but let me know if you want more feedback !

I've filed a potential issue that might be able to make use of this label.
https://github.com/hashicorp/terraform-provider-google/issues/12377

c2thorn (approved):
LGTM

Issue #10492, opened 2021-11-03
google_sql_database_instance permanent diff when disk_type is set to pd-ssd


jcanseco:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Terraform Version

```
Terraform v1.0.2
on linux_amd64
+ provider registry.terraform.io/hashicorp/google v4.0.0
```
### Affected Resource(s)

* google_sql_database_instance

### Terraform Configuration Files

```tf
resource "google_sql_database_instance" "my-sqlinstance" {
  name             = "my-sqlinstance"
  region           = "us-central1"
  database_version = "POSTGRES_11"
  settings {
    tier = "db-f1-micro"
    disk_type = "pd-ssd"
  }
}
```

### Expected Behavior

Running `terraform plan` after having successfully created the resource using `terraform apply` should not produce any diffs.

### Actual Behavior

The `terraform plan` produced a diff. 

It seems this is because the GCP API converts the value of `settings.disk_type` from `pd-ssd` to `PD_SSD`, thereby resulting in permanent diffs every time `terraform plan` / `terraform apply` is executed.

The same might be true for `pd-hdd` -> `PD_HDD`, but I haven't gotten to verifying it myself.

To be fair, the field description does mention that it accepts `PD_SSD` and `PD_HDD`, but I suppose the GCP API is more liberal with what it accepts. 

Perhaps one potential fix is to add a diff suppress?

### Steps to Reproduce

1. `terraform apply`
2. `terraform plan`

Notice how a diff is produced.

### References

* b/204906473

ScottSuarez:
changing to enhancement... resource working in acc with doc.. potential improvement

melinath:
b/237554378

NA2047:
Suggest closing this issue as it has been solved with the following pull request being merged 

[...]

PR #12374, opened 2022-08-24
Adding DiffSuppress for google_sql_database_instance on disk_type

NA2047:
Potential fix for #10492


melinath (changes requested):
I'm not sure if this will be reusable; for now it might be preferable to have a very specific diff suppress func for this field that uses strings.Replace to convert dashes to underscore - unless the API supports any delimiter?

If a specific diff suppress func would work, it should be defined in the resource file instead of the common file.

Either way - please add some unit tests for the diff suppress func. See common_diff_suppress_test.go for examples. Thanks!

PR #12434, opened 2022-08-31
Update cloud_run_service.html.markdown

hrchu:
No cloud run Anthos example anymore


megan07:
Hi @hrchu ! Thanks for your contribution - I think you'll want to make this change [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/cloudrun/terraform.yaml#L64) as this file is generated from the upstream magic-modules repo. Thanks!

PR #12533, opened 2022-09-10
fix(google_cloud_run_service): correct typo

rayrutjes:


rileykarson:
You can edit this [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/a4f6b611271299f634f5bbda73b4197f01315dec/mmv1/products/cloudrun/terraform.yaml#L62-L64) to make the change!

PR #12666, opened 2022-09-28
Update storage_bucket.html.markdown

msampathkumar:
Just re-org content. Move required attribute(`location`) the top of the Argument Reference section


msampathkumar:
Hi @melinath Sorry for the trouble. I will raise a request to Magic Modules itself. Thanks for checking on this one.

PR #12790, opened 2022-10-14
resource - compute_instance: added link to default metadata values

tibeer:
Signed-off-by: Tim Beermann <beermann@osism.tech>


slevenick:
Can you make this change here instead? https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/compute_instance.html.markdown

That file is used to generate both this & the google-beta provider

tibeer:
@slevenick Done!

PR #12828, opened 2022-10-18
Add SOA to root-level record deletion exception

rtokarek-fastly:
When trying to delete record-sets that are managed by terraform, some record types can't be deleted by DNS and API restrictions.

(It doesn't make sense to delete the Start Of Authority (SOA) or root-level NS records by themselves as they're fundamental to the zone construct – their value can be changed, but they must exist for the zone itself to exist.)

The code previously only pretended deletes for root-level NS records. This change adds SOA to that behavior so that terraform delete/destroy can function properly when SOA is managed.

### DIFF
--- a/google/resource_dns_record_set.go
+++ b/google/resource_dns_record_set.go
@@ -336,14 +336,13 @@ func resourceDnsRecordSetDelete(d *schema.ResourceData, meta interface{}) error 
 
 	zone := d.Get("managed_zone").(string)
 
-	// NS records must always have a value, so we short-circuit delete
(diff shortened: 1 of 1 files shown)


rtokarek-fastly:
See issue https://github.com/hashicorp/terraform-provider-google/issues/12827

rtokarek-fastly:
I see this is built/copied in from here https://github.com/GoogleCloudPlatform/magic-modules/blob/276f1c70094bafdf5c5d66e09330a4755f985776/mmv1/third_party/terraform/resources/resource_dns_record_set.go#L339-L357

I think I need to PR from that.

PR #13197, opened 2022-12-07
Fix user-agent bug due to goreleaser ldflags not overriding value in release process

SarahFrench:
# Description

Partially closes https://github.com/hashicorp/terraform-provider-google/issues/13196

This PR fixes a bug where ldflags didn't work to override the default version value (`dev`). Currently this bug means that the user agent used by the provider doesn't reflect the provider version.

A full path is needed because the value defined isn't in the `main` package. You can see this already in action in the Makefile:

https://github.com/hashicorp/terraform-provider-google/blob/766f86a189d8088e555f6f085d09108c6f2e18e1/GNUmakefile#L14-L15

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -29,7 +29,7 @@ builds:
       - goarch: arm64
         goos: windows
     ldflags:
-      - -s -w -X version.ProviderVersion={{.Version}}
+      - -s -w -X github.com/hashicorp/terraform-provider-google/version.ProviderVersion={{.Version}}
     mod_timestamp: '{{ .CommitTimestamp }}'
 checksum:
   extra_files:


SarahFrench:
In draft until I test this

SarahFrench:
Manual testing steps:

- `cd terraform-provider-google`
- `git checkout v4.45.0`
- Manually edit _.goreleaser.yml_ to have the changes in this repo
- Run `goreleaser build --single-target --debug --rm-dist --skip-validate` to generate a release package in a _dist_ folder
- Perform `terraform apply` with a developer override pointing at the folder created in the previous step while using a proxy to intercept outgoing API calls

I can see the expected header given that I was using v4.45.0 of the provider:

Screenshot 2022-12-08 at 13 37 08

I'll mark this PR as ready for review, and this test should be sufficient evidence for the TPGB version of this PR too

rileykarson (approved):
I've confirmed the string works locally as well, with `$ go install -v -ldflags="-s -w -X github.com/hashicorp/terraform-provider-google-beta/version.ProviderVersion=5.0.0"` and examining the installed provider's request with `TF_LOG=DEBUG` rather than a proxy: `User-Agent: Terraform/1.3.2 (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google-beta/5.0.0`

Additionally, the string matches https://github.com/hashicorp/terraform-provider-google-beta/blob/2082b2dd1b46f5b81e86470ef665042c5a458fd0/GNUmakefile#L15 which gives me: `User-Agent: Terraform/1.3.2 (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google-beta/acc`

rileykarson:
(well, I tested in beta locally and this is the ga provider repo. Same thing minus the `-beta` bit)

PR #13415, opened 2023-01-07
Use "Cloud Run" instead of "Cloud Run (2nd gen)"

### LINKED ISSUE #13414: "Cloud Run (2nd gen)" is not an approved name, should be simply "Cloud Run"
In the left nav of the [official terraform docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_v2_service), I can see two categories: 
* Cloud Run
* Cloud Run (2nd gen)

image

I am not sure where "Cloud Run (2nd gen)" comes from, but it is not a name that is used anywhere else or even approved by Google Cloud branding.

The `v2` of Cloud Run Terraform resources (`google_cloud_run_v2_service` and `google_cloud_run_v2_job` should go under the Cloud Run category.

steren:
Fix #13414

### DIFF
--- a/website/docs/r/cloud_run_v2_job.html.markdown
+++ b/website/docs/r/cloud_run_v2_job.html.markdown
@@ -12,7 +12,7 @@
 #     .github/CONTRIBUTING.md.
 #
 # ----------------------------------------------------------------------------
-subcategory: "Cloud Run (2nd gen)"
+subcategory: "Cloud Run"
 page_title: "Google: google_cloud_run_v2_job"
 description: |-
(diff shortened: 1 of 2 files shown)


melinath (changes requested):
The subcategory needs to include the fact that this resource uses the v2 API, but it doesn't need to use the phrasing "2nd gen" since this is not in fact a completely different product in this case.

steren:
https://github.com/GoogleCloudPlatform/magic-modules/pull/7098 went in.

If further changes are required, please follow up.

PR #13664, opened 2023-02-06
Standardized issue template affected resource list

melinath:
Made test failure template include "Affected resources" block so that we can extract that information easily. Also added instructions on using google_*.


melinath:
This is intentional :-)

roaks3 (approved):
LGTM

PR #13753, opened 2023-02-15
Cloud Functions: Doc updates

AmandaHassoun:
- I'm removing the `depends_on` statement on lines 425 and 494 for the **Cloud Functions Gen2 doc** because they're unnecessary since we're implicitly depending on them when we're referencing them on lines 421 and 490 respectively. 
- For the regular **Cloud Functions docs**, it would be nice to have an example using `secret_environment_variables` (it wasn't super straightforward to me when I first read through).


ScottSuarez (approved):
Seems reasonable ! :)

PR #13755, opened 2023-02-15
google_container_node_pool's pod_range should be `computed` to prevent diff

cgetzen:
As an example, if I create a pool with the following network config, it will not be in-state on a subsequent plan, due to the pod_range being generated by GCP:
```
  network_config {
    create_pod_range     = true
    enable_private_nodes = true
    pod_ipv4_cidr_block  = "192.168.80.0/20"
  }
```


ScottSuarez:
Could you create a ticket/issue for this?

The change as is would conflict as Force_New and Computed aren't compatible flags

ScottSuarez:
completed by https://github.com/GoogleCloudPlatform/magic-modules/pull/6860

PR #13773, opened 2023-02-16
issue template: Replace nightly build with test history 

ScottSuarez:
Nightly build links die after some time.. Test history link are much preferable so the data don't disappear.


.github/ISSUE_TEMPLATE/10_test_failure.md
```diff
@@ -23,9 +23,9 @@ title: "Failing test(s): TestAccWhatever"
 
 * google_XXXXX
 
-### Nightly builds
+### Teamcity test history
```
melinath, line 26:
```suggestion
### Nightly build test history
```

melinath (approved):
👍

PR #13953, opened 2023-03-09
Fix headline of google_iam_workload_identity_pool

muffl0n:
Move \ to the correct position.

### DIFF
--- a/website/docs/d/iam_workload_identity_pool.html.markdown
+++ b/website/docs/d/iam_workload_identity_pool.html.markdown
@@ -4,7 +4,7 @@ description: |-
   Get a IAM workload identity pool from Google Cloud
 ---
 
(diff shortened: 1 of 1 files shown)


rileykarson:
Hey @muffl0n! As mentioned by the bot, would you be able to submit this modification directly against https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/d/iam_workload_identity_pool.html.markdown? While you're there, I think https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/d/iam_workload_identity_pool_provider.html.markdown has the same issue as well!

muffl0n:
Sure thing! https://github.com/GoogleCloudPlatform/magic-modules/pull/7449

PR #14048, opened 2023-03-21
Update compute_subnetwork.html.markdown to support Gateway API

dwilliams782:
I'm just following the guide to deploy [Gateway API](https://cloud.google.com/kubernetes-engine/docs/how-to/deploying-gateways#configure_a_proxy-only_subnet) and noticed this is missing in the docs.


dwilliams782:
Just realised this should be updated [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/compute/Subnetwork.yaml) instead.

PR #14108, opened 2023-03-27
Formatting issue in compute_instance.html.markdown

ryanaross:
Simple formatting issue in this document:
image


c2thorn:
Hi @ryanaross,
Changes made directly to this repository will be clobbered by Magic Modules. The upstream file you would want to make this change to can be found here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/compute_instance.html.markdown

PR #14113, opened 2023-03-28
typo: auto_restart -> automatic_restart

samtsai15:


melinath:
This change will need to be contributed via https://github.com/GoogleCloudPlatform/magic-modules. Thanks!

PR #14175, opened 2023-04-01
Illustrate how a Pubsub Subscription `ttl` would be specified as indefinite

ethankent:
Because of the way a human might skim this doc, some are likely to wind up with the wrong default value. Imagine that someone has gone straight to the `ttl` section and missed the reference to a 31 day default in the `expiration_policy` section. They would likely then create an empty `expiration_policy` block and get a validation error. After removing that block altogether the validation error would disappear, but they would be left with a 31 day expiration rather than the indefinite one they had desired.

### DIFF
--- a/website/docs/r/pubsub_subscription.html.markdown
+++ b/website/docs/r/pubsub_subscription.html.markdown
@@ -411,7 +411,7 @@ The following arguments are supported:
   (Required)
   Specifies the "time-to-live" duration for an associated resource. The
   resource expires if it is not active for a period of ttl.
-  If ttl is not set, the associated resource never expires.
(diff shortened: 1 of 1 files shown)


SarahFrench:
Hi @ethankent - could you please open a new PR in the upstream GoogleCloudPlatform/magic-modules repo for this change? I believe [you'd need to update this line](https://github.com/GoogleCloudPlatform/magic-modules/blob/2a32beb359a669581cfad2af5b98d072cc577d41/mmv1/products/pubsub/Subscription.yaml#L383) to cause the documentation to be updated like in this PR. Thanks!

PR #14234, opened 2023-04-05
Move scripts to a Go submodule

sethvargo:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/14218

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,5 @@
 module github.com/hashicorp/terraform-provider-google
+
 go 1.19
 
 require (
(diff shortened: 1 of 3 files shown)


sethvargo:
@slevenick @rileykarson

rileykarson:
@slevenick do you mind taking a look at this? We probably need to update MM in some way otherwise it'll think it owns these files. It'd be nice for this gotcha to go away.

sethvargo:
> ...gotcha to go away

_Go_ away. Good pun, well done.

PR #14568, opened 2023-05-11
Add field path to google_storage_transfer_job for s3

kiwamizamurai:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/12997


melinath:
Thanks for your contribution! However, this change will need to be contributed via https://github.com/GoogleCloudPlatform/magic-modules.

PR #14681, opened 2023-05-23
Update labeler.yml to address node12 deprecation

SarahFrench:
Closes https://github.com/hashicorp/terraform-provider-google/issues/14625

Affected action: https://github.com/hashicorp/terraform-provider-google/actions/workflows/labeler.yml

Updating actions/checkout from v2 to v3.5.2

Updating github/issue-labeler from v2.3 to v2.6

Related: https://github.com/hashicorp/terraform-provider-google/pull/14387

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -8,8 +8,8 @@ jobs:
   triage:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


SarahFrench:
Just seen that v2.6 of github/issue-labeler would also allow moving away from node12 without needing to navigate the breaking change in v3+

PR #14682, opened 2023-05-23
Update lock.yml to address node12 deprecation

SarahFrench:
Closes https://github.com/hashicorp/terraform-provider-google/issues/14624

Affected action : https://github.com/hashicorp/terraform-provider-google/actions/workflows/lock.yml

Updating dessant/lock-threads to v4.0.0 addresses the node12 deprecation - [see CHANGELOG here](https://github.com/dessant/lock-threads/blob/master/CHANGELOG.md#400-2022-12-04).

As we're updating from v2.x to v4.x we [need to change the inputs to address breaking changes between v2.x and v3.x](https://github.com/dessant/lock-threads/blob/master/CHANGELOG.md#300-2021-09-27)

Related: https://github.com/hashicorp/terraform-provider-google/pull/14387

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -8,16 +8,16 @@ jobs:
   lock:
     runs-on: ubuntu-latest
     steps:
-      - uses: dessant/lock-threads@486f7380c15596f92b724e4260e4981c68d6bde6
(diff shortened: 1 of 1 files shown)


megan07 (approved):
lgtm, thanks!

PR #14774, opened 2023-05-31
Update dependencies in issue-comment-created.yml

SarahFrench:
This PR bumps the dependencies in issue-comment-created.yml.

Notably it switches from using actions-ecosystem/action-remove-labels to using hashicorp/action-remove-labels - an internal fork where the necessary updates to use Node 16 are included.

### DIFF
--- a/.github/workflows/issue-comment-created.yml
+++ b/.github/workflows/issue-comment-created.yml
@@ -8,8 +8,8 @@ jobs:
   remove_waiting_response:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
-      - uses: actions-ecosystem/action-remove-labels@556e306654eb4c343ecaff657772edaa3f1add86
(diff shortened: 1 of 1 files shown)


SarahFrench:
Closing because I overlooked that internal forks won't work with a public repo 😢

SarahFrench:
We don't have the ability to create a public fork quickly, so here's a public fork under my account as a temporary solution

SarahFrench:
I did a quick manual test here : https://github.com/SarahFrench/test-remove-label-action/issues/1

After working out issues with how the new repo was set up (GHA workflow permissions) it worked as expected. The Actions UI doesn't show any warnings related to node 12

megan07 (approved):
LGTM, thanks!

SarahFrench:
I'll merge this, but we're working to get a HasiCorp fork of the original action. Once that's happened and we push the upgrade from Node12 => Node16 we will resume using `action-remove-labels` but it'll be `hashicorp/action-remove-labels` instead of `actions-ecosystem/action-remove-labels`.

PR #14847, opened 2023-06-08
Multiple updates

bluPhy:


slevenick:
As noted by the bot, this will need to be targeted at the magic-modules repo to preserve any changes here.

It would be appreciated though, as these look like a lot of typos!

PR #14964, opened 2023-06-23
Fix condition to call api to fetch id token

golemiso:
I use GCE metadata server to authenticate and want to get id token.

```
  data "google_service_account_id_token" "oidc" {
    target_audience = "https://foo.bar/"
  }
```

The former condition `if creds.JSON == nil` prevents me from passing the line `idtoken.NewTokenSource()` and returns incomprehensible error message below: 

```
Error: error calling iamcredentials.GenerateIdToken: googleapi: got HTTP response code 404 with body: 

  with data.google_service_account_id_token.oidc,
  on provider.tf line 6, in data "google_service_account_id_token" "oidc":
   6: data "google_service_account_id_token" "oidc" {
```

I think this can be fixed by that the API is called only if target_service_account is provided.


golemiso:
https://github.com/GoogleCloudPlatform/magic-modules/pull/8195

PR #14991, opened 2023-06-27
Update Replacement Method on Example

joshw123:
Using the Example on the Documentation you get the following error 

`│ Error: Error updating region managed group instances: googleapi: Error 400: Invalid value for field 'resource.updatePolicy.maxSurge': '{  "percent": 20}'. maxSurge must be equal to 0 when replacement method is set to RECREATE, invalid`

I have updated the replacement method so that it can be copied directly


joshw123:
Made PR here instead - https://github.com/GoogleCloudPlatform/magic-modules/pull/8210

Issue #15116, opened 2023-07-10
Support keepers for google_apikeys_key resource


jawnsy:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

The `google_apikeys_key` resource is missing a `keepers` field in the schema, which means it's impossible to automatically re-create the key periodically.

The `google_service_account_key` resource has a `keepers` field, so could serve as a good example: https://github.com/hashicorp/terraform-provider-google/blob/3894440287ec5e194faa594b1eaf30f30953997f/google/services/resourcemanager/resource_google_service_account_key.go#L61-L66

### New or Affected Resource(s)

* google_apikeys_key

### Potential Terraform Configuration

```tf
resource "time_rotating" "ninety_days" {
  rotation_days = 90
}

resource "google_apikeys_key" "key" {
  keepers = {
    rotation_time = time_rotating.ninety_days.rotation_rfc3339
  }
}
```

### References

* #0000

rileykarson:
I don't think resource-specific keepers are required now that https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#replace_triggered_by is available. Have you tried that and has/has not that worked for you?

PR #15118, opened 2023-07-10
Add keepers to apikeys_key resource

jawnsy:
This adds a "keepers" field (copied from service_account_key) to support automatic rotation using the time_rotating resource.

Closes: #15116

### DIFF
--- a/google/services/apikeys/resource_apikeys_key.go
+++ b/google/services/apikeys/resource_apikeys_key.go
@@ -132,6 +132,13 @@ func ApikeysKeyRestrictionsSchema() *schema.Resource {
 				Elem:        ApikeysKeyRestrictionsIosKeyRestrictionsSchema(),
 			},
 
+			"keepers": {
+				Description: "Arbitrary map of values that, when changed, will trigger recreation of resource.",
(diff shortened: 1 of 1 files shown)


google/services/apikeys/resource_apikeys_key.go
```diff
@@ -132,6 +132,13 @@ func ApikeysKeyRestrictionsSchema() *schema.Resource {
 				Elem:        ApikeysKeyRestrictionsIosKeyRestrictionsSchema(),
 			},
 
+			"keepers": {
```
jawnsy, line 135: This service doesn't look like it's being generated in magic-modules, as far as I can tell: https://github.com/GoogleCloudPlatform/magic-modules/blob/f7457595c9214bdafd57a70898b65b948105ed0e/mmv1/third_party/terraform/provider/provider.go.erb#L468-L474

slevenick:
It's generated from here: https://github.com/GoogleCloudPlatform/declarative-resource-client-library/tree/main/services/google/apikeys

jawnsy:
@slevenick Thanks so much for taking a look and the pointer!

PR #15124, opened 2023-07-11
Update TF registry docs to have user_labels nested under settings and…

AmandaHassoun:
… not settings.advanced_machine_features.

I was getting the following failure originally:

```
╷
│ Error: Unsupported argument
│
│   on modules/grafana/sql.tf line 30, in resource "google_sql_database_instance" "instance":
│   30:       user_labels = {
│
│ An argument named "user_labels" is not expected here.
```

Until I looked through the JSON API spec and realized it's supposed to only be nested under `settings` (same level as `tier`).


slevenick:
Can you make this against https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown instead? That's the source of truth for the files in this repo

PR #15129, opened 2023-07-11
Add self service labeler

trodge:


.github/workflows/self-service-labeler.yml
```diff
@@ -0,0 +1,40 @@
+name: Issue Opened Self-Service Labeler
+
+on:
```
roaks3, line 3: We're you able to do any testing with this? It might be helpful for testing/one-offs to add a manual option where the user provides an issue id (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs). No worries if that would make more sense in a separate PR, I just brought it up here for the testing benefit.

.github/workflows/self-service-labeler.yml
```diff
@@ -0,0 +1,40 @@
...
+  issues:
+    types: [opened]
+
+jobs:
+  triage:
```
roaks3, line 8: I believe this is the name of the job. Could maybe change to something like `label` or `add-labels`?

roaks3 (commented):
Overall code looks good 👍 . Since this would immediately start running on all new tickets, I would just want to make sure:
1. We've tested the workflow in some way
2. We alert the team before we merge so they know about the change, and be ready to revert if something goes wrong

.github/workflows/self-service-labeler.yml
trodge: I tested in my fork here: https://github.com/trodge/terraform-provider-google/issues

.github/workflows/self-service-labeler.yml
trodge: Done.

.github/workflows/self-service-labeler.yml
roaks3: 👍 awesome thanks!

roaks3 (approved):

.github/workflows/self-service-labeler.yml
```diff
@@ -0,0 +1,40 @@
...
+        go-version: '>=1.19.0'
+    - name: Pick labels
+      id: pick_labels
+      env:
+        ISSUE_BODY: ${{ github.event.issue.body }}
+      run: |
+        cd magic-modules/tools/issue-labeler
+        echo "labels=$(go run .)" >> $GITHUB_OUTPUT
+    - name: Apply labels
+      uses: actions/github-script@v6
```
SarahFrench, line 31:
```suggestion
      uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410 # v6.4.1
```

This is a different version (6.0.0 => 6.4.1). At HashiCorp the 'approved' versions of this GHA are currently v6.4.0 and v6.4.1. Either I can go through the process of getting v6.0.0 approved or we can upgrade the version we use here to match one that's already approved.

I've just made the same change for a different GHA in this repo, here: https://github.com/hashicorp/terraform-provider-google/pull/15194

SarahFrench (commented):
Nice! This'll be a big help for triage.

The changes here look good to me, but I've made some suggested changes so that actions are referenced via commit hashes instead of via tags. The only change in version number is for `actions/github-script`, which shouldn't result in any changes to your PR's behaviour but might be worth a manual test to be sure?

.github/workflows/self-service-labeler.yml
```diff
@@ -0,0 +1,40 @@
...
+    types: [opened]
+
+jobs:
+  add-labels:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+    - name: Checkout magic modules
+      uses: actions/checkout@v2
```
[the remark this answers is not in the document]
SarahFrench: This version of checkout is used in other GHAs for this repo, so I've bumped it so it matches the others

SarahFrench (approved):
Thanks for the change to referencing actions via commit SHA, LGTM

PR #15270, opened 2023-07-24
Fix MD link

jrdnbradford:
Small fix to a markdown link.


melinath:
Thanks for opening this PR! However, this file is managed by Magic Modules, so you'll need to make the change here instead: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/index.html.markdown

PR #15289, opened 2023-07-25
Add ability to create assets as Managed.

bradmiro:
Add ability to create Dataplex assets as Managed. Taken from existing API spec: https://cloud.google.com/dataplex/docs/reference/rest/v1/projects.locations.lakes.zones.assets#accessmode


melinath:
Thanks for your work on this! Please see https://googlecloudplatform.github.io/magic-modules/ for information about how to contribute code to this provider using Magic Modules.

PR #15467, opened 2023-08-10
Update template to link to new TC project, and provide info about transition period after migration to TC Cloud

SarahFrench:
Related to https://github.com/hashicorp/terraform-provider-google/issues/15461


SarahFrench:
(I'll create an issue assigned to me for removing some of these notes after the transition period is over)

melinath (commented):
lgtm

PR #15573, opened 2023-08-21
client_secret: should be sensitive.

lomkju:
## WHAT

Adds the field `Sensitive` to the client_secret.

## WHY

This field should be a secret and it should not print during a plan.


rileykarson:
This is a generated resource, we'll want to apply this change through Magic Modules: https://googlecloudplatform.github.io/magic-modules/ / https://googlecloudplatform.github.io/magic-modules/develop/resource/

PR #15677, opened 2023-08-30
Retry 403 error on a Service Account creation

### LINKED ISSUE #15042: Error reading service account after creation - 403 Halts Execution
### Community Note

* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Terraform Version

Running in Github Runner:

provider.google v4.69.1

### Affected Resource(s)

google_service_account

```tf
# Copy-paste your Terraform configurations here.
#
# For large Terraform configs, please use a service like Dropbox and share a link to the ZIP file.
# For security, you can also encrypt the files using our GPG public key:
#    https://www.hashicorp.com/security
#
# If reproducing the bug involves modifying the config file (e.g., apply a config,
# change a value, apply the config again, see the bug), then please include both:
# * the version of the config before the change, and
# * the version of the config after the change.
provider "google" {
  project = var.project_id
  region  = var.region
  zone    = var.zone
}

resource "google_service_account" "service_account" {
  account_id   = var.service_account_id
  display_name = "Constellation service account"
  description  = "Service account used inside Constellation"
}
....
```

Here is the complete configuration:
https://github.com/edgelesssys/constellation/blob/main/cli/internal/terraform/terraform/iam/gcp/main.tf

### Debug Output
Not yet able to reproduce with debug output, since only happens sporadically.
```
Error: Error reading service account after creation: googleapi: Error 403: Permission 'iam.serviceAccounts.get' denied on resource (or it may not exist).
Details:
[
  ***
    "@type": "type.googleapis.com/google.rpc.ErrorInfo",
    "domain": "iam.googleapis.com",
    "metadata": ***
      "permission": "iam.serviceAccounts.get"
    ***,
    "reason": "IAM_PERMISSION_DENIED"
  ***
]
, forbidden

  with google_service_account.service_account,
  on main.tf line 16, in resource "google_service_account" "service_account":
  16: resource "google_service_account" "service_account" ***


Attempting to roll back.
Rollback succeeded.
Error: exit status 1
```

### Expected Behavior

Should create account without error.

### Actual Behavior

Halts due to 403.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

Only fails at times, in the CI pipeline. Authentication with GCP is done with Github Action and roles must be correct since it works most of the times. (https://github.com/edgelesssys/constellation/blob/main/.github/actions/login_gcp/action.yml)

Issue is very similar to #10227.

### References

* #10227
* #6758 

b/298050821

b/291928614

sbocinec:
Retry polling for a service account right after its creation if 403 Forbidden is returned. Trying to get a service account immediately after it was created might fail due to the [IAM API being eventually consistent.](https://cloud.google.com/iam/docs/overview#consistency). 

Given the [`GetServiceAccountRequest`](https://cloud.google.com/iam/docs/reference/rpc/google.iam.admin.v1#getserviceaccountrequest) to IAM Admin API  returns 403 if the service account is not found, we must retry when trying to read (confirm) the SA was indeed created.

This fixes #15042

### DIFF
--- a/google/services/resourcemanager/resource_google_service_account.go
+++ b/google/services/resourcemanager/resource_google_service_account.go
@@ -122,8 +122,11 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
 			_, saerr := config.NewIamClient(userAgent).Projects.ServiceAccounts.Get(d.Id()).Do()
 			return saerr
 		},
-		Timeout:              d.Timeout(schema.TimeoutCreate),
-		ErrorRetryPredicates: []transport_tpg.RetryErrorPredicateFunc{transport_tpg.IsNotFoundRetryableError("service account creation")},
+		Timeout: d.Timeout(schema.TimeoutCreate),
+		ErrorRetryPredicates: []transport_tpg.RetryErrorPredicateFunc{
+			transport_tpg.IsNotFoundRetryableError("service account creation"),
+			transport_tpg.IsForbiddenIamServiceAccountRetryableError("service account creation"),
+		},
 	})
 
 	if err != nil {
--- a/google/transport/error_retry_predicates.go
+++ b/google/transport/error_retry_predicates.go
@@ -484,3 +484,18 @@ func IsSwgAutogenRouterRetryable(err error) (bool, string) {
 
 	return false, ""
 }
+
+// IAM API is eventually consistent and returns 403 Forbidden (instead of 404 Not found) for some operations
+// when a newly created resource is attempted to be read right after and not yet found.
+// Retry if getting a resource/operation returns a 403 for specific IAM Admin API Service Account operations.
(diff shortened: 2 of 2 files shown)


slevenick:
We use another repository to hold these resources so they can be shared between the `google` and `google-beta` provider. Can you make these changes against that instead?

Here's the service account file there: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go

sbocinec:
@slevenick thanks for the heads up - I'm closing this PR and created https://github.com/GoogleCloudPlatform/magic-modules/pull/8815 instead.

PR #15754, opened 2023-09-07
Add content-type header to google_cloud_scheduler_job json examples

mkows:
By default scheduler calls with data as `Content-Type: application/octet-stream`. As examples are with json data, specifying `Content-Type: application/json`.

In helped me resolved an issue when calling GCP Cloud Functions (gen2), so I think that it will be valuable addition to the docs.


ScottSuarez:
upstreaming here
https://github.com/GoogleCloudPlatform/magic-modules/pull/8890

PR #16047, opened 2023-09-28
Add testnolint build rule

trodge:


GNUmakefile
```diff
@@ -11,6 +11,10 @@ build: lint
 test: lint
 	go test $(TESTARGS) -timeout=30s $(TEST)
 
+# Used in CI to prevent lint failures from being interpreted as test failures
+testnolint:
```
rileykarson, line 15: Can we call this step from `test`? That will avoid the two diverging.

trodge: Done

rileykarson (approved):

PR #16135, opened 2023-10-05
Don't fail plan if project,region,zone defaults cannot be retrieved

queeno:
The DefaultProviderProject, DefaultProviderRegion and DefaultProviderZone funcs set respectively the project, region and zone in a diff if one is provided, unless they fetch the defaults from config. However, at times, these values may not be available during a plan, as they are dynamically computed at apply time.
We should avoid throwing an error if this is the case, but rather set an empty value in the diff.

Resolves hashicorp/terraform-provider-google#16133


SarahFrench:
Please instead propose this change against this file in the upstream repo: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/tpgresource/utils.go

Thanks!

PR #16362, opened 2023-10-25
Tweak default date and dry run description for the issue labeler action

roaks3:
This is mainly intended to make usage a bit easier (we almost always use `1973-01-01`, to run against all tickets)


roaks3:
cc @trodge

ScottSuarez (approved):

PR #16386, opened 2023-10-27
Update container_cluster.html.markdown

dr3s:
The current documentation is incorrect.  The GPU sharing and multi-instance config are properties of the accelerator not the driverinstall


dr3s:
Replaced by https://github.com/GoogleCloudPlatform/magic-modules/pull/9374

PR #16481, opened 2023-11-09
Allow updating DNSConfig in-place

wewark:


c2thorn:
Hi @wewark,
The provider code is not managed here, it is managed by Magic Modules: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb

Check out our MM documentation: https://googlecloudplatform.github.io/magic-modules/

PR #16544, opened 2023-11-16
Update google_project_iam_custom_role.html.markdown to say custom role

edhgoose:
I believe that the wording for this section is wrong, and should say "custom role". We're not creating a service account.


edhgoose:
Will do - sorry, I did see the github action note but hadn't got round to it yet.

Opened here: https://github.com/GoogleCloudPlatform/magic-modules/pull/9497

PR #16620, opened 2023-11-29
Update storage_bucket.html.markdown cors safelisted

mkly:


rileykarson (commented):
As mentioned by the robot, this will need to be made against `magic-modules`. https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown is the upstream version of that file.

PR #16861, opened 2023-12-22
Update storage_bucket.html.markdown public_access_prevention

880831ian:
After testing, it was found that when using Google Storage to create a Bucket, the default value of public_access_prevention is enforced, not inherited.

截圖 2023-12-22 下午4 44 04


880831ian:
I've gone to GoogleCloudPlatform/magic-modules and created a new PR
https://github.com/GoogleCloudPlatform/magic-modules/pull/9710

PR #16868, opened 2023-12-26
Documentation Fix compute_security_policy

andrew-grasso:
Fix mixed up descriptions between request_uri and request_query_param.


melinath:
Thanks for your contribution! This file is managed by magic modules, so you would need to make the change there. https://googlecloudplatform.github.io/magic-modules/ has information on the contribution procedure.

PR #16893, opened 2024-01-03
Removed double quotes and typo in google_project_iam_member documenta…

aleksvujic:
Previously, documentation for importing `google_project_iam_member` contained a typo in the `id` section.

I changed it from ...

```
id = ""{{project_id}} roles/viewer user:foo@example.com"m"
```

... to ...

```
id = "{{project_id}} roles/viewer user:foo@example.com"
```

### DIFF
--- a/website/docs/r/google_project_iam.html.markdown
+++ b/website/docs/r/google_project_iam.html.markdown
@@ -228,7 +228,7 @@ An [`import` block](https://developer.hashicorp.com/terraform/language/import) (
 
 ```tf
 import {
-  id = ""{{project_id}} roles/viewer user:foo@example.com"m"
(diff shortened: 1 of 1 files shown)


melinath (changes requested):
Thanks for catching this! This change would need to be made in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/google_project_iam.html.markdown

aleksvujic:
I cloned this pull request to [magic-modules](https://github.com/GoogleCloudPlatform/magic-modules) repository: https://github.com/GoogleCloudPlatform/magic-modules/pull/9750.

PR #16899, opened 2024-01-04
fix syntax error in IAM member import documentation

afeld:
I'm assuming this is coming from [Magic Modules](https://googlecloudplatform.github.io/magic-modules/), but it was unclear to me where. Opening this to show the intended change. Thanks!


aleksvujic:
This is a duplicate of my pull request which was created at 3-Jan-2024 09:13 UTC: https://github.com/hashicorp/terraform-provider-google/pull/16893.

PR #17117, opened 2024-01-26
Fix typos in  index_config usage docs

chrisleavoy:


ttrinh42 (approved):

melinath:
Thanks! However, this change will need to be made in https://github.com/GoogleCloudPlatform/magic-modules.

PR #17120, opened 2024-01-26
Added test issue form for bugs

melinath:
Unfortunately I don't think there's a way to test this without merging. However, if it works as expected, it should make our bug reports somewhat more standardized (and we can do the same for enhancements.) I also cleaned up some issues identified in https://github.com/hashicorp/terraform-provider-google/issues/14822 and made some other small language changes.


SarahFrench (approved):
Looks good! And thanks for doing this. If we can't test it other than merging and seeing what happens I think it's ok; we'll just fix it asap.

rileykarson (approved):

PR #17130, opened 2024-01-29
Fixed problem with issue form

melinath:
I think that the problem is that the `[` was interpreted as starting an array instead of being part of a string? But the error message was not super clear: https://github.com/hashicorp/terraform-provider-google/blob/main/.github/ISSUE_TEMPLATE/12_bug_test_form.yaml

The error does not show up on this branch's version of the file. I also moved it to `.yml` to match the example on https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms just in case that somehow is also an issue.


c2thorn (approved):
couldn't hurt, hopefully fixes

PR #17173, opened 2024-02-03
raise provisioned_iops upper limit to 160,000 for hyperdisks.

xwu2intel:
See https://cloud.google.com/compute/docs/disks/hyperdisks#hd-performance-limits for hyperdisk upper limits.


rileykarson:
Superseded w/ https://github.com/GoogleCloudPlatform/magic-modules/pull/9960/files

PR #17184, opened 2024-02-05
Update permissions argument of file-resource to be optional instead of computed

MahmoudNada0:
Fix for https://github.com/hashicorp/terraform-provider-google/issues/16737

/assign @edwardmedia 
/cc @edwardmedia


SarahFrench:
Please make this change to the equivalent file in the upstream repo (https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/osconfig/resource_os_config_os_policy_assignment.go). Thanks!

PR #17282, opened 2024-02-14
Update cloudfunctions2_function.html.markdown `event_type` doc

meldaravaniel:
Main change:
* `event_type` is required, but said both optional and required.

Minor changes:
* removing redundant "optional," "required," and "output" text

### DIFF
--- a/website/docs/r/cloudfunctions2_function.html.markdown
+++ b/website/docs/r/cloudfunctions2_function.html.markdown
@@ -810,8 +810,10 @@ The following arguments are supported:
   Resource name of a KMS crypto key (managed by the user) used to encrypt/decrypt function resources.
   It must match the pattern projects/{project}/locations/{location}/keyRings/{key_ring}/cryptoKeys/{crypto_key}.
 
-* `project` - (Optional) The ID of the project in which the resource belongs.
-    If it is not provided, the provider project is used.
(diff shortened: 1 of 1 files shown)


meldaravaniel:
I noticed an older PR in this repo that directed folks to the magic modules repo, but I don't see a corresponding file over there.

ScottSuarez:
See here
https://github.com/GoogleCloudPlatform/magic-modules/blob/c4260af062bfbc3dd321ad3990ba3a735d537291/mmv1/products/cloudfunctions2/Function.yaml#L598-L600

For example eventType likely needs property 'required'

SarahFrench:
I'm going to close this PR as it's raised against the incorrect repo. I've created an issue related to how the field should be required here: https://github.com/hashicorp/terraform-provider-google/issues/18961

If you'd like to make this change to the resource (and the documentation as a side effect of changing the resource) then please raise a PR against the magic-modules repo (esp. [this file](https://github.com/GoogleCloudPlatform/magic-modules/blob/c4260af062bfbc3dd321ad3990ba3a735d537291/mmv1/products/cloudfunctions2/Function.yaml#L598-L600)) following our [contribution guides](https://googlecloudplatform.github.io/magic-modules/develop/resource/)

PR #17309, opened 2024-02-17
docs: fix import example in google_project_iam

yuki0920:
## What

Fix the example in the `google_project_iam` documentation.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_project_iam#import


yuki0920:
I  created a PR related to this change.
https://github.com/GoogleCloudPlatform/magic-modules/pull/10007

PR #17315, opened 2024-02-19
docs(Cloud Run): Move v2 datasource to correct subcategory

ppuschmann:
There is a separate subcategory in the navigation for resources and datasources of the Cloud Run v2 API.
This commit assigns the correct subcategory for the datasource of `google_cloud_run_v2_job` and `google_cloud_run_v2_service`


slevenick:
This change will need to be made against the https://github.com/GoogleCloudPlatform/magic-modules repository.

PR #17316, opened 2024-02-19
Fix broken datasource google_compute_machine_types example

rwblokzijl:
The query doesn't work with single quotes. This needs to be double.


slevenick:
Can you make this PR against the upstream repository instead: https://github.com/GoogleCloudPlatform/magic-modules

That's the source of truth to keep this repo in sync with the beta provider

PR #17318, opened 2024-02-19
docs: Improve Terraform example for google_vpc_access_connector

ppuschmann:
Improve the Terraform example for `google_vpc_access_connector` with Shared-VPC. The current example does not work properly, because the lookup of the subnet is not working perfectly.

Since the network management in the context of Shared-VPC is mostly done centrally, the new example uses a data source for the serverless subnetwork `google_compute_subnetwork`.

### DIFF
--- a/website/docs/r/vpc_access_connector.html.markdown
+++ b/website/docs/r/vpc_access_connector.html.markdown
@@ -54,22 +54,19 @@ resource "google_vpc_access_connector" "connector" {
 ```hcl
 resource "google_vpc_access_connector" "connector" {
   name          = "vpc-con"
+
   subnet {
-    name = google_compute_subnetwork.custom_test.name
+    name       = google_compute_subnetwork.custom_test.name
+    project_id = "host-project-id"
   }
+
   machine_type = "e2-standard-4"
 }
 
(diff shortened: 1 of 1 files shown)


melinath:
Thanks for the contribution! This change will need to be made via magic-modules.

ppuschmann:
@melinath ... okay, who will do this?
Do _I_ really have to dig into magic modules or will you do this minor fix?

melinath:
Hi @ppuschmann - sorry, but we don't have bandwidth to upstream changes to magic-modules, especially since upstreaming can uncover additional changes that need to be made, even if the original change seems trivial.

It looks like you're wanting to modify https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/vpc_access_connector_shared_vpc.tf.erb - for more information on examples (which are generally used to generate tests as well as docs), see https://googlecloudplatform.github.io/magic-modules/develop/test/test/.

ppuschmann:
Wow, thank you for this friendly reply.
You now gave me a good first idea how to dig into the "magic-modules".

Yes, I can understand your position. 👍🏼

PR #17463, opened 2024-03-01
chore: update Cloud SQL `instances` field description

Markusdreyer:
The documentation is imprecise regarding the correct type for the `instances` field. Especially considering that this value used to be a comma separated string for the cloud run v1 API.


SarahFrench:
Please update that description here (https://github.com/GoogleCloudPlatform/magic-modules/blob/02194045a6eafb6a377acf536c333b5b1c11e857/mmv1/products/cloudrunv2/Service.yaml#L757) instead of in this repo. Thanks!

PR #17465, opened 2024-03-01
Update docscheck.sh to accept provider function docs

SarahFrench:
This script is a good candidate for being replaced with official tools, but for now updating this script to avoid false alarms about bad docs

TODO:
- ~Make same PR for the feature branch on TPGB~
- ~Make same PRs again into main for TPG and TPGB~

New plan:
After merging [this replacement PR](https://github.com/hashicorp/terraform-provider-google/pull/17497) I'll cherry pick the commit from main into the feature branch.

### DIFF
--- a/scripts/docscheck.sh
+++ b/scripts/docscheck.sh
@@ -27,6 +27,15 @@ for doc in $docs; do
       fi
       ;;
 
+    "functions")
+      # Functions require a page_title
+      grep "^page_title: " "$doc" > /dev/null
+      if [[ "$?" == "1" ]]; then
+        echo "Doc is missing a page_title: $doc"
+        error=true
+      fi
+      ;;
+
     *)
       error=true
       echo "Unknown category \"$category\". " \


SarahFrench:
@shuyama1 I requested you as reviewer because I'm not sure how to handle a file that exists only in the downstreams while working on a feature branch. My understanding is that I'll need to update the file in the feature branch here in the downstream, but also I'll need to update the file in the main branch. That's because when the feature branch upstream in Magic Modules is merged into main and changes made in this PR become irrelevant?

shuyama1:
Can this change be made to main branch at this moment, or it will need to wait until the feature branch in Magic Modules gets merged first? 
If the change can be made into the main branch directly now, I'd propose make a PR against main and cherry pick that commit into the downstream feature branch. 
If it needs to wait until the upstream feature branch merges to the main first, you're correct and you'll need to update the file in the main branch later since changes in this PR will not sync into the downstream main branch. Hope this makes sense and let me know if you have any questions.

SarahFrench:
This could go into main currently - I believe what would happen is that as the script iterates over the docs present on main the new section I've added to the script will never run as it'll never encounter any docs files in `website/docs/functions` until the feature branch is merged in MM.

SarahFrench:
Superseded by https://github.com/hashicorp/terraform-provider-google/pull/17497

PR #17509, opened 2024-03-07
Update docscheck.sh to accept provider function docs (#17497)

SarahFrench:
This script is a good candidate for being replaced with official tools, but for now updating this script to avoid false alarms about bad docs


shuyama1 (approved):

SarahFrench:
Thanks!

PR #17643, opened 2024-03-21
Update compute_backend_bucket_signed_url_key.html.markdown

mruediger:
The key needs to be stored in b64_std format or else it doesn't work.


slevenick:
Can you make this change against https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/backend_service_signed_url_key.tf.erb instead?

PR #17743, opened 2024-04-02
Fix incorrect reference to nonexistent resource

andrefilipemm:
The `time_resource` resource type doesn't exist in the time provider, the correct resource type should be `time_sleep`

This error is found [here](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/google_project_service#mitigation---adding-sleeps)

### DIFF
--- a/website/docs/guides/google_project_service.html.markdown
+++ b/website/docs/guides/google_project_service.html.markdown
@@ -63,7 +63,7 @@ resource "google_project" "my_project" {
   billing_account = var.billing_account_id
 }
 
-resource "time_resource" "wait_30_seconds" {
+resource "time_sleep" "wait_30_seconds" {
   depends_on = [google_project.my_project]
 
   create_duration = "30s"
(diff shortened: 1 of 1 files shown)


slevenick:
Weird that we have references to that... Can you make the fix against https://github.com/GoogleCloudPlatform/magic-modules instead? That repo is used to sync both this and terraform-provider-google-beta

andrefilipemm:
I can't seem to find any occurrence of `time_resource` in that repo

slevenick:
Looks like it's here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/guides/google_project_service.html.markdown

andrefilipemm:
Thanks! Here it is: https://github.com/GoogleCloudPlatform/magic-modules/pull/10350

andrefilipemm:
Hello @slevenick, this https://github.com/GoogleCloudPlatform/magic-modules/pull/10350 was merged so probably no need to keep this up since this repo is synced with magic-modules?

slevenick:
Yup, this change should happen automatically.

PR #17794, opened 2024-04-08
Reworked service labeler to act in more cases

melinath:
Specifically, if labels are removed or if the first comment is edited - but only if there is not a service label yet.

google_alloydb_instance

### DIFF
--- a/.github/workflows/service-labeler.yml
+++ b/.github/workflows/service-labeler.yml
@@ -25,14 +25,14 @@ jobs:
     - name: Check for service labels
       id: service_labels
       run: |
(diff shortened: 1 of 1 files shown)


ScottSuarez (approved):

melinath:
Successfully detected existing service label: https://github.com/hashicorp/terraform-provider-google/actions/runs/8606506951/job/23585089624?pr=17794

melinath:
Successfully computed new service labels: https://github.com/hashicorp/terraform-provider-google/actions/runs/8606515668/job/23585111965?pr=17794#step:5:16 - wasn't able to update because PRs only get read permissions.

PR #17883, opened 2024-04-17
Docs: Update container_cluster with MPS strategy on gpu_sharing_strategy

jeanbrunogabriel:
Adding MPS strategy on gpu_sharing_strategy docs which is already working.


rileykarson:
Per the robot, can you update the upstream version of the file at https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown instead? Thanks!

PR #17995, opened 2024-05-01
chore(ci): Update .goreleaser.yml to include LICENSE.txt file

BBBmau:
Adding LICENSE.txt file to Terraform Provider Release Archives


SarahFrench (changes requested):
I think this file is controlled by Magic Modules, so changes should instead be made here : https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/.goreleaser.yml.erb

PR #18003, opened 2024-05-02
Update bug issue template to include "provider version" in section title

SarahFrench:
In some bug issues users report their Terraform version but not the provider version - by updating the title of the section we might be able to correct this a little?


melinath (approved):
Alternatively / additionally, we could set a "template value" that includes Terraform and provider lines

SarahFrench:
@melinath thanks for the suggestion - I've added that value and [the preview shown here](https://github.com/hashicorp/terraform-provider-google/blob/SarahFrench-patch-1/.github/ISSUE_TEMPLATE/00_bug.yml) looks ok

Issue #17040, opened 2024-01-18
Error when using CLOUD_IAM_GROUP as google_sql_user type


haraldhaas:
The [google_sql_user](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_user) resource in the provider document explains the example of using the the type [CLOUD_IAM_GROUP](https://cloud.google.com/sql/docs/mysql/add-manage-iam-users#add-iam-group-db) for [google_sql_user](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_user). However, creating this resource always fails with either of the errors that can be seen below.

### Terraform Version
```
Terraform v1.5.4
on darwin_arm64
+ provider registry.terraform.io/hashicorp/google v5.12.0
+ provider registry.terraform.io/hashicorp/random v3.6.0
```

### Affected Resource(s)
* google_sql_user

### Terraform Configuration Files
```
resource "google_sql_database_instance" "main" {
  name             = "main-instance"
  project          = "<project-id>"
  region           = "europe-west1"
  database_version = "MYSQL_8_0"

  settings {
    tier = "db-f1-micro"

    database_flags {
      name  = "cloudsql_iam_authentication"
      value = "on"
    }
  }
}

resource "google_sql_user" "iam_group_user" {
  name     = "<group@domain.com>"
  project  = "<project-id>"
  instance = google_sql_database_instance.main.name
  type     = "CLOUD_IAM_GROUP"
}
```

### Debug Output
```
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to google_sql_user.iam_group_user, provider
│ "provider[\"registry.terraform.io/hashicorp/google\"]" produced an unexpected new value: Root resource was present,
│ but now absent.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

```
google_sql_user.iam_group_user: Creating...
╷
│ Error: Error, failure waiting for insertion of group@domain.com into main-instance-3c1f9de9:
│
│
│   with google_sql_user.iam_group_user,
│   on main.tf line 21, in resource "google_sql_user" "iam_group_user":
│   21: resource "google_sql_user" "iam_group_user" {
```

### Expected Behavior

Group should have been added to the database, as per [documentation](https://cloud.google.com/sql/docs/mysql/add-manage-iam-users#add-iam-group-db).

### Actual Behavior

Terraform apply failed.

### Steps to Reproduce

1. `terraform apply`

[...]

PR #18101, opened 2024-05-10
[CloudSQL] Handle CLOUD_IAM_GROUP username edgecases

jackweinbender:
Although some of the issues surrounding the use of `CLOUD_IAM_GROUP` users with cloudsql I believe are google's own problems, in this case, I think this PR will address some of the problems with TF not being able to locate `CLOUD_IAM_GROUP` type resources, as discussed in #17040.

The issue, I believe, is that group users retain the domain as a part of their username while standard users and service accounts have the domain stripped. You can see on L369 of the PR that this fact is already accounted for in part, but the same exception wasn't included in the `diffSuppress` function (L21ff.).

I've also included comments to make some of this explicit, since it's a bit counterintuitive.

### DIFF
--- a/google/services/sql/resource_sql_user.go
+++ b/google/services/sql/resource_sql_user.go
@@ -19,12 +19,25 @@ import (
 )
 
 func diffSuppressIamUserName(_, old, new string, d *schema.ResourceData) bool {
-	strippedName := strings.Split(new, "@")[0]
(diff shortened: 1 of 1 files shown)


slevenick (commented):
Can you make this against https://github.com/GoogleCloudPlatform/magic-modules instead? That's the source of truth to keep this repo and the google-beta repo in sync.

jackweinbender:
No problem, @slevenick . That bot message didn't register for me at 5 on a friday. See https://github.com/GoogleCloudPlatform/magic-modules/pull/10666

PR #18227, opened 2024-05-23
`google_compute_instance` - add validation for `name`

AlexPykavy:
To fail during the plan phase if the name violates the following rules:
https://cloud.google.com/compute/docs/naming-resources#resource-name-format.


AlexPykavy:
Closing the PR as this change has to be moved to https://github.com/GoogleCloudPlatform/magic-modules

PR #18539, opened 2024-06-25
Fix a typo in pubsub_subscription web manual page

wdzeng:
Fix that the field `bigquery_config.table` should be of format `{project}.${dataset}.${table}`, not `${project}:${dataset}.${table}`.

### DIFF
--- a/website/docs/r/pubsub_subscription.html.markdown
+++ b/website/docs/r/pubsub_subscription.html.markdown
@@ -615,7 +615,7 @@ The following arguments are supported:
 
 * `table` -
   (Required)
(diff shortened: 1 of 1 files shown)


shuyama1:
@wdzeng Thanks for making change. The code in this repo is generated via https://github.com/GoogleCloudPlatform/magic-modules, would mind making the change in the https://github.com/GoogleCloudPlatform/magic-modules? For this change specifically, its source code is located at https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/pubsub/Subscription.yaml#L143

wdzeng:
Got it. File [there](https://github.com/GoogleCloudPlatform/magic-modules/pull/11068) and close this one.

PR #18571, opened 2024-06-27
Fix docs link to predefined metadata keys

serpro69:
The link to default metadata keys seems to have changed since the old one leads to 404.


serpro69:
Opened a PR in https://github.com/GoogleCloudPlatform/magic-modules/pull/11065

PR #18675, opened 2024-07-08
Replace labelKeys/Values for tagKeys/Values in constraints

rubensf:
GCP does not support labels as iam or org policy constraints. This is documented in multiple locations, such as
https://cloud.google.com/compute/docs/labeling-resources#labels_and_tags.


rubensf:
Closing since I have to update magic-modules instead

rubensf:
https://github.com/GoogleCloudPlatform/magic-modules/pull/11125

PR #18810, opened 2024-07-23
fix typo

tomasmota:


c2thorn:
This needs to be made here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/container_node_pool.html.markdown?plain=1#L268

PR #18858, opened 2024-07-26
Docs: Add ingress_from documentation for VPC Service Controls resources for using networks as an incoming source.

melissa-mo:
This change addresses the issue at https://github.com/hashicorp/terraform-provider-google/issues/18142.

I could not find `access_context_manager_service_perimeter.html.markdown` file in `magic-modules/mmv1/third_party/terraform/website/docs/r`.


melinath:
Those files are generated from https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/accesscontextmanager/ServicePerimeter.yaml and https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/accesscontextmanager/IngressPolicy.yaml, so that's where you would need to make the updates.

PR #18875, opened 2024-07-27
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18874 - Adding DELTA_LAKE format to source_format optional values

ik8:


melinath (changes requested):
Thanks! This change needs to be made in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go and https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/bigquery_table.html.markdown rather than in this repository.

PR #18920, opened 2024-07-31
Add missing special member identifiers

kustodian:
`projectOwners`, `projectWriters` and `projectReaders` were missing from the documentation.


slevenick:
Can you make this against the upstream repo?

SarahFrench:
Please update this file instead : https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/bigquery_dataset_iam.html.markdown

Thanks!

PR #19023, opened 2024-08-07
update `teamcity-nightly` workflows to use `nightly-test` as static branchFilter

BBBmau:
This PR would simplify the way teamcity checkouts of main in order to ensure that the same commit is being used across multiple builds. Previously the name included the date which relied on renaming of the branch being used for the build. Although this still occurs we now only use `nightly-test` instead of the constantly changing `UTC-nightly-test-<DATE>`. This allows the builds to always appear in the teamcity UI without having to explicitly change the `activeBuildBranch.age`.

An example of how the UI appears can be seen here:
image

With workflow runs being found here:
[teamcity-nightly-workflow.yaml](https://github.com/BBBmau/terraform-provider-google/actions/workflows/teamcity-nightly-workflow.yaml)
[teamcity-nightly-sweeper.yaml](https://github.com/BBBmau/terraform-provider-google/actions/workflows/teamcity-nightly-sweeper.yaml)


SarahFrench (approved):
Looks good, glad the edit was pretty simple!

Issue #19096, opened 2024-08-11
PubSub Subscription filter length not validated on plan - causes deletion of subscription without creation when applied


omrikiei:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to a user, that user is claiming responsibility for the issue.
* Customers working with a Google Technical Account Manager or Customer Engineer can ask them to [reach out internally](https://github.com/hashicorp/terraform-provider-google/wiki/Customer-Contact#raising-gcp-internal-issues-with-the-provider-development-team) to expedite investigation and resolution of this issue.

I plan to submit a (tiny) fix for the issue

### Terraform Version & Provider Version(s)

Terraform v1.9.3
on mac m1
+ provider registry.terraform.io/hashicorp/google v5.4.0

### Affected Resource(s)

google_pubsub_subscription

### Terraform Configuration

```tf
provider "google" {
  project = "test"
  region  = "eu-west-3"
}

resource "google_pubsub_topic" "foo" {
  name = "foo_topic"
}

resource "google_pubsub_subscription" "foo" {
  name   = "foo_subscription"
  topic  = google_pubsub_topic.foo.id
  filter = "attributes.foo = \"barbarbarbarbarbarbarbararbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbar\""
  labels = {
    foo = "foo_subscription"
  }
  retry_policy {
    minimum_backoff = "60.0s"
  }
  ack_deadline_seconds = 30
  enable_exactly_once_delivery = true
}
```

### Expected Behavior

`tf plan` should fail if the filter's length exceeds 256 bytes

### Steps to reproduce

1. `terraform plan` / `terraform apply`

charan-kumar-137:
Since the Subscription argument `filter` has **ForceNew** set. Any modification to the argument will cause deletion and creation.

[...]

PR #19097, opened 2024-08-11
[#19096] limit filter byte size in subscription

omrikiei:
This PR limits the size of the subscription filter to 256 bytes


omrikiei:
Should be changed in magic-modules, closing this PR

PR #19127, opened 2024-08-14
refactor `teamcity-nightly-workflows` to be just one yaml

BBBmau:
some refactoring was necessary in order to always have a `nightly-test` branch for TeamCity to use for nightly builds.

Workflow triggers can be found here as well as each case:

- nightly test not present -> rename fails but `nightly-test` creation succeeds: https://github.com/BBBmau/terraform-provider-google/actions/runs/10380514976

- nightly-test is present -> rename is successful and is followed with a nightly-test branch creation off of main: https://github.com/BBBmau/terraform-provider-google/actions/runs/10380522355

- nightly-test is present -> rename is not successful due to branch name used for renaming is already used: https://github.com/BBBmau/terraform-provider-google/actions/runs/10380445853


SarahFrench (commented):
Looks good, and thanks for testing how different failure modes behave.

Does this PR need to remove the https://github.com/hashicorp/terraform-provider-google/blob/main/.github/workflows/teamcity-nightly-sweeper.yaml file?

.github/workflows/teamcity-nightly-workflow.yaml
```diff
@@ -22,7 +22,32 @@ jobs:
           retry-exempt-status-codes: 400, 401, 403, 404, 422
           script: |
                   let dateToday = new Date().toJSON().slice(0, 10);
+                  const oldBranchName = "nightly-test";
+                  const newBranchName = "UTC-nightly-tests-" + dateToday;
```
SarahFrench, line 26: Here the date needs to be dateToday - 1 day

SarahFrench, line 26:
```suggestion
                  // Get yesterday's date, when the branch was originally made
                  const date = new Date()
                  date.setDate(date.getDate() -1);
                  
                  // Get YYYY-MM-DD string
                  dateString = date.toJSON().slice(0, 10)
                  
                  const oldBranchName = "nightly-test";
                  const newBranchName = "UTC-nightly-tests-" + dateString;
```

SarahFrench (approved):

PR #19186, opened 2024-08-20
Docs: Add missing quote

sgatewood-da:
changing the world 1 character at a time


slevenick:
Oh, can you make this against the linked magic-modules repository instead? That repo keeps this and terraform-provider-google-beta in sync.

PR #19435, opened 2024-09-11
fix: Match all whitespace and non-whitespace for Pub/Sub subscription filter validation

### LINKED ISSUE #19296: Pubsub subscription filters with newlines are no longer allowed by validator
### Community Note

* If an issue is assigned to a user, that user is claiming responsibility for the issue.
* Customers working with a Google Technical Account Manager or Customer Engineer can ask them to [reach out internally](https://github.com/hashicorp/terraform-provider-google/wiki/Customer-Contact#raising-gcp-internal-issues-with-the-provider-development-team) to expedite investigation and resolution of this issue.

### Terraform Version & Provider Version(s)

Terraform v1.9.5
on <architecture>
+ provider registry.terraform.io/hashicorp/google v5.43.0
+ provider registry.terraform.io/hashicorp/google-beta v5.43.0

### Affected Resource(s)

google_pubsub_subscription

### Terraform Configuration

```tf
resource "google_pubsub_subscription" "test-subscriptions" {
  filter = <<EOF
attributes.foo = "foo"
AND attributes.bar = "bar"
EOF
}
```

### Debug Output

```
│ Error: "filter" ("attributes.foo = \"foo\"
AND attributes.bar = \"bar\"
") doesn't match regexp "^.{1,256}$"
```

### Expected Behavior

This should work just fine -- newlines are valid characters accepted by the API, these subscription filters have been working just fine, and using heredocs in terraform should be acceptable in this case.

### Actual Behavior

```
│ Error: "filter" ("attributes.foo = \"foo\"
AND attributes.bar = \"bar\"
") doesn't match regexp "^.{1,256}$"
```

### Steps to reproduce

1. `terraform apply`

### Important Factoids

_No response_

### References

Validation was introduced [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11421) and seems to have caused [other issues as well](https://github.com/hashicorp/terraform-provider-google/issues/19269).

michaelpri10:
This change the regular expression validation for Pub/Sub subscriptions to use `[\s\S]` instead of `.` when matching characters to match both whitespace and non-whitespace characters. Because we allow empty filters (see [this issue](https://github.com/hashicorp/terraform-provider-google/issues/19269)), having only whitespace characters (which would be matched by this filter) is valid as well.

This fixes #19296.

### DIFF
--- a/google/services/pubsub/resource_pubsub_subscription.go
+++ b/google/services/pubsub/resource_pubsub_subscription.go
@@ -366,7 +366,7 @@ Example - "3.5s".`,
 				Type:         schema.TypeString,
 				Optional:     true,
 				ForceNew:     true,
-				ValidateFunc: verify.ValidateRegexp(`^.{0,256}$`),
+				ValidateFunc: verify.ValidateRegexp(`^[\s\S]{0,256}$`),
 				Description: `The subscription only delivers the messages that match the filter.
 Pub/Sub automatically acknowledges the messages that don't match the filter. You can filter messages
 by their attributes. The maximum length of a filter is 256 bytes. After creating the subscription,


michaelpri10:
Per the warning, I've created a PR in the magic_modules repository to fix this issue: https://github.com/GoogleCloudPlatform/magic-modules/pull/11676

PR #19674, opened 2024-09-27
Changelog fix for 6.4.0

melinath:
`google_scc_v2_organization_scc_big_query_export` was added but not captured in the release notes. I've added it to the releases page already. Beta fix: https://github.com/hashicorp/terraform-provider-google-beta/pull/8303


wyardley:
I think also a missing bullet on line 37 before `container: fixed a bug where specifying node_pool_defaults.node_config_defaults` maybe (same in release notes)?

melinath:
indeed; fixed that as well

ScottSuarez (approved):

PR #19677, opened 2024-09-30
add docs about reserved labels of cluster labels

takahyon:
add docs about limitation of cluster labels

Background:
When I add custom compute class, I have to add node label llike
`Node label: cloud.google.com/compute-class=COMPUTE_CLASS`
https://cloud.google.com/kubernetes-engine/docs/concepts/about-custom-compute-classes#manual-node-pools

Plan didn't show any errors, but when I apply it I recognized it reserved.
> │ Error: googleapi: Error 400: Node labels with key "[cloud.google.com/compute-class](http://cloud.google.com/compute-class)" are managed by GKE or Kubernetes and must not be manually specified.

I think this implement make sense, but it should written in docs.

### DIFF
--- a/website/docs/r/container_cluster.html.markdown
+++ b/website/docs/r/container_cluster.html.markdown
@@ -890,8 +890,8 @@ gvnic {
 * `image_type` - (Optional) The image type to use for this node. Note that changing the image type
     will delete and recreate all nodes in the node pool.
 
-* `labels` - (Optional) The Kubernetes labels (key/value pairs) to be applied to each node. The kubernetes.io/ and k8s.io/ prefixes are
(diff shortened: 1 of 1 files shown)


ScottSuarez:
hi @takahyon , this file is managed by an upstream repository. Would you be able to make this change there?

https://github.com/GoogleCloudPlatform/magic-modules/blob/2d34af8bcf47b390ede98ace828c98e80cb3acd0/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown?plain=1#L893

takahyon:
@ScottSuarez OK, Thanks.

PR #19887, opened 2024-10-16
Schedule daily run for backfill labeler

shuyama1:
This fixes https://github.com/hashicorp/terraform-provider-google/issues/19888

The approach of handling defaults for input parameters in schedule GHA is inspired by https://stackoverflow.com/questions/72539900/schedule-trigger-github-action-workflow-with-input-parameters

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -3,6 +3,8 @@ name: Issue Opened Triage
 on:
   issues:
     types: [opened]
+  schedule:
+        - cron: '0 7 * * *' # UTC 7AM (-7)-> 12AM PST
 
 jobs:
   triage:
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -3,8 +3,6 @@ name: Issue Opened Triage
 on:
   issues:
     types: [opened]
-  schedule:
-        - cron: '0 7 * * *' # UTC 7AM (-7)-> 12AM PST
 
 jobs:
   triage:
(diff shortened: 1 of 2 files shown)


.github/workflows/labeler.yml
```diff
@@ -3,6 +3,8 @@ name: Issue Opened Triage
 on:
   issues:
     types: [opened]
+  schedule:
+        - cron: '0 7 * * *' # UTC 7AM (-7)-> 12AM PST
```
roaks3, line 7: I believe this is the job that adds the `crash` label, and for service labels we would instead want to change `.github/workflows/backfill-issue-labels.yml`. Schedule config LGTM though.

roaks3 (changes requested):

.github/workflows/labeler.yml
shuyama1: oops, you're right! I'll update it. Thanks for catching it

roaks3 (commented):
The only thing I'm still wondering about is the handling of the 2 variables. Do we need to set those for the run, or make sure that the correct defaults are used when they aren't provided?

shuyama1:
> The only thing I'm still wondering about is the handling of the 2 variables. Do we need to set those for the run, or make sure that the correct defaults are used when they aren't provided?

oh, good point. I'll double check if schedule will pick up default values for input.

shuyama1:
I manually tested this in a local repo and I think it works as expected. 

Config: https://github.com/shuyama1/mm-ci-test/blob/main/.github/workflows/test_job.yml

Manual triggered with updated input values: https://github.com/shuyama1/mm-ci-test/actions/runs/11447489514/job/31848933300

Scheduled with default values: https://github.com/shuyama1/mm-ci-test/actions/runs/11447565178/job/31849174447

roaks3 (approved):
Awesome, LGTM

PR #20061, opened 2024-10-28
Remove reference to deleted etiquette guide from CONTRIBUTING.md

SarahFrench:
Relates to (but does not close) https://github.com/hashicorp/terraform-provider-google/issues/20047#event-14966480533

This doc references this file, which was deleted: https://github.com/hashicorp/terraform/pull/32741/files


SarahFrench:
We don't need to make a similar change in the TPGB repo, as the CONTRIBUTING.md doc there only redirects people to Magic Modules

SarahFrench:
I'm going to bypass branch protections given that this is a change to a markdown file, not provider code, and it's a md file that we don't expect users to reference during the contribution process.

PR #20318, opened 2024-11-13
Documentation Update: Revise Text Styling

tajpouria:
I came across this in the Terraform documentation, and it struck me as rather bold:

image


SarahFrench:
Yeah, I agree the markdown should be updated to remove that formatting!

Please make an equivalent PR to our upstream repo, editing this file there: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown

PR #20330, opened 2024-11-14
Clarify `google_storage_bucket` `response_header` in `storage_bucket.html.markdown`

philipp-tailor:
The description of `response_header` prior to this change seemed to suggest that simple response headers were always implicitly included in  the `response_header` list. However, the link to the fetch standard on simple response headers was broken, and searching for the term "simple response" doesn't yield the result. [According to MDN](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header), the `Content-Type` would be considered such a simple response header. However on my bucket, preflight requests on a bucket failed until `response_header` was set to `["Content-Type"]`, countering the docs. The wrong docs cost me hours, until I finally decided to go try adding going with code analogous to the [basic CORS example configuration](https://cloud.google.com/storage/docs/cors-configurations). Now my assumption is that at least `Content-Type` is not automatically returned in responses to cross origin preflight requests, and - since this header supposedly is a simple response header - none of those is automatically returned. This PR updates the documentation accordingly, removing the claim that simple response headers are always included, and instead adds the relatively common example of the `Content-Type` header.

### DIFF
--- a/website/docs/r/storage_bucket.html.markdown
+++ b/website/docs/r/storage_bucket.html.markdown
@@ -241,7 +241,7 @@ The following arguments are supported:
 
 * `method` - (Optional) The list of HTTP methods on which to include CORS response headers, (GET, OPTIONS, POST, etc) Note: "*" is permitted in the list of methods, and means "any method".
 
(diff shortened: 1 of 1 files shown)


SarahFrench:
Replaced by https://github.com/GoogleCloudPlatform/magic-modules/pull/12326

PR #20500, opened 2024-11-27
Update org_policy_policy documentation to make resource definition clearer

juliorenner:
This PR:
- Updates the documentation to make it explicit that what is required for the resource definition is the project id, not the project name.

The project name may differ from the project id in google. For example, you may have a project ID `sandbox-123123` while the project name is `sandbox`. If in this case using the project name the role creation will fail.

### DIFF
--- a/website/docs/r/org_policy_policy.html.markdown
+++ b/website/docs/r/org_policy_policy.html.markdown
@@ -34,8 +34,8 @@ To get more information about Policy, see:
 
 ```hcl
 resource "google_org_policy_policy" "primary" {
-  name   = "projects/${google_project.basic.name}/policies/iam.disableServiceAccountKeyUpload"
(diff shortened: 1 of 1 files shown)


SarahFrench:
Hi @juliorenner - this repo is [generated](https://github.com/hashicorp/terraform-provider-google/pull/20500#issuecomment-2504012839) from the Magic Modules repo, so could you please open a PR to update files in that repo with your diffs? These are the example files used to generate the docs you're editing:

- https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/org_policy_policy_enforce.tf.tmpl
- https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/org_policy_policy_project.tf.tmpl

PR #20501, opened 2024-11-27
Remove sweeper information from CONTRIBUTING.md

SarahFrench:
Removing this data from here because
- sweeper info is also currently present here: https://github.com/hashicorp/terraform-provider-google/wiki/Developer-Best-Practices#sweepers
    - Note: I just updated this to include more up to date info and add some copy deleted in this PR
- it's going to be added to the official contributor docs: https://github.com/hashicorp/terraform-provider-google/issues/17891
- this old CONTRIBUTING.md doc was intended to be replaced by the contribution site, so we should remove these lingering bits of information


SarahFrench:
Looks like branch protections are enforcing the need for code owner review, so I'll just bypass branch protections this time :shipit:  I posted in the shared channel about this in case it affects releases in the near-future

PR #20624, opened 2024-12-08
Add `POSTGRES_16` and `POSTGRES_17`

rinarakaki:


roaks3:
@rinarakaki you'll need to make this change in https://github.com/GoogleCloudPlatform/magic-modules

PR #20678, opened 2024-12-12
`teamcity-nightly-workflow`: only create nightly-test branch on upstream provider repo

BBBmau:
prevents `nightly-test` branch from being created on forked repos since they are only used on upstream terraform-provider-google

### DIFF
--- a/.github/workflows/teamcity-nightly-workflow.yaml
+++ b/.github/workflows/teamcity-nightly-workflow.yaml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   rename-TC-nightly-branch:
-    if: github.repository == 'hashicorp/terraform-provider-google'
+    if: github.repository == 'hashicorp/terraform-provider-google' || github.repository == 'hashicorp/terraform-provider-google-beta'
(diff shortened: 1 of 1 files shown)


.github/workflows/teamcity-nightly-workflow.yaml
```diff
@@ -14,7 +14,7 @@ on:
 
 jobs:
   rename-TC-nightly-branch:
-
+    if: github.repository == 'hashicorp/terraform-provider-google'
```
melinath, line 17: I think we need it in beta too?

```suggestion
    if: github.repository == 'hashicorp/terraform-provider-google' || github.repository == 'hashicorp/terraform-provider-google-beta'
```

melinath (changes requested):

.github/workflows/teamcity-nightly-workflow.yaml
BBBmau: good catch, thanks!

melinath (approved):

jrhouston (approved):

PR #20813, opened 2025-01-06
docs(README): intended guidance links

guspan-tanadi:


zli82016:
Thanks for the fix. Do you mind making the same change in terraform-provider-google-beta provider? Thanks.

PR #21326, opened 2025-02-10
Correct typo in documentation for google_container_cluster

taleh007:
`google_container_cluster.config.wokloads_config.worker.min_count`


roaks3:
@taleh007 thanks for the fix! Could you please make the contribution over at https://github.com/GoogleCloudPlatform/magic-modules? Per the above message, it will be overwritten if only applied to this repo, because the google provider is generated.

PR #21415, opened 2025-02-14
Update sql_database_instance.html.markdown

rebecca-pete:
Update `preferred_zone` definition.


rileykarson:
Per the robot, you'll want to make the change against Magic Modules, at https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown instead. Thanks!

PR #21421, opened 2025-02-14
Suppressing diff for database_version for MySQL 8.0

ankushkhanna1998:
Suppressing diff for database_version for MySQL 8.0. With introduction of auto version upgrade for MySQL 8.0.35 and beyond, customers will be auto upgraded to a newer version which will lead to change in database version.

Hence, to avoid breakage of terraform scripts suppressing this diff while maintaining the sanity that the major version does not change, suppression is done only if there is a diff in minor version.


ankushkhanna1998:
Raised this PR here instead: https://github.com/GoogleCloudPlatform/magic-modules/pull/13090

PR #21459, opened 2025-02-18
Update website documentation about Composer 3 and the recovery_config block

StevenSalazarM:
Update documentation for  ```recovery_config``` and ```scheduled_snapshots_config``` for Composer 3.

The blocks are already supported but the documentation does not include them for Composer 3.

Impacted page: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/composer_environment#argument-reference---cloud-composer-3


trodge:
I think this change will need to be made in [magic modules](https://github.com/GoogleCloudPlatform/magic-modules).

trodge (approved):

PR #21472, opened 2025-02-18
Improve provider credentials docs. Fixes #21471

### LINKED ISSUE #21471: Documentation: credentials for google cloud provider require the `file()` function which is not documented
### Community Note

* If an issue is assigned to a user, that user is claiming responsibility for the issue.
* Customers working with a Google Technical Account Manager or Customer Engineer can ask them to [reach out internally](https://github.com/hashicorp/terraform-provider-google/wiki/Customer-Contact#raising-gcp-internal-issues-with-the-provider-development-team) to expedite investigation and resolution of this issue.

### Description

The [credentials parameter](https://github.com/hashicorp/terraform-provider-google/blob/5d380eac6531809fb18198eab7435ed3e3a67df9/website/docs/guides/provider_reference.html.markdown?plain=1#L119) to the google provider requires using the terraform `file()` function. This is not documented. In contrast, the same argument in the storage backend should NOT use that function and instead be a string with the path. This had me stumped for a long time and go down rabbit holes of doubting the way  i made the credentials files. 
Documenting this more clearly would be useful.

* provider.google

```tf 
provider "google" {
  alias       = "myotherproject"
  project     = var.project_id
  region      = var.region
  credentials = var.gcloud_dac_file
}
```
doesnt work but

```tf 
provider "google" {
  alias       = "myotherproject"
  project     = var.project_id
  region      = var.region
  credentials = file(var.gcloud_dac_file)
}
```
does.
While
```
terraform {
  backend "gcs" {
    bucket = "my-state-bucket"
    prefix = "terraform/state"
    credentials = file("/path/to/my/file.json")
  }
```
doesn't work but 

```
terraform {
  backend "gcs" {
    bucket = "my-state-bucket"
    prefix = "terraform/state"
    credentials = "/path/to/my/file.json"
  }
```
does.

### References

_No response_

dolfandringa:
Fixes #21471: Adds more docs on the credentials argument to the google provider.

### DIFF
--- a/website/docs/guides/provider_reference.html.markdown
+++ b/website/docs/guides/provider_reference.html.markdown
@@ -116,8 +116,8 @@ impersonated service account regardless of the primary identity in use.
 
 ## Authentication Configuration
 
-* `credentials` - (Optional) Either the path to or the contents of a
-[service account key file] in JSON format. You can
+* `credentials` - (Optional) Either the path (as returned by the [file terraform function])
+to or the contents of a [service account key file] in JSON format. You can
 [manage key files using the Cloud Console]. Your service account key file is
 used to complete a two-legged OAuth 2.0 flow to obtain access tokens to
 authenticate with the GCP API as needed; Terraform will use it to reauthenticate
@@ -136,6 +136,15 @@ by precedence.
     the path of your service account key file in the
     `GOOGLE_APPLICATION_CREDENTIALS` environment variable, or configure
     authentication through one of the following;
+  
+     ```hcl
+     provider "google" {
+         alias       = "other"
+         project     = var.my_other_project_id
+         region      = var.region
+         credentials = file(pathexpand(var.gcloud_other_dac_file))
+     }
+     ```
 
 * If you're running Terraform from a GCE instance, default credentials
 are automatically available. See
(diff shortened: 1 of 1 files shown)


website/docs/guides/provider_reference.html.markdown
```diff
@@ -136,6 +136,15 @@ by precedence.
     the path of your service account key file in the
     `GOOGLE_APPLICATION_CREDENTIALS` environment variable, or configure
     authentication through one of the following;
+  
+     ```hcl
+     provider "google" {
+         alias       = "other"
```
slevenick, line 142: I don't think we want to add docs with an alias unless there's a particular reason to do so

website/docs/guides/provider_reference.html.markdown
```diff
@@ -136,6 +136,15 @@ by precedence.
...
+         project     = var.my_other_project_id
```
slevenick, line 143: Please change the values in all of these fields to be example text rather than references to variables. It will be consistent with other config snippets in this doc

website/docs/guides/provider_reference.html.markdown
```diff
@@ -136,6 +136,15 @@ by precedence.
...
+         region      = var.region
+         credentials = file(pathexpand(var.gcloud_other_dac_file))
```
slevenick, line 145: Prefer using a hard-coded example string if possible
```suggestion
         credentials = file(pathexpand("/path/to/file.json"))
```

slevenick (changes requested):

PR #21692, opened 2025-03-04
fix issue-19051 max retention period for GCS bucket

mjogdand:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19051

### DIFF
--- a/google/services/storage/resource_storage_bucket.go
+++ b/google/services/storage/resource_storage_bucket.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"math"
 	"regexp"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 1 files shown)


mjogdand:
@trodge Can you please advice on the change?

trodge (approved):

trodge:
This change will need to be made in magic modules rather than here.

mjogdand:
> This change will need to be made in magic modules rather than here.

Thank you for the review @trodge 
As per your suggestion, new PR is created https://github.com/GoogleCloudPlatform/magic-modules/pull/13297

mjogdand:
Raised PR against magic-modules https://github.com/GoogleCloudPlatform/magic-modules/pull/13297

PR #21783, opened 2025-03-10
Add the release note for the skipped commit

zli82016:
Add the release note for the skipped commit, which is included in https://github.com/hashicorp/terraform-provider-google/pull/21782


melinath:
per discussion, we believe that updating the PRs should fix this (instead of needing new files.)

PR #22271, opened 2025-04-08
Netapp volume documentation correction

hyacin75:
Documentation for "restore_parameters" contains an error - it describes both source_snapshot and source_backup as "Full name of the snapshot to use..." - this corrects that to "backup" where applicable.

### DIFF
--- a/google/services/netapp/resource_netapp_volume.go
+++ b/google/services/netapp/resource_netapp_volume.go
@@ -245,7 +245,7 @@ Only the volume with largeCapacity will be allowed to have multiple endpoints.`,
 							Type:     schema.TypeString,
 							Optional: true,
 							ForceNew: true,
-							Description: `Full name of the snapshot to use for creating this volume.
+							Description: `Full name of the backup to use for creating this volume.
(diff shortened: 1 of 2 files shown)


hyacin75:
> ** WARNING **: This repository is generated by https://github.com/GoogleCloudPlatform/magic-modules. Any changes made directly to this repository will likely be overwritten. If you have further questions, please feel free to ping your reviewer or, internal employees, reach out to one of the engineers. Thank you!

I would assume this means the markdown will be regenerated from the golang source, but as I updated both locations it should be a non-issue if it just clobbers my changes to the markdown file with the exact same thing.

roaks3:
Technically yes, but then we'd need to be concerned about managing any sort of change that could be coming from either source. It's simpler for us to have everything come from magic-modules, and there is really no advantage to skipping that process and applying the change directly to this repo.

roaks3:
I think you're looking for https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/netapp/Volume.yaml#L296

PR #22385, opened 2025-04-16
Fix TestAccDNSRecordSet_routingPolicy

nbodepudi:
Should fix https://github.com/hashicorp/terraform-provider-google/issues/21770. Added unique names for all the proxy subnets.


nbodepudi:
My bad, will open a pull request in Magic modules.

PR #22428, opened 2025-04-19
Use the correct form of `sha256.Sum256`.

mcsaucy:
`sha256.New().Sum(foo)` appends the SHA-256 hash for nil to foo. See https://go.dev/play/p/vSW0U3Hq4qk


melinath:
This change would need to be made in https://github.com/GoogleCloudPlatform/magic-modules, specifically https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/constants/compute_certificate.go.tmpl

PR #22767, opened 2025-05-13
Update using_gke_with_terraform.html.markdown

kianjones9:
Small irrelevant typo in docs, thought I'd fix.  

Let me know if this needs to be made through magic modules.


ScottSuarez:
hi, yes can you make the change here
https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/guides/using_gke_with_terraform.html.markdown#L3

PR #22856, opened 2025-05-19
Update go.mod

shrishty:
Upgrade is needed to unblock the change: GoogleCloudPlatform/magic-modules#13848


go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	golang.org/x/exp v0.0.0-20240409090435-93d18d7e34b8
 	golang.org/x/net v0.39.0
 	golang.org/x/oauth2 v0.29.0
-	google.golang.org/api v0.229.0
```
BBBmau, line 35: this looked good but can you just resolve the conflicts that were introduced recently?

rileykarson:
This should be redundant with https://github.com/GoogleCloudPlatform/magic-modules/pull/13848 in theory, since that should be authoritative for the `go.mod` file.

shrishty:
Right closed the pull request

PR #23548, opened 2025-07-09
Added wait after service enablement in tests :Bug 394920780

Lavanya-Polipalli:


trodge:
This change will need to be made in https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/services/apigee/resource_apigee_endpoint_attachment_test.go instead, otherwise it will be overwritten.

PR #23989, opened 2025-08-14
Fix two typos in schema description for certificate map entries

naheedtayab:


slevenick:
This repo is generated, please make the change against this file instead! https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/certificatemanager/CertificateMapEntry.yaml

PR #24134, opened 2025-08-25
doc: fix expample in data google_artifact_registry_package

GRBurst:
We have a data source that shows an example for a resource. I assume there was a simple oversight.


GRBurst:
will open the PR in [GoogleCloudPlatform/magic-modules](https://github.com/GoogleCloudPlatform/magic-modules)

PR #24257, opened 2025-09-03
Update TestAccCloudRunV2WorkerPool_cloudrunv2WorkerPoolGpuExample to updated launch stage.

kminsu-google:
Update launch stage value from ALPHA to BETA for the gpu test case as we've promoted the GPU feature within WorkerPools to BETA


melinath:
Please make this change in magic-modules. Thanks!

PR #24834, opened 2025-10-24
bucket_id uses name instead of id

bollenn:
Spent a couple of hours to figure that `bucket_id` uses name instead of id


ScottSuarez:
You can make the relevant change here

https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/firebasestorage_bucket_basic.tf.tmpl#L11

PR #26109, opened 2026-02-15
New version support for postgresql 18

superarvind007:
https://docs.cloud.google.com/sql/docs/db-versions#version-support-policy The supported versions have been changed, and thus their supported tier too.


melinath:
Thanks for this contribution! However, this change will need to be made to https://github.com/GoogleCloudPlatform/magic-modules, as mentioned above.

PR #26393, opened 2026-03-10
Add Location Hint Support

LAVEEN:
Adding location hint support.


melinath:
This repository is generated code - you'll need to make this change in https://github.com/GoogleCloudPlatform/magic-modules. I think all the relevant files are in https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/services/compute

PR #26496, opened 2026-03-18
Add ForceNew attribute to name field in mesh resource

zhangzxjzxj:


melinath:
This repository is generated by https://github.com/GoogleCloudPlatform/magic-modules, so you'll need to make this change there (specifically by marking the field [immutable](https://googlecloudplatform.github.io/magic-modules/reference/field/#immutable) in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/networkservices/Mesh.yaml)

melinath:
Thanks!

PR #26896, opened 2026-04-13
`teamcity-nightly-workflow`: update cronjob to use timezone: `America/Los_Angeles'

BBBmau:
locks us into 7PM PST, avoiding potentially triggering at 6PM PST on DST causing merge overlaps on branch creation


shuyama1 (approved):
Thanks!

Issue #20073, opened 2024-10-29
google_tags_tag_key import does not support purpose_data


joekohlsdorf:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to a user, that user is claiming responsibility for the issue.
* Customers working with a Google Technical Account Manager or Customer Engineer can ask them to [reach out internally](https://github.com/hashicorp/terraform-provider-google/wiki/Customer-Contact#raising-gcp-internal-issues-with-the-provider-development-team) to expedite investigation and resolution of this issue.

### Terraform Version & Provider Version(s)

Terraform v1.9.6
on darwin_amd64
+ provider registry.terraform.io/hashicorp/google v6.8.0
+ provider registry.terraform.io/hashicorp/google-beta v6.8.0

### Affected Resource(s)

google_tags_tag_key

### Terraform Configuration

```tf
resource "google_tags_tag_key" "tag_key" {
  parent = "projects/123"
  short_name = "test-tag"
  purpose = "GCE_FIREWALL"
  purpose_data = {
    network = "123/vpc-us-west1"
  }
}
```

### Expected Behavior

After creating the resource, removing it from state and then importing it again should not cause any changes to be detected.

### Actual Behavior

After `terraform import`, the state does not contain `purpose_data` leading to a plan which will replace the resource:

```
 -/+ destroy and then create replacement
 Terraform will perform the following actions:
   # google_tags_tag_key.tag_key must be replaced
 -/+ resource "google_tags_tag_key" "tag_key" {
       ~ create_time     = "2024-10-29T15:21:57.059029Z" -> (known after apply)
       ~ id              = "tagKeys/281477391791474" -> (known after apply)
       ~ name            = "281477391791474" -> (known after apply)
       ~ namespaced_name = "test-project-123/test-tag" -> (known after apply)
       ~ parent          = "projects/123" -> "projects/test-project-123"
       + purpose_data    = { # forces replacement
           + "network" = "test-project-123/vpc-us-west1"
         }
       ~ update_time     = "2024-10-29T15:21:57.059029Z" -> (known after apply)
[...]

PR #26980, opened 2026-04-17
Fix issue #20073

faustroxas:
Fix to issue: https://github.com/hashicorp/terraform-provider-google/issues/20073

### DIFF
--- a/google/services/tags/resource_tags_tag_key.go
+++ b/google/services/tags/resource_tags_tag_key.go
@@ -625,7 +625,7 @@ func flattenTagsTagKeyPurposeData(v interface{}, d *schema.ResourceData, config 
 	return v
 }
 
-func flattenTagsTagKeyAllowedValuesRegex(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{}
+func flattenTagsTagKeyAllowedValuesRegex(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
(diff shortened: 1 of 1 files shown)


melinath:
@faustroxas Thanks! However, this repository is generated code; you'll need to make this contribution in magic-modules; specifically, it looks like you want to remove `ignore_read: true` from purposeData: https://github.com/GoogleCloudPlatform/magic-modules/blob/ef8fcbf855873cc09674d6a7332f2afdbdd63689/mmv1/products/tags/TagKey.yaml#L128

I'm not sure whether the reasons for including that have changed; sometimes they do!

Here's the Magic Modules contribution guide: https://googlecloudplatform.github.io/magic-modules/

faustroxas:
@melinath 
If you read the issue attached, there is a bug in importing this resource. Imports of tag key resources are ignoring the `purposeData` tf block which is causing replacements of the resource. In that case, can you point me to the code where I can fix that as I see this issue has been active for more than 18 months already?

melinath:
@faustroxas As I said in my previous message:

> it looks like you want to remove ignore_read: true from purposeData: https://github.com/GoogleCloudPlatform/magic-modules/blob/ef8fcbf855873cc09674d6a7332f2afdbdd63689/mmv1/products/tags/TagKey.yaml#L128

I'm pretty sure that making that change would generate the equivalent code to this PR. I hope that helps!

PR #27539, opened 2026-05-26
ci: update dependencies for GoTest workflow

BBBmau:
Mirrors updates to go.yml workflow file that were introduced in:
- https://github.com/hashicorp/terraform-provider-google-beta/pull/12312

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -13,10 +13,10 @@ jobs:
     steps:
 
     - name: Check out code
-      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v5.0.0
 
     - name: Set up Go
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -13,10 +13,10 @@ jobs:
     steps:
 
     - name: Check out code
-      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v5.0.0
 
     - name: Set up Go
-      uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.31.0
+      uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.4.0
```
melinath, line 19: This is different than in the other PR; which is correct?

melinath (changes requested):

rileykarson:
Stale

PR #27606, opened 2026-05-29
Release 7.34.1 changelog

BBBmau:
changelog for missing resource identity in `google_storage_bucket` resource


rileykarson (changes requested):
Discussed offline; we're rolling out this fix in `7.35.0` on Tuesday and not releasing a patch.

Issue #17999, opened 2024-05-01
Support for Persistence Configuration on Memory Store Redis Cluster


CharlKlein:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to a user, that user is claiming responsibility for the issue.
* Customers working with a Google Technical Account Manager or Customer Engineer can ask them to [reach out internally](https://github.com/hashicorp/terraform-provider-google/wiki/Customer-Contact#raising-gcp-internal-issues-with-the-provider-development-team) to expedite investigation and resolution of this issue.

### Description

Persistence Configuration has been made available in GCP. Would it be appropriate to add it to the Module while the feature is still pre-GA or only after the GA Release 

Pre-GA release date: 5 April 2024
Article: https://cloud.google.com/memorystore/docs/cluster/persistence-overview

Thanks

### New or Affected Resource(s)

* google_redis_cluster

### Potential Terraform Configuration

```tf 
resource "google_redis_cluster" "cluster-ha" {
  name = "its-a-cluster"
  # Omitted for brevity

  persistence_config { # One of the Options must be specified either rdb_config or aof_config

    rdb_config {
      rdb_snapshot_period    = "TWELVE_HOURS"         # Possible values are: ONE_HOUR, SIX_HOURS, TWELVE_HOURS, TWENTY_FOUR_HOURS.
      rdb_next_snapshot_time = "2014-10-02T15:01:23Z" #Timestamp in RFC3339 UTC "Zulu" format
    }

    aof_config {
      append_fsync = "ALWAYS" # Possible values are: NO, ALWAYS, EVERYSEC.
    }
  }
}
```

### References

- https://cloud.google.com/memorystore/docs/cluster/reference/rest/v1/projects.locations.clusters#ClusterPersistenceConfig
- https://cloud.google.com/memorystore/docs/cluster/persistence-overview

StarpTech:
Hi, any updates on this? What's missing to add support for it?

ashokkavuri:
any update on this ?

kenzik:
[...]

Issue #15767, opened 2023-09-07
container_cluster.node_config.kubelet_config.cpu_cfs_quota value is set in cluster request even when no value is set in config


hoskeri:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to the `modular-magician` user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If an issue is assigned to a user, that user is claiming responsibility for the issue. If an issue is assigned to `hashibot`, a community member has claimed the issue already.

### Terraform Version

cloudtop:~$ terraform version
Terraform v1.5.6
on linux_amd64

### Affected Resource(s)
google_container_node_pool
google_container_cluster

### Terraform Configuration Files

```hcl
resource "google_container_cluster" "primary" {
  name     = "some-cluster"
  location = "us-central1-a"
  initial_node_count       = 1
  node_config {
    kubelet_config {
      cpu_manager_policy = "none"
      pod_pids_limit = 8192
    }
  }
}


```

### Debug Output

```
023-09-07T20:43:49.384Z [DEBUG] provider.terraform-provider-google_v4.81.0_x5: ---[ REQUEST ]---------------------------------------
2023-09-07T20:43:49.384Z [DEBUG] provider.terraform-provider-google_v4.81.0_x5: POST /v1/projects/hoskeri-gke-dev/locations/us-central1-a/clusters?alt=json&prettyPrint=false HTTP/1.1
2023-09-07T20:43:49.384Z [DEBUG] provider.terraform-provider-google_v4.81.0_x5: Host: container.googleapis.com
2023-09-07T20:43:49.384Z [DEBUG] provider.terraform-provider-google_v4.81.0_x5: User-Agent: google-api-go-client/0.5 Terraform/1.5.6 (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google/4.81.0

...
023-09-07T20:43:49.384Z [DEBUG] provider.terraform-provider-google_v4.81.0_x5:   "networkConfig": {},
[...]

PR #28194, opened 2026-07-01
fix:  panic in  when creating  resources programmatically via the SDK…

### LINKED ISSUE #28190: Panic: "value is null" in expandNodeConfig when creating `google_container_cluster` without a `node_config` block
Since v7.0.0 (commit `d9313fa07`, PR https://github.com/hashicorp/terraform-provider-google/pull/15268 - which fixed the semantic issue reported in https://github.com/hashicorp/terraform-provider-google/issues/15767), the provider panics with `value is null` inside `expandNodeConfig` whenever `google_container_cluster` is created without a `node_config` block

Its a common pattern when using `remove_default_node_pool = true` and defining nodes exclusively via separate `google_container_node_pool` resources

The fix for #15767 traverses `RawConfig` starting from the resource root. When called from a `google_container_cluster` code path with `prefix="node_pool.N."`, it does `rawConfigNPRoot.GetAttr("node_pool").Index(cty.NumberIntVal(int64(npIndex)))` without first checking that the `node_pool` attribute exists and has at least `npIndex+1` elements in the raw config. When the caller is a programmatic bridge (upjet / SDK-generated wrappers used by Crossplane, Config Connector, etc.) that doesn't populate `node_pool` in the raw config, `GetAttr("node_pool")` returns `cty.NullVal(...)` and `.Index(...)` on that panics with "value is null"

The result is a **runtime nil-dereference that kills the provider process** on every cluster create attempt of this shape

### Impact / Blast Radius

- Any Terraform module or downstream tool (Crossplane provider upjet-provider-gcp, Config Connector, etc.) that creates google_container_cluster without a node_config block is broken on v7.0.0+.
- Common pattern: remove_default_node_pool = true with node pools defined via separate google_container_node_pool resources — panics on every create.
- Regression versus v6.x, which handled the missing block gracefully.
- Silent — no upgrade guide warning, no doc note

### Workaround
- Add a stub node_config { kubelet_config { cpu_cfs_quota = true } } to every cluster resource.

### Terraform Version & Provider Version(s)

Terraform Terraform v1.9.x (any version supporting plugin protocol 5/6) `hashicorp/google v7.0.0` through current (v7.37.x)

Terraform v1.9.5
on darwin_arm64
+ provider registry.terraform.io/hashicorp/google v7.35.0

Also reproducible on:
- Terraform v1.7.x, v1.8.x, v1.9.x
- hashicorp/google v7.0.0 through v7.37.x (all v7.x releases since d9313fa07 landed on 2025-10-01)
- hashicorp/google-beta same version range (same code path)

### Affected Resource(s)

**google_container_cluster**

In `google/services/container/node_config.go`, at the block added by PR 15268 (line 1777 in v7.35.0), inside `expandNodeConfig`, inside the `if v, ok := nodeConfig["kubelet_config"]; ok` branch:

```go
// Added by 15268 to fix the semantic issue in 15767
rawConfigNPRoot := d.GetRawConfig()
if prefix != "" {
    parts := strings.Split(prefix, ".")
    npIndex, err := strconv.Atoi(parts[1])
    if err != nil {
        panic(fmt.Errorf("unexpected format for node pool path prefix: %w. value: %v", err, prefix))
    }
    rawConfigNPRoot = rawConfigNPRoot.GetAttr("node_pool").Index(cty.NumberIntVal(int64(npIndex)))  // ← PANIC HERE
}

if vNC := rawConfigNPRoot.GetAttr("node_config"); vNC.LengthInt() > 0 {
    ...
}
```

The LengthInt() > 0 guards on line 1780 correctly protect against a missing node_config block, but the earlier rawConfigNPRoot.GetAttr("node_pool").Index(...) on line 1777 is not guarded. When the raw config lacks a node_pool attribute (e.g. programmatic / upjet-bridge callers that fill node pools via ForProvider rather than declared HCL), GetAttr("node_pool") returns cty.NullVal and .Index(N) on it panics with "value is null".

### Terraform Configuration

```tf
resource "google_container_cluster" "repro" {
  name               = "panic-repro"
  location           = "us-central1"
  initial_node_count = 1
  network            = google_compute_network.n.id
  subnetwork         = google_compute_subnetwork.s.id
  # NOTE: intentionally no `node_config { ... }` block
}
```

### Debug Output

```shell
panic: value is null

github.com/hashicorp/terraform-provider-google/google/services/container.expandNodeConfig(0x…, {0x0, 0x0}, {0x…, 0x…})
    github.com/hashicorp/terraform-provider-google/google/services/container/node_config.go:1781 +0x1f85
github.com/hashicorp/terraform-provider-google/google/services/container.resourceContainerClusterCreate(...)
    github.com/hashicorp/terraform-provider-google/google/services/container/resource_container_cluster.go
```

### Expected Behavior

Provider should either (a) create the cluster with GKE-default node config (as it did in v6.x), or (b) return a clear validation error at plan time. Instead it panics at apply time

The panic originates inside the cpu_cfs_quota block added by PR 15268: the raw-config traversal at rawConfigNPRoot.GetAttr("node_pool").Index(...) is not guarded against the case where node_pool is absent from RawConfig. This case is trivially reachable from any programmatic caller of the underlying SDK (Crossplane's upjet-based provider, Config Connector, etc.) where node pools are supplied out-of-band rather than declared as raw HCL

### Actual Behavior

`terraform apply` fails on cluster creation with the provider panicking. The GKE Cluster is never created

### Steps to reproduce

1. `terraform apply`

### Important Factoids

_No response_

### References

- Semantic bug that motivated the change: #15767
- Introducing PR: hashicorp/terraform-provider-google#15268 (mirror: GoogleCloudPlatform/magic-modules#15268)
- Introducing commit: d9313fa07e302ddee45a4c182e039e8a2038eba7
- Affected upstream consumers:
    - upbound/provider-gcp (Crossplane) — pending v7 bump blocked on this
    - Any HCL config using remove_default_node_pool = true without an explicit node_config block

b/530561047

aayushrangwala:
Fixes #28190

### Summary

`expandNodeConfig` traverses `d.GetRawConfig()` to determine whether the caller explicitly set `node_config.kubelet_config.cpu_cfs_quota` and `shutdown_grace_period_seconds` / `shutdown_grace_period_critical_pods_seconds`

When called on a `google_container_node_pool` block that belongs to a `google_container_cluster` (prefix `node_pool.N`), the traversal indexes into node_pool without first checking that node_pool exists in raw config

Programmatic SDK callers (Crossplane's upjet-based provider, Config Connector, terraform-plugin-mux consumers, etc.) do not populate the node_pool attribute in raw config, so `GetAttr("node_pool").Index(N)` panics with value is null on every create attempt of any google_container_cluster that uses inline node pools without a raw HCL node_pool block

### Root Cause

Introduced in commit d9313fa07 (PR #15268), which fixed the semantic issue in #15767. The fix added raw-config reads for cpu_cfs_quota but did not guard against the case where the parent node_pool list is absent in raw config. A second, similarly-unguarded block was later added for shutdown_grace_period_seconds in PR #17999 (28015), inheriting the same latent bug

### Fix

Guard the raw-config traversal: if `node_pool is null,` unindexable, or shorter than npIndex, skip both raw-config sub-fixes cleanly. The API-default behaviour then applies (`cpu_cfs_quota=true`, `shutdown_grace_period` unset), matching v6.x behaviour when the block was absent.

```go
npAttr := rawConfigNPRoot.GetAttr("node_pool")
if npAttr.IsNull() || !npAttr.CanIterateElements() || npAttr.LengthInt() <= npIndex {
    skipRawConfigSubFixes = true
} else {
    rawConfigNPRoot = npAttr.Index(cty.NumberIntVal(int64(npIndex)))
}

if !skipRawConfigSubFixes {
    // ... existing cpu_cfs_quota + shutdown_grace_period_seconds blocks
}
```

### Tests

- New acceptance test: TestAccContainerCluster_withoutNodeConfigBlock — regression guard. Creates a google_container_cluster with remove_default_node_pool = true and no node_config block, then attaches a google_container_node_pool resource. Before the fix, this panics on create. After the fix, it succeeds.

### Release Note

container: fixed a panic in expandNodeConfig when creating google_container_cluster resources programmatically via the SDK without a raw node_config HCL block (impacts Crossplane's upjet-based provider, Config Connector, and similar callers)

### Risk Assessment

Low. The change is narrow (one new guard clause) and preserves all existing HCL behavior. Only affects the previously-panicking programmatic path.

### DIFF
--- a/.changelog/28194.txt
+++ b/.changelog/28194.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+container: fixed a panic in `expandNodeConfig` when creating `google_container_cluster` resources programmatically via the SDK without a raw `node_config` HCL block (impacts Crossplane's upjet-based provider, Config Connector, and similar callers)
+```
--- a/google/services/container/node_config.go
+++ b/google/services/container/node_config.go
@@ -1827,6 +1827,14 @@ func expandNodeConfig(d *schema.ResourceData, prefix string, v interface{}) *con
 		rawConfigNPRoot := d.GetRawConfig()
 		// if we have a prefix, we're in `node_pool.N.` in GKE Cluster. Traverse the RawConfig object to reach that
 		// root, at which point local references work going forwards.
+		//
+		// Guard: some callers (programmatic SDK bridges such as Crossplane's
+		// upjet-generated provider, Config Connector, etc.) do not populate the
+		// `node_pool` attribute in raw config even though `prefix` is set. In
+		// that case GetAttr("node_pool") returns cty.NullVal and Index(N)
+		// panics with "value is null". Skip the raw-config sub-fixes below
+		// cleanly in that case. See hashicorp/terraform-provider-google#28190.
+		skipRawConfigSubFixes := false
 		if prefix != "" {
 			parts := strings.Split(prefix, ".") // "node_pool.N." -> ["node_pool" "N", ""]
 			npIndex, err := strconv.Atoi(parts[1])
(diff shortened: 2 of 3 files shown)


aayushrangwala:
should be resolved by this PR
https://github.com/GoogleCloudPlatform/magic-modules/pull/18148

PR #28415, opened 2026-07-21
sql: send password_policy on google_sql_user updates

eldad:
Changes to password_policy were silently dropped because the update path never attached the expanded policy to the User object sent to the Cloud SQL Users.Update API. Wire up expandPasswordPolicy in the update path, mirroring the create path.


eldad:
Closing since the source is indeed in https://github.com/GoogleCloudPlatform/magic-modules.

danilo-rocha-unifique:
> Closing since the source is indeed in https://github.com/GoogleCloudPlatform/magic-modules.

https://github.com/GoogleCloudPlatform/magic-modules/pull/18350

